PR #17, opened 2019-06-27
Make sure to render as public if the user is not logged in

rullzer:
Else we will render the apps, contacts menu etc. Which is undesireable.


jancborchardt:
@skjnldsv check it :)

PR #33, opened 2019-07-15
Update max-version for current development version

jancborchardt:
Please review @nextcloud/forms


affan98:
I don't have write access so I am not able to approve.  I'd love to continue to be involved so getting access would be appreciated

jancborchardt:
@affan98 you are in the Nextcloud org (as that you have write access) and I was able to set you as a reviewer, so it should work? :)

We only set the master branch to protected, so all changes have to go through pull requests. That’s what we do always so changes are reviewed.

skjnldsv:
Tests are still failing.
We should check this. :)

PR #40, opened 2019-07-23
Don't allow access if registered user has already taken part in survey

marcelklehr:
It might be debatable if this should be optional on form creation, but since I hacked this in for my installation, I didn't want to withhold this from you :)

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -438,6 +438,7 @@ class ApiController extends Controller {
 		$newEvent->setDescription($event['description']);
 
 		$newEvent->setIsAnonymous($event['isAnonymous']);
+		$newEvent->setUnique($event['unique']);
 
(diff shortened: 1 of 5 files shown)


affan98:
I think this would be better as an optional setting.  Could you add some kind of button to the settings that lets you choose to 'only allow one response per user'

marcelklehr:
There's two possibilities for implementing this, it could either be an authorization option: ie. instead of 'registered' the access restriction would be 'registered_once' -- this is not too clean, IMHO. A different approach would be to add a new option that is only allowed when using registered-only access.

marcelklehr:
I've added an option in a new column on the event table.

affan98:
So I'm not able to test right now but the code looks good.  If you've tested this well enough I'll merge.

marcelklehr:
I've tested it, but I can't guarantee that it won't break anything else.

PR #60, opened 2019-08-30
Pimp entities and models

rullzer:
* Strict
* Extend proper class
* Typehint
* Return types


lib/Db/Question.php
```diff
@@ -35,7 +36,7 @@
  * @method integer getTimestamp()
  * @method void setTimestamp(integer $value)
  */
-class Question extends Model {
```
skjnldsv: Was it working? Seems like a serious typo to me!

lib/Db/Model.php
```diff
@@ -1,39 +0,0 @@
...
- *  You should have received a copy of the GNU Affero General Public License
- *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *
- */
-
-namespace OCA\Forms\Db;
-
-use OCP\AppFramework\Db\Entity;
-
-abstract class Model extends Entity {
```
skjnldsv: ah ok :)

PR #64, opened 2019-08-30
Use a proper DELETE to delete the form in a more resty way

rullzer:

### DIFF
--- a/src/views/List.vue
+++ b/src/views/List.vue
@@ -63,9 +63,9 @@
 </template>
 
 <script>
-// import moment from 'moment'
-// import lodash from 'lodash'
+
 import formListItem from '../components/formListItem'
+import axios from 'nextcloud-axios'
 
 export default {
 	name: 'List',
@@ -119,7 +119,7 @@ export default {
 				buttonConfirmText: t('forms', 'Yes, delete form.'),
 				onConfirm: () => {
 					// this.deleteForm(index, event)
-					this.$http.post(OC.generateUrl('apps/forms/remove/form'), event)
+					axios.delete(OC.generateUrl('apps/forms/forms/{id}', {id: event.id}))
 						.then((response) => {
 							this.forms.splice(index, 1)
 							OC.Notification.showTemporary(t('forms', 'Form "%n" deleted', 1, event.title))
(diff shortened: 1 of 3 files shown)


skjnldsv:
Could be nice to also have more beautiful apis for the other requests

``` php
['name' => 'api#write_form',	'url' => '/write/form',			'verb' => 'POST'],
['name' => 'api#get_form',		'url' => '/forms/{id}',			'verb' => 'GET'],
['name' => 'api#get_options',	'url' => '/forms/{id}/options',	'verb' => 'GET'],
['name' => 'api#get_votes',		'url' => '/forms/{id}/votes',	'verb' => 'GET'],
['name' => 'api#get_shares',	'url' => '/forms/{id}/shares',	'verb' => 'GET'],
['name' => 'api#get_event',		'url' => '/forms/{id}/event',	'verb' => 'GET'],
['name' => 'api#remove_form',	'url' => '/forms/{id}',			'verb' => 'DELETE'],
['name' => 'api#get_forms',		'url' => '/forms',				'verb' => 'GET'],
```

rullzer:
Yes baby steps

src/views/List.vue
```diff
@@ -119,7 +119,7 @@ export default {
 				buttonConfirmText: t('forms', 'Yes, delete form.'),
 				onConfirm: () => {
 					// this.deleteForm(index, event)
-					this.$http.post(OC.generateUrl('apps/forms/remove/form'), event)
+					axios.delete(OC.generateUrl('apps/forms/forms/{id}', {id: event.id}))
```
marcelklehr: I think this.$http is an axios instance, already.

skjnldsv: Nope :)
https://vuejs.org/v2/cookbook/adding-instance-properties.html#Real-World-Example-Replacing-Vue-Resource-with-Axios

marcelklehr: https://github.com/nextcloud/forms/blob/master/src/main.js#L60

skjnldsv: Ah sorry, I thought you said that vuejs was shipping axios at this.$http :p
Indeed!

Well, I see no reason to bind this.$http to axios if we can just use axios instead. Confusing otherwise as we still assume it's the default vue request handler ;)

rullzer: Yes it is. But I prefer to import it where needed as well. Cleaner. Easier tree shaking per chunk etc.

PR #71, opened 2019-09-03
Update configs and removed unused dependencies

skjnldsv:
- removed unused dependencies
- fixed incompatible dependencies
- use appropriate apps file schema
- use unified makefile
- add proper linting (css/js)
- removed yarn and switch to npm as we do it on all of our apps


affan98:
Actually one thing.  Could you update the readme to reflect that we aren't using yarn

skjnldsv:
Done :)

Issue #27, opened 2019-07-13
Replace "View Results" text with icon


bpcurse:
**Which Forms version are you running?** (see apps page)
1.0.3

**Nextcloud or ownCloud?:**
Nextcloud

**Nextcloud/ownCloud version:** (see Nextcloud admin page)
15.0.5

### Actual behaviour
The large "View Results" text with blue background is not in line with the other icons.
![Bildschirmfoto_2019-07-13_15-09-44](https://user-images.githubusercontent.com/29312856/61172103-54495600-a580-11e9-9b0c-bc32767751d3.png)

### Expected behaviour
It should be replaced by a meaninguful icon like "icon-toggle" (https://docs.nextcloud.com/server/16/developer_manual/_images/toggle.png), glasses or a combination of icons.

luixxiul:
Closed with #118

PR #118, opened 2019-11-10
Replace "View Results" text with icon

justin-sleep:
Implements #27.


justin-sleep:
I've added text for accessibility to all of the icons.

PR #123, opened 2019-11-20
Fix submitting answers

ordaolmayanadam:
Events should be passed with "@" not ":" which makes it impossible to add answers to questions.


jospoortvliet:
That does fix it for me, indeed ;-) :+1: 
Nice catch!

PR #130, opened 2019-11-29
Unnecessary $questions assignment removed

ordaolmayanadam:
$questions assignment is not used in the function so removed

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -271,8 +271,7 @@ class PageController extends Controller {
 		}
 
 		$votes = $this->voteMapper->findByForm($form->getId());
-		$participants = $this->voteMapper->findParticipantsByForm($form->getId());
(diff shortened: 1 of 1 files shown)


jancborchardt:
The build fails with PHP 7.1, but I assume that’s a separate issue? cc @skjnldsv

jancborchardt:
Can you check this @skjnldsv? :)

skjnldsv:
> Can you check this @skjnldsv? :)

I have no clue ^^

marcelklehr:
> `CORE_BRANCH=master`

Does master still support php v7.1?

skjnldsv:
> Does master still support php v7.1?

guess not, we also support 7.4

jancborchardt:
> > Can you check this @skjnldsv? :)
> 
> I have no clue ^^

Do you know who has a clue? :) @rullzer maybe?

rullzer:
Rebase on master and all should be good

PR #183, opened 2020-01-24
Perform CSRF checks

rullzer:


rullzer:
I did some smoke testing and it seemed good. If we need to add something back later then so be it. But the majority of those calls should have proper CSRF protection!

Issue #179, opened 2020-01-21
Use the IP address of the user if userId is empty


githubDante:
**Which Forms version are you running?** (see apps page)
git [master]
**Nextcloud or ownCloud?:**
Nextcloud
**Nextcloud/ownCloud version:** (see Nextcloud admin page)
17.0.2

#### Description
If the patch proposed in #178 is applied, the form is **not** configured as _Anonymous Form_ and Access is set to _Public access_ the _Name_ column in the results view will be left empty. The patch included below is using the IP address from where the request is received as userId and as a result the IP address will be shown in _Name_ column.

```
diff --git a/lib/Controller/PageController.php b/lib/Controller/PageController.php
index dcb60f0..5ba188b 100644
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -278,7 +278,11 @@ class PageController extends Controller {
                $questions = $this->questionMapper->findByForm($form->getId());

                try {
-                       $notification = $this->notificationMapper->findByUserAndForm($form->getId(), $this->userId);
+                       if($this->userId === null) {
+                               $notification = $this->notificationMapper->findByUserAndForm($form->getId(), '');
+                       } else {
+                               $notification = $this->notificationMapper->findByUserAndForm($form->getId(), $this->userId);
+                       }
                } catch (DoesNotExistException $e) {
                        $notification = null;
                }
@@ -400,6 +404,9 @@ class PageController extends Controller {
                                if($form->getIsAnonymous()){
                                                $vote->setUserId($anonID);
                                }else{
+                                               if ($userId === '') {
+                                                       $userId = $this->request->getRemoteAddress();
+                                               }
                                                $vote->setUserId($userId);
                                }
                                $vote->setVoteOptionText(htmlspecialchars($questions[$i]['text']));
```

oscaropenness:
I like your suggestion of using some identifier for unidentfied users.
I tried to apply your suggestion in my Nextcloud instance. My userId column is still empty.
There is no error in the log or anything. Am I missing something?

Another thought:
From a privacy point of view it would be more appropriate to store some hash of the RemoteIp since it contains already some information about the user (location, workplace, country...)

githubDante:
I just tested again with the master branch and the IP address is displayed correctly in the Name column. I'm sending a pull request.

jotoeri:
Things are mixed up here a bit.
The internal server Error itself is removed on master, just a release is pending.

[...]

Issue #178, opened 2020-01-21
Internal server error when public link is used 


githubDante:
**Which Forms version are you running?** (see apps page)
git [master] - see #166 why
**Nextcloud or ownCloud?:**
Nextcloud
**Nextcloud/ownCloud version:** (see Nextcloud admin page)
17.0.2
### Steps to reproduce
1. Create a form, set the access type to Public.
2. Open the form link from private window.

### Actual behaviour
Internal server error.

### Expected behaviour
The link should provide access to the form.

### Logs

* Log from the admin panel:
```
[index] Error: Exception: Argument 2 passed to OCA\Forms\Db\NotificationMapper::findByUserAndForm() must be of the type string, null given, called in
<removed>/apps/forms/lib/Controller/PageController.php on line 281 at <<closure>>

0. <removed>/lib/private/AppFramework/App.php line 126
   OC\AppFramework\Http\Dispatcher->dispatch(OCA\Forms\Controller\PageController {}, "gotoForm")
1. <removed>/lib/private/AppFramework/oRuting/RouteActionHandler.php line 47
   OC\AppFramework\App::main("OCA\\Forms\\Controller\\PageController", "gotoForm", OC\AppFramework\ ... {}, {hash: "2tk0hQCg     ... "})
2. <<closure>>
   OC\AppFramework\Routing\RouteActionHandler->__invoke({hash: "2tk0hQCg ... "})
3. <removed>/lib/private/Route/Router.php line 297
   undefinedundefinedcall_user_func(OC\AppFramework\ ... {}, {hash: "2tk0hQCg ... "})
4. <removed>/lib/base.php line 997
   OC\Route\Router->match("/apps/forms/form/2tk0hQCgNLExJ9xK")
5. <removed>/index.php line 42
   OC::handleRequest()
GET /index.php/apps/forms/form/2tk0hQCgNLExJ9xK
```
### Reason
userID will remain null in the scenario described above, which in turn causes the reported error.

### Fix
The following patch contains a fix.

[PageController_notifMapper_patch.txt](https://github.com/nextcloud/forms/files/4091661/PageController_notifMapper_patch.txt)

schniepp:
Worked for me! Hope this will get merged! For 1.1.2?

jotoeri:
Solved with #235 .

riverboboy:
How can I change it?

PR #195, opened 2020-02-12
Fix for the internal server error reported in #178

githubDante:
Additionally the vote will use the IP address from the request if the form
is not _Anonymous_ and the user is not logged in. See #179

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -274,7 +274,11 @@ class PageController extends Controller {
 		$participants = $this->voteMapper->findParticipantsByForm($form->getId());		
 
 		try {
-			$notification = $this->notificationMapper->findByUserAndForm($form->getId(), $this->userId);
+			if($this->userId === null) {
+				$notification = $this->notificationMapper->findByUserAndForm($form->getId(), '');
+			} else {
+				$notification = $this->notificationMapper->findByUserAndForm($form->getId(), $this->userId);
+			}
 		} catch (DoesNotExistException $e) {
 			$notification = null;
 		}
@@ -394,6 +398,9 @@ class PageController extends Controller {
 				if($form->getIsAnonymous()){
 						$vote->setUserId($anonID);
 				}else{
+						if ($userId === '') {
+							$userId = $this->request->getRemoteAddress();
+						}
 						$vote->setUserId($userId);
 				}
 				$vote->setVoteOptionText(htmlspecialchars($questions[$i]['text']));


JanDragon:
Hey,

thanks for your first PR. Code looks good and is working for me. Please add a DCO and we are ready. More about DCO? Check out this link: https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for

githubDante:
Fixed. Thanks for the advice.

lib/Controller/PageController.php
```diff
@@ -274,7 +274,11 @@ public function gotoForm($hash) {
 		$participants = $this->voteMapper->findParticipantsByForm($form->getId());		
 
 		try {
-			$notification = $this->notificationMapper->findByUserAndForm($form->getId(), $this->userId);
+			if($this->userId === null) {
+				$notification = $this->notificationMapper->findByUserAndForm($form->getId(), '');
+			} else {
+				$notification = $this->notificationMapper->findByUserAndForm($form->getId(), $this->userId);
+			}
```
jotoeri: This part should suffice to solve the error itself.

lib/Controller/PageController.php
```diff
@@ -394,6 +398,9 @@ public function insertVote($id, $userId, $answers, $questions) {
 				if($form->getIsAnonymous()){
 						$vote->setUserId($anonID);
 				}else{
+						if ($userId === '') {
+							$userId = $this->request->getRemoteAddress();
+						}
```
jotoeri: Why not using the already existing anonymous ID?

githubDante: Because this only applies when the form is not marked as anonymous.

jotoeri: Sure. But does the ip of an unidentified user contain helpful information? If i wanted to know who answered, i would make the form only visible to registered users...?!

jotoeri: ......in fact, its not that big thing. Both works. ;)

jotoeri:
Btw.: If you just amend an empty commit, the Travis-check should run again. Seems like the failing was not due to your code, but a wget() failed within this moment.

lib/Controller/PageController.php
githubDante: I didn't want to change anything in the current implementation, just to add on top of it. :)

jotoeri: I'd say just use a hash of the ip, as @oscaropenness proposed on #179. Then push again and i would give it a go... ;)

jotoeri: Ah, and maybe better place it in line 376 below the anonID. Then it also applies to Checkbox-Questions.

jotoeri:
I think this PR can be closed, as the main-topic (internal server error) became obsolete with #235 .
If you continue working on, just open a new PR for the IP-Address-Topic.

Issue #176, opened 2020-01-19
Error on saving form, see console


nooblag:
**Which Forms version are you running?**
1.1.0

**Nextcloud or ownCloud?:**
Nextcloud

**Nextcloud/ownCloud version:**
17.0.1

### Steps to reproduce
1. Fresh install of Forms plugin
2. Create new form
3. Add fields to form, click "Done."

### Actual behaviour
Upon clicking "Done," message appears: "Error on saving form, see console."
Form does not save.

### Expected behaviour
Form should be added, saved successfully.

**In case of a bug, please answer the following questions**

### Server configuration

**Where did you install Nextcloud from:**
https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh

**Operating system**:
Ubuntu 18.04 LTS

**Web server:**
Default (Apache 2.4.29)

**Database:**
Default (PostgreSQL 10.10)

**PHP version:**
Default (PHP 7.2.24)

**Signing status:**
Don't known what that means.

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.
```
No errors have been found.

**List of activated apps:**

AppOrder
0.8.0
Bookmarks
2.3.1
Calendar
1.7.1
Official
Camera RAW Previews
0.7.2
Extract
1.2.2
Full text search - Files - Tesseract OCR
1.3.1
Group folders
5.0.4
Official
Mail
0.20.0
Official
Notes
3.0.3
Official
ONLYOFFICE
3.0.2
Official
Suspicious Login
3.0.0
Accessibility
1.3.0
Official
Activities for shared file downloads
1.6.0
Activity
2.10.1
Official
Announcement center
3.6.1
Official
Auditing / Logging
1.7.0
Official
Collaborative tags
1.7.0
Official
Comments
1.7.0
Official
Contacts
3.1.6
Official
Deleted files
1.7.0
Official
DropIt
0.3.0
External sites
3.4.1
Official
Favorites Quickaccess Sorting
1.0.2
Federation
1.7.0
Official
File access control
1.7.0
Official
File sharing
1.9.0
Official
Files automated tagging
1.7.0
Official
First run wizard
2.6.0
Official
Forms
1.1.0
Full text search
1.3.6
Full text search - Files
1.3.6
Gallery
18.4.0
Official
Guests
1.3.1
Impersonate
1.4.0
Official
Log Reader
2.2.0
Official
Metadata
0.10.0
Monitoring
1.7.0
[...]

Issue #166, opened 2020-01-15
Answering options cannot be added in v1.1.0


DBJRdev:
**Which Forms version are you running?**
1.1.0, installed today from app store via Nextcloud web interface

**Nextcloud/ownCloud version:**
Nextcloud 17.0.2

### Steps to reproduce
1. Start to create form
2. Pick any clickable answer type (radio, checkboxes, dropdown)
3. Any given answer option will not be accepted by the Forms app
![Bildschirmfoto von 2020-01-15 15-29-42](https://user-images.githubusercontent.com/13522599/72442101-31586280-37ac-11ea-9793-f68a3a014a54.png)

Tested on Firefox 72.0.1 and Chromium 79, both on Ubuntu 19.10

### Actual behaviour
There seem to be no way to enter answering options. The visible input field does not do anything. If you enter the answer options there, you won't be able to finish creating – when clickin on "Fertig" (Finish?): there is a message that all questions need answers.

### Expected behaviour
Some possibility to enter answering options for checkboxes, radio buttons and dropdown questions.

schniepp:
Same problem here. NC17, apache, chrome, server:ubuntu-18.04, client:ubuntu-18.04

joseams86:
I have the same problem too. I tried installing from the store, downloading from the web and it does not work, I also tried to install the previous version of forms but it is not compatible with nextcloud 17.

I hope someone knows the solution or update the error soon.

fdrgit:
Same problem. Nextcloud 17.0.2 server on Debian 9 (stretch), forms 1.1.0 installed from the store. Tested with Chrome Version 79.0.3945.117 running on win10. 
Was working fine with previous release of forms...

born2webdesign:
Same here.
Is there anyone who _doesn’t_ see this with NC 17(.0.2) and forms 1.1.0?
I think the problem is with Forms and/or NC; but, FWIW: server is PHP 7.3, MySQL, Apache; client is Firefox 68.4.1 on Debian 10.

pwepwe973:
Sam Problem for me.

mikweb2017:
Sam Problem for me NC 18 and forms 1.1.0

jotoeri:
Didn't try now, but a few weeks ago it was possible to add answers by typing and pressing Enter.
Indeed, it's just kind of a workaround for now, a button or so would be helpful.

githubDante:
Same here. Nextcloud 17 and the latest version of the app installed through the admin panel. 

[...]

PR #199, opened 2020-02-17
Make 'unique' an integer

jotoeri:
Solves #176 #189

Greets,
Jonas

PS: Somebody able to release a V1.1.1? I think this would make the app at least usable as-is for many users again, since aboves Issues, as well as #166 would be solved...?!

### DIFF
--- a/lib/Db/Event.php
+++ b/lib/Db/Event.php
@@ -46,7 +46,7 @@ use OCP\AppFramework\Db\Entity;
  * @method integer getIsAnonymous()
  * @method void setIsAnonymous(integer $value)
  * @method integer getUnique()
- * @method void setUnique(boolean $value)
+ * @method void setUnique(integer $value)
  */
 class Event extends Entity {
 	protected $title;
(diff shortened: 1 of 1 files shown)


jotoeri:
And sorry for this number of force-pushed versions of the commit. I was just struggling with the DCO-Check.

jotoeri:
That't true. I chose this way, as the two Anonymous-Columns are handled as integer, too.
But indeed, it would make sense to alter all three to proper boolean columns. Maybe as part of #206 ? ;)

juliushaertl:
> But indeed, it would make sense to alter all three to proper boolean columns. Maybe as part of #206 ? ;)

Yes, if there is more to cleanup, it makes sense to do this separately and properly planned :+1:

skjnldsv:
@jotoeri so you'll tackle the rest in a followup pr?

jotoeri:
I'm not sure if i can do it all on my own, as its quite much to do and i'm not very experienced yet. But if a few people are working together on it, i will gladly participate or start with it.
Would be nice if some of you could give their opinion on #206 .

rullzer:
@skjnldsv maybe setup a talk room for forms?

skjnldsv:
https://cloud.nextcloud.com/call/im4cw68i :)

PR #215, opened 2020-02-28
Move navigation entry to info.xml

rullzer:
Just to cleanup a bit


appinfo/info.xml
```diff
@@ -34,4 +34,12 @@
     <dependencies>
         <nextcloud min-version="17" max-version="19" />
     </dependencies>
+	<navigations>
+		<navigation>
+			<name>Forms</name>
+			<route>forms.page.index</route>
+			<icon>app.svg</icon>
+			<order>77</order>
```
skjnldsv: Can we use 66 ?
https://www.youtube.com/watch?v=lb7kdPYfsm8

appinfo/app.php
```diff
@@ -1,25 +0,0 @@
...
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- *  GNU Affero General Public License for more details.
- *
- *  You should have received a copy of the GNU Affero General Public License
- *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *
- */
-
-$app = new \OCA\Forms\AppInfo\Application();
-$app->registerNavigationEntry();
```
skjnldsv: Don't we still need this to properly init the app via query?

rullzer: Nah happens automagically :) Basically this was a hack that is not needed now

skjnldsv: Oh really? So no log spamming? 
So we did all those crazy fixes for 18 for nothing? :see_no_evil:

skjnldsv: All those https://github.com/nextcloud/server/pull/17829
https://github.com/nextcloud/server/pull/17805/files ?

rullzer: No we didn't. Because those need to be initialized in app.php
But this app doesn't need to be.

skjnldsv: Why doesn't this one need to?

Issue #206, opened 2020-02-23
Alter Database


jotoeri:
Hi all,

i've got a quite heavy proposal, about altering the Database. But maybe its worth to do it now, rather than in a further state of development. **I would appreciate your opinions.**

### Background
Looking into the Database, it took me quite a while to understand the sense of the tables. And especially for the tables of questions and votes, i would have expected some relation, but except the form_id it was hard to find some:
- In the questions-table, the questions are stored just with the corresponding form_id, their text and a *app-wide* question-id.
- Within the votes-table however (containing the submittted answers), the form_id is still related to the form, while a "vote-option-id" seems to indicate a *form-wide* question-id, which is not related to the questions-table. Instead, the Questions Text & Type are just stored a second time with the answer, producing kind of a duplicate of the question-table for each submission of the form.

[oc_forms_questions.png](https://github.com/jotoeri/share/blob/master/oc_forms_questions.PNG)
[oc_forms_votes.png](https://github.com/jotoeri/share/blob/master/oc_forms_votes.PNG)

### My Proposal
So - my Proposal would be to alter the database - and now that one is on to do so - to produce a **relational database** with **intuitive naming**. Of course - it would be a heavy change to do so, but doing it now, would make further development quite easier, while in a later state of development, there would be more functionality to alter.

#### A possible target-database could be:
- _oc_forms_forms_
-> [form_id, hash, title, description, owner, created, access, expire, is_anonymous, full_anonymous, ans_unique]
-> unique(form_id)
- _oc_forms_questions_
-> [form_id, question_id, type, text]
-> unique(form_id.question_id)
- _oc_forms_options_
-> [form_id, question_id, option_id, text]
-> unique(form_id.question_id.option_id)
- _oc_forms_answers_
-> [form_id, submit_id, question_id, user_id, answer]
-> (form_id.submit_id) forms a unique identifier.
- _oc_forms_notif_
-> [id, form_id, user_id] (just as until now)

#### oc_forms_forms
- Rename table '_oc_forms_events_' to '_oc_forms_forms_'
-> The table contains in fact the forms, so why calling it events?
- In new '_oc_forms_forms_' rename column '_id_'  to '_question_id_'
-> Just to use the same naming withing db
- In new '_oc_forms_forms_' rename column '_unique_'  to '_ans_unique_'
-> unique seems to be an (even deprecated) mysql-keyword. The column indicates if a user is allowed to answer the form once or multiple times.
#### oc_forms_questions
[...]

Issue #34, opened 2019-07-16
Forms app review


jancborchardt:
@skjnldsv @rullzer & @jancborchardt at Contributor Week June 2019

We also looked at [Google Forms](https://www.google.com/forms/about/) and [Typeform](https://www.typeform.com/) for comparison. We think it’s best to mainly model it after Google Forms as it’s quite widely used.

Below are detail review points of the current state so we can make small steps in the right direction based on the current app. :)

cc @nextcloud/forms 

## Initial list view
* [x] Forms should be listed in the left navigation instead of the main content. The main content is to edit the form.
* [x] Help button too present
* [x] Home icon is confusing and not necessary?
* [x] List header shows even though no forms
* [x] Emptycontent message missing?

## Form creation
* [x] No way to submit/confirm answer with mouse
* [x] Misalignment of answers
* [x] Too many delete buttons visible at the same time
* [x] Too many different font-sizes
* [x] Interface doesn’t reflect how the form will look for people afterwards
* [x] No way to reorder answers or put things inbetween
* [x] Replies are not autosaved

Answer types:
* [x] Icons for answer types
* [ ] No way to change the answer type
* [x] "Radio button" → Multiple choice
* [x] "Checkboxes" → Checkboxes
* [x] "Short response" → Short answer
* [x] "Long response" → Long text
* [x] "Drop Down" is the same as Single choice?

## Sidebar
* [x] Only 1 tab, which is not necessary → use sidebar component
* [ ] Spacing on owner is strange
* [ ] Settings are in 2 columns – unusual
* [ ] Settings like "Anonymous form" need explanation
* [ ] Use main Nextcloud sharing – default is public access which means everyone on the instance can see it

## List after creation
* [ ] Icon not clear (with green dot)
* [x] Clicking form should open editing configuration
* [x] Delete too present
* [x] "Created" date unnecessary

## Share view
* [x] Header should not show apps or contacts menu <- this is the logged-in view
* [ ] Header should not show "Nextcloud" placed a bit to the top <- this is unauthenticated view
* [ ] Header should have the survey title
* [ ] Input fields and textareas don’t look like inputs cause the other answer types have a container like that around it too
* [ ] Asterisk `*` before questions is confusing? Mandatory? Not possible to set or unset. Should instead say "mandatory" in a subline
* [x] "Thank you for completing he survey!" page is very bland

## Results view
[...]

PR #217, opened 2020-03-02
Add Submit button and alignments on form creation

jotoeri:
Heyhey,

according to the review issue i now added a button to add options on form creation, and aligned the form-creation elements.
I'm not sure if this is now, what you would call a good style. Do you have any Coding-Styleguide or so, i could have a look into?

Greets,
Jonas

PS: As i said - i'm not that experienced yet, so please feel free to tell me, what shit i'm doing here. ;)

grafik

### DIFF
--- a/src/components/quizFormItem.vue
+++ b/src/components/quizFormItem.vue
@@ -20,32 +20,34 @@
   -
   -->
 <template>
-	<li class="questionList">
-		<div id="text">
-			{{ question.text }}
-		</div>
-		<div v-show="(question.type != 'text') && (question.type != 'comment')"
-			id="add-option">
-			<input
-				v-model="newQuizAnswer"
-				:placeholder=" t('forms', 'Add Answer')"
-				@keyup.enter="emitNewAnswer(question)">
-			<button class="symbol icon-add"
-				@click="emitNewAnswer(question)" />
-		</div>
-		<div id="optionList">
-			<transitionGroup
-				name="transitionList"
-				tag="ul"
-				class="form-table">
-				<li
-					is="text-form-item"
-					v-for="(ans, index) in formQuizAnswers"
-					:key="ans.id"
-					:option="ans"
-					@remove="emitRemoveAnswer(question, index)"
-					@delete="question.answers.splice(index, 1)" />
-			</transitionGroup>
+	<li id="questionList">
+		<div id="questionListData">
+			<div id="text">
+				{{ question.text }}
+			</div>
+			<div v-show="(question.type != 'text') && (question.type != 'comment')"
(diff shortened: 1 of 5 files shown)


jancborchardt:
@jotoeri do you have a link to the issue this fixes, or a before/after screenshot for context? :)

jotoeri:
It is part of the Review-Issue #34, tackling the three points of 

>  **Form creation**
> - No way to submit/confirm answer with mouse
> - Misalignment of answers
> - Too many delete buttons visible at the same time


Old Version was:
old


New one is:
new

- Recyclebins are visible on hover
- The sidebar only scales down to the smaller width on long question texts as the lorem ipsum. Normally it has the same width as in old picture. This is a feature that was already existing.
- Maybe it is not perfect yet if you only input short questions, but otherwise it is hard to align the Option-Inputs for even long and short question-texts...

jancborchardt:
Looks good, much better! :)

> * Recyclebins are visible on hover

Here make sure that they are also accessible via keyboard-only when people go through the entries with the Tab key. :)

jotoeri:
Took me quite a while, but now they are also well usable on keyboard... :)

marcelklehr:
I tend to avoid on-hover solutions, because they're not very accessible on mobile.

jotoeri:
Well - i tested on two Smartphones and my tablet, the hover shouldn't be a problem. Just tapping onto the Item, the bin shows up. Otherwise you would have to show all delete-icons again.

Anyways i recognized, the view on mobile devices was not usable until now. I corrected that, so it would be usable. For now it would work, even that it might be necessary to have a look onto the design again after including the sidebar.

jotoeri:
Hm, anybody else any opinion?
Is it worth to rebase and care for or should i close the PR, as these changes are not wanted? Some request for changes?
@jancborchardt @skjnldsv

skjnldsv:
Hey @jotoeri 
Sorry about this! :)

So there is a lot of restructuring that is happening in the forms app.
Unfortunately, your pr conflicts a lot with what will happen.

I would suggest we wait until i'm done with the cleaning, and then we will be able to properly work on separate areas of forms, what do you think? :)

jotoeri:
Ok, tank you for the information. :)

I'm currently also working on the Database-Thing #206. Sounds like i better stop this work until you are done? ([Branch alter_Db](https://github.com/jotoeri/forms/tree/alter_Db/develop) Branch is a bit dirty until now.... )

skjnldsv:
I do not touch much in the backend, so we should be safe here :)

jotoeri:
Ok :)

Issue #229, opened 2020-03-11
Forms 1.1.1 is completely broken


pzehle:
### Steps to reproduce
1. Install forms app (if it is installed, delete it and re-install).
2. As an admin, click on forms, then new form, and reload page.
3. As a regular user click on forms.

### Expected behaviour
To be able to create a form, edit it, add answers, share, as an admin and as a normal user.

### Actual behaviour
As an admin, after clicking on create form, and reloading the page, it says _Access forbidden, CSRF check failed_. No forms can be created.
As a normal user, clicking on forms result on a redirection to / and a message in the console:

`TypeError: Cannot read property 'call' of undefined
    at o (forms.js?v=30498835-10:1)
    at Object.<anonymous> (1.js:68)
    at n (1.js:1)
    at Module.<anonymous> (1.js:68)
    at n (1.js:1)
    at 1.js:1
    at Object.<anonymous> (1.js:1)
    at o (forms.js?v=30498835-10:1)
    at Object.227 (5.js:1)
    at o (forms.js?v=30498835-10:1)`

## Server configuration detail

**Operating system:** Linux 4.15.0-66-generic #75-Ubuntu SMP Tue Oct 1 05:24:09 UTC 2019 x86_64

**Webserver:** Apache/2.4.29 (Ubuntu) (apache2handler)

**Database:** mysql 10.4.12

**PHP version:** 

7.2.24-0ubuntu0.18.04.3
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, sodium, session, standard, apache2handler, mysqlnd, PDO, xml, apcu, apc, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, iconv, imagick, intl, json, exif, mysqli, pdo_mysql, Phar, posix, readline, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, zip, Zend OPcache

**Nextcloud version:** 18.0.1 - 18.0.1.3

**Updated from an older Nextcloud/ownCloud or fresh install:** 

**Where did you install Nextcloud from:** unknown

<details><summary>Signing status</summary>

Array
(
)

</details>

<details><summary>List of activated apps</summary>

```
Enabled:
 - accessibility: 1.4.0
 - admin_audit: 1.8.0
 - announcementcenter: 3.7.0
 - bruteforcesettings: 1.5.0
[...]

PR #237, opened 2020-03-20
Cleanup and change routes

skjnldsv:
First batch of cleanup

Capture d’écran_2020-03-20_16-35-35

- [x] Cleanup routes
- [x] Cleanup app structure
- [x] Emptycontent(s)
- [x] Add Navigation
- [x] Navigation form actions
- [x] Navigation icons
- [ ] QA


lib/Controller/PageController.php
```diff
@@ -86,40 +87,63 @@ public function __construct(
...
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 *
+	 * @return TemplateResponse
+	 */
 	public function index(): TemplateResponse {
-		return new TemplateResponse('forms', 'forms.tmpl',
-		['urlGenerator' => $this->urlGenerator]);
+		Util::addScript($this->appName, 'forms');
+		Util::addStyle($this->appName, 'icons');
```
rullzer: I assume the final plan is to include the scss in the bundle right?

skjnldsv: Not icons, as it uses the server scss functions.
But al the other style: yes :)

jotoeri:
Wow. Kind of seems like you are writing a completely new app. Respect! :) 👍

skjnldsv:
> Wow. Kind of seems like you are writing a completely new app. Respect! :)

Mostly moving things around and using our vue components :)

skjnldsv:
@jotoeri can you do a review and see if everything is still working for you?
We'll do the design updates/cleanup and fixes on followup prs :)
As well as tackle Jan's comments here and here: https://github.com/nextcloud/forms/issues/34

jotoeri:
Sure, here is what i got on a newly installed NC18.0.1:

Functionality:
- How can i respond/submit to a form? Or how to share a link to a form? I didn't find a way... :(
- When changing between editing of two forms, the already added answers do not show up. Seems to be connected to the database-issues below.
- Show results gives me an "Access denied" Notification, even that i am the author of the form.

Looking into the database:
- Each change of the form produces the database to be changed. Even selecting a different form, or while typing a longer description, the whole form, questions and answers are changed permanently on Db. This leads not only to show up these many saved-notifications but also to a quite large database-traffic?!
- These savings always produce new copies of the questions inside the database. Old Questions are deleted, so storage-consumption should remain, but the question-id's inside Db explode quickly.
- Each change of one question produces newly saved copies of all answers, too. But old answers are not deleted in this case, they just remain with an question-id corresponding to a deleted question.
- When just selecting a different form in navigation, the questions are newly stored, but the answers are not. The new Questions remain without answers.

Corresponding to Commit "Fix gitignore"
- I think the files in js-Folder were intended to be within the repo. survey.jquery.min.js is the file containing the surveyjs/Form-Functionality and vote.js was used as an interface between surveyjs and the forms-app. We could move to the survey-vue or survey-jquery Package on npm somewhen, but that could be one more large project...
- The two more files within js folder i am not sure what they did or if they were still used on the project... (delete should be controlled by Page/Api-Controller, so app.js could be deprecated)

---
**So for now**:
- What about introducing a save-button again, making people only able to create and save the form once and we will have a look into editing the form and correspondingly updating Db in a next step?
- For now, the results (and a Share-Link?) could be shown, when selecting a saved form?

This would enable to merge this new design and adding some more functionality step by step...?!

Greets,
Jonas

skjnldsv:
> * I think the files in js-Folder were intended to be within the repo. survey.jquery.min.js is the file containing the surveyjs/Form-Functionality and vote.js was used as an interface between surveyjs and the forms-app. We could move to the survey-vue or survey-jquery Package on npm somewhen, but that could be one more large project...
> * The two more files within js folder i am not sure what they did or if they were still used on the project... (delete should be controlled by Page/Api-Controller, so app.js could be deprecated)

Yep, we realised there was some uncompiled files in the js folder :)
I'll restore them.



> * Each change of the form produces the database to be changed. Even selecting a different form, or while typing a longer description, the whole form, questions and answers are changed permanently on Db. This leads not only to show up these many saved-notifications but also to a quite large database-traffic?!

Good for now.
This is like any other change in contacts, calendar, mail. It will trigger database writing. Let's leave it like that and optimise later on. It's not that big data anyway :)

> * These savings always produce new copies of the questions inside the database. Old Questions are deleted, so storage-consumption should remain, but the question-id's inside Db explode quickly.

This needs to be fixed though, @rullzer ?

> * When just selecting a different form in navigation, the questions are newly stored, but the answers are not. The new Questions remain without answers.

I'm not following here, you mean displayed? The questions are displayed but not the new answers? I did not touched the answer view, maybe we need a timer and fetch new answers if that is what you mean?


> * How can i respond/submit to a form? Or how to share a link to a form? I didn't find a way... :(


I'll add a copy link buttonon the sidebar and navigation menu :)

jotoeri:
> I'm not following here, you mean displayed?

I meant stored in Db. This is connected to the updating thing. Questions are updated, resulting in new Question-Ids, but answers are not updated. So the answers still contain the old Question-Id.
-> In fact, for already existing Questions & Answers, the questions should be updated instead of inserted into Db.

I can have a look into tomorrow, if you want. Just today i'm busy.

skjnldsv:
Voting restored

skjnldsv:
> * I think the files in js-Folder were intended to be within the repo. survey.jquery.min.js is the file containing the surveyjs/Form-Functionality and vote.js was used as an interface between surveyjs and the forms-app. We could move to the survey-vue or survey-jquery Package on npm somewhen, but that could be one more large project...

Let's try to move voting to vue as well :)
That way we can have all files in `js/` to be compiled files. And we can drop the old jQuery. Would you be up for this?
Which part do you want to focus on? :smiley:

jotoeri:
Puhhh, sorry to tell that directly, please don't feel personally attacked, for me it's just a matter of a working-principle:
Is it good to merge changes, if this results in a in-fact non-working master branch?
Just clicking a bit through the app, my NC Error-log gets longer and longer, and as i said - the app is in this state not really usable on editing forms, as already inserted answers do not show up on a already saved form.

Shouldn't the master-branch be kept clean on a usable and working app? So changes to be tested and corrected before merging? :/

skjnldsv:
> Shouldn't the master-branch be kept clean on a usable and working app? So changes to be tested and corrected before merging? :/

Fine line is how we ease the dev here.
Sure we could have gone for a new branch, doesn't matter too much here for me, master is broken anyway (see #229)

If we create one pr for all the changes this will become impossible to review pretty fast. (this pr is already 1450/1000 lines changed, let's ease up a bit on ourselves)


> Just clicking a bit through the app, my NC Error-log gets longer and longer, and as i said - the app is in this state not really usable on editing forms, as already inserted answers do not show up on a already saved form.

Ah right, I missed this one, didn't realized it was that much of an issue and couldn't reproduce. Anyway, let's fix it today!

jotoeri:
I think #229 is just an issue of the non-updated js-files in browser. The app itself is not broken. The title is a bit misleading...

Looking at this answers-issue it seems like a complete change already from loading questions from db, showing it in transitions-groups, back to storing it would be useful...

skjnldsv:
Let's not argue and be productive, fix incoming :)

jotoeri:
So you're on to correct it?

skjnldsv:
> Ah right, I missed this one, didn't realized it was that much of an issue and couldn't reproduce. Anyway, let's fix it today!

Yes, ui fix is done, backend patch is worked on

PR #252, opened 2020-03-27
Fix questions display

skjnldsv:
- [ ] Questions gets duplicated in the db

### DIFF
--- a/lib/Db/AnswerMapper.php
+++ b/lib/Db/AnswerMapper.php
@@ -95,4 +95,27 @@ class AnswerMapper extends QBMapper {
 
 	   $qb->execute();
 	}
+
+	public function findById(int $answerId): Answer {
+		$qb = $this->db->getQueryBuilder();
+
+		$qb->select('*')
+			->from($this->getTableName())
+			->where(
+				$qb->expr()->eq('id', $qb->createNamedParameter($answerId))
+			);
+
+		return $this->findEntity($qb);
+	}
+
+	public function deleteByQuestion(int $questionId): void {
+		$qb = $this->db->getQueryBuilder();
+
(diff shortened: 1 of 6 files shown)


jotoeri:
@skjnldsv How far is your plan grown? Should i have a look into, too somehow?

skjnldsv:
> @skjnldsv How far is your plan grown? Should i have a look into, too somehow?

You can if you want :)
I was waiting for @rullzer to do/finish th backend part, he was caught up in a customer call I think :thinking:

rullzer:
Yes feel free to tackle this. I'm rather busy today

jotoeri:
Ok... 
Would an Ansatz like you wrote the API->NewForm be reasonable to continue in NewQuestion & NewAnswer? (So one on each table)
Would result further in UpdateEvent, RemoveQuestion & Remove Answer, which alltogether replace the current writeForm?

jotoeri:
Adding questions and answers works, but deleting does not. Does one of you see my problem?
In Create.vue->deleteQuestion the variables are ok, but between the axios.post-call and ApiController->deleteQuestion somehow the id gets lost and i just receive an id of 0?!

lib/Db/AnswerMapper.php
```diff
@@ -95,4 +95,16 @@ public function deleteByForm(int $formId): void {
 
 	   $qb->execute();
 	}
+
+	public function findById(int $answerId): Answer {
```
jotoeri: Isn't this the same as just find($id)?

jotoeri: Sorry, my Error. I'm sometimes confused, which function is Base-Class and which one is Specialisation.

PR #249, opened 2020-03-23
New Database-Structure

jotoeri:
As already stated in #206 i would like to change the database-structure to a more relational and properly named database.
The new tables are shown in [TargetTables.md](https://github.com/jotoeri/forms/blob/alter_Db/temp/TEMP_AlterDb_TargetTables.md).
Most of the changes are then to adapt the namings.

Will now be interesting to rebase to all changes, made in #237 and what came in inbetween.

Still to solve:
- How to Migrate old data from old Structure to new Structure. @rullzer Any hint to this from your side?

Issue #92, opened 2019-10-04
Increase length for description and show proper errors


Sohalt:
**Which Forms version are you running?**
1.0.3

**Nextcloud or ownCloud?:**
Nextcloud

**Nextcloud/ownCloud version:** (see Nextcloud admin page)
16.0.4

### Steps to reproduce
1. Try to submit a form with a description with more than 1024 characters

### Actual behaviour
Form does not get saved and shows a generic "Error on saving form, see console".
Request gives a 500 Internal Server Error.

### Expected behaviour
Form does get saved (or at least an error message saying that descriptions with more than 1024 characters are not supported).

skjnldsv:
@jotoeri done, right?

jotoeri:
Hm, i set on DB maxlengths to:

'title' => 256,
'description' => 2048,
'questiontext' => 2048,
'optiontext' => 1024,
'answertext' => 2048,

I think the input-fields on the vue-components are not adapted yet. Just a maxlength everywhere should do that, as it is not possible to produce the error on db then...?

skjnldsv:
> should do that, as it is not possible to produce the error on db then...?

It is, but not properly.
We should make this available as initialstate and use them int he app
https://github.com/nextcloud/forms/blob/e2f503e189b0e588668f42e5389fd3ff02724a5f/src/components/Questions/QuestionLong.vue#L38
https://github.com/nextcloud/forms/blob/e2f503e189b0e588668f42e5389fd3ff02724a5f/src/components/Questions/QuestionShort.vue#L38

jotoeri:
Hm, don't know, what you mean with initialstate... :/

skjnldsv:
https://github.com/nextcloud/forms/blob/38f4f1aa03440655fd7a4df02db9953aaeb10572/lib/Controller/PageController.php#L184
https://github.com/nextcloud/forms/blob/38f4f1aa03440655fd7a4df02db9953aaeb10572/src/views/Submit.vue#L69
https://github.com/nextcloud/forms/blob/38f4f1aa03440655fd7a4df02db9953aaeb10572/src/views/Submit.vue#L99

Our api to provide static initial data from php to any js application :)

PR #256, opened 2020-03-28
Enh/design fixes

skjnldsv:
dev skjnldsv com_apps_forms_j9qYSJXXjnbESw63_edit

- [x] Top right toolbar
	- [x] Show sidebar
	- [x] Show results
- [x] Main content
	- [x] Narrow page
	- [x] Title & description header
	- [x] New question toolbar
	- [x] Emptycontent


jotoeri:
So, i did a rebase onto current version of alterDb. Maybe i/we will have to adapt, if you request changes on Db-PR, but you can have a look into already.
Three things i did erroneously already while resolving merge-conflicts instead of afterwards, so they got merged into your commits. I'm sorry for that, just revert these changes if you don't want them:
- Included a maxlength to the input-fields of title & description, corresponding to the max Db-Table lengths. Should solve #92, well not by showing errors, but by just blocking longer texts.
- Removed the variables nextOptionId & nextQuestionId, as they were not used anymore since merging #252.
- Making the tests checkAllHaveOpt/checkQuestionText/checkOptionText return-functions, instead of storing the value into 'global' variable. I thought you would have done this somewhere already, but seems like my mind was wrong on this... :/

To not just override your commits here and not before alterDb is merged, i first pushed to a [branch on my fork](https://github.com/jotoeri/forms/tree/enh/design-fixes). Let me know, if you want me to push them here after merging alterDb.

And by that way...: That add-questions menu is awesome! :heart_eyes: 
Unbenannt

skjnldsv:
Thanks a lot @jotoeri !!!
So, currently this pr did not touched the methods, or at least, barely.
I still have to implement the methods for the new question button/menu.

I really did not looked nor do I really get the whole vision of the app yet.
So to answer your questions about `checkAllHaveOpt`, `checkQuestionText` & `checkOptionText`; I have no idea ^^'


> And by that way...: That add-questions menu is awesome!

Thanks :hugs:

skjnldsv:
> To not just override your commits here and not before alterDb is merged, i first pushed to a [branch on my fork](https://github.com/jotoeri/forms/tree/enh/design-fixes) [](/jotoeri/forms/tree/HEAD@{2020-04-01T15:02:51Z}/design-fixes). Let me know, if you want me to push them here after merging alterDb.

Yes, let's be careful here. Let's merge the alterDB pr, then we can focus on the other tasks. I'll manage the rebase of this pr after the others are merged :)

skjnldsv:
Rebased on latest master after #249

skjnldsv:
Multiple
Peek 07-04-2020 18-03

jotoeri:
So, how far is this one grown for now? Feels a bit like quite much depends on this one to be finished...?

skjnldsv:
> So, how far is this one grown for now? Feels a bit like quite much depends on this one to be finished...?

No, I guess this one should be the latest one to be merged or something, as the others are cleanup,s I will adapt this pr to the new data every time I can, but yes, we should get this one as soon as we can sure :)

jotoeri:
Hm, ok. API i did now, see PR. The form-object can be done, yes, but for Page-Controller i don't really know where to continue....

src/views/Create.vue
```diff
@@ -394,7 +395,7 @@ export default {
 </script>
 
 <style lang="scss">
-.app-content {
+#app-content {
```
skjnldsv, line 398: I'm using latest vue components master, here this is changed to a class.
This was on purpose :)

src/views/Create.vue
```diff
@@ -239,7 +240,7 @@ export default {
 		async saveFormProperty(key) {
 			try {
 				// TODO: add loading status feedback ?
-				await axios.post(OC.generateUrl('/apps/forms/api/v1/question/update'), {
+				await axios.post(generateUrl('/apps/forms/api/v1/form/update'), {
```
skjnldsv, line 243: ... I have some excuses, I was very tired xD :see_no_evil:

src/views/Create.vue
```diff
@@ -475,94 +476,94 @@ export default {
...
+#shiftDates {
+	min-width: 16px;
+	min-height: 16px;
+	margin: 0;
+	padding: 10px;
+	padding-left: 34px;
+	text-align: left;
+	background-repeat: no-repeat;
+	background-position: 10px center;
+} */
```
skjnldsv, line 567: Lol, this prevent any good code editor to do quick comments toggle :wink:

skjnldsv:
Thanks for the fix @jotoeri  :)
Avoid running fixes for syntax and other stuff though, if I did not cleaned it, it's not by mistake ;)

src/views/Create.vue
jotoeri: :D 👍

src/views/Create.vue
jotoeri: :D ok. I just used the package so it looked quite nasty... ;)

src/views/Create.vue
jotoeri: Depends on good code editor i think...? :D VS-Code did this by block-comment-toggle. And with just the line-comments i got 20/30 Build-Errors due to indentation...

jotoeri:
> Avoid running fixes for syntax and other stuff though, if I did not cleaned it, it's not by mistake ;)
> [5bd4a5b](https://github.com/nextcloud/forms/commit/5bd4a5b053aa67ab6b5eb136b2608a786eb21563)

Don't worry to omit the commit, i don't do now by myself, for not to force-push... I was just not able to build at all without at least the comment & showSuccess changes...
Same applies to the last commit to Save the Form-Properties. I don't know if this is a proper way or the way you want to go to save changed properties from Sidebar. I just tried it, so if it's not your way, don't worry to just throw it away....

skjnldsv:
> Don't worry to omit the commit, i don't do now by myself, for not to force-push... I was just not able to build at all without at least the comment & showSuccess changes...

Are you building production?
Building dev doesn't care ;)

skjnldsv:
> Same applies to the last commit to Save the Form-Properties. I don't know if this is a proper way or the way you want to go to save changed properties from Sidebar. I just tried it, so if it's not your way, don't worry to just throw it away....

No no, what you did looks good! Nice call to move it to the mixin :)

jotoeri:
> Are you building production?
> Building dev doesn't care ;)

Hm, no, i normally use `npm run watch`, which returns the list of errors just like `npm run build`. But yes, i just tested it and it seems like the build itself still works, even that i see the errors. Once again i learned sth. 🤔 😉

appinfo/routes.php
```diff
@@ -27,25 +27,32 @@
...
+
+		// Questions
+		['name' => 'api#newQuestion', 'url' => '/api/v1/question', 'verb' => 'POST'],
+		['name' => 'api#updateQuestion', 'url' => '/api/v1/question/update', 'verb' => 'POST'],
+		['name' => 'api#reorderQuestions', 'url' => '/api/v1/question/reorder', 'verb' => 'POST'],
+		['name' => 'api#deleteQuestion', 'url' => '/api/v1/question/{id}', 'verb' => 'DELETE'],
+
+		// Answers
+		['name' => 'api#newOption', 'url' => '/api/v1/option', 'verb' => 'POST'],
+		['name' => 'api#updateOption', 'url' => '/api/v1/option/update', 'verb' => 'POST'],
```
skjnldsv, line 52: I added this @jotoeri

src/components/Questions/AnswerInput.vue
```diff
@@ -0,0 +1,161 @@
...
+
+			this.$emit('delete', this.answer.id, this.index)
+		},
+
+		/**
+		 * Create an unsynced answer to the server
+		 *
+		 * @param {Object} answer the answer to sync
+		 */
+		createAnswer: pDebounce(async function(answer) {
```
skjnldsv, line 119: Not properly working, I'll have to add this to a pQueue to make sure the creation always gets triggered first.
Then it should be fine!

src/components/Questions/AnswerInput.vue
```diff
@@ -0,0 +1,161 @@
...
+			required: true,
+		},
+	},
+
+	methods: {
+		/**
+		 * Focus the input
+		 */
+		focus() {
+			this.$refs.input.focus()
```
skjnldsv, line 59: Auto focus seems a bit broken after moving this into the AnswerInput component, will fix later.
Other than this it includes the following UX:
- Adding a new answer focus the input automatically
- Deleting an answer automatically select the previous answer

src/components/Questions/AnswerInput.vue
jotoeri: Hm, i already thought about maybe moving the options to json, too. So it would be just an array of user-text-input and could be created/updated/delted by updateQuestion. I think if only doing this, when leaving the options-edit mode, the traffic & writing shouldn't be too much...?!

skjnldsv: Nah, I debounce the requests, so we only save when the user is done typing (+500ms delay).
So it will not matter what implementation you choose on the db, it will be the same :)

It's fine like that, don't bother!

skjnldsv: Also, we're talking about one request with a text as param.
There won't be 1k users creating answers simultaneously on the same instance. This is just for the edit mode, we'll be super fine, you did a great job with the cleanup and simplification!!

If you find anything that I did poorly in the php api, feel free to help, I'm out for the day ;)

src/views/Create.vue
```diff
@@ -58,16 +58,17 @@
 				:required="true"
 				autofocus
 				type="text"
-				@change="onTitleChange"
```
jotoeri: Hm, to be honest i really liked the @change, as it only saved when leaving the component, but not on nearly each single keyup...

skjnldsv: This is debounced, it doesn't matter :)

jotoeri: I know it's debounced... But from typing just one sentence within the description (where users usually input a longer text, than just one sentence), i get a quite long list of saving intermediate states...^^
Indeed, it works, but the other one was more beautiful in my opinion. 😉 
grafik

skjnldsv: It will be the same with change, I don't see your point?
I'm fine with change to be honest, no Idea why I put keyUp here  :p
But it will behave exactly the same in terms of request count

skjnldsv: Ah right, not with change. I forgot it was using the focus/unfocus to trigger.
I think I prefer having the whole data saved even when the user keeps focus.

skjnldsv: Nonetheless, We can set the debounce to 500, it doesn't really matter too much here ;)

src/components/Questions/AnswerInput.vue
jotoeri: No, looks good. Just two copy-paste typos i found... ;)

src/views/Create.vue
jotoeri: Hm... :man_shrugging:

jotoeri:
Rebased & squashed the commits... Backup of unsquashed ones is still on a backup-branch, there is no diff except the three changes on master.

jancborchardt:
There’s a lot of conflicts towards https://github.com/nextcloud/forms/pull/297 now, I assume because of the squashing? Is the squashing necessary in this case, can we just go with a rebase?

skjnldsv:
Okay, please test again both of you.
I think we got everything clear here. Despite the various fixes in #297 of course.

skjnldsv:
> * When you create a multiple choice or checkbox question, type in a question, and then use Tab twice to go to the first answer and type one in, there neither appears a second answer option nor does it get saved. When you click out of it and in again, the answer text disappears

It only works on click, right?

jotoeri:
> It only works on click, right?

Yes, but only for the first time. Once you are in edit mode, the tab works...

skjnldsv:
> Yes, but only for the first time. Once you are in edit mode, the tab works...

Yes, then it's the issue I had in mind.
Should be an easy fix, but i's not a blocker for this one :)

src/utils/CancelableRequest.js
```diff
@@ -0,0 +1,57 @@
...
+
+import axios from '@nextcloud/axios'
+
+/**
+ * Creates a cancelable axios 'request object'.
+ *
+ * @param {function} request the axios promise request
+ * @returns {Object}
+ */
+const CancelableRequest = function(request) {
```
jotoeri, line 32: Not used yet i think but planned for somewhere?

skjnldsv: Yes, when switching from one form to the other, cancel any ongoing form loading, forgot to remove this

PR #260, opened 2020-03-29
Fix/only own forms

rullzer:


jotoeri:
I'll rebase this one...

src/Forms.vue
```diff
@@ -120,9 +120,11 @@ export default {
 		 * Initial forms load
 		 */
 		async loadForms() {
+			console.debug('Loading forms')
```
skjnldsv, line 123: Hello there !

skjnldsv:
Feel free to review now @jotoeri

lib/Controller/ApiController.php
```diff
@@ -109,44 +90,6 @@ public function __construct(
...
 	}
 
-	/**
-	 * Transforms a string with user and group names to an array
-	 * of nextcloud users and groups
-	 * @param string $item
-	 * @return Array
-	 */
-	private function convertAccessList($item) : array {
-		$split = [];
```
jotoeri: I think we still need this one, as it converts the Shared-List from string to array. We should change this, but this will be a different thing than display only own forms...

lib/Controller/ApiController.php
```diff
@@ -339,31 +282,21 @@ public function getFullForm($formIdOrHash) {
...
-			$form = $this->getFullForm($formElement->id);
-			//if ($form['grantedAs'] !== 'none') {
-				$formsList[] = $form;
-			//}
+		$result = [];
+		foreach ($forms as $form) {
+			$result[] = [
+				'id' => $form->getId(),
+				'form' => $form->read(),
+				'questions' => $this->getQuestions($form->getId())
```
jotoeri, line 339: Added Shares again...

jotoeri:
@skjnldsv If you are good with this, i think we could merge it...

lib/Controller/ApiController.php
skjnldsv: Was there an issue?
Also we don't need this 'mode' entry I think :)

jotoeri: Yes, shares were not loaded, and then overwritten by the first writeForm.
The mode i would have removed when rewriting the writeForm to an updateForm. Until now, it is still used in writeForm.

PR #264, opened 2020-04-01
Fix/expirations

jotoeri:
This PR is a draft, as it is already based on the AlterDB PR.

- [x] The expiration Date was not shown when loading a form. -> ExpirationDatePicker didn't like the Variable-Format, so didn't interpret the variable
- [x] When disabling an expiration, this was not stored to Db. -> Changed null in Db to date(null), which is 1970-01-01
- [x] The expired-Variable is not updated, when changing the Expiration-Date. -> Update on ExpirationChange.

@skjnldsv What i need some help on the Design/Frontend side, maybe you understand why & have an easy idea for this:
- [x] The expired-Variable is now updated, when changing the Date. Still, the Bullet in AppNavigation is not updated. Is this just due to the formatForm-Script and will disappear when resolving the flattened structure? => Edit: Seems like my tests went wrong. Suddenly this works for me.
- [ ] Do you know a good way for not to show the 1970-Date in ExpirationDatePicker? When expiraitonDate corresponds to the Null-Date, better the placeholder would be shown.

### DIFF
--- a/src/views/Sidebar.vue
+++ b/src/views/Sidebar.vue
@@ -58,7 +58,6 @@
 				v-model="form.form.expirationDate"
 				v-bind="expirationDatePicker"
 
-				:time-picker-options="{ start: '00:00', step: '00:05', end: '23:55' }"
 				style="width:170px" />
 		</div>
 
@@ -140,13 +139,15 @@ export default {
 				editable: true,
 				minuteStep: 1,
 				type: 'datetime',
+				valueType: 'YYYY-MM-DD HH:mm:ss',
+				defaultValue: moment(),
 				format: moment.localeData().longDateFormat('L') + ' ' + moment.localeData().longDateFormat('LT'),
-				lang: this.lang.split('-')[0],
+				lang: moment.localeData()['_config'],
 				placeholder: t('forms', 'Expiration date'),
 				timePickerOptions: {
 					start: '00:00',
-					step: '00:30',
-					end: '23:30',
+					step: '00:15',
+					end: '23:45',
 				},
 			}
 		},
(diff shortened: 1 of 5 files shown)


src/views/Sidebar.vue
```diff
@@ -140,13 +139,15 @@ export default {
 				editable: true,
 				minuteStep: 1,
 				type: 'datetime',
+				valueType: 'YYYY-MM-DD HH:mm:ss',
+				defaultValue: moment(),
 				format: moment.localeData().longDateFormat('L') + ' ' + moment.localeData().longDateFormat('LT'),
-				lang: this.lang.split('-')[0],
+				lang: moment.localeData()['_config'],
 				placeholder: t('forms', 'Expiration date'),
 				timePickerOptions: {
```
skjnldsv, line 147: We'll manage this directly with the DateTimepicker from the components later I guess.
We have a working example of it in calendar or in the server sharing sidebar to inspire us on :)

jotoeri: This comment just for information? Or better remove this line within this PR? Or leave as is?

skjnldsv: Well, don't bother, we're probably going to have to update the sidebar design as well so it's cleaner, so this code might :boom: anyway. We'll see if time allows :thinking:

jotoeri:
How about this one? You didn't want to merge because of the sidebar-thing?

skjnldsv:
I'm fine to get this in, 

> You didn't want to merge because of the sidebar-thing?

Yeah, but also because I wanted to check why we were using date and not just handling timestamps ?

skjnldsv:
> We can move to timstamps, should even be easier to handle, no?

Yes, I think too :)

jotoeri:
Well, i just looked into. This is too much of mixed things now. I'll do the timestamp thing, the access thing and then i can come back to fix this...

PR #265, opened 2020-04-03
Enh/mandatory-questions

jotoeri:
This PR is a draft, as it is already based on the alterDB-PR.

This includes the functionality to set Questions mandatory or not.
The UI on creating/editing forms was just roughly implemented due to the ongoing design-changes on #256.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -531,6 +531,38 @@ class ApiController extends Controller {
 
 	/**
 	 * @NoAdminRequired
+	 * @param int $id QuestionId of question to update
+	 * @param array $keyvalues Array of key=>value pairs to update.
+	 */
+	public function updateQuestion(int $id, array $keyvalues): Http\JSONResponse {
+		$this->logger->debug('Updating question: questionId: {id}, values: {keyvalues}', [
+			'id' => $id,
+			'keyvalues' => $keyvalues
+		]);
+
+		try {
+			$question = $this->questionMapper->findById($id);
+			$form = $this->formMapper->find($question->getFormId());
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find question or form');
+			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+		}
+
+		if ($form->getOwnerId() !== $this->userId) {
+			$this->logger->debug('This form is not owned by the current user');
+			return new Http\JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		$question = Question::fromParams($keyvalues);
+		$question->setId($id);
+
+		$this->questionMapper->update($question);
+
+		return new Http\JSONResponse($question->getId());
+	}
+
+	/**
+	 * @NoAdminRequired
 	 */
 	public function deleteQuestion(int $id): Http\JSONResponse {
 		$this->logger->debug('Delete question: {id}', [
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -46,6 +46,7 @@ return [
 
 		['name' => 'api#newForm', 'url' => 'api/v1/form', 'verb' => 'POST'],
 		['name' => 'api#deleteForm', 'url' => 'api/v1/form/{id}', 'verb' => 'DELETE'],
+		['name' => 'api#updateQuestion', 'url' => 'api/v1/question/update/', 'verb' => 'POST'],
 		['name' => 'api#newQuestion', 'url' => 'api/v1/question/', 'verb' => 'POST'],
 		['name' => 'api#deleteQuestion', 'url' => 'api/v1/question/{id}', 'verb' => 'DELETE'],
 		['name' => 'api#newOption', 'url' => 'api/v1/option/', 'verb' => 'POST'],
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -277,20 +277,23 @@ class PageController extends Controller {
 
 		//Insert Answers
 		foreach($questions as $question) {
-			if($question['type'] === "checkbox"){
-				foreach(($answers[$question['text']]) as $ansText) {
+			// If question is answered, the questionText exists as key in $answers. Does not exist, when a (non-mandatory) question was not answered.
+			if (array_key_exists($question['text'], $answers)) {
+				if($question['type'] === "checkbox"){
+					foreach(($answers[$question['text']]) as $ansText) {
+						$answer = new Answer();
+						$answer->setSubmissionId($submissionId);
+						$answer->setQuestionId($question['id']);
+						$answer->setText($ansText);
+						$this->answerMapper->insert($answer);
+					}
+				} else {
 					$answer = new Answer();
 					$answer->setSubmissionId($submissionId);
 					$answer->setQuestionId($question['id']);
-					$answer->setText($ansText);
+					$answer->setText($answers[$question['text']]);
 					$this->answerMapper->insert($answer);
 				}
-			} else {
-				$answer = new Answer();
-				$answer->setSubmissionId($submissionId);
-				$answer->setQuestionId($question['id']);
-				$answer->setText($answers[$question['text']]);
-				$this->answerMapper->insert($answer);
 			}
 		}
 


lib/Controller/ApiController.php
```diff
@@ -529,6 +529,38 @@ public function newQuestion(int $formId, string $type, string $text): Http\JSONR
 		return new Http\JSONResponse($question->getId());
 	}
 
+	/**
+	 * @NoAdminRequired
```
skjnldsv, line 533: Better comments on what this function is about please :)

src/components/quizFormItem.vue
```diff
@@ -21,7 +21,15 @@
   -->
 <template>
 	<li>
-		<div>{{ question.text }}</div>
+		<div>
+			<input :id="'questionId_' + question.id + '_mandatory'"
```
skjnldsv, line 25: No change required in the front for this pr if possible please.
I'll manage it with the #256 as there is a new ui.
Please just add the backend functionalities :)

lib/Controller/ApiController.php
```diff
@@ -529,6 +529,38 @@ public function newQuestion(int $formId, string $type, string $text): Http\JSONR
...
+	 * @param int $id QuestionId of question to update
+	 * @param array $keyvalues Array of key=>value pairs to update.
+	 */
+	public function updateQuestion(int $id, array $keyvalues): Http\JSONResponse {
```
skjnldsv, line 537: What is allowed in those? We could change anything? Title too?

src/components/quizFormItem.vue
skjnldsv: I removed it ;)

skjnldsv:
@jotoeri 
> So, I'm wondering, how bad is it to keep settings of something as a json in the db?
> I mean for the forms_v2_questions, we have one entry per config
> e.g 'mandatory' to define if a question is, well, mandatory or not
> But it means we'll have to change the db layout to add a new column every time we want to add a new config
> I would (maybe stupidly) just have a config longtext in here so we can add whatever we want?

lib/Controller/ApiController.php
jotoeri: Yes, this was my goal, to have one procedure in API, enabling to change single properties on questions, that can be used for ordering/mandatory/title. As for now, you can change everything except the id.

skjnldsv: I'm on it :)

lib/Controller/ApiController.php
jotoeri: Better like this...? ;)

skjnldsv:
Don't touch this pr please, i'm working on it :)

jotoeri:
> So, I'm wondering, how bad is it to keep settings of something as a json in the db?

Well, from my understanding of Databases or my point of view i would interpret it as non-beautiful Db-Style. But to seriously answer this on a knowledge of current Paradigms, i'm maybe not experienced enough. Maybe better aks @rullzer on this...?

skjnldsv:
> Maybe better aks @rullzer on this...?

Already done, got green light ;)
Just wait for my commits :)

skjnldsv:
All good! @jotoeri please have a look :)

lib/Controller/ApiController.php
```diff
@@ -529,6 +529,62 @@ public function newQuestion(int $formId, string $type, string $text): Http\JSONR
...
+	 * @param array|string $value 
+	 */
+	public function updateQuestion(int $id, string $key, $value): Http\JSONResponse {
+		$this->logger->debug('Updating {key} of question {id} with value: {value}', [
+			'id' => $id,
+			'key' => $key,
+			'value' => $value
+		]);
+
+		if ($key !== 'title' && $key !== 'attributes') {
```
skjnldsv, line 548: Later we can add the option to switch types.
Currently on #256 I'm using a single array to store answers. For single questions We'll only use the index 0 and multiple questions we'll use the full array.

That should allow us to easily switch from one entry to the other :)

jotoeri:
Well, as i said, honestly it's not the way i like with the json-string.
Having single columns is much more clear in my opinion (database as well as this update-procedure) and you don't need to always store, read&rewrite all the attributes, if just changing a single property.
The advantage of needing one update-script less to add more data/a column is not that big to add this complexity in my opinion.

skjnldsv:
> The advantage of needing one update-script less to add more data/a column is not that big to add this complexity in my opinion.

Well, we usually want to avoid touching the database as much as possible.
Migrations are really expensive for the system. So I would rather avoid adding multiple steps over the releases.

Nonetheless, we can also take a second approach if you want. The same we do in server, we can have a forms_v2_questions_attributes table and insert paris of id/key/value. Like we do in the preferences or appconfig tables.

Keep in mind though that we're not stupidly outsmarting mysql, storring JSON is entirely supported by the database as a dedicated type. This is not as dirty as it looks :)

jotoeri:
Is a single Migration when updating an app that expensive, so it is better to read&store double/tree/four times the data on each single activity? :worried: Changing the database does not happen too often i think...?

>Keep in mind though that we're not stupidly outsmarting mysql, storring JSON is entirely supported by the database as a dedicated type.
Have this in mind, but doen't change my opinion. ;)

skjnldsv:
> Is a single Migration when updating an app that expensive, so it is better to read&store double/tree/four times the data on each single activity?  Changing the database does not happen too often i think...?
> 
> >

It's not about the size of it as far as I understood but more that avoiding touching the db as much as possible is a good thing to have.
I guess if you have multiple replicas or multiple instances, it can maybe lead to issues? I'm just repeating what I remember reading. :)
@rullzer, @ChristophWurst an opinion? I'm really not the best to ask here! :see_no_evil:

js/submit.js
```diff
@@ -49,7 +49,7 @@ $(document).ready(function () {
 		q.options.forEach(o => {
 			qChoices.push(o.text);
 		});
-		surveyJSON.questions.push({type: q.type, name: q.text, choices: qChoices, isRequired: q.mandatory});
+		surveyJSON.questions.push({type: q.type, name: q.text, choices: qChoices, isRequired: 'true'});
```
jotoeri: Why don't you replace with sth., that keeps functionality of this PR?!

skjnldsv: Because I'd like to implement this entirely.
The goal is to nuke the submit.js file to replace with proper vue layout.

lib/Controller/ApiController.php
skjnldsv: @jotoeri 
Well, afterthoughts, this is what I went for to try the new questions design. But I'm fine with using better results
Though something like this would be much cleaner, there is far too many informations.

```js
// Form
{
	id: 2,
	hash: 'sLKsLaByy2NeiEia',
	title: 'New form 2',
	description: 'Lorep ipsum',
	created: '2020-04-04 07:27:14',
	attributes: {
		anonymous: boolean,
		expires: Date,
		access: 'public|registered|hidden',
		onlyOnce: boolean
	}
	questions: [],
}
```
```js
// Question
{
	id: 1,
	type: 'short|long|multiple|checkbox',
	attributes: {
		required: boolean,
	},
	title: 'Do you agree ?',
	options: [
		{
			id: 1,
			text: 'Yes',
		},
		{
			id: 2,
			text: 'No',
		},
		{
			id: 3,
			text: 'Maybe',
		},
	],
},
```

jotoeri: > Though something like this would be much cleaner, there is far too many informations. 

Here i can't follow from translations, what you exactly mean with your sentence... Informtion won't get less, all information, that we store is somehow used, i think.... :/

- I can imagine quite well to use the JSON-Type e.g. within the access-column, to ease the access-management of selected users&groups. Or if we should ever implement the issue of translated forms to store different translations of the question-text into a 'text'-JSONColumn. So in fact to store variable user-information as JSON instead of a manually created string.
- But i don't see the advantage of packing known-type&-length information into a sub-structure within Db, on change to write on change instead of 1bit boolean a pack of [12byte of text (anonymous: 1) + x*XXbyte of all other attributes as text] and to manage the data manually, instead of letting the DB-system do this. I think this is job of a database-system to manage the data.
- And where ist the difference of a created- & an expiration-date? Why is one an attribute, but the other one is not? Why is a title not an attribute? We won't have to search for it... In consequence, all data instead of id & hash would be attributes...?! 

- Cleaning up the form-data-structure within Code makes sense, yes. We talked about the form.form thing already. And i think having a closer look, there will be even more things, that could be cleaned up within this form-structure. But i don't think it gets more clear if we introduce a new sub-structure there...

skjnldsv: Okay, let's do it your way then :)
My main concern is having a clean javascript object to work with.

So I would like lots of other unneeded properties to be removed and to move the forms/questions config (or attributes, option, whatever you want to call it) into its own dedicated object key, so it's much simpler.

Would you be up for that?
As long as the db is clean as well, I'm all in!
Are you available this week? :)

jotoeri: Hm, how to store into the js-object is just a kind of definition. I think i would have put it into the single properties directly, maybe some computed data into a specific sub-structure. But if you like more to put some of the properties into sub-structures inside the object as above, i'm fine with.
I didn't see a rule yet, on which ones are attributes and which ones are not, expiration & creation-dates still confuse me a bit, but i think i could still work with your form-structure as listed above.
Yes, i think i can be more available now again, than the last two days... ;)

skjnldsv: Well, about the forms objects, it's easy, as long as the data returned to me by the apicontroller is clean, then I'll manage.

Regarding your previous implementation of the updateQuestion, you said: `you don't need to always store, read&rewrite all the attributes, if just changing a single property`, But your implementation update all the rows every time, no?

jotoeri: No, it just writes the column you specify.
The Question-Entity gets created with the key-value pair, that was given, and then the id is set. So the Question-Entity just cotains id & the updated value. When giving this to the update-procedure, only the given columns are updated, the other ones remain. I think this was also the problem on resetting the expiration-date to null, as attributes beeing null are just not written by the update-procedure.

rullzer:
> It's not about the size of it as far as I understood but more that avoiding touching the db as much as possible is a good thing to have.
I guess if you have multiple replicas or multiple instances, it can maybe lead to issues? I'm just repeating what I remember reading. :)
@rullzer, @ChristophWurst an opinion? I'm really not the best to ask here! see_no_evil

Most of th time we try to avoid it for big tabes like the filecaache. As adding a column there might take ages on some DB setups. For forms it should be fine.

lib/Controller/ApiController.php
jotoeri: If you want we can change the interface to the way you called it with (id, key, value). The old way enabled to write multiple attributes at the same time, but if we don't need this, we can change...

skjnldsv: Isn't `$this->questionMapper->update($question);` updating the whole row?

jotoeri: I can have a look into the mapper definition, but the call with only the specific properties worked fine.

skjnldsv: Let me revert my two last commits then :)

jotoeri: So, yes, as far as i see, the Entity stores which fields got updated and the mapper then only inserts those into the query...

skjnldsv: Reverted!
Tested the update, only updates what is provided! :+1: 
```sql
db_1 | UPDATE `forms_v2_questions` SET `text` = 'Hello' WHERE `id` = 6
```

skjnldsv: Resolving this discussion thread then :grin:

skjnldsv:
> Most of th time we try to avoid it for big tabes like the filecaache. As adding a column there might take ages on some DB setups. For forms it should be fine.

@jotoeri let's go then! :)
Let's integrate our discussion about cleaning the json outputed in another pr!

Issue #262, opened 2020-03-31
Reorder & Delete Questions


jotoeri:
- Possibility to reorder Questions
- Delete Questions by marking them as deleted, but keep until Form is deleted
- Can be stored within one order-column into DB (order-index 0 as deleted)

jancborchardt:
Reordering and deleting questions now works :+1:

PR #270, opened 2020-04-07
Order Questions

jotoeri:
Ok, so next thing i started already a few days ago, was to include the possibility to reorder questions:
- I first used 'order' as key, until phpmyadmin told me, that order would be a sql-keyword. I'm not completely happy with the naming 'rank', if you know sth. better, just tell me.
- Rank can be updated with updateQuestion. Changing the Ranking of two questions needs to call update with rank on both questions separately.
- Delete Question only changes the Rank to 0, deleteForm really deletes all Questions & corresponding options.
- newQuestion takes the last highest Rank +1 and returns not only the question-id but also the rank of the new question.
- On simple questionMapper->findByForm, only the non-deleted questions are loaded, an optional switch makes it possible to load all Questions.

@skjnldsv:
- On frontend i didn't change anything, except the return on newQuestion, as this one changed in API. Like this, this PR doesn't destroy the functionality to add Questions. I hope this works for you?
- I now changed the 'old' new Migration-Script, as it has not been released yet, so users will only get one migration. For your installation, you maybe have to add the Column rank manually with notnull true and default 1. Then insert the values manually or just delete all forms and insert new ones if you want to test a ranking functionality... Are you ok with this?

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -495,19 +495,124 @@ class ApiController extends Controller {
 			return new Http\JSONResponse([], Http::STATUS_FORBIDDEN);
 		}
 
+		// Retrieve all active questions sorted by Order. Takes the order of the last array-element and adds one.
+		$questions = $this->questionMapper->findByForm($formId);
+		$lastQuestion = array_pop($questions);
+		if ($lastQuestion) {
+			$questionOrder = $lastQuestion->getOrder() + 1;
+		} else {
+			$questionOrder = 1;
+		}
+
 		$question = new Question();
 
 		$question->setFormId($formId);
+		$question->setOrder($questionOrder);
 		$question->setType($type);
 		$question->setText($text);
 
 		$question = $this->questionMapper->insert($question);
 
-		return new Http\JSONResponse($question->getId());
+		$response = [
+			'id' => $question->getId(),
+			'order' => $question->getOrder()
+		];
+
+		return new Http\JSONResponse($response);
 	}
 
 	/**
 	 * @NoAdminRequired
+	 * Updates the Order of all Questions of a Form.
+	 * @param int $formId Id of the form to reorder
+	 * @param int $newOrder Array of Question-Ids in new order.
+	 */
+	public function reorderQuestions(int $formId, array $newOrder): Http\JSONResponse {
+		$this->logger->debug('Reordering Questions on Form {formId} as Question-Ids {newOrder}', [
+			'formId' => $formId,
+			'newOrder' => $newOrder
+		]);
+
+		try {
+			$form = $this->formMapper->find($formId);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
+			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+		}
+
+		if ($form->getOwnerId() !== $this->userId) {
+			$this->logger->debug('This form is not owned by the current user');
+			return new Http\JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		// Check if array contains duplicates
+		if ( array_unique($newOrder) !== $newOrder ) {
+			$this->logger->debug('The given Array contains duplicates.');
+			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+		}
+
+		// Check if all questions are given in Array.
+		$questions = $this->questionMapper->findByForm($formId);
+		if ( sizeof($questions) !== sizeof($newOrder) ) {
+			$this->logger->debug('The length of the given array does not match the number of stored questions');
+			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+		}
+
+		$questions = []; // Clear Array of Entities
+		$response = []; // Array of ['questionId' => ['order' => newOrder]]
+
+		// Store array of Question-Entities and check the Questions FormId & old Order.
+		foreach($newOrder as $arrayKey => $questionId) {
+
+			try {
+				$questions[$arrayKey] = $this->questionMapper->findById($questionId);
+			} catch (IMapperException $e) {
+				$this->logger->debug('Could not find question. Id:{id}', [
+					'id' => $questionId
+				]);
+				return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+			}
+
+			// Abort if a question is not part of the Form.
+			if ($questions[$arrayKey]->getFormId() !== $formId) {
+				$this->logger->debug('This Question is not part of the given Form: questionId: {questionId}', [
+					'questionId' => $questionId
+				]);
+				return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+			}
+
+			// Abort if a question is already marked as deleted (order==0)
+			$oldOrder = $questions[$arrayKey]->getOrder();
+			if ( $oldOrder === 0) {
+				$this->logger->debug('This Question has already been marked as deleted: Id: {id}', [
+					'id' => $questions[$arrayKey]->getId()
+				]);
+				return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+			}
+
+			// Only set order, if it changed.
+			if ($oldOrder !== $arrayKey + 1) {
+				// Set Order. ArrayKey counts from zero, order counts from 1.
+				$questions[$arrayKey]->setOrder($arrayKey + 1);
+			}
+		}
+
+		// Write to Database
+		foreach($questions as $question) {
+			$this->questionMapper->update($question);
+
+			$response[$question->getId()] = [
+				'order' => $question->getOrder()
+			];
+		}
+
+		return new Http\JSONResponse($response);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * Writes the given key-value pairs into Database.
+	 * Key 'order' should only be changed by reorderQuestions() and is not allowed here.
 	 * @param int $id QuestionId of question to update
 	 * @param array $keyvalues Array of key=>value pairs to update.
 	 */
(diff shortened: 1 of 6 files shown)


skjnldsv:
> * On frontend i didn't change anything, except the return on newQuestion, as this one changed in API. Like this, this PR doesn't destroy the functionality to add Questions. I hope this works for you?

Yes, perfect! We're free to add as many as we want in the backend, no regressions as this is new features, so no loss here :)

src/views/Create.vue
```diff
@@ -211,10 +211,11 @@ export default {
 			} else {
 				if (this.newQuestion !== null & this.newQuestion !== '' & (/\S/.test(this.newQuestion))) {
 					const response = await axios.post(generateUrl('/apps/forms/api/v1/question/'), { formId: this.form.id, type: this.selected, text: this.newQuestion })
-					const questionId = response.data
+					const respData = response.data
 
 					this.form.questions.push({
-						id: questionId,
+						id: respData.questionId,
+						rank: respData.questionRank,
```
skjnldsv, line 218: I would prefer more straightforward keywords like `order` :)

jotoeri: Yes, me normally too. @rullzer Can you estimate where problems could occur with sql-keyword, when using order here?

skjnldsv: > I first used 'order' as key, until phpmyadmin told me, that order would be a sql-keyword. I'm not completely happy with the naming 'rank', if you know sth. better, just tell me.

Sorry, I should have read your first post better! :see_no_evil:

jotoeri:
> @skjnldsv skjnldsv linked an issue that may be closed by this pull request now
Reorder & Delete Questions #262 

:D this one i removed again, as the frontend is not done with this PR yet...? ;)

skjnldsv:
> * Rank can be updated with updateQuestion. Changing the Ranking of two questions needs to call update with rank on both questions separately.

Doesn't scale super well, but I guess that's o.k. :thinking: 
What about replacing the associated question to swap the two automatically? Is it a bad idea?

skjnldsv:
> * Delete Question only changes the Rank to 0, deleteForm really deletes all Questions & corresponding options.

Why? I feel like I've missed something :thinking:

jotoeri:
> Doesn't scale super well, but I guess that's o.k. 🤔
> What about replacing the associated question to swap the two automatically? Is it a bad idea?

Well we could implement a `changeQuestion(id1, id2)` on top, that calls the update for both?

jotoeri:
> Why? I feel like I've missed something 🤔

As to show the Answers that were submitted before changing the form, it would be good to still know the question-text.

skjnldsv:
> As to show the Answers that were submitted before changing the form, it would be good to still know the question-text.

Ah right, this is a nice catch!

> Well we could implement a `changeQuestion(id1, id2)` on top, that calls the update for both?

If the function is only changing the order, we still should use proper variables names :)
`swapQuestions(id1, id2)` ? `/api/v1/forms/question/swap` ?
Just so it's clear to understand when people will take over maintaining this app. This was complicated enough, I don't want to create more confusion :see_no_evil:

jotoeri:
:D 👍 
I will include this one tomorrow...

lib/Controller/ApiController.php
```diff
@@ -546,6 +546,61 @@ public function newQuestion(int $formId, string $type, string $text): Http\JSONR
...
+		$question1->setRank($rank2);
+		$question2->setRank($rank1);
+		$this->questionMapper->update($question1);
+		$this->questionMapper->update($question2);
+
+		// Response as Array of ['id' => ['rank'=>value]]
+		$response = [];
+		$response[$id1] = [
+			'rank' => $question1->getRank(),
+		];
```
jotoeri, line 593: This response is a bit peculiar, but i think it will be easier to read it like this?
`response.data[$id]['rank']`

src/views/Create.vue
jotoeri: [SQL-Keywords](https://www.w3schools.com/SQl/sql_ref_keywords.asp)
Seems like order is only a keyword in combination as `ORDER BY`...?

marcelklehr:
I think with the query builder you can usually name your columns as you like, as column names are escaped/delimited.

Regarding the form of the endpoint, if you implement it as swap and someone reorders question 2 to slot 7, you'd need 5 requests to the swap endpoint. I think it makes more sense to send a complete  new permutation to the server i.e. something like an array of question IDs: `[1,3,4,5,6,2,7]`.

skjnldsv:
> Regarding the form of the endpoint, if you implement it as swap and someone reorders question 2 to slot 7, you'd need 5 requests to the swap endpoint. I think it makes more sense to send a complete new permutation to the server i.e. something like an array of question IDs: `[1,3,4,5,6,2,7]`.

As long as the front only have to do one request I'm fine for this.
I'll leave the implementation up to @jotoeri (and you @marcelklehr if you want ;) )

jotoeri:
Well, depends on what the front will enable users to do, no? ;)
If you just use Buttons for up&down move of questions, the swap will suffice and will write less to Db, while if you enable users to shift questions around by mouse, even more than just one order, the permutation array is a good idea. - So what is in your mind for the frontend @skjnldsv ? ;)

skjnldsv:
> So what is in your mind for the frontend @skjnldsv ? ;)

Current implementation includes drag and drop :)

marcelklehr:
> and you @marcelklehr if you want ;) 

heh, I'm fine with weighing in with my two cents from time to time, for now, if that's ok. I've got my plate with bookmarks and floccus :D

jotoeri:
> heh, I'm fine with weighing in with my two cents from time to time, for now, if that's ok. I've got my plate with bookmarks and floccus :D

:D Were quite useful two cent, i think. ;) I'll give it a try later...

jotoeri:
Changed now to a `reorderQuestion($formId, $[arrayofallQuestionIds])`.
Didn't test yet, this i will do tomorrow.

jotoeri:
So. With all checks, the reorder became a quite lenghty thing, but i think now all erroneous order-arrays should be catched...^^
Reorder can be called with
`await axios.post(generateUrl('/apps/forms/api/v1/question/reorder/'), { formId: 1, newOrder: [1, 9, 8, 3] })`

lib/Controller/ApiController.php
jotoeri: @skjnldsv Are you ok with this response? (except it is called `['order']` now....)

lib/Controller/ApiController.php
```diff
@@ -518,19 +518,119 @@ public function newQuestion(int $formId, string $type, string $text): Http\JSONR
...
+				return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+			}
+
+			// Set Order. ArrayKey counts from zero, order counts from 1.
+			$questions[$arrayKey]->setOrder($arrayKey + 1);
+		}
+
+		// Write to Database
+		foreach($questions as $question) {
+			$this->questionMapper->update($question);
```
skjnldsv, line 621: So it will only update the one that change, super clever!
I really like this approach!

src/views/Create.vue
```diff
@@ -211,10 +211,11 @@ export default {
 			} else {
 				if (this.newQuestion !== null & this.newQuestion !== '' & (/\S/.test(this.newQuestion))) {
 					const response = await axios.post(generateUrl('/apps/forms/api/v1/question/'), { formId: this.form.id, type: this.selected, text: this.newQuestion })
-					const questionId = response.data
+					const respData = response.data
 
 					this.form.questions.push({
-						id: questionId,
+						id: respData.id,
+						order: respData.order,
```
skjnldsv, line 218: Is it really needed?

jotoeri: To change Create.vue?
It will be necessary afterwards to add questions. If you want to include it on your branch, i can remove here. As long as i won't have to test anything on the order property it should work...

lib/Controller/ApiController.php
jotoeri: Not completely... It still changes the order-columns on all questions of the respective form. But nice idea, i can still change to update only the changed order-columns, i read the old order anyways. :D 👍

src/views/Create.vue
jotoeri: But keep in mind, that if you rebase your branch, adding questions won't work until you include this part, as the response from API changed...

jotoeri: Ahh shit, it won't even work on master due to the changed API... So yes, i think this change would be good to have a usable master on which we can continue development.

skjnldsv: You need to provide an order to create a question?
This should most likely be an optional parameter :thinking:

jotoeri: The newQuestion() looks up on Db, which order would be the next one an returns this order together with the questionId to the frontend. So the response.data changed from just an integer to an array of two key-integer-pairs.

skjnldsv: But it's fine to not give the information then no? Do I have to POST the full form object every time I update a question?

jotoeri: Now i got confused...^^
Each question-object contains a variable called order. So it would be good, if new questions even get their order and as it is calculated from API, i think it shouldn't be recalculated in Create, but returned. Updating a Question only needs a specific key-value-pair, updating a form updates all variables inside form-table (until now).
(We could maybe remove the order-variable from question-objects when cleaning up the object-structures saying that the question-arrays will be ordered anyways from Db-Request, but this different thing...)

jotoeri: Well - After sleeping about... It's not my favourite, but i can remove it. So order will only be stored inside Db, but not within the question-object. Questions are loaded in an ordered array, afterwards the order is only contained within the order of the array, but not as property...?!

skjnldsv: Nono, I guess it's fine then.
Just needs to make sure I cannot update the keyvalue pair `order` of a single question and break the unique order number

jotoeri: So you mean to block the order-key on updateQuestion? Or within frontend?

skjnldsv: Well we should make sure we don't allow the api to break the questions order :)

jotoeri: Included a block in updateQuestion. Don't know, where else it should be of interest...

PR #271, opened 2020-04-08
Ignore Compile-Cache

jotoeri:
Can we include this one? I'm not sure if you see this too, but i run quite often into the problem to commit the compile-cache... :/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ cypress/snapshots
 
 # Compiled files
 js/
+
+# Compile-Cache
+v8-compile-cache-0/


skjnldsv:
Where is it coming from?

jotoeri:
Make build-js produces the directory...

skjnldsv:
Do you have a proper npm setup on your local user?
I guess you should setup npm to use a local folder?
https://github.com/sindresorhus/guides/blob/master/npm-global-without-sudo.md

jotoeri:
Hm, i followed the steps on you link, but still get the folder...
I'm working on a WSL-Machine as root, to be honest (Some poeple would like to kill me for this? :D). So i don't know, if there are any differences due to this... But doesn't matter. The gitignore suffices for now and a compile-cache shouldn't be a bad thing in general, i think... ;)

skjnldsv:
Ah, no idea for your setup then :)

PR #278, opened 2020-04-10
Convert AccessList

jotoeri:
So, this one now converts the Access-List to the JSON-Object.
Quite many things like the `grantedAs` or the `search()` in PageController fell out, as they were never used, so i didn't convert them.
Tests worked so far, the only thing that does not work is to change the access-list, as i didn't change the sharedDiv.vue for now.

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -47,6 +47,7 @@ use OCP\AppFramework\Http\TemplateResponse;
 use OCP\IGroup;
 use OCP\IGroupManager;
 use OCP\IRequest;
+use OCP\ILogger;
 use OCP\IURLGenerator;
 use OCP\IUserManager;
 use OCP\User; //To do: replace according to API
@@ -66,6 +67,9 @@ class PageController extends Controller {
 	private $userMgr;
 	private $groupManager;
 
+	/** @var ILogger */
+	private $logger;
+
 	public function __construct(
 		IRequest $request,
 		IUserManager $userMgr,
@@ -77,7 +81,8 @@ class PageController extends Controller {
 		QuestionMapper $questionMapper,
 		OptionMapper $optionMapper,
 		SubmissionMapper $SubmissionMapper,
-		AnswerMapper $AnswerMapper
+		AnswerMapper $AnswerMapper,
+		ILogger $logger
 	) {
 		parent::__construct(Application::APP_ID, $request);
 		$this->userMgr = $userMgr;
@@ -90,6 +95,7 @@ class PageController extends Controller {
 		$this->optionMapper = $optionMapper;
 		$this->submissionMapper = $SubmissionMapper;
 		$this->answerMapper = $AnswerMapper;
+		$this->logger = $logger;
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


lib/Controller/PageController.php
```diff
@@ -304,131 +310,55 @@ public function insertSubmission($id, $userId, $answers, $questions) {
...
-	/**
 	 * Check if user has access to this form
-	 *
-	 * @param Form $form
-	 * @return bool
 	 */
-	private function hasUserAccess($form) {
-		$access = $form->getAccess();
+	private function hasUserAccess(Form $form): bool {
+		$access = json_decode($form->getAccess(), true); // assoc=true, => Convert to associative Array
```
skjnldsv, line 316: I would prefer have this in the formMapper
That way we can know from the lint that it's already an array and not any string :)

lib/Db/Form.php
```diff
@@ -37,8 +37,8 @@
  * @method void setDescription(string $value)
  * @method string getOwnerId()
  * @method void setOwnerId(string $value)
- * @method string getAccess()
- * @method void setAccess(string $value)
+ * @method string getAccess() // Must be decoded using json_decode($array, true). (assoc=true, => Convert to associative Array)
+ * @method void setAccess(string $value) // String must be json-encoded access-array!
```
skjnldsv, line 41: Not great for linting, see other comment :thinking:

lib/Controller/PageController.php
jotoeri: Yes, my problem was that i didn't find a way to change/override the getAccess, as it is a function of the getter in Entity.
But i think i now found a nice workaround by renaming the Db-Column, so i don't get a conflict when defining `setAccess`&`getAccess`. :) What do you think about?

skjnldsv: Looks good :)

skjnldsv:
So, I rebased #256 on this and there is quite a few discrepancy.
There is an extra `$shares = $this->getShares($form['id']);` line 175 that broke new Forms. 
There is some weird stuff going on, let's fine the time to finish the ui revamp, and let's fix the sharing :)

jotoeri:
Hm, ok. The getFullForm should not have been used anymore, as from the Only-Own-Forms Cleanup, it is not used anymore when loading forms.

@skjnldsv Can you review the API-Cleanup, so the getFullForm falls out anyways? Best also have a look into it within an editor, not only the diff-view, where it looks quite nasty.
The Rebase onto the API-Cleanup then should be easy. `updateForm` can be used as `writeForm` with a slightly different call, to write the full form for the moment, the `form.form` should be easily replaced by `form` with search/replace and the Form-Loading just needs one additional call when selecting a form to edit.

> there is quite a few discrepancy

What else did you find?

Tomorrow i'll be away, but if you could review the API-PR i can have a look into the rebasing & Errors in the next two days. How far was your branch grown? Was everything working down to Db on the old Base?

skjnldsv:
> What else did you find?

Mostly because I was confused on the access/shares change!
But more importantly, because I was very tired ^^

PR #288, opened 2020-04-14
Migrate Timestamps

### LINKED ISSUE #275: Move dates to timestamps
Move all stored dates to timestamps

jotoeri:
- Moving all dates to unix-timestamps.
- Renamed expirationDate now to expiresTimestamp
- Defined `expiresTimestamp==0` as 'form does not expire/`expires==false`'
- Created the procedures getExpires() & getExpired() for the computed properties of the form.
- Didn't take too much effort into the sidebar now due to the pending rewrite, just so it works for now.

closes #275

### DIFF
--- a/lib/Db/Form.php
+++ b/lib/Db/Form.php
@@ -39,10 +39,10 @@ use OCP\AppFramework\Db\Entity;
  * @method void setOwnerId(string $value)
  * @method array getAccess()
  * @method void setAccess(array $value)
- * @method string getCreated()
- * @method void setCreated(string $value)
- * @method string getExpirationDate()
- * @method void setExpirationDate(string $value)
+ * @method integer getCreated()
+ * @method void setCreated(integer $value)
+ * @method integer getExpiresTimestamp()
+ * @method void setExpiresTimestamp(integer $value)
  * @method integer getIsAnonymous()
  * @method void setIsAnonymous(bool $value)
  * @method integer getSubmitOnce()
@@ -56,7 +56,7 @@ class Form extends Entity {
 	protected $ownerId;
 	protected $accessJson;
 	protected $created;
-	protected $expirationDate;
+	protected $expiresTimestamp;
 	protected $isAnonymous;
 	protected $submitOnce;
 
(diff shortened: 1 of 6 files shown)


lib/Db/Form.php
```diff
@@ -101,9 +105,9 @@ public function read() {
 			'ownerDisplayName' => \OC_User::getDisplayName($this->getOwnerId()),
 			'created' => $this->getCreated(),
 			'access' => $this->getAccess(),
-			'expires' => $expires,
-			'expired' => $expired,
-			'expirationDate' => $this->getExpirationDate(),
+			'expires' => $this->getExpires(),
+			'expired' => $this->getExpired(),
+			'expiresTimestamp' => $this->getExpiresTimestamp(),
```
skjnldsv, line 110: What is the diff between expires & expiresTimestamp ?
`expires` should be enough, no?

jotoeri: `expires` is the boolean value, that is modeled by the checkbox. But yes, maybe we could move that in a further cleanup to be watched & calculated just where it is necessary... Same for `expired`, that is also a calculated boolean.

skjnldsv: I think that is enough to store the timestamp too yes :)

jotoeri: Let's do this when cleaning up the form-object anyways....?

lib/Migration/Version010200Date2020323141300.php
```diff
@@ -220,8 +225,8 @@ public function postSchemaChange(IOutput $output, \Closure $schemaClosure, array
 						'description' => $qb_restore->createNamedParameter($event['description'], IQueryBuilder::PARAM_STR),
 						'owner_id' => $qb_restore->createNamedParameter($event['owner'], IQueryBuilder::PARAM_STR),
 						'access_json' => $qb_restore->createNamedParameter($newAccessJSON, IQueryBuilder::PARAM_STR),
-						'created' => $qb_restore->createNamedParameter($event['created'], IQueryBuilder::PARAM_STR),
-						'expiration_date' => $qb_restore->createNamedParameter($event['expire'], IQueryBuilder::PARAM_STR),
+						'created' => $qb_restore->createNamedParameter($this->convertDateTime($event['created']), IQueryBuilder::PARAM_INT),
+						'expires_timestamp' => $qb_restore->createNamedParameter($this->convertDateTime($event['expire']), IQueryBuilder::PARAM_INT),
```
skjnldsv, line 229: So I might be picky here, but I actually don't like using variable names that describe themselves :p
it's like doing `submit_once_boolean`, doesn't make sense to me :see_no_evil: I would go for just simple `expires`, same we do with `created` :grin:

jotoeri: Yes, you're right, but this is the same as we just discussed above. Expires is used by the other expires variable. So when cleaning up the form-object, i will rename this one to `expires` too.

skjnldsv: You'll clean the migration too?

jotoeri: Well, when renaming the object-variable, DB should be changed consistently too, yes.

skjnldsv: Good to me :)

PR #289, opened 2020-04-14
Fix/api cleanup

jotoeri:
This would cleanup the rest of API-Controller.
- Changing writeForm to an updateForm right as we did for questions.
- Removing unused and not useful routes getOptions & getShares.
- Providing the API to load just necessary data for forms-list `getForms()`, while loading the full form only via `getForm()`, when the form is chosen to edit (was proposed somewhere on first design-enhancement...?)
- Renamed `formMapper->find()` to `formMapper->findById()` to have consistent and clear calls.
- Due to reordering the procedures, the diff became quite strange. I tried to split into commit&fixup here.

Depends now a bit on #256, as it would for now destroy functionality of the UI, but would provide a proper interface for this PR...?


skjnldsv:
Found a few bugs, fixing them in the #256

Issue #296, opened 2020-04-23
Design review of current state


jancborchardt:
So @skjnldsv and I just did a design review of the latest state in https://github.com/nextcloud/forms/pull/256, here are items to be done. :)

cc @nextcloud/forms feel free to attach your name to items which you would like to pick up (I will do some design and wording ones). **Bold items are ones we considered more important.**

## Fresh usability test with 3 separate people

Sorted somewhat by importance, duplicates some of the items below, but repeating to emphasize their importance:

- [ ] "Only allow one response per user" is confusing, should only be visible when possible
- [x] Sometimes question title is not saved _-> Seems to be outdated._
- [x] Submission view: Pressing Enter in short text submits whole form (if everything is filled), unexpected #413 
- [x] Questions with either empty titles or empty answer options are not shown – we should communicate that to the form creator and highlight those unfinished questions? #415
- [x] Not clear which question type (multiselect or checkboxes) is used, should show checkboxes / radio buttons in edit mode too #409 
- [x] Not possible to switch question type #352 -> scheduled for V2.1
- [x] Sometimes last response option is not saved when using Tab-key to advance to next field #396 
- [x] Not clear that "New question" is editable, should not be prefilled → pull request https://github.com/nextcloud/forms/pull/367 @jancborchardt 
- [x] Not clear that "New form" is editable, should not be prefilled → pull request https://github.com/nextcloud/forms/pull/367 @jancborchardt 
- [x] Submission view: "People can enter a short answer" for short/long text makes no sense, needs to be differentiated. In create view: "People can enter a short answer" and in submit view "Enter a short answer" #389 _(we can do a quick-fix to make it "Short answer text" and disable the field?) - @jancborchardt for the quick fix_
- [x] Questions with empty titles are possible and will be shown, should be pointed out #397
- [x] Not clear how to get share link, needs to be in sidebar next to radio option too, and ideally in responses view too when empty
- [x] Autosave sometimes messes with input of questions, try typing in answer options fast and sometimes characters disappear https://github.com/nextcloud/forms/pull/327
- [x] "Add question" button position confusing on top, should be below questions https://github.com/nextcloud/forms/pull/328
- [x] Icon of multiple choice question should use radio icon to reduce confusion about what "Multiple choice" is https://github.com/nextcloud/forms/pull/329

## Create

* [ ] Add bottom padding to compensate the add new answer input (optional if looks weird)
[...]

PR #297, opened 2020-04-23
Further design improvements after review

jancborchardt:
Ref #296

- "+" button needs text "Add question" – probably a Vue component thing? → opened issue at https://github.com/nextcloud/nextcloud-vue/issues/1055
- [x] "+ Add question" button should be moved to the left instead (I could look into that)

This is how it looked before:
forms emptycontent

### DIFF
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -131,7 +131,7 @@ export default {
 					? t('forms', 'Form link copied')
 					: t('forms', 'Cannot copy, please copy the link manually')
 			}
-			return t('forms', 'Copy to clipboard')
+			return t('forms', 'Copy share link')
 		},
 	},
 
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -38,9 +38,9 @@
 			</ActionLink>
 			<ActionRouter :close-after-click="true"
 				:exact="true"
-				icon="icon-forms"
+				icon="icon-comment"
 				:to="{ name: 'results', params: { hash: form.hash } }">
-				{{ t('forms', 'Show results') }}
+				{{ t('forms', 'Responses') }}
 			</ActionRouter>
 			<!-- <ActionRouter :close-after-click="true"
 				:exact="true"
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -210,6 +210,7 @@ export default {
 			border-bottom: 1px dotted transparent;
 			border-radius: 0;
 			font-size: 16px;
+			font-weight: bold;
 			line-height: 22px;
 		}
 
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -185,7 +185,7 @@ export default {
 	&__content {
 		flex: 1 1 100%;
 		max-width: 100%;
-		margin: 20px;
+		margin-bottom: 20px;
 		padding: 0;
 	}
 
@@ -195,8 +195,7 @@ export default {
 		flex: 1 1 100%;
 		justify-content: space-between;
 		width: auto;
-		margin: 20px;
-		margin-bottom: 0;
+		margin-top: 20px;
 
 		// Using type to have a higher order than the input styling of server
 		&-title,
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -155,8 +155,6 @@ export default {
 	justify-content: stretch;
 	margin-bottom: 22px;
 	padding-left: 44px;
-	// room for the new question menu
-	padding-right: 44px;
 	user-select: none;
 	background-color: var(--color-main-background);
 
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -167,7 +167,15 @@ export default {
 		left: 0;
 		width: 44px;
 		height: 100%;
+		opacity: .5;
+
+		&:hover,
+		&:focus {
+			opacity: 1;
+		}
+
 		cursor: grab;
+
 		&:active {
 			cursor: grabbing;
 		}
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -423,7 +423,7 @@ export default {
 			top: var(--header-height);
 			display: flex;
 			align-items: center;
-			align-self: flex-end;
+			align-self: flex-start;
 			width: 44px;
 			height: var(--top-bar-height);
 			// make sure this doesn't take any space and appear floating
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -37,8 +37,8 @@
 		<!-- Show results & sidebar button -->
 		<TopBar>
 			<button @click="showResults">
-				<span class="icon-forms" role="img" />
-				{{ t('forms', 'Show results') }}
+				<span class="icon-comment" role="img" />
+				{{ t('forms', 'Responses') }}
 			</button>
 			<button v-tooltip="t('forms', 'Toggle settings')"
 				:aria-label="t('forms', 'Toggle settings')"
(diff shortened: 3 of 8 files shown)


skjnldsv:
> * [ ]  "+" button needs text "Add question" – probably a Vue component thing?

Yes, not supported for now :/
Might have to delay this :smiley:

jancborchardt:
> > * [ ]  "+" button needs text "Add question" – probably a Vue component thing?
> 
> Yes, not supported for now :/
> Might have to delay this :smiley: 

Ok, then I would say we can merge this as-is. :)

src/views/Create.vue
```diff
@@ -371,7 +371,7 @@ export default {
 	header,
 	section {
 		width: 100%;
-		max-width: 900px;
+		max-width: 650px;
```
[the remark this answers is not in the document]
jancborchardt: Google Forms uses 640px at least for questionnaire filling. What do you think @skjnldsv? 
We can also use 750px first though.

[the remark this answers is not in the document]
skjnldsv: I agree with @jotoeri here. 900 looks better :)
We are also using the same as other places in Nextcloud (Talk), I think it's best to keep it the same :)

src/components/Questions/Question.vue
```diff
@@ -154,7 +160,7 @@ export default {
 	&__content {
 		flex: 1 1 100%;
 		max-width: 100%;
-		margin: 20px;
+		margin-bottom: 20px;
```
skjnldsv, line 163: Why those changes?

src/views/Create.vue
```diff
@@ -434,7 +423,7 @@ export default {
 			top: var(--header-height);
 			display: flex;
 			align-items: center;
-			align-self: flex-end;
+			left: 44px;
```
skjnldsv, line 426: Any reasons for this? What were you trying to do here?

src/views/Create.vue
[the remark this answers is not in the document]
jancborchardt: 900 is really a lot for creating a questionnaire – check out the Google Forms width. I’ll change it to 750 as a middle path.

src/components/Questions/Question.vue
jancborchardt: Because the reply options were too far away from the question, they didn’t look like they belong together.

src/views/Create.vue
jancborchardt: It’s moving the "Add question" menu to the left instead of to the right, and always leaving 44px for the navigation toggle.

src/components/Questions/Question.vue
jancborchardt: Yeah – so any margin we would want to the right of the answers should be padding, so it’s still clickable/tappable on the right. Will check that

jancborchardt: Checked and it’s fine, nothing that was not there before this change. Will work further on the whitespace in follow-up

src/views/Create.vue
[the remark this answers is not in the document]
jancborchardt: Done

src/components/AppNavigationForm.vue
```diff
@@ -101,16 +98,15 @@ export default {
...
 		 */
-		bulletColor() {
-			const style = getComputedStyle(document.body)
+		formTypeIcon() {
 			if (this.form.expires && moment().unix() > this.form.expires) {
-				return style.getPropertyValue('--color-error').slice(-6)
+				return 'icon-checkmark'
 			}
-			return style.getPropertyValue('--color-success').slice(-6)
+			return 'icon-forms'
```
jotoeri, line 109: With this change, the total view gets quite grey and monotonous. And the forms icon is now visible quite everywhere... 🤔
I liked the bullets... Maybe sth. like an aria-label could make them a bit more understandable, if this was the intention?

src/views/Sidebar.vue
```diff
@@ -132,7 +132,7 @@ export default {
 
 	data() {
 		return {
-			opened: true,
+			opened: false,
```
jotoeri, line 137: Isn't there a possiblitly to do this only on mobile? As on desktop i think it's good to see the options directly. Doesn't disturb too much...

src/views/Sidebar.vue
```diff
@@ -22,92 +22,94 @@
...
-			</label>
-		</div>
-
-		<ShareDiv v-show="form.access.type === 'selected'"
-			:active-shares="form.shares"
-			:placeholder="t('forms', 'Name of user or group')"
-			:hide-names="true"
-			@update-shares="updateShares"
-			@remove-share="removeShare" />
+		<h3>{{ t('forms', 'Responses') }}</h3>
```
jotoeri, line 25: 'Configurations'/'Settings' & and an icon was better here in my opinion...

src/views/Sidebar.vue
```diff
@@ -22,92 +22,94 @@
...
+				</label>
+				<DatetimePicker v-show="formExpires"
+					id="expiresDatetimePicker"
+					v-model="form.expires"
+					v-bind="expirationDatePicker"
+					@change="onExpiresChange" />
+			</li>
+		</ul>
+
+		<h3>{{ t('forms', 'Sharing') }}</h3>
```
jotoeri, line 65: Same here... An icon for the heading was good, i think...

src/components/AppNavigationForm.vue
jancborchardt: The issue is that the bullets were not understandable. The forms icons help to understand that this is the list of forms.

Aria is for accessibility, it's not something that should be required for basic understanding. I'd say let's work with this first and then see what makes sense for the icons.

src/views/Sidebar.vue
jancborchardt: Needs input from @nextcloud/vuejs – I wouldn't mind the sidebar opening by default if there is enough screen space.

But as it is, if we have to decide between always opening or not, I'd rather go with not as it takes the attention away and on mobile it fully overlaps the main interface.

src/views/Sidebar.vue
jancborchardt: Right, will change it back to "Settings".

src/views/Sidebar.vue
jancborchardt: It's non-standard design here, and the "Sharing" entries all have icons.

At this stage I'd like to get all these design fixes in as it's quite the blocker and then we can work based on master again. :)

src/views/Sidebar.vue
ChristophWurst: yes there is a mixin `isMobile` in the vue lib

src/views/Sidebar.vue
jotoeri: Did it...

src/components/AppNavigationForm.vue
jotoeri: Well, let's push this commit to another branch and discuss there, then i'm fine with merging.

I think having a list with the form-titles is understandable, that this is the list of forms. And a check-symbol for a form, that expired even makes it less understandable to me...

src/views/Create.vue
jotoeri: @skjnldsv Are you fine with this?

src/components/AppNavigationForm.vue
jotoeri: https://github.com/nextcloud/forms/tree/design/navigation-icon

src/views/Create.vue
skjnldsv: No, please use the same syntax that was here before:
```suggestion
			align-self: flex-start
```

jotoeri: It's done. If you're good, you can merge...

src/views/Sidebar.vue
jancborchardt: Thank you @jotoeri! :)

PR #298, opened 2020-04-24
Design/result topbar

jotoeri:
> - [ ] Add "Back to form" in the top navbar

I tried the last days already to create a new Result-View and oriented on the new Create.vue. So the NavBar i included there already. This one i adopted now as kind of quick-fix to have on frontend only the NavBar included. Code is adapted similarly to the Code of new Create.vue.

(Edit: Ok, should have requested merge onto the design-branch. Sorry.)

grafik

_The very first try of a completely new result-view is in [this branch](https://github.com/jotoeri/forms/tree/enh/result_view). Probably i will need some experienced help on the final design, but i think the first idea is visible. Having a sidebar there too, would then enable to switch the block (sorting) between Question-Blocks & Submission-Blocks and provide the Download/StoreInCloud functionality._

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -23,59 +23,85 @@
   -->
 
 <template>
-	<AppContent>
-		<div class="section">
-			<h2>{{ t('forms', 'Results of {title}', { title: form.title }) }}</h2>
-			<button class="button btn primary" @click="download">
-				{{ t('forms', 'Export to CSV') }}
+	<AppContent v-if="loadingResults">
+		<EmptyContent icon="icon-loading">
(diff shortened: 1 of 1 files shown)


skjnldsv:
@jotoeri did you pushed Jan's commits here?

jancborchardt:
You can now change the base branch via the title "Edit" button. :) Did that, looks good now :)

jotoeri:
> You can now change the base branch via the title "Edit" button. :) Did that, looks good now :)

Ahh, thank you Jan. Didn't find the way to do that...

PR #305, opened 2020-04-27
Quick-Fix Restore Voting

jotoeri:
Would enable to restore voting until wednesday and having time to move voting to vue-components...


skjnldsv:
Let's see what I can come up with today :)

PR #308, opened 2020-04-28
Enh/submit

skjnldsv:
- [x] Main default rendering
- [x] Accessibility & use default browser form validation
- [ ] Screen reader (bonus)
- [x] Cleanups old implementations/scripts/templates
- [x] Improve POSTing
- [ ] Add extra padding at the bottom like in creation mode https://github.com/nextcloud/forms/blob/dd06c4671cfcb3377e50e38d430b633944d0f1d0/src/views/Create.vue#L421
- [ ] Find a way to merge the two layouts base for both views (they use the same styling for header and form/section)


src/components/Questions/Question.vue
```diff
@@ -96,6 +97,10 @@ export default {
 			type: Boolean,
 			required: true,
 		},
+		readOnly: {
```
skjnldsv, line 100: Lol, I had something in my head, and now I realise I need to pick a better variable name ^^'
This is meant to disable the edit Mode.
i'll probably go for `canEdit` or something like that ? :thinking:

jotoeri: It's good, i understood directry, what you meant... ;)

jotoeri:
grafik
Text is a bit cut....

lib/Controller/PageController.php
```diff
@@ -156,6 +162,45 @@ public function getResult(): TemplateResponse {
...
+			}
+
+		} catch (DoesNotExistException $e) {
+			//handle silently
+		}finally{
+			return $questionList;
+		}
+	}
+
+	private function getForm(int $id): array {
```
jotoeri, line 199: Can we use Api from here? Copied code from Api is not the best way, no?
What if we take a reference to API in constructor and call the getForm from there? Bad style? ;)

skjnldsv: Well, we can remove everything here I think since it was only for the submission?

lib/Controller/ApiController.php
```diff
@@ -62,12 +63,13 @@ class ApiController extends Controller {
 	/** @var ILogger */
 	private $logger;
 
-	/** @var string */
-	private $userId;
+	/** @var IUserSession */
+	private $userSession;
```
skjnldsv, line 67: :+1:

lib/Controller/ApiController.php
```diff
@@ -603,4 +606,84 @@ public function getSubmissions(string $hash): Http\JSONResponse {
...
+		try {
+			$form = $this->formMapper->findById($formId);
+			$questions = $this->getQuestions($formId);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
+			return new Http\JSONResponse(['message' => 'Could not find form'], Http::STATUS_BAD_REQUEST);
+		}
+
+		$user = $this->userSession->getUser();
+		// TODO check again hasUserAccess?!
```
skjnldsv, line 633: Why? Did you found an issue?

lib/Controller/ApiController.php
```diff
@@ -603,4 +606,84 @@ public function getSubmissions(string $hash): Http\JSONResponse {
...
+			'formId' => $formId,
+			'answers' => $answers,
+		]);
+
+		try {
+			$form = $this->formMapper->findById($formId);
+			$questions = $this->getQuestions($formId);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
+			return new Http\JSONResponse(['message' => 'Could not find form'], Http::STATUS_BAD_REQUEST);
```
skjnldsv, line 629: NOT_FOUND?

jotoeri:
grafik
See chat... ;)

lib/Controller/ApiController.php
jotoeri: No, it was just again the idea, if users would manually post a request, then the access had to be checked again. Just like the issues you said, if users would post in the name of other users. Now it would be the logged-in user, but if he/her is not allowed to submit, it is not checked yet...

lib/Controller/ApiController.php
jotoeri: Just needs to be replaced in the rest of API, but that's a follow-up now...

lib/Controller/PageController.php
skjnldsv: Ah no, we need it for the `this->getForm`. I guess we should maybe have a common view for those? Do you want to add a Trait?

jotoeri: >Do you want to add a Trait?

Don't know what you mean with this...

lib/Controller/ApiController.php
skjnldsv: So it's all good now that we take the userSession, right?

jotoeri: Well, i don't know, what is most clean... I was just wondering that you cleaned the part, that roeland cleaned up.^^

skjnldsv: It's all good then! If we take the userSession we have the current user, no cheating possible :)

lib/Controller/ApiController.php
jotoeri: You mean in the sense of can't happen?
I just copied the standard findForm, that we use everywhere in API. But yes, it's unlikely, that somebody tries to post to a form, that does not exist...

jotoeri:
grafik
I could provide this as fast thing. But it's not fully tested yet and surely not best design...

lib/Controller/ApiController.php
jotoeri: Ahh - you meant Http-Status NOT_FOUND?
Wouldn't say that, as the route itself has been found, but with bad parameters... We use it in the whole controller like this, and i think it was roeland who epxlicitely changed somewhere a not-found to bad-request. ;)

lib/Controller/PageController.php
```diff
@@ -206,114 +188,88 @@ public function getQuestions(int $formId): array {
...
+	 * @return TemplateResponse
 	 */
-	public function insertSubmission($id, $userId, $answers, $questions) {
+	public function gotoForm($hash): ?TemplateResponse {
+		// Inject style on all templates
+		Util::addStyle($this->appName, 'forms');
 
-		$form = $this->formMapper->findById($id);
-		$anonID = "anon-user-".  hash('md5', (time() + rand()));
+		// TODO: check if already submitted
```
jotoeri, line 217: You meant submitOnce? This is done in hasUserAccess...

src/components/Questions/QuestionLong.vue
```diff
@@ -32,7 +32,7 @@
 			<textarea ref="textarea"
 				:aria-label="t('forms', 'A long answer for the question “{text}”', { text })"
 				:placeholder="t('forms', 'Long answer text')"
-				:readonly="edit"
+				:required="true /* TODO: implement required option */"
```
jotoeri, line 35: Maybe use mandatory later to have common naming?

src/components/Questions/QuestionMultiple.vue
```diff
@@ -42,8 +42,9 @@
 							'checkbox question__checkbox': !isUnique,
 						}"
 						:name="`${id}-answer`"
-						:readonly="true"
-						:type="isUnique ? 'radio' : 'checkbox'">
+						:required="isRequired(answer.id)"
+						:type="isUnique ? 'radio' : 'checkbox'"
+						@change="onChange($event, answer.id)">
```
jotoeri, line 47: Still an $event here?

jotoeri: Ok, seems to be a true event. Found it in onChange... :)

lib/Controller/PageController.php
skjnldsv: Yes!
Okay then, a bit confusing but fair enough :)

lib/Controller/ApiController.php
skjnldsv: yes ;)
Both are fine :)

PR #309, opened 2020-04-28
Further design fixes

jancborchardt:
- Fix 'Add question' button overlapping navigation toggle
- Improve vertical spacing
- Fix answer placeholder for long and short answer question
- Remove max-height for description so it is always shown in full 

## Before
All jumbled and spacing not good
before questions
before

## After
All fixed
spacing
Overlap and question types

### DIFF
--- a/src/components/Questions/QuestionLong.vue
+++ b/src/components/Questions/QuestionLong.vue
@@ -31,7 +31,7 @@
 			<!-- TODO: properly choose max length -->
 			<textarea ref="textarea"
 				:aria-label="t('forms', 'A long answer for the question “{text}”', { text })"
-				:placeholder="t('forms', 'Long answer text')"
+				:placeholder="t('forms', 'People can enter a long answer')"
 				:required="true /* TODO: implement required option */"
 				:value="values[0]"
 				class="question__text"
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -25,9 +25,9 @@
 		<AppContent>
 			<!-- Forms title & description-->
 			<header>
-				<h3 id="form-title">
+				<h2 id="form-title">
 					{{ form.title }}
-				</h3>
+				</h2>
 				<p v-if="!loading && !success" id="form-desc">
 					{{ form.description }}
 				</p>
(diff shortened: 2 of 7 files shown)


src/components/Questions/QuestionLong.vue
```diff
@@ -30,8 +30,8 @@
 		<div class="question__content">
 			<!-- TODO: properly choose max length -->
 			<textarea ref="textarea"
-				:aria-label="t('forms', 'A long answer for the question “{text}”', { text })"
-				:placeholder="t('forms', 'Long answer text')"
+				:aria-label="t('forms', 'People can enter a long answer for the question “{text}”', { text })"
```
skjnldsv, line 33: Not sure this is what aria expects. Aria is more like instructions, so I would keep something describing what I (as a person) have to do, not what others can do ?

jancborchardt: Thing is, as a survey creator, this field is useless. It does not do anything, which is why we said we fix the placeholder. ;)

jancborchardt: Rebased and fixed this to not change the aria value.

src/views/Submit.vue
```diff
@@ -153,6 +153,11 @@ export default {
 	align-items: center;
 	flex-direction: column;
 
+	// Force hide navigation toggle as there is no navigation
+	::v-deep #app-navigation-toggle {
+		display: none !important;
+	}
```
skjnldsv, line 159: Will be fixed with upcoming release of vue components.

src/views/Submit.vue
```diff
@@ -163,35 +168,49 @@ export default {
...
+
+		::v-deep ul.question__content {
+			// Left-align multiple choice and checkboxes with question text
+			margin-left: -14px;
+		}
+
+		// Overwrite core styles
+		::v-deep .question__label {
+			text-align: left !important;
+		}
```
skjnldsv, line 208: Why here and not in the appropriate component?
https://github.com/nextcloud/forms/blob/7462be7bfedc8a2d55cdd5ad5a2fa36dfc8dc121/src/components/Questions/Question.vue#L194
https://github.com/nextcloud/forms/blob/7462be7bfedc8a2d55cdd5ad5a2fa36dfc8dc121/src/components/Questions/Question.vue#L194

skjnldsv: Let me fix so we can do merge this

skjnldsv: Ah ok, I get it now!
I only moved the text align then as it make sense for submit & create view :+1: 
Great work!

jancborchardt: Yeah, as the issue only appears in body-public.

PR #310, opened 2020-04-29
Cleanup old src code

skjnldsv:
Signed-off-by: John Molakvoæ (skjnldsv) <skjnldsv@protonmail.com>


src/router.js
```diff
@@ -73,14 +60,5 @@ export default new Router({
 			name: 'results',
 			props: { default: true },
 		},
-		{
```
skjnldsv, line 76: Should be done as a function, not a dedicated view

src/router.js
```diff
@@ -45,19 +45,6 @@ export default new Router({
 			path: '/',
 			name: 'root',
 		},
-		{
-			path: '/new',
```
skjnldsv, line 49: Done as a request already, not a dedicated route

src/router.js
```diff
@@ -45,19 +45,6 @@ export default new Router({
...
-			components: {
-				default: Create,
-				sidebar: Sidebar,
-			},
-			name: 'create',
-		},
-		{
-			path: '/:hash',
```
skjnldsv, line 57: Different entry point for the submit view

src/plugins/ModalDialog.vue
```diff
@@ -1,97 +0,0 @@
-<template>
```
jotoeri, line 1: Is it typical on NC to use the standard-confirm window? Cause this one looked quite good, i think.
And on the standard window, as it is right now, i can prevent the app from showing up the confirm-window, then preventing me to delete forms, as i can't confirm anymore...

skjnldsv: No, it's not.
But we don't use the Modal anywhere anymore here :)

Nonetheless we have a custom dialog already, but we plan to move it to a dedicated vue component & library in @nextcloud/dialogs

skjnldsv: https://github.com/nextcloud/nextcloud-vue/issues/631#issuecomment-616795543

jotoeri: Ok, so maybe to use the custom-dialog instead of the standard one would be good... As i said, i ran into the problem already to not be able to delete forms anymore...

PR #311, opened 2020-04-29
Bump copyrights & add php cs & fixed linting

skjnldsv:
- Added php code styling rules
- Moved all authors into info.xml where we use it for the app store and in the apps management on Nextcloud
- Updated all copyright to reflect what they truely belong to
- Fixed stylelint

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,22 +2,18 @@
 	"name": "nextcloud/forms",
 	"type": "project",
 	"license": "AGPLv3",
-	"authors": [
-		{
-			"name": "Vinzenz Rosenkranz"
-		},
-		{
-			"name": "Kai Schröer"
-		},
-		{
-			"name": "René Gieling"
-		}
-	],
 	"require-dev": {
-		"christophwurst/nextcloud": "^16.0",
-		"phpunit/phpunit": "^8"
-	},
-	"scripts": {
-		"lint": "find . -name \\*.php -not -path './vendor/*' -exec php -l \"{}\" \\;"
-	}
+		"christophwurst/nextcloud": "^17.0",
+		"phpunit/phpunit": "^8",
+        "nextcloud/coding-standard": "^0.3.0"
+    },
+    "config": {
+        "optimize-autoloader": true,
+        "classmap-authoritative": true
+    },
+    "scripts": {
+        "lint": "find . -name \\*.php -not -path './vendor/*' -exec php -l \"{}\" \\;",
+        "cs:check": "php-cs-fixer fix --dry-run --diff",
+        "cs:fix": "php-cs-fixer fix"
+    }
 }
--- a/lib/Controller/SystemController.php
+++ b/lib/Controller/SystemController.php
@@ -2,22 +2,23 @@
 /**
  * @copyright Copyright (c) 2017 Vinzenz Rosenkranz <vinzenz.rosenkranz@gmail.com>
  *
- * @author René Gieling <github@dartcafe.de>
+ * @author affan98 <affan98@gmail.com>
+ * @author Roeland Jago Douma <roeland@famdouma.nl>
  *
  * @license GNU AGPL version 3 or any later version
  *
- *  This program is free software: you can redistribute it and/or modify
- *  it under the terms of the GNU Affero General Public License as
- *  published by the Free Software Foundation, either version 3 of the
- *  License, or (at your option) any later version.
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
  *
- *  This program is distributed in the hope that it will be useful,
- *  but WITHOUT ANY WARRANTY; without even the implied warranty of
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- *  GNU Affero General Public License for more details.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
  *
- *  You should have received a copy of the GNU Affero General Public License
- *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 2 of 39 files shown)


composer.json
```diff
@@ -2,22 +2,18 @@
 	"name": "nextcloud/forms",
 	"type": "project",
 	"license": "AGPLv3",
-	"authors": [
```
skjnldsv, line 5: All moved to info.xml as this is what the appstore displays

package.json
```diff
@@ -2,50 +2,6 @@
 	"name": "forms",
 	"description": "Forms app for nextcloud",
 	"version": "2.0.0-beta1",
-	"authors": [
```
skjnldsv, line 5: All moved to info.xml as this is what the appstore displays

lib/Controller/SystemController.php
```diff
@@ -28,13 +29,10 @@
 use OCP\AppFramework\Http\DataResponse;
 
 use OCP\IGroupManager;
-use OCP\IUser;
```
jotoeri, line 31: Not sure, if this is good... The userManager returns a IUser, where parameters are read from then.
Same applies for the Groups, maybe even the `IGroup` should be included...?

lib/Db/OptionMapper.php
```diff
@@ -1,32 +1,31 @@
 <?php
+
 declare(strict_types=1);
 
 /**
  * @copyright Copyright (c) 2019 Inigo Jiron <ijiron@terpmail.umd.edu>
```
jotoeri, line 6: In a few places you changed the authors, but not the copyright... So seems like particularly people now have copyright, that are not named as authors? Is that reasonable?

lib/Db/OptionMapper.php
```diff
@@ -1,32 +1,31 @@
...
- *  You should have received a copy of the GNU Affero General Public License
- *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
 
 namespace OCA\Forms\Db;
 
-use OCP\DB\QueryBuilder\IQueryBuilder;
```
jotoeri, line 29: Nope, it's used on the whole document...? ;)

lib/Db/QuestionMapper.php
```diff
@@ -28,10 +30,7 @@
 use OCP\IDBConnection;
 use OCP\AppFramework\Db\QBMapper;
 
-use OCA\Forms\Db\OptionMapper;
```
jotoeri, line 31: Used in deleteByForm to delete the corresponding options of a question.

lib/Db/OptionMapper.php
skjnldsv: This is our automated script :)
Copyright is set to the creator of the file.
Author to everyone that contributed afterwards

lib/Db/OptionMapper.php
skjnldsv: Weird, usually it's done properly, this is automated. :thinking:

lib/Db/SubmissionMapper.php
```diff
@@ -27,10 +29,7 @@
 use OCP\IDBConnection;
 use OCP\AppFramework\Db\QBMapper;
 
-use OCA\Forms\Db\AnswerMapper;
```
jotoeri, line 30: Same here - Used to delete corresponding Answers when deleting submissions.

lib/Migration/Version0010Date20190000000007.php
```diff
@@ -6,28 +6,25 @@
...
- *  You should have received a copy of the GNU Affero General Public License
- *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
 
 namespace OCA\Forms\Migration;
 
-use Doctrine\DBAL\Exception\TableNotFoundException;
```
jotoeri, line 26: Not sure if it works without, the $schema->hasTable could return that Exception....?

lib/Migration/Version010200Date20200323141300.php
```diff
@@ -2,29 +2,28 @@
...
- *  You should have received a copy of the GNU Affero General Public License
- *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
 
 namespace OCA\Forms\Migration;
 
-use Doctrine\DBAL\Exception\TableNotFoundException;
```
jotoeri, line 26: Same here - hasTable is used quite often...

tests/bootstrap.php
```diff
@@ -2,22 +2,22 @@
 /**
  * @copyright Copyright (c) 2017 Kai Schröer <git@schroeer.co>
  *
- * @author Kai Schröer <git@schroeer.co>
```
jotoeri, line 5: Are you sure? I think i wouldn't change that. Affan was not the only one developing on the project...

lib/Db/OptionMapper.php
jotoeri: Hm, ok.

tests/bootstrap.php
jotoeri: Ok, seems to be the script again...?

lib/Controller/SystemController.php
skjnldsv: But the class is not used. It's not because we're manipulating the object that we need to use the class :)

lib/Migration/Version010200Date20200323141300.php
skjnldsv: https://github.com/nextcloud/server/blob/5bf3d1bb384da56adbf205752be8f840aac3b0c5/lib/public/DB/ISchemaWrapper.php#L51

hasTable does not throws. We're not catching. We don't need TableNotFoundException

lib/Controller/SystemController.php
jotoeri: Hm, ok. Then i'm not enough within php. I thoght if we call methods of the class onto an instance of the class, the class would be necessary.

lib/Migration/Version010200Date20200323141300.php
jotoeri: ok...

PR #312, opened 2020-04-29
Fix question icons broken in dark theme

skjnldsv:
Capture d’écran_2020-04-29_12-24-49


skjnldsv:
> Don't know how it looked before, but looks good. ;)

Only the two first were inverted
image

jancborchardt:
Just noticed – the primary button icon (and text) should be white in the dark theme as well, as otherwise it's not enough contrast.

skjnldsv:
@jancborchardt can you add that to the list so it's not lost? :)

jancborchardt:
> @jancborchardt can you add that to the list so it's not lost? :)

Isn't it an issue for the Vue repository?

skjnldsv:
No, we're enfocring the icon with ours :)
It won't be an issue anymore with the icon library though ;) (since we'd be able to use css `color`)

PR #313, opened 2020-04-29
Comply to new Actions primary/title standard

skjnldsv:
Needs https://github.com/nextcloud/nextcloud-vue/pull/1062

image

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -356,8 +356,6 @@ export default {
 </script>
 
 <style lang="scss">
-// Replace with new vue components release
-#app-content,
 #app-content-vue {
 	display: flex;
 	align-items: center;
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -71,9 +71,9 @@
 			<!-- Add new questions toolbar -->
 			<div class="question-toolbar" role="toolbar">
 				<Actions ref="questionMenu"
-					v-tooltip="t('forms', 'Add a question to this form')"
-					:aria-label="t('forms', 'Add a question to this form')"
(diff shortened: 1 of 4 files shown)


src/views/Create.vue
```diff
@@ -71,9 +71,10 @@
 			<!-- Add new questions toolbar -->
 			<div class="question-toolbar" role="toolbar">
 				<Actions ref="questionMenu"
-					v-tooltip="t('forms', 'Add a question to this form')"
 					:aria-label="t('forms', 'Add a question to this form')"
```
jancborchardt, line 74: Now that we have a proper button with text, we can remove the additional aria-label, no?

skjnldsv: They're different

skjnldsv:
> * Text in our buttons is always bold, especially in the primary actions

Should be for all right?
https://github.com/nextcloud/nextcloud-vue/pull/1062

src/views/Create.vue
skjnldsv: Ah, you meant accessibility speaking!
Yes we don't need it anymore, I was thinking about the technical, they're different props :stuck_out_tongue_closed_eyes:

skjnldsv:
Done in https://github.com/nextcloud/nextcloud-vue/pull/1062 for the bold part, please review there Jan

jancborchardt:
The Vue fix was merged, can we merge this then or do we need a release of the components?

skjnldsv:
> we need a release of the components?

this

skjnldsv:
Pushed @nextcloud/vue 2.0.0 update

PR #315, opened 2020-04-29
Add title on public page

### LINKED ISSUE #4: Don't show Nextcloud Toolbar with all apps
1.0.3

**Nextcloud or ownCloud?:**
Nextcloud

16.0.1

### Steps to reproduce
1. Create a form, make it public
2. Copy the link
3. Open the link in another browser or in icognito mode, so you're not logged into the Nexcloud instance

### Actual behaviour
On the top, the Nextcloud Toolbar is shown, including all apps that are installed.

### Expected behaviour
The toolbar shouldn't show the installed apps, because it leaks information about the installation.
Instead, a "blank" blue toolbar could be shown, like e.g. the Polls app does for public polls.

skjnldsv:
FInal fix #4

Capture d’écran_2020-04-29_19-06-46


jancborchardt:
> FInal fix #4
> 
> Capture d’écran_2020-04-29_19-06-46

In the subline we usually show the "shared-by" info as per the class name, for additional context: "shared by username"
That’s better than repeating the title of the form which already is shown in very large.

skjnldsv:
> In the subline we usually show the "shared-by" info as per the class name, for additional context: "shared by username"

Not anymore, we removed it for privacy reasons apparently

skjnldsv:
Blocked by https://github.com/nextcloud/server/pull/20744

jancborchardt:
> > In the subline we usually show the "shared-by" info as per the class name, for additional context: "shared by username"
> 
> Not anymore, we removed it for privacy reasons apparently
> EDIT: only when specified on your displayname visibility settings

Blocker is merged, can you adjust the subline then to show the "shared by username" info? If it’s disabled in the displayname visibility settings, just don’t show any subline.

skjnldsv:
Rebased, ready to go @jancborchardt @jotoeri :)

lib/Controller/PageController.php
```diff
@@ -194,12 +204,30 @@ public function gotoForm($hash): ?TemplateResponse {
...
+			$response = new PublicTemplateResponse($this->appName, 'main');
+			$response->setHeaderTitle($form->getTitle());
+
+			// Get owner and check display name privacy settings
+			$owner = $this->userManager->get($form->getOwnerId());
+			if ($owner instanceof IUser) {
+				$ownerAccount = $this->accountManager->getAccount($owner);
+
+				$ownerName = $ownerAccount->getProperty(IAccountManager::PROPERTY_DISPLAYNAME);
+				if ($ownerName->getScope() === IAccountManager::VISIBILITY_PUBLIC) {
```
jotoeri, line 222: Maybe check here for `VISIBILITY_CONTACTS_ONLY`  OR `VISIBILITY_PUBLIC`? User-Logged-in is checked already, so contacts-only should apply?

lib/Controller/PageController.php
```diff
@@ -194,12 +204,30 @@ public function gotoForm($hash): ?TemplateResponse {
...
+					$response->setHeaderDetails($this->l10n->t('Shared by %s', [$owner->getDisplayName()]));
```
jotoeri, line 223: I think i would use here $ownerName->getValue(). Just to keep within the scope of IAccount, instead of switching back to IUser. This is a bit confusing...

lib/Controller/PageController.php
skjnldsv: I'm using the same as we do for public shares.
This was review by our security/privacy devs, so I'm not changing this :p

jotoeri: Yes, sorry, i just saw its if NOT logged in. 👍

lib/Controller/PageController.php
jotoeri: Pushed this as fixup now, so if you're ok with, you can just squash & merge. Hope that's ok for you? ;)

PR #326, opened 2020-05-01
Fix expiration display

skjnldsv:
Kazam_screenshot_00014
Kazam_screenshot_00015
Kazam_screenshot_00016

### DIFF
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -24,10 +24,9 @@
 	<AppNavigationItem
 		ref="navigationItem"
 		:exact="true"
-		:icon="loading ? 'icon-loading-small' : ''"
+		:icon="icon"
 		:title="form.title"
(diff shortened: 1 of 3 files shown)


jotoeri:
- Hm, i still don't think the checkmark is reasonable to mark as expired... 🤔 
- Can we disable typing on the field? When manually changing the text, expires falls back to 1970.

skjnldsv:
> * Hm, i still don't think the checkmark is reasonable to mark as expired... thinking

See with @jancborchardt ;)

> * Can we disable typing on the field? When manually changing the text, expires falls back to 1970.

Good call!

jancborchardt:
> > * Hm, i still don't think the checkmark is reasonable to mark as expired... thinking
> 
> See with @jancborchardt ;)

So @skjnldsv and I both discussed in a design review and it’s already more understandable than the colors. Not perfect, but what is ever? Let’s take small steps as always. :)

PR #328, opened 2020-05-01
Move "Add a question" button to bottom

jotoeri:
Before:
grafik

After: 
grafik

I removed the top-margin of Question.vue. The Questions still get rendered with the same distance due to the bottom-margin, just the white-space between header & first Question is not too large like this. Works for you?

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -68,25 +68,6 @@
 		</header>
 
 		<section>
-			<!-- Add new questions toolbar -->
-			<div class="question-toolbar" role="toolbar">
-				<Actions ref="questionMenu"
-					:open.sync="questionMenuOpened"
-					:menu-title="t('forms', 'Add a question')"
-					:primary="true"
-					:default-icon="isLoadingQuestions ? 'icon-loading-small' : 'icon-add-white'">
-					<ActionButton v-for="(answer, type) in answerTypes"
-						:key="answer.label"
-						:close-after-click="true"
-						:disabled="isLoadingQuestions"
-						:icon="answer.icon"
-						class="question-toolbar__question"
-						@click="addQuestion(type)">
-						{{ answer.label }}
-					</ActionButton>
-				</Actions>
-			</div>
-
 			<!-- Questions list -->
 			<Draggable v-model="form.questions"
 				:animation="200"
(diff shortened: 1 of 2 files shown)


src/views/Create.vue
```diff
@@ -411,16 +411,13 @@ export default {
 		margin-bottom: 250px;
 
 		.question-toolbar {
-			position: sticky;
```
skjnldsv, line 414: I still think it should be sticky so I don't have to scroll at the bottom if I have 10 questions to add a new one  :)

jotoeri: Hm, interesting task... First try was a bit dirty due to the content in background, so made the hole question-toolbar now with background...
To get away a bit from the bottom i further added a padding, using a variable. Can you have a look onto the css in fixup again? :)

grafik
grafik

jancborchardt: > First try was a bit dirty due to the content in background, so made the hole question-toolbar now with background...

We had the same when it was on the top, and this is totally fine. :) We shouldn’t box in / cut a part off the whole content just for the button.

Will check how it looks like with a fat white border – that gives us the separation but not the cutting of content.

src/views/Create.vue
```diff
@@ -411,17 +411,22 @@ export default {
 		margin-bottom: 250px;
 
 		.question-toolbar {
+			$paddingBottom: 10px;
```
jancborchardt, line 414: I wouldn’t use a variable if it’s just a custom value for here anyway, no? Also, we try to convert to a system based on steps of 4 (8, 12, 16, etc :) just for info.

src/views/Create.vue
jancborchardt: Looks totally fine with the floating button, and we do that everywhere else too. Rebased as well :)

src/views/Create.vue
jancborchardt: Adjusted it to 12px

skjnldsv: Actually, we should very similarely use the same we do in the components.
Almost all the custom px sizes we use here are dependent on the 44px icon sizing, so 14px padding, 16px width.
One day 'ill do a pass and index everything based on that so it's cleaner and all aligned as it currently is.

src/views/Create.vue
```diff
@@ -411,17 +411,16 @@ export default {
...
 			display: flex;
 			align-items: center;
 			align-self: flex-start;
-			width: 44px;
-			// To not overlap navigation toggle
-			margin-left: 44px;
-			height: var(--top-bar-height);
+
+			// To align with Drag-Handle
+			margin-left: 16px;
```
skjnldsv, line 423: Most likely 14, no?
But I'm nitpicking! Looks good!

jancborchardt: Really depends what we are aligning with – this is what I was going for and I’d say this is fine so we can focus on more important things for now ;)
forms

skjnldsv:
> > Actually found an issue: Every time you click the "Add a question" button, the whole content view shifts a bit to the top. What’s that? :D

did you narrow it down? a border maybe? How much does it shifts?

jancborchardt:
> > > Actually found an issue: Every time you click the "Add a question" button, the whole content view shifts a bit to the top. What’s that? :D
> 
> did you narrow it down? a border maybe? How much does it shifts?

Yes:
-  It shifts roughly half of the height of the button itself (maybe less, 16px?)
- This happens on master too as said, so no blocker for this
- On master it only happens once you scrolled down so much that the "Add a question" button is sticky on the top

@skjnldsv can you reproduce?

PR #338, opened 2020-05-03
Provide DBs max string lengths as InitialState

### LINKED ISSUE #92: Increase length for description and show proper errors

jotoeri:
Closes #92

I gave a try to the inital-state thing. Hope this is the right place to provide the State? Unfortunately the loadState is not working. Browser console tells me `Error: "Could not find initial state maxStringLengths of forms"`

Can you identify why?

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -64,6 +64,18 @@ class PageController extends Controller {
 	/** @var FormsService */
 	private $formService;
 
+	/** @var Array
+	 *
+	 * Maximum String lengths, the database is set to store.
+	 */
+	private $maxStringLengths = [
+		'formTitle' => 256,
+		'formDescription' => 2048,
+		'questionText' => 2048,
+		'optionText' => 1024,
+		'answerText' => 2048,
+	];
+
 	public function __construct(string $appName,
 								IRequest $request,
 								IGroupManager $groupManager,
@@ -96,6 +108,7 @@ class PageController extends Controller {
 	public function index(): TemplateResponse {
 		Util::addScript($this->appName, 'forms');
 		Util::addStyle($this->appName, 'forms');
+		$this->initialStateService->provideInitialState($this->appName, 'maxStringLengths', $this->maxStringLengths);
 		return new TemplateResponse($this->appName, 'main');
 	}
 
@@ -108,6 +121,7 @@ class PageController extends Controller {
 	public function createForm(): TemplateResponse {
 		Util::addScript($this->appName, 'forms');
 		Util::addStyle($this->appName, 'forms');
+		$this->initialStateService->provideInitialState($this->appName, 'maxStringLengths', $this->maxStringLengths);
 		return new TemplateResponse($this->appName, 'main');
 	}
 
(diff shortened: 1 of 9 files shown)


skjnldsv:
> Can you identify why?

The initial state needs to be provided by the page. So not the api controller.

jotoeri:
OK, thank you, @skjnldsv 

Now it works, i think i tested all cases.

src/components/Questions/QuestionLong.vue
```diff
@@ -54,6 +55,7 @@ export default {
 	data() {
 		return {
 			height: 1,
+			maxStringLengths: loadState('forms', 'maxStringLengths'),
```
skjnldsv, line 58: If all of them have the same data, please use the Question mixin so it doesn't duplicate code :)
Btw, I thought we had different length for short and long answers?

skjnldsv: Ah sorry, we do. I misread. One uses `optionText`, the other `answerText`

skjnldsv: Still, but those into the Mixin for better code styling ;)

lib/Controller/PageController.php
```diff
@@ -182,6 +204,7 @@ public function gotoForm($hash): ?TemplateResponse {
 
 		Util::addScript($this->appName, 'submit');
 		$this->initialStateService->provideInitialState($this->appName, 'form', $this->formsService->getForm($form->getId()));
+		$this->provideStaticData();
```
skjnldsv, line 207: I would rather have this directly here.
Not calling an external function :)

Put the maxStringLengths as constantes too, that is what it's here for 

Regarding the number of duplicate functions like editForm, cloneForm... etc etc, we will put them all into one in a followup pr, we can use the same endpoint for multiple routes with a little trick ;)

src/components/Questions/QuestionLong.vue
jotoeri: Ah, yes, will move this to the mixin...

The option & answer is different thing. Options-Table is the one containing options for multiple-questions (1024 chars), while the answer-table is the one containing user-answers (2048 char as it must be able to take all texts, including long-text answers). Even options are stored within the answer-table, when chosen by the user as answer on submit.
For now i just set the short answer-text to the same max-length, as this is the restriction by DB. If we want, we can restrict short answers to a smaller number, DB won't care^^. But i don't know if it's really necessary...

lib/Controller/PageController.php
jotoeri: Thought we would maybe want to provide more data like this, this was why i chose this way. But yes, for now i can also move it to a constant... 👍

src/components/Questions/QuestionLong.vue
jotoeri: Moved it now up even one more step, so it gets loaded only once on create/submit.

jotoeri:
:D I'm done with this. Took me too much time already. ;)

Just to understand - A mixin is compiled/executed everywhere, where it is included, no? So it would load the state on each vue again, but wouldn't matter, as this is what the StateService is for?^^

PR #341, opened 2020-05-04
New Result View

### LINKED ISSUE #11: Format of results, especially in CSV
A feature request:

1.0.3

**Nextcloud or ownCloud?:**
Nextcloud

16.0.1

### Steps to reproduce
1. Create Form
2. Provide some answers
3. View results

### Actual behaviour

Results are shown with one line per answer:

~~~
Name    Question #   Question             Response
anon1   1            Pineapple on pizza?  Yes
anon1   2            Favorite editor?     Emacs
anon2   1            Pineapple on pizza?  No
anon2   2            Favorite editor?     vim
~~~

### Expected behaviour

I'd actually like the output to be one line per user, so I have all answers in one line, like

~~~
Name    Question #   q1                   r1      q2                 r2
anon1   1            Pineapple on pizza?  Yes     Favorite editor?   Emacs
anon2   1            Pineapple on pizza?  No      Favorite editor?   vim
~~~

or even like

~~~
Name    Question #   Pineapple on pizza?    Favorite editor?
anon1   1            Yes                    Emacs
anon2   1            No                     vim
~~~

That would make the data easier to work with for my use with (using the exported CSV file as a data file for jekyll).

(I'm interested if other users found this useful, too. Currently, I could work around this by doing a custom SQL query as I only have one form).

### LINKED ISSUE #73: Result show sid of ldap user
When i view result of form it just show sid of ldap user, instead of name of ldap user

### LINKED ISSUE #273: New Result-View (incl. show timestamp)
The option to display the date / time the user submitted the form is extremely useful. Currently I only have the answer, I don't know when it was filled out.
Necessary option for a frequent filling form.

It is a form of control of reports / forms, along with the idea of generating PDF or write form responses directly to documents within the cloud.

Great job with the app, congratulations and I hope it always gets better.

### LINKED ISSUE #323: Ability to delete a single response and all respones at once
Hello!

In the results view of a Form you can see all answers made. It would be really nice to remove all responses at once, so clear the results view (e.g. if you overviewed all answeres and don't need them anymore). Also it would be really nice to remove single respones (e.g. if you reviewed a few and don't need them anymore, but the others you still need).

For my work I would use both features regularly.

I know you are improving the results view currently (because it is hard to evaluate them in v2.0 Beta 1), but this seems not to be a point already but would be very useful.

This feature has been discussed already here: #316

jotoeri:
Ok, so this is the current state of my result-view branch.

- Delete works for single as well as all responses at once. Closes #323 
- Export works to csv, using the same format as google does. Closes #11
- Using the displayName of users. Closes #73 
- Shows Timestamp. Closes #273
- As i moved the Options to the top-bar and the actions-button had the 44px, i increased the top-bar now also to 44px. Not my favourite, but i think this would match to what you said with the standard-44px on buttons?
- @jancborchardt Can you please have a look onto the design again? :)
- Next important step would be to include the Summary-View including statistics. (Maybe i do on a separate PR to keep this one small? Then this one would be to review...)
@skjnldsv I think i would not use the Question-vue-files, to make them not too complex with statistics etc., but use separate files for results. What do you think?

grafik


appinfo/routes.php
```diff
@@ -54,8 +54,12 @@
 		['name' => 'api#updateOption', 'url' => '/api/v1/option/update', 'verb' => 'POST'],
 		['name' => 'api#deleteOption', 'url' => '/api/v1/option/{id}', 'verb' => 'DELETE'],
 
+		// Submissions
 		['name' => 'api#getSubmissions', 'url' => '/api/v1/submissions/{hash}', 'verb' => 'GET'],
-		['name' => 'api#insertSubmission', 'url' => '/api/v1/submissions/insert', 'verb' => 'POST'],
+		['name' => 'api#deleteAllSubmissions', 'url' => 'api/v1/submissions/{formId}', 'verb' => 'DELETE'],
```
skjnldsv, line 59:
```suggestion
		['name' => 'api#deleteAllSubmissions', 'url' => '/api/v1/submissions/{formId}', 'verb' => 'DELETE'],
```

jotoeri: Ah sure... 👍

skjnldsv:
@jancborchardt is it time to introduce cards to Nextcloud?

skjnldsv:
> * @skjnldsv I think i would not use the Question-vue-files, to make them not too complex with statistics etc., but use separate files for results. What do you think?

If we're talking abou the  individual view per submission, yes, you should use them, it should look the same as when the user filled the submit form :)

src/components/TopBar.vue
```diff
@@ -51,7 +51,10 @@ $top-bar-height: 60px;
 
 	button {
 		cursor: pointer;
-		&:not(:first-child) {
+		min-height: 44px;
+		margin: 8px;
+
+		&.button-small {
```
skjnldsv, line 57: Please keep it as it was.
Don't use Actions inside, there is no menu, simple buttons are good enough.

src/views/Results.vue
```diff
@@ -31,6 +31,17 @@
 
 	<AppContent v-else>
 		<TopBar>
+			<Actions class="results-menu"
+				:menu-title="t('forms', 'Options')"
```
skjnldsv, line 35: Don't use Actions like this, we use the three dot menu on Nextcloud without text.
This was added for some very specific use case.

src/components/TopBar.vue
skjnldsv: Ah, my bad, you have multiple ActionButton, then you have to use it yes. 
Followup in: https://github.com/nextcloud/forms/pull/341/files#r420056327

appinfo/routes.php
```diff
@@ -54,8 +54,12 @@
 		['name' => 'api#updateOption', 'url' => '/api/v1/option/update', 'verb' => 'POST'],
 		['name' => 'api#deleteOption', 'url' => '/api/v1/option/{id}', 'verb' => 'DELETE'],
 
+		// Submissions
 		['name' => 'api#getSubmissions', 'url' => '/api/v1/submissions/{hash}', 'verb' => 'GET'],
-		['name' => 'api#insertSubmission', 'url' => '/api/v1/submissions/insert', 'verb' => 'POST'],
+		['name' => 'api#deleteAllSubmissions', 'url' => '/api/v1/submissions/{formId}', 'verb' => 'DELETE'],
+
+		['name' => 'api#insertSubmission', 'url' => '/api/v1/submission/insert', 'verb' => 'POST'],
+		['name' => 'api#deleteSubmission', 'url' => '/api/v1/submission/{id}', 'verb' => 'DELETE'],
```
skjnldsv, line 62: cleaner! :+1:

src/views/Results.vue
jancborchardt: Fixed this in my commit :)

src/components/TopBar.vue
jancborchardt: > @jancborchardt is it time to introduce cards to Nextcloud?

Where it makes sense we already use them (Deck, Bookmarks). Here it doesn’t really as they will be full width anyway. Try with an actual survey with real data, that is long questions and possibly long responses. :)

jancborchardt:
> @jancborchardt is it time to introduce cards to Nextcloud?

Where it makes sense we already use them (Deck, Bookmarks). Here it doesn’t really as they will be full width anyway. Try with an actual survey with real data, that is long questions and possibly long responses. :)

jancborchardt:
> Next important step would be to include the Summary-View including statistics. (Maybe i do on a separate PR to keep this one small? Then this one would be to review...)

Yes this should absolutely be separate. :)

jancborchardt:
Added some more detail fixes, this is definitely good to go now. :)

src/components/TopBar.vue
jotoeri: Hm, button was removed now. But i think it won't be the last time we want to add a second button here. So should i change back or are you ok with?

jotoeri:
So - final state?
grafik

Spartachetto:
If I get well this issue, it could be worthwhile to look at #11 and have the feedback from @Rello

PR #343, opened 2020-05-04
Prevent leaking personnal infos on forms

### LINKED ISSUE #342: Privacy issues on v2
- [x] Anyone can get a form even if the form is not set as public https://github.com/nextcloud/forms/blob/fcae747c5fdf3080dacc7f5c5a9f110f1edcbd35/lib/Controller/ApiController.php#L129
- [x] Do not leak the userid on the public pages

skjnldsv:
Fix #342

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -128,13 +128,18 @@ class ApiController extends Controller {
 	 */
 	public function getForm(int $id): Http\JSONResponse {
 		try {
-			$results = $this->formsService->getForm($id);
+			$form = $this->formsService->getForm($id);
 		} catch (IMapperException $e) {
 			$this->logger->debug('Could not find form');
 			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
 		}
+	
+		if (!$this->formsService->hasUserAccess($id)) {
+			$this->logger->debug('User has no permissions to get this form');
+			return new Http\JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
 
-		return new Http\JSONResponse($results);
+		return new Http\JSONResponse($form);
 	}
 
 	/**
@@ -297,7 +302,7 @@ class ApiController extends Controller {
 			$form = $this->formMapper->findById($formId);
 		} catch (IMapperException $e) {
 			$this->logger->debug('Could not find form');
-			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
+			return new Http\JSONResponse(['message' => 'Could not find form'], Http::STATUS_BAD_REQUEST);
 		}
 
 		if ($form->getOwnerId() !== $this->userId) {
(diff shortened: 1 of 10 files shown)


lib/Controller/ApiController.php
```diff
@@ -128,13 +128,18 @@ public function getForms(): Http\JSONResponse {
...
 	public function getForm(int $id): Http\JSONResponse {
 		try {
-			$results = $this->formsService->getForm($id);
+			$form = $this->formsService->getForm($id);
 		} catch (IMapperException $e) {
 			$this->logger->debug('Could not find form');
 			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
 		}
+	
+		if (!$this->formsService->hasUserAccess($id)) {
```
jotoeri, line 137: So this still allows me on a public-access form to request the (non-public) full form, as hasUserAccess returns generally true...

jotoeri: (full form including ownerId)

lib/Service/FormsService.php
```diff
@@ -96,4 +114,79 @@ public function getForm(int $id): array {
...
+		return $form;
+	}
+
+	/**
+	 * Check if user has access to this form
+	 *
+	 * @param integer $id
+	 * @return boolean
+	 */
+	public function hasUserAccess(int $id): bool {
```
jotoeri, line 141: Use formId to clarify?

lib/Controller/ApiController.php
skjnldsv: Ah right, I should then obfuscate if access===public, correct?

lib/Service/FormsService.php
skjnldsv: Good call!

lib/Controller/ApiController.php
jotoeri: Well, then we wouldn't have this information even when editing the form...?

jotoeri: now, that you have getPublicForm for Submit-view, i think we could just use the check for the owner, as in all other procedures below. Then only the owner can use getForm, but only the owner will have it in his form-list anyways...

skjnldsv: AH, no it's actually fine!
You cannot get on public APIs without a csfr token :)
So we don't have to manage the public access :wink:

lib/Controller/ApiController.php
```diff
@@ -128,13 +128,18 @@ public function getForms(): Http\JSONResponse {
...
+			$this->logger->debug('User have no permissions to get this form');
```
jotoeri, line 138: *has ;)

lib/Controller/ApiController.php
jotoeri: Hm, still the hasUserAccess feels a bit deplaced here to me, as it is specifically written for submission (checks e.g. submitOnce). But ok, privacy thing seems to be ok.

skjnldsv: Ah I think it's indeed not ideal.
Maybe I should outsource the getSubmitOnce to a dedicated function like `canSubmit` and we could have a proper error message stating they already voted instead of a 404!

jotoeri: 👍 Yes, then the hasUserAccess gets in a scope where it can be used for getForm.

skjnldsv: Fixed

lib/Controller/ApiController.php
```diff
@@ -606,13 +611,16 @@ public function insertSubmission(int $formId, array $answers) {
 			return new Http\JSONResponse(['message' => 'Could not find form'], Http::STATUS_BAD_REQUEST);
 		}
 
-		$user = $this->userSession->getUser();
-		// TODO check again hasUserAccess?!
```
jotoeri, line 610: hasUserAccess should be checked here, too! Only canSubmit does not suffice.

templates/expired.php
```diff
@@ -1,8 +1,9 @@
 <?php
 /**
- * @copyright Copyright (c) 2019 Inigo Jiron <ijiron@terpmail.umd.edu>
+ * @copyright Copyright (c) 2020 John Molakvoæ (skjnldsv) <skjnldsv@protonmail.com>
```
jotoeri, line 3: Thought this had all been done right, by an automated script...?! ;)

skjnldsv: Not the copyright. Authors yes

PR #346, opened 2020-05-05
Fix users & groups sharing

### LINKED ISSUE #19: Does not respect autocomplete settings
1.0.3
**Nextcloud or ownCloud?:**
Nextcloud
15.0.8
### Steps to reproduce
1. Uncheck checkbox 'Allow autocomplete username in sharing dialog' under '/settings/admin/sharing'
2. Choose 'Only Shared' under Apps 'Configuration -> Access'
3. start typing to find user to share form with 

### Actual behaviour
Full list of users matching partial input shows

### Expected behaviour
Only on full name match it should show match list

skjnldsv:
- [x] Use files_sharing api
- [x] Use same input as in files sharing sidebar
- [x] Cleanup old files
- [x] Add api sanity checks
- [x] Improve api access data

fix #19


skjnldsv:
Please not this doesn't add the same layout as server yet!
Will be easier to add after though.
This is about the sharing providers.

Ideally I would like the app to understand that we 'interacted with' xxx when sharing here too.
So it appear in the recommendation app

lib/Service/FormsService.php
```diff
@@ -205,4 +222,39 @@ public function hasUserAccess(int $formId): bool {
...
+				'displayName' => $user->getDisplayName()
+			];
+		}
+		return [];
+	}
+
+	/**
+	 * Format groups access
+	 *
+	 * @param string $userId
```
jotoeri, line 246: *groupId

lib/Service/FormsService.php
```diff
@@ -205,4 +222,39 @@ public function hasUserAccess(int $formId): bool {
...
 		return false;
 	}
+
+	/**
+	 * Format users access
+	 *
+	 * @param string $userId
+	 * @return array
+	 */
+	private function formatUsers(string $userId): array {
```
jotoeri, line 232: Can we move these two procedures above getForm? Then we have a block of corresponding procedores above getForm. Makes it easier to find correlated functions...

src/components/UserDiv.vue
```diff
@@ -42,37 +36,19 @@ export default {
 		Avatar,
 	},
 	props: {
-		hideNames: {
-			type: Boolean,
-			default: false,
-		},
-		userId: {
+		id: {
```
jotoeri, line 39: userId to clarify?

lib/Service/FormsService.php
```diff
@@ -205,4 +222,39 @@ public function hasUserAccess(int $formId): bool {
...
+	 * @param string $userId
+	 * @return array
+	 */
+	private function formatGroups(string $groupId): array {
+		$group  = $this->groupManager->get($groupId);
+		if ($group instanceof IGroup) {
+			return [
+				'id' => $groupId,
+				'displayName' => $group->getDisplayName(),
+				'isGroup' => true
```
jotoeri, line 255: in UserDiv you used `type=user/group` for this. How about using the same variable here?

src/components/UserDiv.vue
```diff
@@ -82,7 +58,7 @@ export default {
 		computedDisplayName() {
 			let value = this.displayName
 
-			if (this.userId === getCurrentUser().uid) {
+			if (!this.isNoUser && this.userId === getCurrentUser().uid) {
```
jotoeri, line 61: Inverting to isUser makes it easier to read...

src/components/UserDiv.vue
```diff
@@ -82,7 +58,7 @@ export default {
...
 				value = getCurrentUser().displayName
```
jotoeri, line 62: loading the display-name for current-user again, doesn't make sense. Would only apply, if the owner shares with himself?!

src/components/shareDiv.vue
```diff
@@ -23,122 +23,155 @@
...
+		isValidQuery() {
+			return this.query && this.query.trim() !== '' && this.query.length > this.minSearchStringLength
+		},
+
+		/**
+		 * Multiseelct options. Recommendations by default,
+		 * direct search when search query is valid.
+		 * @returns {Array}
+		 */
+		options() {
```
jotoeri, line 127: use `multiselectOptions` as name?

src/components/shareDiv.vue
```diff
@@ -23,122 +23,155 @@
...
-		},
+		/**
+		 * Add a new share and dispatch the change to the parent
+		 * @param {Object} share the new share
+		 */
+		addShare(share) {
+			const users = this.userShares.slice()
+			const groups = this.groupShares.slice()
+			const newShare = {
+				id: share.shareWith,
```
jotoeri, line 162: `share.userId`?

lib/Service/FormsService.php
skjnldsv: Of course!

src/components/UserDiv.vue
skjnldsv: No, it's used for groups as well

lib/Service/FormsService.php
skjnldsv: After thought, I think we should be as close as the original sharing code, so I'd move to sharingType :)

src/components/UserDiv.vue
skjnldsv: Definitely! UserDiv is the part that will be removed anyway :p

src/components/UserDiv.vue
skjnldsv: I mean, this is the olde code, didn't changed anything here :stuck_out_tongue:

src/components/shareDiv.vue
skjnldsv: Nope, this is the files_sharing code here :)

src/components/shareDiv.vue
skjnldsv: rather keep the same as the html prop `options`

src/components/UserDiv.vue
skjnldsv: Ah, no, and it's also because isNoUser is the name of the Avatar prop :)

src/components/UserDiv.vue
jotoeri: 👍

skjnldsv:
Done! Please update review @jotoeri :)

skjnldsv:
> Looks good, but the state of sharing is not reflected in the dropdown.

We don't show the existing shares in the dropdown :)
Same as we do in sharing

lib/Controller/ApiController.php
```diff
@@ -210,6 +212,21 @@ public function updateForm(int $id, array $keyValuePairs): Http\JSONResponse {
...
 		}
 
+		// Make sure we only store id
+		try {
+			if ($keyValuePairs['access']) {
+				$keyValuePairs['access']['users'] = array_map(function (array $user): string {
+					return $user['id'];
+				}, $keyValuePairs['access']['users']);
+				$keyValuePairs['access']['groups'] = array_map(function (array $group): string {
+					return$group['id'];
```
jotoeri, line 222: Typo missing space

skjnldsv: Dammit ! :grin:

skjnldsv: Done

src/components/shareDiv.vue
```diff
@@ -147,6 +178,196 @@ export default {
...
+		 */
+		async getSuggestions(search) {
+			this.loading = true
+
+			const shareType = [
+				this.SHARE_TYPES.SHARE_TYPE_USER,
+				this.SHARE_TYPES.SHARE_TYPE_GROUP,
+			]
+
+			const request = await axios.get(generateOcsUrl('apps/files_sharing/api/v1') + 'sharees', {
```
jotoeri, line 206: One more typo? :D 'sharees'

jotoeri: Ok, seems to be inteded. 🤔

jancborchardt:
> > Looks good, but the state of sharing is not reflected in the dropdown.
> 
> We don't show the existing shares in the dropdown :)
> Same as we do in sharing

They do show for me, I can add people multiple times, and they keep showing in the dropdown even if I close and open again.

jotoeri:
Sorry, thought you would have fixed the Group-Thing.

- [ ] The shown active Shares have the popup, showing (Group) on users, but not on Groups.

> They do show for me, I can add people multiple times, and they keep showing in the dropdown even if I close and open again.

This one works for me, i don't see them... 🤔

jancborchardt:
>     * [ ]  The shown active Shares have the popup, showing (Group) on users, but not on Groups.

Can confirm, seems to be inverse logic.


> > They do show for me, I can add people multiple times, and they keep showing in the dropdown even if I close and open again.
> 
> This one works for me, i don't see them... thinking

Still getting this issue after update and cache clear, but we can also just ignore that then if nobody can reproduce. :)

src/components/UserDiv.vue
```diff
@@ -20,79 +20,47 @@
...
-				value = getCurrentUser().displayName
-			} else {
-				if (!this.displayName) {
-					value = this.userId
-				}
-			}
-			if (this.type === 'group') {
-				value = value + ' (' + t('forms', 'Group') + ')'
+			if (this.shareType !== this.SHARE_TYPES.SHARE_TYPE_GROUP) {
+				return `${this.displayName} (${t('forms', 'Group')})`
```
jancborchardt, line 61: There we have the culprit ;) Will fix

jancborchardt:
Does this actually also fix **Does not respect autocomplete settings #19**? cc @ljo

skjnldsv:
> Does this actually also fix **Does not respect autocomplete settings #19**? cc @ljo

Yes :)

src/components/UserDiv.vue
skjnldsv: Weird, I thought I pushed it yesterday!
Let me push now

jotoeri:
Ok, so now i found Jan's bug, as well as a new one:
- [ ] The one, that Jan mentioned: When adding a share, but not changing/closing the search, the old search is still open, so the User/Group can be added a second time. So checking the currently selected items in multiselect would be good, but needs also possibility to uncheck again. Maybe add  shares only when changing search or closing the multiselect?
- [ ] Groups can not be deleted.

skjnldsv:
Fixed & pushed @jancborchardt @jotoeri 
Please do another review :)

src/components/shareDiv.vue
```diff
@@ -144,8 +148,8 @@ export default {
 	methods: {
 		removeShare(item) {
 			// Filter out the removed item
-			const users = this.userShares.filter(user => !(user.id === item.id && !item.isGroup))
-			const groups = this.groupShares.filter(group => !(group.id === item.id && item.isGroup))
+			const users = this.userShares.filter(user => user.shareWith !== item.shareWith && user.shareType !== item.shareType)
+			const groups = this.groupShares.filter(group => group.shareWith !== item.shareWith && group.shareType !== item.shareType)
```
jotoeri, line 152: The old logic was much easier to understand with the inverted brackets. And as it is now, all shares of the same type will be deleted, when deleting one.
I'd say just use the old code, and check for the item type to be the expected one.
```suggestion
			const users = this.userShares.filter(user => !(user.shareWith === item.shareWith && item.shareType === this.SHARE_TYPES.SHARE_TYPE_USER))
			const groups = this.groupShares.filter(group => !(group.shareWith === item.shareWith && item.shareType === this.SHARE_TYPES.SHARE_TYPE_GROUP))
```

skjnldsv: > And as it is now, all shares of the same type will be deleted, when deleting one.

Ah, weird! Nice catch!

skjnldsv:
Thanks Jonas! :hugs: 
So I restored the default search query persistance like it was, like it is on the current sidebar sharing. Was a mistake :)
Also fixed the delete of same entries! :rocket:

src/components/ShareDiv.vue
```diff
@@ -144,8 +148,8 @@ export default {
 	methods: {
 		removeShare(item) {
 			// Filter out the removed item
-			const users = this.userShares.filter(user => !(user.id === item.id && !item.isGroup))
-			const groups = this.groupShares.filter(group => !(group.id === item.id && item.isGroup))
+			const users = this.userShares.filter(user => !(user.shareWith === item.shareWith && user.shareType === this.SHARE_TYPES.SHARE_TYPE_USER))
+			const groups = this.groupShares.filter(group => !(group.shareWith === item.shareWith && group.shareType === this.SHARE_TYPES.SHARE_TYPE_GROUP))
```
jotoeri, line 152: Ahhhrgh, i think you want to check the item-share type here, no? A user in userShares is already of that type. 😭

skjnldsv: Dammit :crying_cat_face:

skjnldsv: Jesus, this day needs to end! :sweat:

jancborchardt:
Haha just tested and wanted to merge too ;) Works nicely now!

PR #347, opened 2020-05-05
Mandatory option on questions

### LINKED ISSUE #135: Mandatory/non mandatory questions
**Which Forms version are you running?**
1.0.3

**Nextcloud or ownCloud?:**
Nextcloud

17

### Steps to reproduce
1. create a form with non-mandatory fields
2. enter data into mandatory fields (*) leave non-mandatory fields empty

### Actual behaviour
3. cannot submit form without filling some non-mandatory fields with eg „n/a“ or some other non-null value

### Expected behaviour
Option to submit anyway or preset fields with some default text eg „n/a“

### Server configuration
**Where did you install Nextcloud from:**
Nextcloud App store 
**Operating system**:
Linux
**Web server:**
Apache2
**Database:**
MySql/MariaDB
**PHP version:**
7.2

jotoeri:
grafik

When editing the specific question
grafik

Closes #135

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -43,10 +43,13 @@
 				minlength="1"
 				:maxlength="maxQuestionLength"
 				required
-				@input="onInput"
-				@keyup="onTitleChange">
-			<h3 v-else class="question__header-title" v-text="text" />
+				@input="onTitleChange">
+			<h3 v-else class="question__header-title" v-text="computedText" />
 			<Actions v-if="!readOnly" class="question__header-menu" :force-menu="true">
+				<ActionCheckbox :checked="mandatory"
+					@update:checked="onMandatoryChange">
+					{{ t('forms', 'Mandatory') }}
+				</ActionCheckbox>
 				<ActionButton icon="icon-delete" @click="onDelete">
 					{{ t('forms', 'Delete question') }}
 				</ActionButton>
@@ -59,14 +62,10 @@
 </template>
 
 <script>
-import axios from '@nextcloud/axios'
-import debounce from 'debounce'
-import { generateUrl } from '@nextcloud/router'
-import { showError } from '@nextcloud/dialogs'
 import { directive as ClickOutside } from 'v-click-outside'
-
 import Actions from '@nextcloud/vue/dist/Components/Actions'
 import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
+import ActionCheckbox from '@nextcloud/vue/dist/Components/ActionCheckbox'
 
 export default {
 	name: 'Question',
@@ -78,19 +77,20 @@ export default {
 	components: {
 		Actions,
 		ActionButton,
+		ActionCheckbox,
 	},
 
 	props: {
-		id: {
-			type: Number,
-			required: true,
-		},
 		index: {
 			type: Number,
 			required: true,
 		},
 		text: {
 			type: String,
+			required: true,
+		},
+		mandatory: {
+			type: Boolean,
 			required: true,
 		},
 		edit: {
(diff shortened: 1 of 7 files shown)


skjnldsv:
Nice! :)
For the design though, we agreed on using the traditional asterisk `*` and having a `* Mandatory answer` after the description.
This is something web have been doing from the beginning of times as well as others forms app.

skjnldsv:
Regarding the checkbox, I'm not sure. It seems to take out some readability.
What about putting it as an ActionCheckbox in the menu?

jancborchardt:
Good stuff @jotoeri! :)

I’d agree with putting it as an action in the menu, also because on mobile it will cut off the question title too much.
Also agree with Barth’s other points

(And just a note, we generally never use italic/slanted text because it is difficult to read and not accessible.)

src/components/Questions/QuestionMultiple.vue
```diff
@@ -45,6 +45,7 @@
 						}"
 						:name="`${id}-answer`"
 						:required="isRequired(answer.id)"
+						:aria-required="isRequired(answer.id)"
```
skjnldsv, line 48: Should not be needed, required is enough.
Aria is a fallback when there is not the option :)

jotoeri: :D so what did you want to tell me with your comment then?

skjnldsv: That in case you were wondering how accessibility would work, the required property is the one to go for :see_no_evil: 
https://www.youtube.com/watch?v=g9Qff0b-lHk

skjnldsv: Also, I did say 

> TL;DR required attribute in the inputs will be read by screen readers

jotoeri: :D ok.
`TL;DR` is a shortcut for...?

skjnldsv: "too long didn't read"  :)
https://www.urbandictionary.com/define.php?term=tl%3Bdr

jotoeri:
Ok, so second go here - The mandatory information at description only shows up, if at least one question is marked as mandatory.

grafik

grafik

grafik

src/components/Questions/Question.vue
```diff
@@ -107,8 +106,35 @@ export default {
...
+		 * @returns {Boolean}
+		 */
+		isMandatory: {
+			get: function() {
+				return this.mandatory
+			},
+			set: function(mandatory) {
+				this.$emit('update:mandatory', mandatory)
+			},
+		},
```
skjnldsv, line 121: I would prefer having the same syntax as we do for `onTitleChange`, enableEdit... etc
Use `:checked="this.mandatory"` and watch for the event `update:checked="onMandatoryChange"` or something :)

src/components/Questions/Question.vue
```diff
@@ -107,8 +106,35 @@ export default {
...
+		/**
+		 * Extend text with asterisk if question is mandatory
+		 * @returns {Boolean}
+		 */
+		computedText() {
+			if (this.mandatory) {
+				return this.text + ' *'
+			} else {
+				return this.text
+			}
```
skjnldsv, line 132:
```suggestion
			}
			return this.text
```

src/components/Questions/Question.vue
jotoeri: I tried, but didn't get it to work. So this was the example i found used everywhere on server^^
But seems like i found the error now on second try.^^

jotoeri:
@jancborchardt Design of the information text is also good for you?

PR #351, opened 2020-05-06
Revert topbar changes

skjnldsv:
Revert changes
Like the Talk standard.

Capture d’écran_2020-05-06_13-03-33

### DIFF
--- a/src/components/TopBar.vue
+++ b/src/components/TopBar.vue
@@ -51,15 +51,9 @@ $top-bar-height: 60px;
 
 	button {
 		cursor: pointer;
-		min-height: 44px;
 
 		// Fix button having too little spacing left and right of text
-		&:not(.button-small) {
-			padding-left: 16px;
-			padding-right: 16px;
-		}
-
-		&.button-small {
+		&:not(:first-child) {
 			width: 44px;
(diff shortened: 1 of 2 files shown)


jotoeri:
- Well, if we change back here, then let's do everything alright, to adapt all the button-heights to a common height.
- The `&:not(.button-small)` change in top-bar was done by Jan as his design-review. You're sure you want to revert that?
- Will we never use a second large button on the top-bar? I tried that already once before on dev and the top-bar was not very handy with this not-first.

skjnldsv:
> * Will we never use a second large button on the top-bar? I tried that already once before on dev and the top-bar was not very handy with this not-first.

a second vue Actions you mean?

skjnldsv:
> * Well, if we change back here, then let's do everything alright, to adapt all the button-heights to a common height.

Design specs from talk are like those, regular buttons with text are server styling, icon-only button 44x44.
Like it is with this pr

skjnldsv:
> * The `&:not(.button-small)` change in top-bar was done by Jan as his design-review. You're sure you want to revert that?

I'm sure, but Jan might not, let's see with him.

PR #362, opened 2020-05-07
Fix export-button

jotoeri:
Now with small top-bar i would suppose to make the export button the same height. ;)

grafik

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -39,8 +39,8 @@
 
 		<header v-if="!noSubmissions">
 			<h2>{{ t('forms', 'Responses for {title}', { title: form.title }) }}</h2>
-			<div>
-				<button id="exportButton" @click="download">
+			<div class="response_actions">
+				<button class="response_actions__export" @click="download">
 					<span class="icon-download" role="img" />
 					{{ t('forms', 'Export to CSV') }}
 				</button>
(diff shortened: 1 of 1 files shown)


skjnldsv:
I thought I pushed this already :(

src/views/Results.vue
```diff
@@ -39,7 +39,7 @@
 
 		<header v-if="!noSubmissions">
 			<h2>{{ t('forms', 'Responses for {title}', { title: form.title }) }}</h2>
-			<div>
+			<div id="responseActions">
```
skjnldsv, line 42: Please use the BEM format and no id :)
```suggestion
			<div class="response_actions">
```

jotoeri:
Based on current master - maybe you wanted on topbar-branch, but it got lost?

src/views/Results.vue
```diff
@@ -39,7 +39,7 @@
...
 				<button id="exportButton" @click="download">
```
skjnldsv, line 43:
```suggestion
				<button class="response_actions__export" @click="download">
```

src/views/Results.vue
```diff
@@ -234,9 +234,13 @@ h2 {
 	white-space: nowrap;
 }
 
-#exportButton {
-	width: max-content;
-	padding: 13px 16px;
-	margin-left: 16px;
+#responseActions {
```
skjnldsv, line 237:
```suggestion
.response_actions {
```

src/views/Results.vue
```diff
@@ -234,9 +234,13 @@ h2 {
...
+	display: flex;
+	align-items: center;
+
+	#exportButton {
```
skjnldsv, line 241:
```suggestion
	&__export {
```

src/views/Results.vue
jotoeri: Not my change, but yes, your're right.

jotoeri:
Ah - just added you as co-author now... ;)

PR #363, opened 2020-05-07
Include more sharing links

jotoeri:
Fixes
> - [x] Not clear how to get share link, needs to be in sidebar next to radio option too, and ideally in responses view too when empty

- Included a link on EmptyContent in results
- Included a link in Sidebar
- Changed the Clipboard-package to fix closing of Actions-Menu, when copying link (now remains open, showing the nice green checkmark). [Found that package here](https://vuejsexamples.com/a-clipboard-library-with-vue-js/)

grafik
grafik
grafik

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2141,9 +2141,9 @@
 			}
 		},
 		"ansi-regex": {
-			"version": "3.0.0",
-			"resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.0.tgz",
-			"integrity": "sha1-7QMXwyIGT3lGbAKWa922Bas32Zg=",
+			"version": "2.1.1",
+			"resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
+			"integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
 			"dev": true
 		},
 		"ansi-styles": {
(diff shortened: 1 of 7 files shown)


skjnldsv:
Ah, please use the Actions menu on the sidebar then :)
It's dedicated for those uses exactly :wink: 
https://nextcloud-vue-components.netlify.app/#/Components/App%20containers?id=appsidebar-1

```
	<AppSidebar>
	...
		<template #secondary-actions>
			<ActionButton>
				{{ t('forms', 'Copy to clipboard') }}
			</ActionButton>
		</template>
	</AppSidebar>
```

jancborchardt:
> 

The "Copy share link" button should only show when the "Share link" radio button is selected, as otherwise it is not active, no? (And in that case it should also show next to that radio option, in the same line.

(Sidenote: German translation could be improved if you are on Transifex, shouldn’t say "Öffentlich" as it’s not public – maybe something like "Link teilen".)

jotoeri:
As for now, users don't get a message, if the form is shared to them. So the link is still necessary on all sharing-options.
But yes, for when notifications are running, it will be good to do sharing like this.

jotoeri:
This is when i use the secondary-actions-slot.
@jancborchardt Different design works for you?

grafik

jancborchardt:
Hmm, since this _should_ be the main way to get the share link, I’m not sure about it. Let’s get it in for now since the pull request improves it a lot, but it being icon-only and separated from the actual "Sharing" settings below, there will be issues.

jotoeri:
Well i didn't finish the work & commit yet. This is why i ask.
Alternative would be to merge as it was [above](https://user-images.githubusercontent.com/47433654/81272708-1efd7f80-904e-11ea-8267-061ef54a89ea.png)

skjnldsv:
> Let’s get it in for now since the pull request improves it a lot, but it being icon-only and separated from the actual "Sharing" settings below, there will be issues.

I'd guess we should also think about others app, Talk is also having the copy to clipboard there?

skjnldsv:
> This is when i use the secondary-actions-slot. (Less visible...?)
> @jancborchardt Different design works for you?
> 
> grafik

Looks good to me :)

jotoeri:
grafik
grafik

PR #366, opened 2020-05-09
Fix overlapping text of long answers

jancborchardt:
Ref https://github.com/nextcloud/forms/issues/296#issuecomment-623120930

The vertical spacing of course changes when switching to edit mode on longer answers. On shorter, one-line ones, it stays the same as before. :)

Before | After
-|-
Form answers broken overlapping | Fixed form answers short clip

### DIFF
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -283,14 +283,16 @@ export default {
 	position: relative;
 	display: inline-flex;
 	align-items: center;
-	height: 44px;
+	min-height: 44px;
 
 	.question__label {
 		flex: 1 1 100%;
 		// Overwrite guest page core styles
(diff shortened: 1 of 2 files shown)


jancborchardt:
@jotoeri fixed, thanks for the pointer! :) Used 11px since then it comes out at 44px height, and also used padding instead of margin so it helps for clickability.

jotoeri:
ERROR in
src/components/Questions/QuestionMultiple.vue
 292:9  ✖  Unknown word   CssSyntaxError

jotoeri:
Shit, sorry. Seems like my pull/merge was not completed.

skjnldsv:
Yeah, github is weirdly linting the QuestionMultiple file! :thinking: 
Isn't there a typo somewhere? Or an issue in the script/style tag open/close?

PR #367, opened 2020-05-09
Do not prefill form or question title for less confusion

jancborchardt:
People thought "New form" and "New question" were headings, not fields. This is fixed now by not prefilling them so the placeholders show, and fixing the wording.

Before | After
-|-
Forms before | Forms after

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -633,7 +633,7 @@ class ApiController extends Controller {
 			// Append Display Name
 			if (substr($submission['userId'], 0, 10) === 'anon-user-') {
 				// Anonymous User
-				$submission['userDisplayName'] = $this->l10n->t('anonymous user');
+				$submission['userDisplayName'] = $this->l10n->t('Anonymous response');
 			} else {
 				$userEntity = $this->userManager->get($submission['userId']);
 
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -148,7 +148,7 @@ class ApiController extends Controller {
 			$this->logger->debug('Could not find form');
 			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
 		}
-	
+
 		if (!$this->formsService->hasUserAccess($id)) {
 			$this->logger->debug('User has no permissions to get this form');
 			return new Http\JSONResponse([], Http::STATUS_FORBIDDEN);
@@ -171,7 +171,7 @@ class ApiController extends Controller {
 			16,
 			ISecureRandom::CHAR_HUMAN_READABLE
 		));
-		$form->setTitle('New form');
+		$form->setTitle('');
 		$form->setDescription('');
 		$form->setAccess([
 			'type' => 'public'
@@ -279,7 +279,7 @@ class ApiController extends Controller {
 			$this->logger->debug('Invalid type');
 			return new Http\JSONResponse(['message' => 'Invalid type'], Http::STATUS_BAD_REQUEST);
 		}
-		
+
 		try {
 			$form = $this->formMapper->findById($formId);
 		} catch (IMapperException $e) {
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -25,7 +25,7 @@
 		ref="navigationItem"
 		:exact="true"
 		:icon="icon"
-		:title="form.title"
+		:title="formTitle"
 		:to="{ name: 'edit', params: { hash: form.hash } }">
 		<template v-if="!loading" #actions>
 			<ActionLink
@@ -125,6 +125,17 @@ export default {
 		},
 
 		/**
+		 * Return form title, or placeholder if not set
+		 * @returns {string}
+		 */
+		formTitle() {
+			if (this.form.title) {
+				return this.form.title
+			}
+			return 'New form'
+		},
+
+		/**
 		 * Return the form share link
 		 * @returns {string}
 		 */
@@ -148,7 +159,7 @@ export default {
 
 	methods: {
 		async onDeleteForm() {
-			if (!confirm(t('forms', 'Are you sure you want to delete the form “{title}”?', { title: this.form.title }))) {
+			if (!confirm(t('forms', 'Are you sure you want to delete {title}?', { title: this.formTitle }))) {
 				return
 			}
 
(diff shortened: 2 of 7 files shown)


jancborchardt:
> Still needs a fallback for the navigation?

Right, will look into if I can figure that out, otherwise I'll have to hand over to you. :)

jancborchardt:
> Still needs a fallback for the navigation?

Fixed, and also fixed it in loading messages, on the responses page, and the submission page. :)

lib/Controller/ApiController.php
```diff
@@ -633,7 +633,7 @@ public function getSubmissions(string $hash): Http\JSONResponse {
 			// Append Display Name
 			if (substr($submission['userId'], 0, 10) === 'anon-user-') {
 				// Anonymous User
-				$submission['userDisplayName'] = $this->l10n->t('anonymous user');
+				$submission['userDisplayName'] = $this->l10n->t('Anonymous response');
```
jotoeri, line 636: That will produce a 'Response by Anonymous response' ? :D

grafik

src/components/Results/Submission.vue
```diff
@@ -23,9 +23,7 @@
 <template>
 	<div class="section submission">
 		<div class="submission-head">
-			<h3>
-				{{ t('forms', 'Response by {userDisplayName}', { userDisplayName: submission.userDisplayName }) }}
-			</h3>
+			<h3>{{ submission.userDisplayName }}</h3>
```
jotoeri, line 26: Ok, that clarifies aboves change. But like this you just get the Display-Name as header per submission...? I think i would leave the `response by` ....

jotoeri: ... if you would like to have sth. different for the Anonymous Responses, maybe just create a computed text here, that shows in case `anonymous response` and otherwise the 'response by'....?

jancborchardt: There was too much text before which obscured the actual info. The heading says "Responses for Form name", so in the individual entries it only needs to state the names.

lib/Controller/ApiController.php
jotoeri: Hm, ok. I just looked into. Looks ok...
Still i would use `Anonymous user` here? Thus it would match the other blocks...

src/components/AppNavigationForm.vue
```diff
@@ -124,6 +124,17 @@ export default {
...
 
+		/**
+		 * Return form title, or placeholder if not set
+		 * @returns {string}
+		 */
+		formTitle() {
+			if (this.form.title) {
+				return this.form.title
+			}
+			return 'New form'
```
jotoeri, line 135: Translation!?

lib/Controller/ApiController.php
jancborchardt: We try to avoid the word "user" for several reasons:
- It is very unspecific – user of what? In this case, "Anonymous response" is very clear about what this is.
- "user" is a very common word to refer to drug users

src/views/Results.vue
```diff
@@ -33,12 +33,12 @@
 		<TopBar>
 			<button @click="showEdit">
 				<span class="icon-forms" role="img" />
-				{{ t('forms', 'Back to form') }}
+				{{ t('forms', 'Back to questions') }}
```
jotoeri, line 36: Hm, not sure.

lib/Controller/ApiController.php
skjnldsv: Also Jan, this have nothing to do with this pr, please don't mix them up :)

src/components/AppNavigationForm.vue
skjnldsv, line 135: Translation

jancborchardt: WHOOOPS! Good catch, will fix! :+1:

skjnldsv: You were faster @jotoeri ;)

src/views/Results.vue
skjnldsv: I like it with "questions", give the focus to the important part :)

jancborchardt: This is basically how it is in Google Forms as well, with the 2 tabs "Questions" and "Responses". We are in the form, just in the responses part. The other part is questions.

src/components/AppNavigationForm.vue
jancborchardt: Fixed! Also in the other 2 files.

lib/Controller/ApiController.php
jotoeri: - Well - on all Blocks, the username is on top. So why is the username suddenly a response...? ;)
- I'm no native english speaker - but 'user' i would not primarily connect to drugs. :D Feels like it is used on the internet everywhere.

src/views/Results.vue
jancborchardt: That is, eventually this will be a connected button or tabs to switch between `( Questions | Responses )` anyway. :)

src/components/AppNavigationForm.vue
jancborchardt: Alerady fixed :)

src/views/Results.vue
jotoeri: Hm, ok. I'm fine with this change for now.

To do it connected...? I think it's really good as it is...
And still... I always get crazy if i see that Google F***. This app is independent and i would choose it just to get away of Google! Sure, Google does have the UI research, but we don't need to just copy Google! Take it particularly as inspiration, yes. But we can even make it better, if we think it's good! ;) ❤️

lib/Controller/ApiController.php
jancborchardt: > Also @jancborchardt, this have nothing to do with this pr, please don't mix them up :)

I know, I was working on further wording and just wanted to get these other fixes over with.

jancborchardt: > * I'm no native english speaker - but 'user' i would not primarily connect to drugs. :D Feels like it is used on the internet everywhere. And in german it's far away from beeing connected to drugs^^

See https://en.wikipedia.org/wiki/User – half of it is negatively connotated. In German it will be differently translated anyway. "Anonyme Antwort" is totally fine here too.
Another relevant read: https://spin.atomicobject.com/2019/06/11/there-is-no-user/

PR #368, opened 2020-05-09
Update window title

jotoeri:
- [x] The title is just "Forms - Nextcloud" but should be "Form name - Forms - Nextcloud" so you can see the relevant part even when the tab is short. Same for share view

### DIFF
--- a/src/mixins/WindowTitleMixin.js
+++ b/src/mixins/WindowTitleMixin.js
@@ -1,0 +1,32 @@
+/**
+ * @copyright Copyright (c) 2020 Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @author Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


src/mixins/WindowTitleMixin.js
```diff
@@ -0,0 +1,32 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export default {
+	methods: {
+		setWindowTitle(formTitle) {
```
skjnldsv, line 24: I think  I would prefer this as a function to import, rather than a mixin. Because it really doesn't need the full component context :)
@jotoeri

jotoeri: :D shit. I see, i should wait for two approvals even on small changes. :man_facepalming:
I'll change it.

skjnldsv: Usually we do aim for 2 in nextcloud, but there isn't too much devs here (3), so after release we could bump it to 2 to make sure ;)

PR #369, opened 2020-05-09
Focus title after form load

jancborchardt:
Makes it even more obvious that you can fill it, especially in combination with **Do not prefill form or question title for less confusion #367**.

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -47,6 +47,7 @@
 			<label class="hidden-visually" for="form-title">{{ t('forms', 'Form title') }}</label>
 			<input
 				id="form-title"
+				ref="title"
 				v-model="form.title"
 				:minlength="0"
 				:maxlength="maxStringLengths.formTitle"
@@ -229,7 +230,19 @@ export default {
 				this.errorForm = true
 			} finally {
 				this.isLoadingForm = false
+				if (this.form.title === '') {
+					this.focusTitle()
+				}
 			}
+		},
+
+		/**
+		 * Focus title after form load
+		 */
+		focusTitle() {
+			this.$nextTick(() => {
+				this.$refs.title.focus()
+			})
 		},
 
 		/**


src/views/Create.vue
```diff
@@ -229,9 +230,19 @@ export default {
 				this.errorForm = true
 			} finally {
 				this.isLoadingForm = false
+				this.focusTitle()
```
jotoeri, line 233: Maybe doing this only if the title is empty? This it was thought for, no?
It's quite pleasant & esp. less intrusive to not having focused, when opening an existing form...

skjnldsv: (Especially on mobile when keyboard open :)

jancborchardt: I thought that too, but decided for focusing it for accessibility: If you have a long form list otherwise you have to tab in first. This way you are directly inside the form.

jancborchardt:
> Doesn't `autofocus` line 56 already works for you?
> It does for me thinking
> https://caniuse.com/#feat=autofocus

No, doesn’t work in Firefox here at least, otherwise I wouldn’t have made this pull request. ;)

skjnldsv:
> No, doesn’t work in Firefox here at least, otherwise I wouldn’t have made this pull request. ;)

Maybe you would have kept this in the list and didn't noticed :p

jancborchardt:
> > No, doesn’t work in Firefox here at least, otherwise I wouldn’t have made this pull request. ;)
> 
> Maybe you would have kept this in the list and didn't noticed :p

I don’t get what you are saying – I saw it, saw that it apparently didn’t work, and tried to fix it. It works here now, and doesn’t break for you, correct? Then we can merge? :D

src/views/Create.vue
jancborchardt: > (Especially on mobile when keyboard open :)

Good point – will change that

jancborchardt: Fixed now to only focus the title when it’s empty.

PR #370, opened 2020-05-09
Set max-version to 20 for new development version of Nextcloud

jancborchardt:
Quick review @nextcloud/forms – this also might fix CI which said:

> Nextcloud was successfully installed
> App "Forms" cannot be installed because it is not compatible with this version of the server.

See https://github.com/nextcloud/forms/pull/367/checks?check_run_id=659162028

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -39,7 +39,7 @@
 	<screenshot>https://raw.githubusercontent.com/nextcloud/forms/master/screenshots/Results.PNG</screenshot>
 
     <dependencies>
-        <nextcloud min-version="17" max-version="19" />
+        <nextcloud min-version="17" max-version="20" />
     </dependencies>
 
 	<navigations>


skjnldsv:
Well, be careful when you release then! :)

jancborchardt:
> Well, be careful when you release then! :)

What do you mean by that?

skjnldsv:
That releasing the next beta will have to make sure max version is 19 :)
We don't know if 20 will have breaking changes yet :)

It was a reminder for me or you if you have the release keys ;)

jotoeri:
Hm, we could alternatively change PHPunit tests to server/stable19 branch and set our version back to 19? As far as i see, it would need three small changes in phpunit and needs then to be kept in track with server-maxversion, but if it helps to avoid 'bad' releases (even for future) ....
Your decision, @skjnldsv, if you want.  ;)

https://github.com/nextcloud/forms/blob/a3079b5008fc30a4a1a87054fc39a72522f99c66/.github/workflows/phpunit.yml#L23
Same on Line 89 & Line 159

skjnldsv:
Ideally we should test against all supported releases & dev, no need for all database tests, but we could add stable17, 18 & 19 to the mysql matrix?

jotoeri:
Hm, all supported versions, yes, sounds useful.
Not all databases i'm not sure about. I had already something that worked on Mysql but not on Pgsql. (Ok, it was strangely done anyways...^^) Maybe ask Roeland about this, if it would work to test only mysql?
But you're right, we should also see, it doesn't get too much. Only increases the chance that one test fails due to that strange Db-Lost thing.

jancborchardt:
We officially support MySQL/MariaDB, SQLite and PostgreSQL – we should make sure it's tested and working with all of those.
Do we know more about the issue with PostgreSQL?

jotoeri:
> Do we know more about the issue with PostgreSQL?

You mean the one, that i mentioned? That was just on my first try to migrate the database to v2. Nothing that ever got onto our master branch... ;)

PR #375, opened 2020-05-10
Correct fix window title

jotoeri:
Like this, @skjnldsv ? ;)

- Also changed to the new computed formTitle property, that Jan introduced. Can one of you check, if the title()-property in Create.vue really was deprecated? I looked through and didn't find a usage, but just to make sure i didn't miss anything.


jotoeri:
Jup, was for easier review. ;)
Squashed & pushed. 👍

PR #377, opened 2020-05-10
Fix redirects

jotoeri:
- [x] When deleting a form, the Main-Content (editing Form) is not updated.
(+ Renamed computed `hash` to `routerHash` for clarity)
- [x] When within a result-view and selecting another result-view from sidebar, the Result-Content is not updated.

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -139,6 +139,13 @@ export default {
 		},
 	},
 
+	watch: {
+		// Reload results, when form changes
+		hash() {
+			this.loadFormResults()
+		},
+	},
+
 	beforeMount() {
 		this.loadFormResults()
 		SetWindowTitle(this.formTitle)
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -32,7 +32,7 @@
 		</AppNavigation>
 
 		<!-- No forms & loading emptycontents -->
-		<AppContent v-if="loading || noForms || (!hash && $route.name !== 'create')">
+		<AppContent v-if="loading || noForms || (!routeHash && $route.name !== 'create')">
 			<EmptyContent v-if="loading" icon="icon-loading">
 				{{ t('forms', 'Loading forms …') }}
 			</EmptyContent>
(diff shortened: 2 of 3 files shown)


src/views/Results.vue
```diff
@@ -139,6 +139,13 @@ export default {
 		},
 	},
 
+	watch: {
+		// Reload results, when form changes
+		'form.hash'() {
+			this.loadFormResults()
+		},
```
skjnldsv, line 146: `form.hash` is not supposed to ever change in the logic :)
`form` does change. `hash` too. ;)
This syntax is confusing :)

```suggestion
		hash() {
			this.loadFormResults()
		},
```

jotoeri: You're right. Strange logic^^ 👍

jotoeri:
@jancborchardt , that one was not yet based onto #376. Seems to me like you experienced this bug here. I rebased now on current master, can you check again, if you still see this?

Your second thing is sth. different, i would say. It's connected to the AppNavigation Component and when it is selected. I'll write it into the bug-list...

src/views/Results.vue
jotoeri: Ahh, but now it is not working anymore... 🤔

jotoeri: What do you say about using `$route.hash`? That is in fact the intention behind..

jotoeri:
Ok, @skjnldsv i dived into the router-thing a bit to understand it more and pushed one more fixup now.

Is it that on the router you wanted to create the `props: true` to update the hash in the mixin? But accidentially mixed up boolean-mode vs. object-mode? So pushing the object `{ default: true }`, instead of pushing `{ hash: [anyhash] }`? (See [router-doc: modes](https://router.vuejs.org/guide/essentials/passing-props.html#boolean-mode))

With the last pushed fixup it works now on only watching `hash()`, which now has essentially the same effect as watching `$router.params.hash`, as the _param_ `hash` is now pushed to the _prop_ `hash` through the router.n :)

skjnldsv:
> is it that on the router you wanted to create the `props: true` to update the hash in the mixin? But accidentially mixed up boolean-mode vs. object-mode?

Nope, this make sure we only pass hash to the default router-view :)
So not the sidebar. https://router.vuejs.org/guide/essentials/passing-props.html#passing-props-to-route-components

But it's also fine like that sure :)

jotoeri:
Puhh, ok seems to have been a bit of both. The multi-component route needs the props-object to work, while the single-component route needs the single prop. Reverted the one on the edit route.
Now it should work and should be good to merge from my side.

PR #378, opened 2020-05-11
Clean loading results

jotoeri:
Now using formsService, as well as the Views-Mixin for storage.

### DIFF
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -78,7 +78,12 @@ class FormsService {
 		$this->userSession = $userSession;
 	}
 
-
+	/**
+	 * Load options corresponding to question
+	 *
+	 * @param integer $questionId
+	 * @return array
+	 */
 	public function getOptions(int $questionId): array {
 		$optionList = [];
 		try {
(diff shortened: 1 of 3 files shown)


src/views/Results.vue
```diff
@@ -117,14 +117,12 @@ export default {
...
 			loadingResults: true,
-			submissions: [],
-			questions: [],
 		}
 	},
 
 	computed: {
 		noSubmissions() {
-			return this.submissions && this.submissions.length === 0
+			return this.form.submissions && this.form.submissions.length === 0
```
skjnldsv, line 125:
```suggestion
			return this.form.submissions?.length === 0
```

lib/Service/FormsService.php
```diff
@@ -93,10 +93,10 @@ public function getOptions(int $questionId): array {
 		}
 	}
 
-	public function getQuestions(int $formId): array {
+	public function getQuestions(int $formId, bool $loadDeleted = false): array {
```
skjnldsv, line 96: phpdocs? :p

jotoeri: Haha, there were none before already! :D
But yes, can do that...

jancborchardt:
Which bug does it fix, or how to test @jotoeri? :)

jotoeri:
None, it's just a bit of code-cleanup in results.
So as long as results are still shown, deleted Questions are shown in Results if answered but not on edit, then it should work^^

jotoeri:
Sorry. Didn't change anything here, just accidentially pushed a separate commit. It's corrected again.

jotoeri:
Closing in favor of #386

PR #379, opened 2020-05-11
Fix Navigation active on results

jotoeri:
- [x] The active class on the navigation is cleared when you are in the response view, but should keep having the relevant form highlighted as active.

I just gave it a try, as i was on to learn new stuff about routes anyway. ;) I didn't find another way to tell the router-link a different active-path than the link, so i chose the form-root, here. Shouldn't interfer with the submission route, as this is done on php already, but will be interesting again, when we want to include the submission-view into the logged-in navigation...

I don't know if this is the proper or best way, so feedback is appreciated, @skjnldsv ;)

### DIFF
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -23,10 +23,9 @@
 <template>
 	<AppNavigationItem
 		ref="navigationItem"
-		:exact="true"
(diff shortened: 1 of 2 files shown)


jotoeri:
Rebased.

PR #380, opened 2020-05-11
Close navigation on mobile on new form

jotoeri:
- [x] On mobile, the navigation doesn’t close when you click on a form or create new

### DIFF
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -28,6 +28,7 @@
 			<AppNavigationForm v-for="form in forms"
 				:key="form.id"
 				:form="form"
+				@mobile-close-navigation="mobileCloseNavigation"
 				@delete="onDeleteForm" />
 		</AppNavigation>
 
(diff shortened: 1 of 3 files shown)


jotoeri:
Ah -  and edit: Ammended a very small fix, to center the emptyContent-Description-Text on mobile.

jotoeri:
Oups, seems like i read just half of the text.^^

jotoeri:
Still doesn't close on Click on Answers.

@skjnldsv :
- Feels like the ActionRouter does not properly emit the @click event. Can you have a look onto this?
- Maybe the whole process 'On mobile, close navigation on click' could be a property of ActionElements? Indeed, it's very specific, but might be very commonly useful...?

jancborchardt:
> * Maybe the whole process 'On mobile, close navigation on click' could be a property of ActionElements? Indeed, it's very specific, but might be very commonly useful...?

Might be good – if it’s possible to set per action. For example the navigation should be closed when clicking "Responses", but it should stay open when clicking "Copy share link" and "Delete form".

jotoeri:
> Feels like the ActionRouter does not properly emit the @click event. Can you have a look onto this?

https://github.com/nextcloud/nextcloud-vue/pull/1097

jancborchardt:
> nextcloud/nextcloud-vue#1097

That one is now merged :tada:

jotoeri:
Still needs a release, but yes 👍 :)

PR #386, opened 2020-05-14
Fix result order

### LINKED ISSUE #381: Order of answer incoherent with order of questions
**Describe the bug**
[Not sure if bug or feature request]
After questions are created, their order can be manually adjusted by using drag and drop, which is very useful. However, in the answer page, the order is not adapted accordingly.

**To Reproduce**
Create a new form, with two questions.
Move the second question above the first one.
Fill in your form and go to the answers page.

**Expected behavior**
The order of answers should be the same as the order of questions.

**Screenshots**
Form creation:
Capture d’écran 2020-05-11 à 21 57 04

Answers:
Capture d’écran 2020-05-11 à 21 57 20

- Nextcloud-Version: 18.0.4
- Forms-Version: 2.0.0-beta2

 - OS: MacOs Mojave 10.14.6
 - Browser Firefox 
 - Version 76
(But same situation on other devices with different configurations)

jotoeri:
Fixes #381 

Now showing submissions in current order, not showing deleted questions. Same applies to export.

As therefore some changes of the results-cleanup are not needed anymore, i included this commit here, while changing the respective things.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -649,16 +649,8 @@ class ApiController extends Controller {
 			$submissions[] = $submission;
 		}
 
-		// Load question-texts, including deleted ones.
-		try {
-			$questionEntities = $this->questionMapper->findByForm($form->getId());
-		} catch (DoesNotExistException $e) {
-			//handle silently
-		}
-		$questions = [];
-		foreach ($questionEntities as $questionEntity) {
-			$questions[] = $questionEntity->read();
-		}
+		// Load currently active questions
+		$questions = $this->formsService->getQuestions($form->getId());
 
 		$response = [
 			'submissions' => $submissions,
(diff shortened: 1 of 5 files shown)


jancborchardt:
Also cc @theosche for review as it fixes an issue you reported. :)

jotoeri:
> It does work, but only after a refresh. Ideally it would also work directly, but not sure how difficult that is. :)

@rullzer, that sounds a bit like the one we had already here: https://github.com/nextcloud/forms/issues/229#issuecomment-601911799 ?

jancborchardt:
This actually fixes the export issue which makes it so that any time there are 2 or more answers checked on a checkbox question, everything but the first one is duplicated in the export.

We need to review, merge and deploy this on our own instance, as otherwise the data from our own surveys is wrong. cc @rullzer @jospoortvliet @skjnldsv

PR #388, opened 2020-05-14
Delete empty options from Db

jotoeri:
> - [x] When emptying an answer and leaving the edit mode, it remove the empty one (on purpose) but not removed from the database

### DIFF
--- a/src/components/Questions/AnswerInput.vue
+++ b/src/components/Questions/AnswerInput.vue
@@ -114,21 +114,8 @@ export default {
 
 			// Dismiss delete key action
 			e.preventDefault()
-			const answer = Object.assign({}, this.answer)
-			const index = this.index
 
-			if (!answer.local) {
-				// let's not await, deleting in background
-				axios.delete(generateUrl('/apps/forms/api/v1/option/{id}', { id: this.answer.id }))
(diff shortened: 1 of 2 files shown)


jancborchardt:
(Can’t review here as it’s mostly code review, cc'ing other Vue people.)

jotoeri:
Well we need to delete them somehow. If you don't do here, you will have an inconsitency between Db and UI, which falls back, the moment you reload the form.
The case, that this gets triggered on very many answers within one moment is very unlikely. If people want to delete the answers, they will trigger the delete (also from Db) anyways by delete button or backspace-key (not just clearing every answer-text manually and leaving the field there as empty answer). The case that there are empty answers left, that get deleted within one moment won't be used on really many answers, i think.

jancborchardt:
Can you recheck @skjnldsv? :)

skjnldsv:
Ah, I missed the original point, sorry.
I was thinking about deleted options that we keep so that submitted answers still have the original option shown. My bad!

I'm not fond of doing a function call within a filter.
But it's ok :)

jotoeri:
Second review necessary? Or just go ahead?

skjnldsv:
Let's go

PR #389, opened 2020-05-14
Question-specific placeholders

jotoeri:
> - [x] Default question title is not great, put more details on chat question title type "New multiple choice question"

@jancborchardt Wording in `src/models/AnswerTypes.js` is up to you, i just included placeholders. ;)

### DIFF
--- a/src/models/AnswerTypes.js
+++ b/src/models/AnswerTypes.js
@@ -32,24 +32,38 @@ import QuestionMultiple from '../components/Questions/QuestionMultiple'
  * @property {string} long
  */
 export default {
+	/**
+	 * Specifying Question-Models in a common place
+	 * Further type-specific parameters are possible.
+	 * @prop component The vue-component this answer-type relies on
+	 * @prop icon The icon corresponding to this answer-type
+	 * @prop label The answer-type label, that users will see as answer-type.
+	 * @prop validate *optional* Define conditions where this question is not ok
+	 *
+	 * @prop titlePlaceholder The placeholder users see as empty question-title in edit-mode
+	 * @prop createPlaceholder *optional* The placeholder that is visible in edit-mode, to indicate a submission form-input field
+	 * @prop submitPlaceholder *optional* The placeholder that is visible in submit-mode, to indicate a form input-field
+	 */
 
 	multiple_unique: {
 		component: QuestionMultiple,
 		icon: 'icon-answer-multiple',
 		label: t('forms', 'Multiple choice'),
-		titlePlaceholder: t('forms', 'Multiple choice question title'),
-		unique: true,
-		// Define conditions where this questions is not ok
 		validate: question => question.options.length > 0,
+
+		titlePlaceholder: t('forms', 'Multiple choice question title'),
+
+		// Using the same vue-component as multiple, this specifies that the component renders as multiple_unique.
+		unique: true,
 	},
 
 	multiple: {
 		component: QuestionMultiple,
 		icon: 'icon-answer-checkbox',
 		label: t('forms', 'Checkboxes'),
-		titlePlaceholder: t('forms', 'Checkbox question title'),
-		// Define conditions where this questions is not ok
 		validate: question => question.options.length > 0,
+
+		titlePlaceholder: t('forms', 'Checkbox question title'),
 	},
 
 	short: {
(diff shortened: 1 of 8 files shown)


jotoeri:
And ammending this one, as it is very similar imo. So would be the non-quick-fix variant. ;)

> - [x] Submission view: "People can enter a short answer" for short/long text makes no sense, needs to be differentiated. In create view: "People can enter a short answer" and in submit view "Enter a short answer" (we can do a quick-fix to make it "Short answer text" and disable the field?) - @jancborchardt for the quick fix

src/models/AnswerTypes.js
```diff
@@ -54,12 +56,20 @@ export default {
...
 	},
 
 	long: {
 		component: QuestionLong,
 		icon: 'icon-answer-long',
 		label: t('forms', 'Long text'),
+
+		titlePlaceholder: t('forms', 'Long text question title'),
+		submitPlaceholder: t('forms', 'Enter a long text'),
+		createPlaceholder: t('forms', 'People can enter a long text'),
```
skjnldsv, line 72: If we complicate this part, it would be nice to have a comment header to explain what all of those are on this file I think :)
For future references and new contributors?

src/views/Create.vue
```diff
@@ -87,7 +87,7 @@
 					v-for="(question, index) in form.questions"
 					ref="questions"
 					:key="question.id"
-					:model="answerTypes[question.type]"
+					:answer-type="answerTypes[question.type]"
```
skjnldsv, line 90: Why this?
model is usually a very specific uses js-keyword :)

jotoeri: Might be usual and ist understandable here from outside the question, as it shows, that the question models an answer-type. From within the questions, it is quite confusing to get the parameters (e.g. placeholder) from a model. The js-file is called `answertypes`, so if i get the placeholder from `answertype.placeholder` i directly know, what it relies on, instead of having to search which parameter is anywhere given to `model`.
"For future references and new contributors" ;)

src/models/AnswerTypes.js
jotoeri: See commit

src/views/Create.vue
skjnldsv: I would rather be consistent on the coding naming keywords.
newcomers need to learn how to call things if they don't know it already.
:wink:

jotoeri: Well - then leave the newcomers out of my comment. The first part still applies for everybody. Having a file, that contains and is called `answerTypes`, it is just much more logical having a variable, that contains the specific object out of that file, called `answerType`, than `model`.

skjnldsv: Don't have time to argue here. Let's please revert it to what I put before and move on. :rocket:

jotoeri: Wtf. "Just do what i say"?

jancborchardt: Would it help to get a 3rd opinion from one of the other Vue people? :) Maybe @juliushaertl @ChristophWurst can help resolve this.

Please both understand each other’s situation: Yes it’s not good that the pull request is already open since 26 days, but also Barth was out sick so that’s not nice either. :) Let’s try to keep things civil and productive.

ChristophWurst: What you are discussing here is indeed [one of the two hardest things in computer science](https://www.martinfowler.com/bliki/TwoHardThings.html).

On a serious note I do not know any of this code, so I can try to be neutral on this. A *model* in CS terms is typically is a limited representation of something from the real world. So this could be *anything*. Hence I do agree that for me as someone who sees the code for the very first time, a more specific name makes it easier to process the code without having to look up all the details.

Again, good naming is really hard. Don't worry if you don't think it's right the first time. You can always iterate over the code and improve it incrementally. Nothing is set in stone :v:

juliushaertl: Agree with @ChristophWurst just from reading the snippet here i would expect a `model` property to take the question object (since it is passed to the Question component), but since the answer types are passed, having a named property seems like the more logic approach.

skjnldsv: > Wtf. "Just do what i say"?

It was more about you changing my code :shrug: 
But like I said, no time to argue, I'm not on forms lately

jancborchardt: As said I can understand the frustration on both sides. Thanks to Julius and Christoph for helping resolve the deadlock, we just go with one thing and can adjust in the future. :) → Merging

:heart: to you all

PR #391, opened 2020-05-15
Replace userId with userSession->UID

jotoeri:
Cleaning a Code-Todo, using the userSession instead of userId.
Old code was a bit better to read imo. @skjnldsv ;)

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -74,8 +74,8 @@ class ApiController extends Controller {
 	/** @var IL10N */
 	private $l10n;
 
-	/** @var IUserSession */
-	private $userSession;
+	/** @var IUser */
+	private $currentUser;
 
 	/** @var IUserManager */
 	private $userManager;
@@ -85,7 +85,6 @@ class ApiController extends Controller {
 
 	public function __construct(string $appName,
 								IRequest $request,
-								$userId, // TODO remove & replace with userSession below.
 								IUserSession $userSession,
 								IUserManager $userManager,
 								FormMapper $formMapper,
@@ -98,8 +97,6 @@ class ApiController extends Controller {
 								FormsService $formsService) {
 		parent::__construct($appName, $request);
 		$this->appName = $appName;
-		$this->userId = $userId;
-		$this->userSession = $userSession;
 		$this->userManager = $userManager;
 		$this->formMapper = $formMapper;
 		$this->questionMapper = $questionMapper;
@@ -111,6 +108,8 @@ class ApiController extends Controller {
 		$this->logger = $logger;
 		$this->l10n = $l10n;
 		$this->formsService = $formsService;
+
+		$this->currentUser = $userSession->getUser();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


lib/Controller/ApiController.php
```diff
@@ -207,7 +205,7 @@ public function updateForm(int $id, array $keyValuePairs): Http\JSONResponse {
 			return new Http\JSONResponse([], Http::STATUS_BAD_REQUEST);
 		}
 
-		if ($form->getOwnerId() !== $this->userId) {
+		if ($form->getOwnerId() !== $this->userSession->getUser()->getUID()) {
```
skjnldsv, line 208: What if not logged in? :thinking:

lib/Controller/ApiController.php
```diff
@@ -98,7 +97,6 @@ public function __construct(string $appName,
 								FormsService $formsService) {
 		parent::__construct($appName, $request);
 		$this->appName = $appName;
-		$this->userId = $userId;
```
jancborchardt, line 101: Just reinstate that, but as 
`$this->userSessionId = userSession->getUser()->getUID();`
to make it easier to read again? :)

lib/Controller/ApiController.php
jotoeri: :D You tell me - userId was, what Roeland wrote here - you wanted to change to userSession, as you said it's cleaner... What does userSession return, if not logged in? These comparisons should just return true to return the forbidden-status within the if...

skjnldsv: getUser returns null I guess?

jotoeri: Yep - So it should work, as it has to...
grafik

grafik

kesselb: Question: The user must be logged in to use any of the ApiController methods?

kesselb: https://github.com/nextcloud/mail/blob/5f744699ca36aefa919668cb717ae15c045be89a/lib/Controller/AliasesController.php#L41-L50

That ($currentUser) might be a bit cleaner and $this->currentUser->getUID() is easier to read.

jotoeri: > Question: The user must be logged in to use any of the ApiController methods?

Mostly, yes. Currently only the form-owner is allowed to see responses & edit forms. Plans to extend this do exist. `insertSubmission()` is also usable without beeing logged in, but uses the currentUser for storing userId.

> That ($currentUser) might be a bit cleaner and $this->currentUser->getUID() is easier to read.

Hm, yes somehow similar to what Jan said. Did that now, and even changed on FormsService for similar usage.

jotoeri:
Just pushed again for checks to run.

jotoeri:
Ok, @jancborchardt so i think we can go here? It's just a bit of standardized code-replacement. I just systematically went through all functions, functionality seems te be working everywhere.

jancborchardt:
> Ok, @jancborchardt so i think we can go here? It's just a bit of standardized code-replacement. I just systematically went through all functions, functionality seems te be working everywhere.

Sounds good to me. Can you check out the file conflict? Then afterwards we can merge. :)

jotoeri:
:D that was the logger now.^^

PR #392, opened 2020-05-15
Fix key-exists error

jotoeri:
Just recognized, there was an error on logger since a few days...

grafik


jancborchardt:
@jotoeri how can you trigger the error?

jotoeri:
Just change anything on form properties, except access. So title or description.

PR #393, opened 2020-05-16
Use cancelable request

jotoeri:
Including the planned cancelable request on fetchFullForm.

Also solves bug
> - [x] Fastly changing forms a few times, somwhen runs into an error, that overwrites one form with another one, and logs TypeError: "this.form.form is undefined" on console

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -130,6 +130,7 @@ import Actions from '@nextcloud/vue/dist/Components/Actions'
 import AppContent from '@nextcloud/vue/dist/Components/AppContent'
 
 import answerTypes from '../models/AnswerTypes'
+import CancelableRequest from '../utils/CancelableRequest'
 import EmptyContent from '../components/EmptyContent'
 import Question from '../components/Questions/Question'
 import QuestionLong from '../components/Questions/QuestionLong'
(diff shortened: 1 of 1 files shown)


jancborchardt:
This needs Vue reviewers :) @skjnldsv @juliushaertl @ChristophWurst

juliushaertl:
I'm not a big fan on having the whole fetch logic inside of the component itself, having state management and just obtaining the form from there would be a bit cleaner imo, but I have no idea about the form code so @skjnldsv can probably judge better than me. :wink:

jotoeri:
@ma12-co Added you for review, as you wrote the cancelable request?! Is this proper usage here?

@juliushaertl In fact @skjnldsv newly wrote this fetch logic here, so i can't tell you why he did. But we load the different forms on changes of the vue-router and can change their data. So php (&state-management?) is not involved in between. Would mean we would need to provide all possible data in state-management, and when writing changes to db, the states needed to be adapted each time...?!

jotoeri:
Nice. Thank you both for reviewing! :)

skjnldsv:
> I'm not a big fan on having the whole fetch logic inside of the component itself, having state management and just obtaining the form from there would be a bit cleaner imo, but I have no idea about the form code so @skjnldsv can probably judge better than me.

Yeah, no need for a full state here as we only keep a stupid array oj objects.
I thought it would clearly be overpowered to have a full vuex app framework implemented.

I guess we can still split things into services later on as cleanups :)

PR #394, opened 2020-05-16
Fix some small MultipleInput issues

jotoeri:
_This one is now based onto the delete empty options PR #388 , therefore i marked it as draft for now. I will rebase to master, when the other one is done._

Some small issues somehow nicely fall together in this commit.
- [x] When fastly deleting answers by keyboard, a few answers get deleted, then the backspace suddenly applies to the browser, therefore switching pages
- Navigating from Title to create options is now possible by tab-keys (Didn't find the issue?!)
- Pressing Backspace to delete an option now focuses on the element above, wich is imo the logical order to jump to the line above.
- Deleting the first Option Element is now prevented, just gets emptied. Still gets deleted, when leaving edit mode, but pressing backspace on the single empty answer twice does not completely navigate away anymore, as the focus stays on the text-field.

### DIFF
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -73,7 +73,8 @@
 					:maxlength="maxStringLengths.optionText"
 					minlength="1"
 					type="text"
-					@click="addNewEntry">
+					@click="addNewEntry"
+					@focus="addNewEntry">
 			</li>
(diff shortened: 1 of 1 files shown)


skjnldsv:
still a draft?

jotoeri:
Content is done, but still depends onto an unmerged PR.

PR #396, opened 2020-05-17
Fix saving options on fast proceed

jotoeri:
> - [x] Sometimes last response option is not saved when using Tab-key to advance to next field

Fastly entering answers, didn't save to Db, as the same debounce applied for all answers. Having a separate debounce for each answer solves that.

### DIFF
--- a/src/components/Questions/AnswerInput.vue
+++ b/src/components/Questions/AnswerInput.vue
@@ -59,6 +59,11 @@ export default {
 	data() {
 		return {
 			queue: new PQueue({ concurrency: 1 }),
+
+			// As data instead of Method, to have a separate debounce per AnswerInput
+			debounceUpdateAnswer: pDebounce(function(answer) {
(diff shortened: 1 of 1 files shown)


jancborchardt:
Hmm, both issues still exist, here in a gif.

Basically, whenever the placeholder of the field is "Add a new answer" and not "Answer number X", you know that the text you write will not be saved.

Peek 2020-05-18 14-03

jotoeri:
@jancborchardt I think the one you have in mind is the one solved in #394. Maybe i projected your issue-text then onto another one.

This one here solves that if fastly proceeding forward by pressing enter and continue typing within the debounced 500ms, then the previous answer was not saved.

jancborchardt:
> @jancborchardt I think the one you have in mind is the one solved in #394. Maybe i projected your issue-text then onto another one.

You’re right, just found that one, will test. :)

PR #397, opened 2020-05-17
Fix remove empty questions on submit

jotoeri:
- [x] Questions with empty titles are possible and will be shown, should be pointed out

That one was implemented already, just with the problem, that empty strings are interpreted as false already...


jotoeri:
👍 You include that into the design-review? ;)

jancborchardt:
> +1 You include that into the design-review? ;)

Yup, added to https://github.com/nextcloud/forms/issues/296 :)

PR #402, opened 2020-05-20
Sort Navigation newest forms first

### LINKED ISSUE #401: Sort Navigation
> The fact that new forms are created on the bottom is always a bit strange

jotoeri:
Now showing newest forms on top and adding new forms on top of the forms-list.

Closes #401


jotoeri:
Hm, we don't store such a date yet...

PR #409, opened 2020-05-24
Show MultipleInput icons in Edit-Mode

jotoeri:
Quite some css fun, as from server-side the visible input-items are just dummy-css fields, while the real input is hidden. Therefore i now included some dummy-div and just copied the server-style to show the same elements.

As i had to adjust a few things on the input-fields for alignment, i had to adjust that on all question-types and recognized, that the `.question__input[type="text"]` was located in three places. To have it in one place, i shifted it to the question-element, but unfortunately now needed there an unscoped css. But as we use the BEM-Model on this, it should be ok, i think.

Unfortunately on Firefox, the line-height still changes a bit between showing & editing a question, as firefox overrides the line-height with some own style. That seems a bit like a firefox-bug to me. :/ On Edge it works perfectly.

Solves
> - [x]  Still show checkbox when editing

And therefore also finishes this one? @jancborchardt 
> - [x] In edit mode, not enough difference between choices types:

grafik
grafik


jotoeri:
> There’s a tiny 1px vertical shift of the answers, and also a slight size change of the radio buttons and checkboxes, but we can also look at that afterwards. :)

You're working on firefox, no? Thats why i wrote that part above: 😉 
> Unfortunately on Firefox, the line-height still changes a bit between showing & editing a question, as firefox overrides the line-height with some own style. That seems a bit like a firefox-bug to me. :/ On Edge it works perfectly.

jancborchardt:
Ooops, missed that part. :D Also just tested in Chrome and there’s a similar issue though:
Firefox | Chrome
-|-
forms switch firefox|forms switch

jotoeri:
Argh. On very large size i found it on edge now, too.
Could you have a look onto the css? I don't find the clue at the moment... :/

jancborchardt:
> Could you have a look onto the css? I don't find the clue at the moment... :/

No worries, I’ll look into this. :) Should be something which is fixable with some slight padding-top for example.

jancborchardt:
I tried some things for fixes but even after `npm run dev` it doesn’t show when testing – does it work for you?

In any case, there seem to be a lot of unrelated changes which could create the layout issues, also different issues in different browsers. I would recommend we pick only the absolutely necessary parts from this pull request first, to pin down exactly where the issue occurs?

jotoeri:
Ok, so. This version is now working no MS Edge, as well as Opera on 100% Window-Size. Increasing this Window-zoom, it breaks. On Firefox, it still changes.
Taking the version two commit above, https://github.com/nextcloud/forms/pull/409/commits/7eedadacb089baaccc5aecef9ca61c6750d5f2ba, it works perfectly on firefox, but not on Edge/Opera.

To be honest - i'm done with this now. Took me already two nights. @jancborchardt if you're motivated, just keep on trying. The commit-history now has the first commit that implements the basic functionality, then i documented each small step i changed on css.

jancborchardt:
@skjnldsv could you check this? :)

jotoeri:
:D ok. Just thought about these days, it is some kind of rounding-issue. Now found a solution, that works at least on 100%-view on firefox and edge. Still a strange and unstable thing...

skjnldsv:
> Now found a solution, that works at least on 100%-view on firefox and edge. Still a strange and unstable thing...

Please add lots of comments for the css then :p

src/components/Questions/QuestionMultiple.vue
```diff
@@ -282,15 +289,47 @@ export default {
...
+	}
+
 	.question__label {
 		flex: 1 1 100%;
 		// Overwrite guest page core styles
 		text-align: left !important;
-		padding: 11px 0 11px 30px;
+		padding: 6.5px 0 0 30px;
+		line-height: 22px;
+		min-height: 34px;
```
jancborchardt, line 322: Yeah, maybe here a comment would be good especially because of the half value for the padding. :) Just to make sure someone else doesn’t edit it afterwards. Also the 18 commits could be squashed to less, right @jotoeri?

And @skjnldsv could you give it a review?

jotoeri: Sure, yes. 👍  (for both.^^)

jotoeri: Done... 👍

jancborchardt: Good stuff, thanks! Last check @skjnldsv @juliushaertl @ma12-co maybe? :)

jancborchardt:
@jotoeri could you check out the conflicts arisen from the other merges? :) Thanks!

PR #412, opened 2020-05-26
Add favicons

jancborchardt:
It’s better to create these icons manually to ensure they are present even when the imagick PHP extension is missing. Same as for the Notes app at https://github.com/nextcloud/notes/pull/522


img/app.svg
```diff
@@ -1,8 +0,0 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32">
```
jotoeri, line 1: Referred to in info.xml

jancborchardt: Thanks for the catch – changed that to forms.svg :)

jancborchardt:
php7.4-mysql-master keeps failing but it seems unrelated to this?

PR #413, opened 2020-05-26
Avoid Submit on Enter

jotoeri:
> - [x] Submission view: Pressing Enter in short text submits whole form (if everything is filled), unexpected

Now just heads over to next input on Enter. As the last input is the submit-button, submit on enter still is possible like this.
Just Question-type longtext shows different behaviour, but i think would be dumb to navigate away there on enter.

### DIFF
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -38,7 +38,9 @@
 			</header>
 
 			<!-- Questions list -->
-			<form v-if="!loading && !success" @submit.prevent="onSubmit">
+			<form v-if="!loading && !success"
+				ref="form"
+				@submit.prevent="onSubmit">
 				<ul>
 					<Questions
 						:is="answerTypes[question.type].component"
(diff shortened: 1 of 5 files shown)


jotoeri:
Hm, even on long text fields?
On normal fields i got it by changing my changes. But on long text it even didn't fire before on ctrl+enter. So this would now need special handling...

jancborchardt:
Yeah, usually any web form can be submitted by hitting Ctrl-Enter anywhere in the form. Also e.g. any reply field like in Nextcloud Talk, Twitter, etc.

jotoeri:
Should also work on Longtext now.

src/views/Submit.vue
```diff
@@ -50,9 +52,12 @@
 						:index="index + 1"
 						:max-string-lengths="maxStringLengths"
 						v-bind="question"
-						:values.sync="answers[question.id]" />
+						:values.sync="answers[question.id]"
+						@on-keydown-enter="onKeydownEnter"
+						@on-keydown-ctrl-enter="onKeydownCtrlEnter" />
```
skjnldsv, line 57: I'm not fond of those event names :)
Can we just use @keydown.ctrl.enter if we emit the same event back ?

jotoeri: Just tried this, see fixup. Doesn't work.

jotoeri: we could maybe just do with dashes without the 'on' - So `@keydown-ctrl-enter`

skjnldsv: > Just tried this, see fixup. Doesn't work.

Arf, too good to be true :wink:

skjnldsv: Ah no!
You cannot do `this.$emit('keydown.enter', event)` with modifiers.
Just emit the keydown event, `this.$emit('keydown', event)`, the event already contains the necessary info for vue to detect if this is an enter key event or now (that's how it works initially)

jotoeri: Ah, nice. That's it. Now it works 👍

PR #414, opened 2020-05-26
Fix expiration editable

jotoeri:
> -[x] Disable typing on expiration field. When manually changing the text, expires falls back to 1970.


jancborchardt:
Isn’t this even something that should be fixed in the datepicker component? That is you should be able to also change the numbers of the date and time with typing, but currently you can type _text_ in the field, which makes no sense.

PR #415, opened 2020-05-26
Enh/invalid warning

jotoeri:
_This commit is a draft, as it is based on #389 for similar handling and to avoid conflicts. I will rebase to master, when the other one is done._

> - [x] Questions with either empty titles or empty answer options are not shown – we should communicate that to the form creator and highlight those unfinished questions?

- A not too intrusive way of marking invalid questions. I normally even had in mind to use just a fat red exclamation mark, but the now used sign is the server standard error-icon.

- @jancborchardt The text-size of the tooltip i would like to increase to 13px, as i had the feeling it is quite hard to read normally on the given 12px. The corresponding commit here is just a Dummy, if you're good with, maybe a commit on vue-components would be better, then it would also affect other apps, as e.g. contacts.

font-size 13px
grafik

standard font-size 12px
grafik

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -46,6 +46,10 @@
 				required
 				@input="onTitleChange">
 			<h3 v-else class="question__header-title" v-text="computedText" />
+			<div v-if="!edit && !questionValid"
+				v-tooltip.auto="warningInvalid"
+				class="question__header-warning icon-error-color"
+				tabindex="0" />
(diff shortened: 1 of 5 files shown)


jotoeri:
- Note to myself: Include Documentation when rebasing.

jancborchardt:
>     * Note to myself: Include Documentation when rebasing.

#389 is merged, so feel free to rebase this and include the documentation. :)

PR #420, opened 2020-05-27
Fix public template header

jotoeri:
- [x] In non-logged-in view, only the template to fill the form showed the public header, while the templates for not-found and expired still showed the internal NC-Header.

- [x] Further solves an internal server error, when accessing a form from non-logged in view, that is non-public and allows only one response per user. (quite special case^^)

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -46,6 +46,11 @@ use OCP\IUserSession;
 use OCP\Util;
 
 class PageController extends Controller {
+	private const TEMPLATE_EXPIRED = 'expired';
+	private const TEMPLATE_MAIN = 'main';
+	private const TEMPLATE_NOSUBMIT = 'nosubmit';
+	private const TEMPLATE_NOTFOUND = 'notfound';
+
 	protected $appName;
 
 	/** @var FormMapper */
@@ -124,7 +129,7 @@ class PageController extends Controller {
 		Util::addScript($this->appName, 'forms');
 		Util::addStyle($this->appName, 'forms');
 		$this->initialStateService->provideInitialState($this->appName, 'maxStringLengths', $this->maxStringLengths);
-		return new TemplateResponse($this->appName, 'main');
+		return new TemplateResponse($this->appName, self::TEMPLATE_MAIN);
 	}
 
 	/**
@@ -137,7 +142,7 @@ class PageController extends Controller {
 		Util::addScript($this->appName, 'forms');
 		Util::addStyle($this->appName, 'forms');
 		$this->initialStateService->provideInitialState($this->appName, 'maxStringLengths', $this->maxStringLengths);
-		return new TemplateResponse($this->appName, 'main');
+		return new TemplateResponse($this->appName, self::TEMPLATE_MAIN);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/Controller/PageController.php
```diff
@@ -195,47 +195,67 @@ public function gotoForm($hash): ?TemplateResponse {
...
+			$response = new PublicTemplateResponse($this->appName, $template);
 
-			// Get owner and check display name privacy settings
-			$owner = $this->userManager->get($form->getOwnerId());
-			if ($owner instanceof IUser) {
-				$ownerAccount = $this->accountManager->getAccount($owner);
+			// Set Header
+			if ($template === 'notfound') {
+				$response->setHeaderTitle($this->l10n->t('Forms'));
+				$response->setHeaderDetails($this->l10n->t('Not found'));
```
skjnldsv, line 240: Not sure about this cc @jancborchardt

lib/Controller/PageController.php
```diff
@@ -195,47 +195,67 @@ public function gotoForm($hash): ?TemplateResponse {
...
-			$response->setHeaderTitle($form->getTitle());
+			$response = new PublicTemplateResponse($this->appName, $template);
 
-			// Get owner and check display name privacy settings
-			$owner = $this->userManager->get($form->getOwnerId());
-			if ($owner instanceof IUser) {
-				$ownerAccount = $this->accountManager->getAccount($owner);
+			// Set Header
+			if ($template === 'notfound') {
+				$response->setHeaderTitle($this->l10n->t('Forms'));
```
skjnldsv, line 239: I guess instead of hardcoded `notfound` we can just set to 'Forms' by default, and if `form->getTitle` is not empty, we set it again? Would be cleaner

lib/Controller/PageController.php
jancborchardt: Yeah, I’d also say we don’t need a "headerDetails" part here, as the emptycontent view is very clear already. :)

image

jotoeri: First looked a bit strange to me on desktop, that's why i added it. But np, i'll remove it. 👍

jotoeri: ...ok could be it looked dumb, as i was still working on NC-18.04. Now it looks good :sweat_smile:

lib/Controller/PageController.php
jotoeri: Done with removing the details 👍

lib/Controller/PageController.php
```diff
@@ -195,47 +195,65 @@ public function gotoForm($hash): ?TemplateResponse {
...
-			$response->setHeaderTitle($form->getTitle());
+			$response = new PublicTemplateResponse($this->appName, $template);
 
-			// Get owner and check display name privacy settings
-			$owner = $this->userManager->get($form->getOwnerId());
-			if ($owner instanceof IUser) {
-				$ownerAccount = $this->accountManager->getAccount($owner);
+			// Set Header
+			$response->setHeaderTitle($this->l10n->t('Forms'));
+			if ($template !== 'notfound') {
```
skjnldsv, line 239: Not fan of hardcoded values.
Can you check against `$form` ?

jotoeri: Hm, would be less failsafe? It's not that we show this only as we don't have a form to show. But we show this too, if there is a form, but the user does not have access. So it is specifically, that we want to show this on the specific template, but not when no form is available.

juliushaertl: If keeping those, please move them to a class constant.

jotoeri: See fixup, would that be good?

skjnldsv: Good for me :)

jotoeri:
Squashed & ready to go 👍

PR #424, opened 2020-05-28
Fix linebreak in description

jotoeri:
Now shows line-breaks also in submit-view.

@jancborchardt Found a branch, that increased the line-height, but was no PR, yet. Just cherry-picked your commit from there. I think this is in your intention? ;) Only thing is that i removed your !important again from font-size. Do you still see any issue here on the font-size? I don't...

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -445,6 +445,7 @@ export default {
 		}
 		.form-desc {
 			font-size: 100%;
+			line-height: 150%;
 			padding-bottom: 20px;
 			resize: none;
 		}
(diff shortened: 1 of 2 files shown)


jancborchardt:
> Found a branch, that increased the line-height, but was no PR, yet. Just cherry-picked your commit from there. I think this is in your intention? ;) Only thing is that i removed your !important again from font-size. Do you still see any issue here on the font-size? I don't...

Heh, thanks a lot! :) I only didn’t open a pull request for that yet since I wanted to test out if the `!important` was needed, but then I promptly forgot. ;)

jotoeri:
:D 👍

PR #427, opened 2020-05-30
Allow navigation through edit via Tab-Key

### LINKED ISSUE #416: Impossible to edit questions/answers with keyboard only after initial writing
1. Some a questions with some answers
2. Now don’t use your mouse anymore
3. Using the `Tab` key, try navigating using just your keyboard and edit the existing questions.

It’s not possible as you can not focus the titles or answers, and focusing them does not change it to editing mode.

@jotoeri @skjnldsv any idea how we can trigger that? We can for sure put `tabindex="0"` on each of the question and answer elements, but it needs to then trigger a switch to the edit view.

jotoeri:
This one enables to navigate between questions via tab-key and enables edit-mode when entering a question.
Fixes #416 partially, second part for proper navigation on the actions-menu needs to be done on vue (https://github.com/nextcloud/nextcloud-vue/pull/1126).

### DIFF
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -37,7 +37,10 @@
 		@delete="onDelete">
 		<ul class="question__content">
 			<template v-for="(answer, index) in options">
-				<li v-if="!edit" :key="answer.id" class="question__item">
+				<li v-if="!edit"
+					:key="answer.id"
+					class="question__item"
+					:class="{'question__item--last': (index === options.length-1),}">
 					<!-- Answer radio/checkbox + label -->
 					<!-- TODO: migrate to radio/checkbox component once available -->
 					<input :id="`${id}-answer-${answer.id}`"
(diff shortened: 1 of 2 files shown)


jotoeri:
Copied from vue-PR:
- If the question title is empty (but there are answers) it’s not possible to Tab into the answers, only into the question title field. Guess because the question is already editable, somehow the edit mode doesn’t get triggered for the whole set?
- Shift-Tabbing into answers from a previous question doesn’t seem to work?

jancborchardt:
@nextcloud/accessibility if anyone is interested in giving this another review, please check. :)

jancborchardt:
Ooops, conflicts because of another merge again @jotoeri, so sorry about that. :( Luckily the mountain of pull requests is soon scaled. :)

jotoeri:
:sweat_smile: Arghh.

Some code-review would also be good... @juliushaertl maybe? :upside_down_face::blush:

src/components/Questions/QuestionMultiple.vue
```diff
@@ -150,6 +153,15 @@ export default {
...
 				this.updateOptions(options)
+			} else {
+				// If edit becomes true by tabbing to last element, focus newAnswer-input
+				if (document.activeElement?.parentNode?.classList.contains('question__item--last')) {
+					this.$nextTick(() => {
+						if (this.$refs.inputNewAnswer) {
+							this.$refs.inputNewAnswer.focus()
+						}
+					})
+				}
```
skjnldsv, line 164: I'm sorry, I'm really not ok with this kind of code here.
Querying classes is really an anti-pattern to vue apps here as any change have a high risk of getting forgotten.

skjnldsv:
Break the multiple choice question for me.
I cannot add a new one and add the first question.

I will revert this pr I think.

PR #428, opened 2020-05-30
Keep focus when copy Share-Link

jotoeri:
Just recognized, that while navigating around with keyboard, that when choosing copyLink in ActionsMenu, the focus got removed, so Keyboard-Navigation jumped to main-content. Fixed such, that action-element gets focused back.

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -161,13 +161,15 @@ export default {
 			})
 		},
 
-		copyShareLink() {
+		copyShareLink(event) {
 			const $formLink = window.location.protocol + '//' + window.location.host + generateUrl(`/apps/forms/${this.form.hash}`)
 			if (this.$clipboard($formLink)) {
 				showSuccess(t('forms', 'Form link copied'))
 			} else {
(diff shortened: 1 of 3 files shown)


jotoeri:
... same applies of course on Results and Sidebar.

src/views/Results.vue
```diff
@@ -161,13 +161,15 @@ export default {
...
-		copyShareLink() {
+		copyShareLink(event) {
 			const $formLink = window.location.protocol + '//' + window.location.host + generateUrl(`/apps/forms/${this.form.hash}`)
 			if (this.$clipboard($formLink)) {
 				showSuccess(t('forms', 'Form link copied'))
 			} else {
 				showError(t('forms', 'Cannot copy, please copy the link manually'))
 			}
+			// Set back focus as clipboard removes focus
+			event.originalTarget.focus()
```
skjnldsv, line 172: https://caniuse.com/#feat=mdn-api_event_originaltarget

jotoeri: :D oups.
Changed to event.target...

PR #452, opened 2020-06-11
Invert submitOnce on UI

### LINKED ISSUE #436: Fix setting "Only allow one response per user"
Currently we have a setting "Only allow one response per user" and it’s not enabled by default. This is a bit confusing since the default being off looks like just a disabled empty checkbox, and it’s not clear why it’s empty etc.

This is why we should invert it:
- "Allow multiple responses per person" as wording
- When either "Share link" is selected or "Anonymous responses" is enabled, the checkbox will be checked and the option disabled. This makes clear that with these options, you can not prevent multiple responses.
- When switching to "Show to all users" or "Choose users", the checkbox should ideally be set by default to prevent multiple submissions on accident. It can still be unset.

jotoeri:
Just more or less as a quick-fix, changing UI for submitOnce.
If this will remain as is, should maybe be inverted downto Db and all other usage, but that will be larger change.
Is this good for you @jancborchardt ?

(Closes #436)

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -174,6 +174,7 @@ class ApiController extends Controller {
 		$form->setAccess([
 			'type' => 'public'
 		]);
+		$form->setSubmitOnce(true);
 
 		$this->formMapper->insert($form);
 
--- a/src/views/Sidebar.vue
+++ b/src/views/Sidebar.vue
@@ -33,6 +33,53 @@
 			</ActionLink>
 		</template>
 
+		<h3>{{ t('forms', 'Sharing') }}</h3>
+		<ul>
+			<li>
+				<input id="public"
+					v-model="form.access.type"
+					type="radio"
+					value="public"
+					class="radio"
+					@change="onAccessChange">
+				<label for="public">
+					<span class="icon-public">
+						{{ t('forms', 'Share link') }}
+					</span>
+				</label>
+			</li>
+			<li>
+				<input id="registered"
+					v-model="form.access.type"
+					type="radio"
+					value="registered"
+					class="radio"
+					@change="onAccessChange">
+				<label for="registered">
+					<span class="icon-group">
+						{{ t('forms', 'Show to all users of this instance') }}
+					</span>
+				</label>
+			</li>
+			<li>
+				<input id="selected"
+					v-model="form.access.type"
+					type="radio"
+					value="selected"
+					class="radio"
+					@change="onAccessChange">
+				<label for="selected">
+					<span class="icon-shared">
+						{{ t('forms', 'Choose users to share with') }}
+					</span>
+				</label>
+				<ShareDiv v-show="form.access.type === 'selected'"
+					:user-shares="userShares"
+					:group-shares="groupShares"
+					@update:shares="onSharingChange" />
+			</li>
+		</ul>
+
 		<h3>{{ t('forms', 'Settings') }}</h3>
 		<ul>
 			<li>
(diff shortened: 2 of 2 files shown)


jancborchardt:
Really nice! :) I would only say that when you switch to "Show to all users of this instance" or "Choose users to share with" then the checkbox should get unchecked to only allow 1 response per user by default.

(Cause the other way around this also happens – if you check "Share link", then the checkbox gets checked automatically.)

jancborchardt:
Added 2 related commits for the sidebar sharing settings:
- Fix order of sidebar settings, because they were logically switched top/bottom. Also now the default "Share via link" is on top"
- Fix discoverability of 'Share link' functionality, cause that little icon in the corner just does not work.

Before | After
-|-
share sidebar before | sidebar after

jotoeri:
Ok, so i changed my Ansatz a bit. Now the box is unchecked by default (on new forms), shows up on UI as we talked about, but keeps its state in background. So esp. when changing between the two internal sharing-options, it keeps its state. What do you think, @jancborchardt ? :)


Two more things i recognized on the sidebar now (@jancborchardt?):
- Isn't the 'Share via Link' and sharing Header quite confusing, as we currently always need to use the link? As it is currently, would be a nice UI if we have further stuff implemented (like the visibility of shared forms on sidebar and email-notifications on sharing). But currently, the setting just changes access, while we need the link anyways. (I think we talked about somewhere already...?^^)
- The sidebar-title still shows the bare form-title. Either we insert the computedTitle here, like in the navigation, or we just use a general sidebar-title (my preference, as the form-title is already visible twice in this view).

jancborchardt:
> Isn't the 'Share via Link' and sharing Header quite confusing, as we currently always need to use the link?

Yeah I agree it’s a bit weird. I’d say it will go away once we implement the sharing interface just like we have it in Files? (Of course a big task, but with that design we solved these issues. :)

> The sidebar-title still shows the bare form-title. Either we insert the computedTitle here, like in the navigation, or we just use a general sidebar-title

What do you think about using "Sharing" as the sidebar title, and we just cut the first h3 between button and radio settings?

jancborchardt:
> Ok, so i changed my Ansatz a bit. Now the box is unchecked by default (on new forms), shows up on UI as we talked about, but keeps its state in background. So esp. when changing between the two internal sharing-options, it keeps its state. What do you think, @jancborchardt ? :)

And that is perfect! Didn’t know that was possible, but yeah that’s really the best behavior! Great job :) Will add a commit to fix the title and then it’s good to go

jancborchardt:
Hmm, not sure if compared with the screenshot above it’s _too_ simplified. @jotoeri what do you think, or maybe the wording or order could be changed otherwise?
image

jotoeri:
> Yeah I agree it’s a bit weird. I’d say it will go away once we implement the sharing interface just like we have it in Files?

Hm, yeah, so maybe we should change the wording until this is done? Better have a proper wording now and change later, than having a strange behaviour now, just to show what we want to do somewhen?^^

> Hmm, not sure if compared with the screenshot above it’s too simplified.

Hm, indeed, that looks a bit strange, the sidebar does not only contain sharing. So other way round: What does the sidebar contain? 'Form properties'/'Settings'? So this as overall header? And then maybe sth. different for the current 'Settings'? Maybe split up into 'Responses' and 'Expiration'? Just that 'Expiration' would then only contain a single checkbox+setting, if enabled.... :/ I'm also not sure about... 🤔

jancborchardt:
> Hm, indeed, that looks a bit strange, the sidebar does not only contain sharing.

I’d say that’s not necessarily the problem though, as it is indeed all related to sharing. Anonymous means if people can fill out the share anonymously, expiration means when the share will expire etc. :) It’s more like the layout looks a bit off.

Anyway, I adjusted the wording a bit and think it’s now good to go. We can always check back on that, as we tweak the sidebar every now and then. :)

PR #454, opened 2020-06-12
Cleanup old files

jotoeri:
As stated on chat - we can remove the old Support-File.
Further, the issue-template is not used anymore as there is a new one in .github-folder. And the .l10nignore is out of date, as we don't have the survey.jquery-file anymore.


stefan-niedermann:
ust wanted to say, that i yesterday installed the forms app (i guess it was version 1) and there is a big fat "help"-button in the top left which has a direct link to this file.... so as long as you don't release version 2 i would double check this...

PR #460, opened 2020-06-16
Add summary response visualization, fix #314

### LINKED ISSUE #314: Different Result-views
@jancborchardt @skjnldsv 
Ok, so to clarify, these were the views, that i had in mind. As i understood, i think we agreed, that we need both UI-views?

The radio-buttons now were not the way to go i heard from you., So i think the main discussion-point now was how to switch between the two main-contents? Google does with the three tabs at the top...?

Sorry, @jancborchardt  for that strange discussion, after talking to John i think we just talked particularly on completely different directions & things... ;)

grafik
grafik
grafik

jancborchardt:
Fix #314 by introducing nice response visualizations. For text questions, "no response" is counted and responses listed as they are.

Took me some nights of fighting with Vue.js, but was a nice experience. ;) It works, but can likely be optimized and needs a thorough code review.

stats
stats more

### DIFF
--- a/src/components/Results/Summary.vue
+++ b/src/components/Results/Summary.vue
@@ -1,0 +1,245 @@
+<!--
+  - @copyright Copyright (c) 2020 Jan C. Borchardt https://jancborchardt.net
+  -
+  - @author Jan C. Borchardt https://jancborchardt.net
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div class="section question-summary">
+		<h3>{{ question.text }}</h3>
+		<p class="question-summary__detail">
+			{{ answerTypes[question.type].label }}
+		</p>
+
+		<!-- Answers with countable results for visualization -->
+		<ol v-if="question.type === 'multiple' || question.type === 'multiple_unique' || question.type === 'dropdown'"
+			class="question-summary__statistic">
+			<li v-for="option in questionOptions"
+				:key="option.id">
+				<label :for="`option-${option.questionId}-${option.id}`">
+					{{ option.count }}
+					<span class="question-summary__statistic-percentage">
+						({{ option.percentage }}%):
+					</span>
+					<span class="question-summary__statistic-text">{{ option.text }}</span>
+				</label>
+				<meter :id="`option-${option.questionId}-${option.id}`"
+					min="0"
+					:max="submissions.length"
+					:value="option.count" />
+			</li>
+		</ol>
+
+		<!-- Text answers are simply listed for now, could be automatically grouped in the future -->
+		<ul v-else class="question-summary__text">
+			<li v-for="answer in textAnswers"
+				:key="answer.id">
+				{{ answer }}
+			</li>
+		</ul>
+	</div>
+</template>
+
+<script>
+import answerTypes from '../../models/AnswerTypes'
+
+export default {
+	name: 'Summary',
+
+	props: {
+		submissions: {
+			type: Array,
+			required: true,
+		},
+		question: {
+			type: Object,
+			required: true,
+		},
+	},
+
+	data() {
+		return {
+			answerTypes,
+		}
+	},
+
+	computed: {
+		// For countable questions like multiple choice and checkboxes
+		questionOptions() {
+			// Build list of question options
+			const questionOptionsStats = this.question.options.map(option => ({
+				...option,
+				count: 0,
+				percentage: 0,
+			}))
+
+			// Also record 'No response'
+			questionOptionsStats.unshift({
+				text: t('forms', 'No response'),
+				count: 0,
+				percentage: 0,
+			})
+
+			// Go through submissions to check which options have how many responses
+			this.submissions.forEach(submission => {
+				const answers = submission.answers.filter(answer => answer.questionId === this.question.id)
+				if (!answers.length) {
+					// Record 'No response'
+					questionOptionsStats[0].count++
+				}
+
+				// Check question options to find which needs to be increased
+				answers.forEach(answer => {
+					const optionsStatIndex = questionOptionsStats.findIndex(option => option.text === answer.text)
+					if (optionsStatIndex < 0) {
+						questionOptionsStats.push({
+							'text': answer.text,
+							'count': 1,
+							'percentage': 0,
+						})
+					} else {
+						questionOptionsStats[optionsStatIndex].count++
+					}
+				})
+			})
+
+			// Sort options by response count
+			questionOptionsStats.sort((object1, object2) => {
+				if (object1.count === object2.count) {
+					return 0
+				}
+				return object1.count < object2.count ? 1 : -1
+			})
+
+			// Fill percentage values
+			questionOptionsStats.forEach(questionOptionsStat => {
+				questionOptionsStat.percentage = Math.round((100 * questionOptionsStat.count) / this.submissions.length)
+			})
+
+			return questionOptionsStats
+		},
+
+		// For text answers like short answer and long text
+		textAnswers() {
+			const textAnswers = []
+
+			// Also record 'No response'
+			let noResponseCount = 0
+
+			// Go through submissions to check which options have how many responses
+			this.submissions.forEach(submission => {
+				const answers = submission.answers.filter(answer => answer.questionId === this.question.id)
+				if (!answers.length) {
+					// Record 'No response'
+					noResponseCount++
+				}
+
+				// Add text answers
+				answers.forEach(answer => {
+					textAnswers.push(answer.text)
+				})
+			})
+
+			// Calculate no response percentage
+			const noResponsePercentage = Math.round((100 * noResponseCount) / this.submissions.length)
+			textAnswers.unshift(noResponseCount + ' (' + noResponsePercentage + '%): ' + t('forms', 'No response'))
+
+			return textAnswers
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+.question-summary {
+	padding-left: 16px;
+	padding-right: 16px;
+
+	h3 {
+		font-weight: bold;
+	}
+
+	&__detail {
+		color: var(--color-text-lighter);
+		margin-top: -8px;
+	}
+
+	&__text,
+	&__statistic {
+		margin-top: 8px;
+	}
+
+	&__text {
+		list-style-type: initial;
+
+		li {
+			padding: 4px 0;
+
+			&:first-child {
+				font-weight: bold;
+			}
+		}
+	}
+
+	&__statistic {
+		list-style-type: none;
+
+		li {
+			position: relative;
+			padding: 8px 0;
+
+			label {
+				cursor: default;
+			}
+
+			&:first-child .question-summary__statistic-text {
+				font-weight: bold;
+			}
+
+			.question-summary__statistic-percentage {
+				color: var(--color-text-maxcontrast);
+			}
+
+			meter {
+				display: block;
+				width: 100%;
+				margin-top: 4px;
+				background: var(--color-background-dark);
+				height: calc(var(--border-radius) * 2);
+				border-radius: var(--border-radius);
+
+				&::-webkit-meter-bar {
+					height: calc(var(--border-radius) * 2);
+				}
+
+				// The pseudo-classes of -moz and -webkit have to stay separated even with SCSS, otherwise they don’t work
+				&::-webkit-meter-optimum-value {
+					background: linear-gradient(40deg, var(--color-primary-element) 0%, var(--color-primary-element-light) 100%);
+					border-radius: var(--border-radius);
+				}
+
+				&::-moz-meter-bar {
+					background: linear-gradient(40deg, var(--color-primary-element) 0%, var(--color-primary-element-light) 100%);
+					border-radius: var(--border-radius);
+				}
+			}
+		}
+	}
+}
+</style>
--- a/src/components/Results/Summary.vue
+++ b/src/components/Results/Summary.vue
@@ -121,10 +121,7 @@ export default {
 
 			// Sort options by response count
 			questionOptionsStats.sort((object1, object2) => {
-				if (object1.count === object2.count) {
-					return 0
-				}
-				return object1.count < object2.count ? 1 : -1
+				return object2.count - object1.count
 			})
 
 			// Fill percentage values
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -38,15 +38,36 @@
 		</TopBar>
 
 		<header v-if="!noSubmissions">
-			<h2>{{ t('forms', 'Responses for {title}', { title: formTitle }) }}</h2>
-			<div class="response_actions">
-				<button class="response_actions__export" @click="download">
-					<span class="icon-download" role="img" />
-					{{ t('forms', 'Export to CSV') }}
-				</button>
+			<h2>{{ t('forms', 'Responses for {title}', { title: form.title }) }}</h2>
+			<div class="response-actions">
+				<div class="response-actions__radio">
+					<input id="show-summary--true"
+						v-model="showSummary"
+						type="radio"
+						:value="true"
+						class="hidden">
+					<label for="show-summary--true"
+						class="response-actions__radio__item"
+						:class="{ 'response-actions__radio__item--active': showSummary }">
+						{{ t('forms', 'Summary') }}
+					</label>
+					<input id="show-summary--false"
+						v-model="showSummary"
+						type="radio"
+						:value="false"
+						class="hidden">
+					<label for="show-summary--false"
+						class="response-actions__radio__item"
+						:class="{ 'response-actions__radio__item--active': !showSummary }">
+						{{ t('forms', 'Responses') }}
+					</label>
+				</div>
 				<Actions class="results-menu"
 					:aria-label="t('forms', 'Options')"
 					:force-menu="true">
+					<ActionButton icon="icon-download" @click="download">
+						{{ t('forms', 'Export to CSV') }}
+					</ActionButton>
 					<ActionButton icon="icon-delete" @click="deleteAllSubmissions">
 						{{ t('forms', 'Delete all responses') }}
 					</ActionButton>
@@ -69,7 +90,11 @@
 			</EmptyContent>
 		</section>
 
-		<section v-else>
+		<section v-if="!noSubmissions && showSummary">
+			SUMMARY-CONTENT
+		</section>
+
+		<section v-if="!noSubmissions && !showSummary">
 			<Submission
 				v-for="submission in form.submissions"
 				:key="submission.id"
@@ -117,6 +142,7 @@ export default {
 	data() {
 		return {
 			loadingResults: true,
+			showSummary: true,
 		}
 	},
 
@@ -273,13 +299,25 @@ h2 {
 	white-space: nowrap;
 }
 
-.response_actions {
+.response-actions {
 	display: flex;
 	align-items: center;
+	padding-left: 14px;
 
-	&__export {
-		width: max-content;
-		margin-left: 16px;
+	&__radio {
+		margin-right: 8px;
+
+		&__item {
+			border-radius: 2px;
+			padding: 8px 16px;
+			font-weight: bold;
+			background-color: var(--color-background-dark);
+
+			&--active {
+				background-color: var(--color-primary);
+				color: var(--color-primary-text)
+			}
+		}
 	}
 }
 </style>
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -38,7 +38,10 @@
 		</TopBar>
 
 		<header v-if="!noSubmissions">
-			<h2>{{ t('forms', 'Responses for {title}', { title: form.title }) }}</h2>
+			<h2>{{ formTitle }}</h2>
+			<p>{{ t('forms', '{amount} responses', { amount: form.submissions.length }) }}</p>
+
+			<!-- View switcher between Summary and Responses -->
 			<div class="response-actions">
 				<div class="response-actions__radio">
 					<input id="show-summary--true"
@@ -62,6 +65,8 @@
 						{{ t('forms', 'Responses') }}
 					</label>
 				</div>
+
+				<!-- Action menu for CSV export and deletion -->
 				<Actions class="results-menu"
 					:aria-label="t('forms', 'Options')"
 					:force-menu="true">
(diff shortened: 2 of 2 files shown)


src/components/Results/Summary.vue
```diff
@@ -0,0 +1,258 @@
...
+<template>
+	<div class="section question">
+		<h3>{{ question.text }}</h3>
+		<p class="question-detail">
+			{{ questionType }}
+		</p>
+		<ol v-if="question.type == 'multiple' || question.type == 'multiple_unique'" class="question-multiple">
+			<li v-for="questionOption in questionOptions"
+				:key="questionOption.id"
+				:class="['percentage-' + questionOption.percentage]">
```
juliushaertl, line 32: How about using https://developer.mozilla.org/en/docs/Web/HTML/Element/progress instead of adding 100 css classes? :wink:

jancborchardt: lol, seems like the JS coding overwrote some HTML knowledge in my brain

jancborchardt: `progress` seemed off for this case as it is _not_ a progress, but sure enough there’s `meter`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter

juliushaertl: Ah nice, I wasn't even aware this exists :grin:

juliushaertl: Not working in IE11, but might be ok for the forms app, we drop support anyways soon :partying_face:

jancborchardt: Fixed now

src/views/Results.vue
```diff
@@ -38,15 +38,36 @@
 		</TopBar>
 
 		<header v-if="!noSubmissions">
-			<h2>{{ t('forms', 'Responses for {title}', { title: formTitle }) }}</h2>
-			<div class="response_actions">
-				<button class="response_actions__export" @click="download">
-					<span class="icon-download" role="img" />
-					{{ t('forms', 'Export to CSV') }}
-				</button>
+			<h2>{{ t('forms', '{amount} responses for {title}', { amount: form.submissions.length, title: form.title }) }}</h2>
```
jotoeri, line 41: Hm, that became quite long now on the title...?!
Maybe set the amount as subheader h3 or so?

src/views/Results.vue
```diff
@@ -261,23 +293,48 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-h2 {
+.app-content header h2 {
 	font-size: 2em;
 	font-weight: bold;
 	margin-top: 32px;
+	margin-bottom: 8px;
```
jotoeri, line 300: Use padding, to avoid cutting descenders again?
grafik

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,255 @@
...
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div class="section question">
+		<h3>{{ question.text }}</h3>
+		<p class="question-detail">
+			{{ questionType }}
```
jotoeri, line 27: If you include the answerTypes, you should be able to use their label here directly. Just like it's done on create.vue & submit.vue. Having the advantage to have the labels defined only in one place and don't have to adapt this computed property here, if we change a word or even include a new Question-Type.

So would be ```answerTypes[question.type].label``` here.

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,255 @@
...
+					{{ questionOption.percentage }}%, {{ questionOption.count }}:
+					<span>{{ questionOption.text }}</span>
+				</label>
+				<meter :id="questionOption.text.replace(/\s+/g, '')"
+					min="0"
+					:max="submissions.length"
+					:value="questionOption.count" />
+			</li>
+		</ol>
+		<ul v-if="question.type == 'short' || question.type == 'long'" class="question-text">
```
jotoeri, line 42: Using v-else here, to have some fallback-behaviour?

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,255 @@
...
+		</ul>
+	</div>
+</template>
+
+<script>
+export default {
+	name: 'Summary',
+
+	components: {
+	},
```
jotoeri, line 56: You can just leave out, if empty? ;)

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,255 @@
...
+	<div class="section question">
+		<h3>{{ question.text }}</h3>
+		<p class="question-detail">
+			{{ questionType }}
+		</p>
+		<ol v-if="question.type == 'multiple' || question.type == 'multiple_unique'" class="question-multiple">
+			<li v-for="questionOption in questionOptions"
+				:key="questionOption.id">
+				<label :for="questionOption.text.replace(/\s+/g, '')">
+					{{ questionOption.percentage }}%, {{ questionOption.count }}:
```
jotoeri, line 33: Can we mark/describe the count somehow ('13 responses')? Looks a bit like some number without Context to me...?!

src/components/Results/Summary.vue
jancborchardt: Yeah, I assumed this is possible → thanks for showing me how! :)

src/views/Results.vue
jancborchardt: Should be fixed but let’s see with the polishing when we fixed the error as well.

jancborchardt:
This now needs more code review @juliushaertl @ChristophWurst @georgehrke :)

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,256 @@
...
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div class="section question">
```
jotoeri, line 24: Stumbled a bit about this once more... Maybe use ```summary``` as class? Or ```question-summary```? Shouldn't be a problem on UI later (as they are not shown at the same time), just to get a separation on code from the questions on edit & submit.
(Correspondingly `question-summary__detail`, etc as BEM-Naming then)

src/views/Results.vue
```diff
@@ -117,6 +149,9 @@ export default {
 	data() {
 		return {
 			loadingResults: true,
+			submissions: [],
+			questions: [],
```
jotoeri, line 153: Did my fixup get lost...? 👀

jotoeri: Hm, came back with one of the ``` fixup! fixup! Add Summary response visualization, fix #314 ```. So might need a ```fixupfixupfixup```... 😅

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,256 @@
...
+		<h3>{{ question.text }}</h3>
+		<p class="question-detail">
+			{{ answerTypes[question.type].label }}
+		</p>
+		<ol v-if="question.type == 'multiple' || question.type == 'multiple_unique' || question.type == 'dropdown'"
+			class="question-multiple">
+			<li v-for="questionOption in questionOptions"
+				:key="questionOption.id">
```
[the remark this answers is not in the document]
jancborchardt: Fixed! :)

src/components/Results/Summary.vue
jancborchardt: Yep, good point – fixed :)

src/views/Results.vue
jancborchardt: Also fixed :)

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,256 @@
...
+
+	ol {
+		list-style-type: none;
+	}
+
+	ul,
+	ol {
+		margin-top: 8px;
+
+		&__text {
```
jotoeri, line 193: Unfortunately css broke on your last commit, as the `&__text` is no full class.
If you close the ul,ol directly after the margin, then `&__text` and `&__statistic` are children of `question-summary` and it should work again...

jancborchardt: Yep, noticed as well and working on it :D

jancborchardt: Fix pushed! :) Also simplified the CSS, not using the ul/ol now that we have proper classes

skjnldsv:
> [@jancborchardt ](/jancborchardt)  [fixup! fixup! fixup! fixup! Add Summary response visualization,](https://github.com/nextcloud/forms/pull/460/commits/8cf9c2dbada2723c21cdeccb9d960fdd0b6fdd28) [fix](https://github.com/nextcloud/forms/pull/460/commits/8cf9c2dbada2723c21cdeccb9d960fdd0b6fdd28) #314

image

jancborchardt:
@skjnldsv your review will tell if we need to do another one ;D

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,254 @@
...
+  -->
+
+<template>
+	<div class="section question-summary">
+		<h3>{{ question.text }}</h3>
+		<p class="question-summary__detail">
+			{{ answerTypes[question.type].label }}
+		</p>
+		<ol v-if="question.type == 'multiple' || question.type == 'multiple_unique' || question.type == 'dropdown'"
+			class="question-summary__statistic">
```
skjnldsv, line 30: Thanks for using BEM <3

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,254 @@
...
+  -
+  -->
+
+<template>
+	<div class="section question-summary">
+		<h3>{{ question.text }}</h3>
+		<p class="question-summary__detail">
+			{{ answerTypes[question.type].label }}
+		</p>
+		<ol v-if="question.type == 'multiple' || question.type == 'multiple_unique' || question.type == 'dropdown'"
```
skjnldsv, line 29:
```suggestion
		<ol v-if="question.type === 'multiple' || question.type === 'multiple_unique' || question.type === 'dropdown'"
```

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,254 @@
...
+			class="question-summary__statistic">
+			<li v-for="(questionOption, index) in questionOptions"
+				:key="index">
+				<label :for="questionOption.text.replace(/\s+/g, '')">
```
skjnldsv, line 33: Not fond of having this here, please create a dedicated method 
We aim to have the minimum computation within templates

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,254 @@
...
+					{{ questionOption.count }} <span class="question-summary__statistic-percentage">({{ questionOption.percentage }}%):</span>
+					<span class="question-summary__statistic-text">{{ questionOption.text }}</span>
+				</label>
+				<meter :id="questionOption.text.replace(/\s+/g, '')"
+					min="0"
+					:max="submissions.length"
+					:value="questionOption.count" />
+			</li>
+		</ol>
+		<ul v-else class="question-summary__text">
```
skjnldsv, line 43:
```suggestion

		<!-- Don't hesitate to add comments to describe sections like those (and with this comment syntax on templates) -->
		<ul v-else class="question-summary__text">
```

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,254 @@
...
+			<li v-for="(textAnswer, index) in textAnswers"
+				:key="index">
```
skjnldsv, line 45: Our answers have keys already, let's not use index :)

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,254 @@
...
+			})
+
+			// Build list of question options
+			this.question.options.forEach(option => {
+				questionOptionsStats.push({
+					'text': option.text,
+					'count': 0,
+					'percentage': 0,
+				})
+			})
```
skjnldsv, line 99: You can use Array.map directly :)

```suggestion
			// Build list of question options
			const questionOptionsStats = this.question.options.map(option => ({
				text: option.text,
				count: 0,
				percentage: 0,
			}))

			// Also record 'No response'
			questionOptionsStats.unshift({
				text: t('forms', 'No response'),
				count: 0,
				percentage: 0,
			})
```

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,254 @@
...
+				animation: percentage-animation 1s linear infinite;
+				background: linear-gradient(40deg, var(--color-primary-element), var(--color-primary-element-light) 33%, var(--color-primary-element) 67%, var(--color-primary-element-light));
+				background-size: 300% 100%;
+			}
+		}
+	}
+
+	@keyframes percentage-animation {
+		0% {
+			background-position: 100%;
```
skjnldsv, line 247: Looks awesome :heart_eyes:

src/views/Results.vue
```diff
@@ -38,15 +38,37 @@
...
 		<header v-if="!noSubmissions">
-			<h2>{{ t('forms', 'Responses for {title}', { title: formTitle }) }}</h2>
-			<div class="response_actions">
-				<button class="response_actions__export" @click="download">
-					<span class="icon-download" role="img" />
-					{{ t('forms', 'Export to CSV') }}
-				</button>
+			<h2>{{ formTitle }}</h2>
+			<p>{{ t('forms', '{amount} responses', { amount: form.submissions.length }) }}</p>
+			<div class="response-actions">
```
skjnldsv, line 43: Please spread a bit more the sections for template readability. Personal preferences, but I tend to add a new line and a quick comment between the big divs/sections :)

src/components/Results/Summary.vue
jancborchardt: Thanks to @jotoeri for reviewing to tell me to use BEM ;)

src/components/Results/Summary.vue
jotoeri: These don't have, as they are just a computed array of the text of all corresponding answers...

skjnldsv:
Capture d’écran_2020-06-18_12-00-06
Weirdly, the design doesn't work on brave (chromium)

src/components/Results/Summary.vue
jancborchardt: Of course, thanks – fixed!

src/components/Results/Summary.vue
jancborchardt: Done!

src/components/Results/Summary.vue
jancborchardt: Thank for the nudge, added comments, also in the Results.vue

src/components/Results/Summary.vue
jancborchardt: Ok, so I didn’t change anything here?

src/components/Results/Summary.vue
jancborchardt: Hehe, yep that’s the advanced stuff I didn’t get my head wrapped around yet. Thanks a lot, fixed! :)

src/views/Results.vue
jancborchardt: Done too and added the comments. :)

jancborchardt:
Looking into the visualization issues with Chrome now.

jancborchardt:
> Weirdly, the design doesn't work on brave (chromium)

Fixed now. Turns out the pseudo-classes of -moz and -webkit have to stay separated even with SCSS, otherwise they don’t work. Strange, I thought it would not be necessary since they are built out into separate rules?

skjnldsv:
> Strange, I thought it would not be necessary since they are built out into separate rules?

Weird, I was also not aware of it!
What did the output looked like?

jancborchardt:
> > Strange, I thought it would not be necessary since they are built out into separate rules?
> 
> Weird, I was also not aware of it!
> What did the output look like?

Didn’t check the code, but the result was your screenshot → not applying the styles.

jancborchardt:
So is this good to go, or any further review points?

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,272 @@
...
+			}
+
+			&:hover meter::-webkit-meter-optimum-value,
+			&:hover meter::-moz-meter-bar,
+			&:focus meter::-webkit-meter-optimum-value,
+			&:focus meter::-moz-meter-bar {
+				animation: percentage-animation 1s linear infinite;
+				background: linear-gradient(40deg, var(--color-primary-element), var(--color-primary-element-light) 33%, var(--color-primary-element) 67%, var(--color-primary-element-light));
+				background-size: 300% 100%;
+			}
```
skjnldsv, line 259: So it works here?

But wasn't working on the like 237?

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,272 @@
...
+			answerTypes,
+		}
+	},
+
+	computed: {
+		// For countable questions like multiple choice and checkboxes
+		questionOptions() {
+			// Build list of question options
+			const questionOptionsStats = this.question.options.map(option => ({
+				text: option.text,
```
skjnldsv, line 89:
```suggestion
				...options,
```

skjnldsv: That way you extend all the id/text data to the new object.

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,272 @@
...
+		<h3>{{ question.text }}</h3>
+		<p class="question-summary__detail">
+			{{ answerTypes[question.type].label }}
+		</p>
+
+		<!-- Answers with countable results for visualization -->
+		<ol v-if="question.type === 'multiple' || question.type === 'multiple_unique' || question.type === 'dropdown'"
+			class="question-summary__statistic">
+			<li v-for="(questionOption, index) in questionOptions"
+				:key="index">
```
skjnldsv, line 34:
```suggestion
			<li v-for="questionOption in questionOptions"
				:key="questionOption.id">
```

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,272 @@
...
+				<label :for="removeSpaces(questionOption.text)">
+					{{ questionOption.count }} <span class="question-summary__statistic-percentage">({{ questionOption.percentage }}%):</span>
+					<span class="question-summary__statistic-text">{{ questionOption.text }}</span>
+				</label>
```
skjnldsv, line 38:
```suggestion
				<label :for="`option-${option.questionId}-${option.id}`">
					{{ option.count }}
					<span class="question-summary__statistic-percentage">
						({{ option.percentage }}%):
					</span>
					<span class="question-summary__statistic-text">{{ option.text }}</span>
				</label>
```

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,272 @@
...
+				<meter :id="removeSpaces(questionOption.text)"
```
skjnldsv, line 39:
```suggestion
				<meter :id="`option-${option.questionId}-${option.id}`"
```

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,272 @@
...
+					min="0"
+					:max="submissions.length"
+					:value="questionOption.count" />
```
skjnldsv, line 42:
```suggestion
					:value="option.count" />
```

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,272 @@
...
+			</li>
+		</ol>
+
+		<!-- Text answers are simply listed for now, could be automatically grouped in the future -->
+		<ul v-else class="question-summary__text">
+			<li v-for="(textAnswer, index) in textAnswers"
+				:key="index">
```
skjnldsv, line 49: Same here, you should not use index as keys in vue :)
They doesn't change, so vue is not aware of any data change too

src/components/Results/Summary.vue
jancborchardt: > So having the two browser specific properties works here?

Is this a question or did you test and the animation works for you? :)

jancborchardt: I don’t know how to get the animation work for Chrome, tested a lot now … will just rip that out again as it doesn’t really add much.

src/components/Results/Summary.vue
jancborchardt: Guess you meant `...option` here, since otherwise I’m getting `'options' is not defined`?

jancborchardt:
@skjnldsv fixed all your feedback! :)

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,246 @@
...
+					:max="submissions.length"
+					:value="option.count" />
+			</li>
+		</ol>
+
+		<!-- Text answers are simply listed for now, could be automatically grouped in the future -->
+		<ul v-else class="question-summary__text">
+			<li v-for="answer in textAnswers"
+				:key="answer.id">
+				<template v-if="index === 0">
```
jotoeri, line 53: Still an index remained...

jancborchardt: Fixed :)

rullzer:
I glanced over the code all good. Also I like the styling in general. However there is one bug (there always is at least one!)

x

No response should not be inbetween the normal responses.
Also it might be good to style it different.

@jancborchardt ^^

jancborchardt:
> No response should not be inbetween the normal responses.
> Also it might be good to style it different.

I’d disagree on that it should not be inbetween the normal responses. If e.g. a lot of people didn’t reply, it should be directly visible. "No response" is also an important type of response. (This is e.g. why we show it first on the free text fields: Usually most people don’t reply, and if we list it last it goes under and too much value is assigned to the replies from individual people.)

Yes, it would be good to style it differently but I would handle that in a separate pull request as this is not really blocking this already large pull request.

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,245 @@
...
+
+<template>
+	<div class="section question-summary">
+		<h3>{{ question.text }}</h3>
+		<p class="question-summary__detail">
+			{{ answerTypes[question.type].label }}
+		</p>
+
+		<!-- Answers with countable results for visualization -->
+		<ol v-if="question.type === 'multiple' || question.type === 'multiple_unique' || question.type === 'dropdown'"
```
skjnldsv, line 31: Maybe this should not be listed here but have its own dedicated key in answerTypes ?

src/components/Results/Summary.vue
```diff
@@ -0,0 +1,245 @@
...
+					}
+				})
+			})
+
+			// Sort options by response count
+			questionOptionsStats.sort((object1, object2) => {
+				if (object1.count === object2.count) {
+					return 0
+				}
+				return object1.count < object2.count ? 1 : -1
```
skjnldsv, line 127:
```suggestion
				return object2.count - object1.count
```
Cleaner,

src/views/Results.vue
```diff
@@ -38,15 +38,41 @@
...
+			<h2>{{ formTitle }}</h2>
+			<p>{{ t('forms', '{amount} responses', { amount: form.submissions.length }) }}</p>
+
+			<!-- View switcher between Summary and Responses -->
+			<div class="response-actions">
+				<div class="response-actions__radio">
+					<input id="show-summary--true"
+						v-model="showSummary"
+						type="radio"
+						:value="true"
```
skjnldsv, line 50: Clever!!

src/views/Results.vue
```diff
@@ -261,23 +300,54 @@ export default {
...
-	text-overflow: ellipsis;
-	white-space: nowrap;
-}
+.app-content header {
+	h2 {
+		font-size: 2em;
+		font-weight: bold;
+		margin-top: 32px;
+		padding-left: 14px;
+		padding-bottom: 8px;
```
skjnldsv, line 309: Lots of random units here, I'm not really fond and would prefer if we'd use variables so that we can reuse stuff :wink:

jancborchardt: The units are for alignment with the core styles – as you see it was like that before as well, and only the padding-bottom got added. So I’d say it’s unrelated to this pull request here → separate topic.

skjnldsv: Definitely unrelated :)
Let's tackle it later :rocket:

jancborchardt:
Rebased – this should be all ready to go now.

(@skjnldsv the sign-off on your commit was missing, fixed it.)

PR #478, opened 2020-06-20
Include version on feature-request template

jotoeri:
For easier processing of requests, especially if we think they are done already.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,11 +6,7 @@ labels: 0. Needs triage, bug
 assignees: ''
 
 ---
-### How to use GitHub
-
-* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
-* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
-* Subscribe to receive notifications on status change and new comments. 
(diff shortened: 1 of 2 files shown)


jancborchardt:
I'd say put the current server version in (19.0.?) and also the anticipated new Notes release (2.0.0), that way people see what they ideally should be using. :)

jotoeri:
Thats good idea, yes :+1:
Would you mind to have a look onto both templates again (feature-request & bug-report) from side of UX? 
How do we get users to really use the templates and provide us the information, we need? It happens (quite often), that users don't use esp. the bug-template...

jancborchardt:
> Would you mind to have a look onto both templates again (feature-request & bug-report) from side of UX?
> How do we get users to really use the templates and provide us the information, we need? It happens (quite often), that users don't use esp. the bug-template...

The biggest issue usually is that the template is too long, but that’s not the case here. Might just be a matter of if anything is missing to ask people to use the template (and not individually asking "what’s the version" etc.).

jotoeri:
Hm, ok... So then this one should be good to go now...?

jotoeri:
Ah - or how do you think about shortening the first text as in the fixup? I think i personally would also wonder, if i first get told 'how to use the website'.

jancborchardt:
> Ah - or how do you think about shortening the first text as in the fixup? I think i personally would also wonder, if i first get told 'how to use the website'.

Yeees! So much better, and it's not so confusing. I was also always wondering if that is something to remove, but it's there for peopel _reading_ the issue. :D

PR #484, opened 2020-06-22
Revert "Allow navigation through edit via Tab-Key"

skjnldsv:
Re-open #416

For me #427 leaves the app in a broken state. I concur that keyboard nav is extremely important, but not more than leaving the broken state :stuck_out_tongue_winking_eye: 
Let's find a better solution altogether please for rc1

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -25,9 +25,7 @@
 		:class="{ 'question--edit': edit }"
 		:aria-label="t('forms', 'Question number {index}', {index})"
 		class="question"
-		@click="enableEdit"
-		@focusin="onFocusIn"
-		@focusout="onFocusOut">
+		@click="enableEdit">
 		<!-- Drag handle -->
 		<!-- TODO: implement arrow key mapping to reorder question -->
 		<div v-if="!readOnly"
@@ -38,7 +36,6 @@
 		<!-- Header -->
 		<div class="question__header">
 			<input v-if="edit || !text"
-				ref="titleInput"
 				:placeholder="titlePlaceholder"
 				:aria-label="t('forms', 'Title of question number {index}', {index})"
 				:value="text"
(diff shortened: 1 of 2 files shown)


jotoeri:
Why broken? And where?
For me it works, for Jan it worked...?!

skjnldsv:
> Edit: OK, just saw your comments, will check that tomorrow.

Thanks :)
Maybe I did something wrong, keep me up to date! :hugs:

jotoeri:
Ok. So seems not to work on chromium engine, as the timing of events & nextTick() is a bit strange there.
But let's try to fix that, instead of just reverting the PR.

I gave it a try already. @skjnldsv Do you know a way, how to first process default-Behaviour and then a specific handler afterwards? So ShiftTab first moves to previous tab-element and i can trigger disableEdit afterwards?

skjnldsv:
> I gave it a try already. @skjnldsv Do you know a way, how to first process default-Behaviour and then a specific handler afterwards? So ShiftTab first moves to previous tab-element and i can trigger disableEdit afterwards?

Hum, capture?
https://developer.mozilla.org/fr/docs/Web/API/EventTarget/addEventListener
Not sure, you can try and play a bit with the [event modifiers](https://vuejs.org/v2/guide/events.html#Event-Modifiers) :shrug:

skjnldsv:
@jancborchardt this is a breaking change, I'd rather have this in and have #416 in another following release, do you agree?

jancborchardt:
> @jancborchardt this is a breaking change, I'd rather have this in and have #416 in another following release, do you agree?

As it’s an accessibility issue and there’s also the start of a fix pending at https://github.com/nextcloud/forms/pull/487 (and cause we have some more time for the planned release end of July), isn’t this something that could be checked out a bit more? Did you see the draft fix by @jotoeri?

phonon112358:
that is an important issue in my opinion... 
please review it.

jancborchardt:
So what's the call here @skjnldsv - @jotoeri? :) This one, or #487?

cc @jospoortvliet @rullzer cause it seems a release blocker.

jancborchardt:
> @jancborchardt this is a breaking change, I'd rather have this in and have #416 in another following release, do you agree?

Merging this here now to fix it for the initial release and following up with @jotoeri’s #487 in 2.1. :)

PR #485, opened 2020-06-22
Fix questions & submissions assignment

### LINKED ISSUE #475: Show empty Content, after deleting last response
Deleting all or deleting the last response, does not show the empty-content afterwards, but still the normal view without data, which esp. on summary looks quite strange then.

skjnldsv:
Fix #475 

@jotoeri you cannot just assign data to an object qith Vue, otherwise Vue cannot know when to inject its observer :)
You need to use `this.$set` or `Vue.set`, Then the observer will work. Same if you delete a key btw (not relevant here of course) ;)

:rocket:

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -172,17 +172,19 @@ export default {
 
 		async loadFormResults() {
 			this.loadingResults = true
-			console.debug('Loading Results')
+			console.debug('Loading results for form', this.form.hash)
 
 			try {
 				const response = await axios.get(generateUrl('/apps/forms/api/v1/submissions/{hash}', {
 					hash: this.form.hash,
 				}))
-				this.form.submissions = response.data.submissions
-				this.form.questions = response.data.questions
+
+				// Append questions & submissions
+				this.$set(this.form, 'submissions', response.data.submissions)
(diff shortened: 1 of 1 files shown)


jotoeri:
Ok. 👍  That vue.set i found, too, but only tried on deleting the content...:man_shrugging:

However, as you talk about deleting a key, shouldn't the `this.form.submissions = []` in deleteAll then be replaced, too? (the splice in single-delete can be detected according to my recherche, no?)

rullzer:
> Ok. +1 That vue.set i found, too, but only tried on deleting the content...man_shrugging
> 
> However, as you talk about deleting a key, shouldn't the `this.form.submissions = []` in deleteAll then be replaced, too? (the splice in single-delete can be detected according to my recherche, no?)

^^ @skjnldsv

skjnldsv:
> However, as you talk about deleting a key, shouldn't the `this.form.submissions = []` in deleteAll then be replaced, too? (the splice in single-delete can be detected according to my recherche, no?)

Once the observer is injected by Vue, it should be fine. :)
Also arrays behave differently than object in Vue iirc. https://it.vuejs.org/v2/guide/reactivity.html

Issue #177, opened 2020-01-21
Validation of singleline/short response, url, regex...


genofire:
On a single line answer type it would be nice to have validation:
- E-Mail
- IBAN
- own Regular Expression

---
Edit (@jotoeri):
- Numeric only #605

jancborchardt:
Yes, that would be good.

"Date" should be a separate question type though just like it is in Google Forms: https://github.com/nextcloud/forms/issues/14

ei8fdb:
Crowdsignal (a service I use and pay for) implements this as predefined form fields. See below:

![Screenshot from 2020-05-06 12-54-59](https://user-images.githubusercontent.com/1899506/81174041-01221300-8f99-11ea-92a3-9501e6f2c304.png)

The labels, and format are then customisable:

![Screenshot from 2020-05-06 12-56-55](https://user-images.githubusercontent.com/1899506/81174160-39295600-8f99-11ea-8da8-b58b02f0e7a5.png)

for me, this approach makes sense as it focuses on *what I want to gather* - email, telephone, date, etc.

MrLoverLoverMMMM:
I would like to see this feature in future as well :)

andrewj720:
Would be nice to have regex allowed for individual questions, for example, I'd like to limit character length for certain questions using `[\w]{1,140}`

PR #499, opened 2020-07-10
Add types of questions

simmstein:
* add `date` question
* add `datetime` question
* add `email` question


skjnldsv:
Hey!!
Thanks for your pull request!!
I think we should re-use the existing components though. As the code is duplicated a lot, it would make more sense to pass a proper type date/datetime or whatever in the [QuestionShort component](https://github.com/nextcloud/forms/blob/master/src/components/Questions/QuestionShort.vue#L37) than having to duplicate it.

Also, ideally, for the date/time, we should use our existing datetime picker component :wink: 
https://nextcloud-vue-components.netlify.app/#/Components/DatetimePicker

simmstein:
Date and datetime may be able to have many parameters like a range. I think that they must be a different component than the QuestionShort.

I will work on it.

simmstein:
Last commits resolved a part of your comment. That could be improved but I think that it's a good start :)

skjnldsv:
> Date and datetime may be able to have many parameters like a range. I think that they must be a different component than the QuestionShort.

We could just use an `options` key or something to bind to the component in `src/models/AnswerTypes.js` ?

skjnldsv:
If you add `v-bind="answerType.attributes"` to https://github.com/nextcloud/forms/blob/8779ded0dc14f71cfb36b0e05e16e3b4b978e4c6/src/components/Questions/QuestionShort.vue#L37

And add an `attributes` key to the AnswerTypes, you can just pass whatever you want. You do not need to use new components like `QuestionDatetime.vue`, you can bind whatever you want. Range, type, etc etc

jancborchardt:
Since we run the risk of putting too much in one pull request, can we split up the question types a bit?

- Date and Time are fine to have, they are also for example in Google Forms: https://support.google.com/docs/answer/7322334
- Email and Phone number are a bit too specific, they are basically just formats of "Short answer". We could do this via details/validation of the types: #177 (or look at how Google Forms handles this).

In any case, let’s split Email/Phone out so they don’t block Date and Time. What do you think?

jancborchardt:
Also, this happens now when you click on the "Add a question" button ;)
image

jancborchardt:
By the way @simmstein, we have a "Forms team" chat on our Nextcloud Talk instance. If you like, I can invite you there with a guest account using the email address and name from your commit messages?

Also just invited you to the Nextcloud organization and the Forms team – for the upcoming pull requests, you can simply create branches of this repository and don’t need your own fork anymore. :) That’s much easier for testing and collaboration.

simmstein:
Yes, why not @jancborchardt :)

simmstein:
QuestionDate and QuestionDatetime now use @nextcloud/DatetimePicker

skjnldsv:
> If you add `v-bind="answerType.attributes"` to
> 
> https://github.com/nextcloud/forms/blob/8779ded0dc14f71cfb36b0e05e16e3b4b978e4c6/src/components/Questions/QuestionShort.vue#L37
> 
> And add an `attributes` key to the AnswerTypes, you can just pass whatever you want. You do not need to use new components like `QuestionDatetime.vue`, you can bind whatever you want. Range, type, etc etc

Don't forget this @simmstein, let's not have duplicate code :)

simmstein:
I advance step by step so a PR can me merged and then be improved.

simmstein:
This PR only contains date and datetime types. Tel and email are removed.

skjnldsv:
> I advance step by step so a PR can me merged and then be improved.

You're right, it's probably easier to not edit this pr too much :rocket:

jotoeri:
May we get those two also into one component, changing Date/Time/DateTime by the same way, as we will change validation on QuestionShort? Like this we could keep the Question-List short, to avoid (also in future, having more question-types) what Jan showed above...

I thought about maybe to have a Dropdown in Actions-Menu, to choose Validation and/or here Type of the DateTime Component? Might of course need a Actions-Dropdown Component, but that should be doable. @skjnldsv @jancborchardt @simmstein ?

simmstein:
I think that this optimization of the code is a very good way to get a complex component, hard to improve because of the idea that everything is a short question.
My point of view is quite simple: let's do improvements then refactor theme to get better code. Add a new component file (like I did) is easier that update the very same one at each new idea of field.

muppeth:
Sorry for jumping in here instead of making another feature request, but perhaps this could be achieved in this one. Could there be type to input only numbers? I run into an issue where required answer is number but some people would add words which breaks further process the csv.

skjnldsv:
> My point of view is quite simple: let's do improvements then refactor them to get better code. Add a new component file (like I did) is easier than update the very same one at every single new idea of field.

I have a different view on this, but let's also move things on shall we ;)
Will you be able to do the followups @simmstein ? :bow: 

If so, let's get this in @jotoeri :rocket:

skjnldsv:
@simmstein please rebase and auto-squash your commits into one for cleaner git history though! :wink:

skjnldsv:
> Sorry for jumping in here instead of making another feature request, but perhaps this could be achieved in this one. Could there be type to input only numbers? I run into an issue where required answer is number but some people would add words which breaks further process the csv.

COuld you search if this doesn't exists already and if not open a dedicated issue? :)
So we can keep pull requests simpler and easy to review :wink:

simmstein:
It the firs time I do rebase. I hope it's OK.

skjnldsv:
No, you did a merge, not a rebase.
Please also autosquash all your commits into one. having 20+ commits is not good for a clean git history :)

`git rebase --autosquash -i HEAD~5` 5 = number of commits in your branch.
Make sure the first shown commit is the first of this pull request.

Then you can remove the 'pick' and replace by `f`, alias for `fixup`, meaning it will drop the comment and fuse the two commits.
Capture d’écran_2020-09-15_17-31-39

Then it should (hopefully) not create conflicts (you should never do a `git pull master` into your branch, it's 100% cleaner to rebase to another branch (aka, update changes without adding those changes to this pull request)

skjnldsv:
Honestly, this is quite messy, i'd start with a new branch and cherry-pick each commits I need here.

simmstein:
see https://github.com/nextcloud/forms/pull/604.

PR #513, opened 2020-07-21
Add screenshot of response visualization, adjust readme

jancborchardt:
Some last polishing before the release, making sure people can directly see what Nextcloud Forms has to offer.

Please review @jotoeri @skjnldsv @jospoortvliet


README.md
```diff
@@ -7,10 +7,18 @@
...
 
+### Simple sharing and responding
+![](screenshots/forms2.png)
+
+### Response visualization and exporting
+![](screenshots/forms3.png)
+
 - **📝 Simple design:** No mass of options, only the essentials. Works well on mobile of course.
-- **💾 Export results:** Results can be exported as CSV in the same format used by Google Forms.
+- **📊 View & export results:** Results are visualized and can also be exported as CSV in the same format used by Google Forms.
```
jotoeri, line 20: I would leave Google out here. Ne need to stand behind them, telling we do the same. We just Export as CSV and thats good. ;)

README.md
```diff
@@ -7,10 +7,18 @@
...
+- **🔒 Data under your control!** Unlike in Google Forms, Typeform, Doodle and others, the survey info and responses are kept private on your instance.
```
jotoeri, line 21: Maybe without Doodle? Despite there seems to be some forms on doodle i think, in my mind doodle is more the Polls thing, so we should not refer to from forms-app, but refer to the polls app, if it comes to that point.

README.md
jancborchardt: The point is that we _should_ mention Google Forms, Typeform, Doodle etc so a) people know what they can compare it to and b) it will be findable. :) And also, the point of that line is that they can use it with their existing workflows which might be based on exporting from Google Forms.

README.md
jancborchardt: According to our survey, lots of people actually use Doodle for forms as well (they do have some features for it). So I'd rather we mention it. Again, also for findability.

PR #528, opened 2020-07-29
Fix substring utf8

### LINKED ISSUE #527: 2.0.0 update loose forms history
**Describe the bug**
I have update the forms app and I loose all the existing forms.
I just found a log warning message “Answer-text is too long for new Database:”
I saw that the new forms add new database tables and do a migration from the old one and I can find my old datas. The data seems to be in the new table. I can create a new form but olders are not visible.

I try to delete the new version and manually install the v1.1.1, my forms came back on my instance, but there is and error on accessing due to oc_forms_notif table not found.

**Nextcloud:**
- Nextcloud-Version: 18.0.7 stable branch
- Forms-Version: 2.0.0

jotoeri:
There are still some substr occurences on license.php, which i left out for the moment... Not sure if it is good to change anything there?

Should fix #527

### DIFF
--- a/lib/Migration/Version010200Date20200323141300.php
+++ b/lib/Migration/Version010200Date20200323141300.php
@@ -259,7 +259,7 @@ class Version010200Date20200323141300 extends SimpleMigrationStep {
 				//In case the old Question would have been longer than current possible length, create a warning and shorten text to avoid Error on upgrade.
 				if (strlen($question['form_question_text']) > 2048) {
 					$output->warning("Question-text is too long for new Database: '" . $question['form_question_text'] . "'");
-					$question['form_question_text'] = substr($question['form_question_text'], 0, 2048);
+					$question['form_question_text'] = mb_substr($question['form_question_text'], 0, 2048);
 				}
 
 				$qb_restore->insert('forms_v2_questions')
@@ -285,7 +285,7 @@ class Version010200Date20200323141300 extends SimpleMigrationStep {
 				//In case the old Answer would have been longer than current possible length, create a warning and shorten text to avoid Error on upgrade.
 				if (strlen($answer['text']) > 1024) {
 					$output->warning("Option-text is too long for new Database: '" . $answer['text'] . "'");
-					$answer['text'] = substr($answer['text'], 0, 1024);
+					$answer['text'] = mb_substr($answer['text'], 0, 1024);
 				}
 
 				$qb_restore->insert('forms_v2_options')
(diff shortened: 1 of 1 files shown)


lib/Migration/Version010200Date20200323141300.php
```diff
@@ -396,9 +396,9 @@ private function convertAccessList($accessString) : string {
 		$stringExplode = explode(';', $accessString);
 		foreach ($stringExplode as $string) {
 			if (strpos($string, 'user_') === 0) {
-				$accessArray['users'][] = substr($string, 5);
+				$accessArray['users'][] = mb_substr($string, 5);
```
nickvergessen, line 399: this is not really necessary, as you already checked that the first 5 chars are `user_` which are not utf8 sensible anyway.
Same below

lib/Controller/ApiController.php
```diff
@@ -632,7 +632,7 @@ public function getSubmissions(string $hash): Http\JSONResponse {
 			$submission['answers'] = $this->getAnswers($submission['id']);
 
 			// Append Display Name
-			if (substr($submission['userId'], 0, 10) === 'anon-user-') {
+			if (mb_substr($submission['userId'], 0, 10) === 'anon-user-') {
```
nickvergessen, line 635: same here

lib/Migration/Version010200Date20200323141300.php
jotoeri: Ah, ok, so it is only necessary, where i count and only if my text needs to be reasonable afterwards...^^ 👍

jotoeri: Any drawback of using it anyways?

jotoeri:
@skjnldsv Maybe a patch-release is useful? 🤔

skjnldsv:
> @skjnldsv Maybe a patch-release is useful?

Agreed!

lib/Migration/Version010200Date20200323141300.php
nickvergessen: its a lot slower (factor 4), but well I guess in form (migration) it doesn't matter at the moment:
```php
$start = microtime(true);

for ($i = 0; $i < 100000; $i++) {
    substr('Лорем', 2);
}

var_dump(microtime(true) - $start); // float(0.008073091506958008)

$start = microtime(true);

for ($i = 0; $i < 100000; $i++) {
    mb_substr('Лорем', 2);
}

var_dump(microtime(true) - $start); // float(0.035883188247680664)
```

Well you should use it whenever the string part you are counting/cutting contains such characters. see this:
https://3v4l.org/licri

```php
// No difference:
var_dump(strlen('user_')); // int(5)
var_dump(mb_strlen('user_')); // int(5)
var_dump(substr('user_', 0, 2)); // string(2) "us"
var_dump(mb_substr('user_', 0, 2)); // string(2) "us"
var_dump(substr('user_', 2)); // string(2) "er_"
var_dump(mb_substr('user_', 2)); // string(2) "er_"


// Difference:
var_dump(strlen('Лорем')); // int(10)
var_dump(mb_strlen('Лорем')); // int(5)
var_dump(substr('Лорем', 0, 2)); // string(2) "Л"
var_dump(mb_substr('Лорем', 0, 2)); // string(4) "Ло"
var_dump(substr('Лорем', 2)); // string(8) "орем"
var_dump(mb_substr('Лорем', 2)); // string(6) "рем"


// This breaks your database because of the invalid char
var_dump(substr('Лорем', 3)); // string(7) "�рем"
var_dump(mb_substr('Лорем', 3)); // string(4) "ем"
```

jotoeri: Interesting. Thank you! :)

PR #533, opened 2020-07-29
Increase description length

### LINKED ISSUE #529: Increase the size of the form description field (allow longer text)
- Nextcloud-Version: 19.0.0
- Forms-Version: 2.0.0

Please increase the field size of the form description. My use case is putting a legal agreement text into a form (a contributor license agreement, like the [Software Freedom Conservancy](https://docs.google.com/forms/d/e/1FAIpQLSd2FsN12NzjCs450ZmJzkJNulmRC8r8l8NYwVW5KWNX7XDiUw/viewform?hl=en_US&formkey=dFFjXzBzM1VwekFlOWFWMjFFRjJMRFE6MQ#gid=0) uses). 

Therefore I want to put the main contract text somewhere in the form, and the description seems like a good place, except that the content length is too great, and the text is currently truncated due to the arbitrary limit.

jotoeri:
Ok, just did it now... 😋 
Fixes #529

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -86,10 +86,10 @@ class PageController extends Controller {
 	 */
 	private $maxStringLengths = [
 		'formTitle' => 256,
-		'formDescription' => 2048,
+		'formDescription' => 8192,
 		'questionText' => 2048,
 		'optionText' => 1024,
-		'answerText' => 2048,
+		'answerText' => 4096,
 	];
 
 	public function __construct(string $appName,
--- a/lib/Migration/Version010200Date20200323141300.php
+++ b/lib/Migration/Version010200Date20200323141300.php
@@ -91,7 +91,7 @@ class Version010200Date20200323141300 extends SimpleMigrationStep {
 			]);
 			$table->addColumn('description', Type::STRING, [
 				'notnull' => false,
-				'length' => 2048,
+				'length' => 8192,
 			]);
 			$table->addColumn('owner_id', Type::STRING, [
 				'notnull' => true,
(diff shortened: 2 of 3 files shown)


nickvergessen:
Just make sure there is no attempt of an index on that column.

jotoeri:
There is no index on description 👍 

@skjnldsv @jancborchardt Just thought about, as we are changing lengths anyways, if we should also increase the length of the answers, too? As in #527 there was already somebody having a answer on long-text-question, longer than 2048 chars? Or even look at all sizes again? 🤔

lib/Controller/PageController.php
```diff
@@ -86,7 +86,7 @@ class PageController extends Controller {
 	 */
 	private $maxStringLengths = [
 		'formTitle' => 256,
-		'formDescription' => 2048,
+		'formDescription' => 8192,
 		'questionText' => 2048,
 		'optionText' => 1024,
 		'answerText' => 2048,
```
jotoeri, line 92: ...these lengths are currently set.

skjnldsv:
> As in #527 there was already somebody having a answer on long-text-question, longer than 2048 chars? Or even look at all sizes again?

Sure :+1: 
4096?

jotoeri:
Done 👍  The other lengths are good for you?
Also changed the old migration now, to not cut the answers and then increase the lengths again for those who did not update yet. Should work in this case to backward-change the migration.

skjnldsv:
> Done  The other lengths are good for you?

Seems good to me :shrug:

jancborchardt:
@jotoeri good stuff! :) And yes, if there are any further issues with sizes, people will report it and we can adjust again.

PR #556, opened 2020-08-11
Move to OCS API

skjnldsv:
Move to proper ocs api endpoint

Anything I should edit/improve? Any good habits to have in such case?

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -42,7 +42,9 @@ use OCA\Forms\Service\FormsService;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\IMapperException;
-use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\OCS\OCSBadRequestException;
+use OCP\AppFramework\OCS\OCSForbiddenException;
 use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IRequest;
@@ -84,6 +86,9 @@ class ApiController extends Controller {
 	/** @var FormsService */
 	private $formsService;
 
+	/** @var ISecureRandom */
+	private $secureRandom;
+
 	public function __construct(string $appName,
 								IRequest $request,
 								IUserSession $userSession,
@@ -95,7 +100,8 @@ class ApiController extends Controller {
 								OptionMapper $optionMapper,
 								ILogger $logger,
 								IL10N $l10n,
-								FormsService $formsService) {
+								FormsService $formsService,
+								ISecureRandom $secureRandom) {
 		parent::__construct($appName, $request);
 		$this->appName = $appName;
 		$this->userManager = $userManager;
@@ -109,6 +115,7 @@ class ApiController extends Controller {
 		$this->logger = $logger;
 		$this->l10n = $l10n;
 		$this->formsService = $formsService;
+		$this->secureRandom = $secureRandom;
 
 		$this->currentUser = $userSession->getUser();
 	}
@@ -118,7 +125,7 @@ class ApiController extends Controller {
 	 *
 	 * Read Form-List only with necessary information for Listing.
 	 */
-	public function getForms(): Http\JSONResponse {
+	public function getForms(): DataResponse {
 		$forms = $this->formMapper->findAllByOwnerId($this->currentUser->getUID());
 
 		$result = [];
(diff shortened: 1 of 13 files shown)


src/components/Questions/AnswerInput.vue
```diff
@@ -171,7 +171,7 @@ export default {
 		 */
 		async updateAnswer(answer) {
 			try {
-				await axios.post(generateUrl('/apps/forms/api/v1/option/update'), {
+				await axios.post(generateOcsUrl('apps/forms/api/v1/option', 2) + 'update', {
```
jotoeri, line 174: Hm, this string-concattenation to avoid the final slash is a bit strange, no? I see the reason why you did like this, but in fact it looks a bit like the base-url changes depending on what comes behind...?! GenerateUrl seems to have a better interface here... (Router-Issue!?)

-> To have a common usage within our stuff: How about using generateOcsUrl always to `'apps/forms/api/v1'` and doing everything afterwards then in the concattenation?

So in this file it would be: 
```
axios.post(generateOcsUrl('apps/forms/api/v1', 2) + 'option', {
...
axios.post(generateOcsUrl('apps/forms/api/v1', 2) + 'option/update', {
```
Indeed one would need two concattenations if using a variable, but in total we would have a common coding of generateOcsUrl and concattenation...
```
axios.delete(generateOcsUrl('apps/forms/api/v1', 2) + 'option/' + option.id)
```

jotoeri (commented):
See comments. But seems to work in total.

I stumbled a bit as we moved away from DataResponse to Http\JsonResponse and now go back. But seems to be just what is recommended in docs for ocs api.

lib/Controller/ApiController.php
```diff
@@ -277,19 +308,19 @@ public function newQuestion(int $formId, string $type, string $text): Http\JSONR
...
-			return new Http\JSONResponse(['message' => 'Invalid type'], Http::STATUS_BAD_REQUEST);
+			throw new OCSBadRequestException('Invalid type');
 		}
 
 		try {
 			$form = $this->formMapper->findById($formId);
 		} catch (IMapperException $e) {
 			$this->logger->debug('Could not find form');
-			return new Http\JSONResponse(['message' => 'Could not find form'], Http::STATUS_BAD_REQUEST);
+			throw new OCSBadRequestException();
```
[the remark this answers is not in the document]
skjnldsv: I removed it to not give insight wether the form exists or not :thinking: 
The other `OCSBadRequestException` after `findById` had no message, so I also removed those ones.
Do you think all of them should have the message?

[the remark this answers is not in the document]
jotoeri: Hm, ok. You're right. Just thought it would have been by accident...
Not sure - Might be nice to get a usable error when accessing API, and i'm not sure if it is that security-relevant to hide that an form-id exists, as you can't do much with until you aren't authorized. (At least if we don't have a bug in there... ;))
In total, to be honest, it's not that relevant for the moment, if there is a message or not...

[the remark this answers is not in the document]
skjnldsv: I'm fine with keeping the message tbh, but if we output it, better use a OCSNotFoundException then :)

src/components/Questions/AnswerInput.vue
skjnldsv: > Hm, this string-concattenation to avoid the final slash is a bit strange, no?

it is awful ^^'
It's basically what we use in lots of other apps.
I'm fine using your take on this :)

> `axios.delete(generateOcsUrl('apps/forms/api/v1', 2) + 'option/' + option.id)`

Could even be 
```js
axios.delete(generateOcsUrl('apps/forms/api/v1', 2) + `option/${option.id}`)
```

skjnldsv:
@jotoeri done :)

src/views/Results.vue
```diff
@@ -233,7 +231,7 @@ export default {
 			this.loadingResults = true
 
 			try {
-				await axios.delete(generateUrl('/apps/forms/api/v1/submission/{id}', { id }))
+				await axios.delete(generateOcsUrl('apps/forms/api/v1', 2) + `submissions/${id}`)
```
jotoeri, line 234:
```suggestion
				await axios.delete(generateOcsUrl('apps/forms/api/v1', 2) + `submission/${id}`)
```

jotoeri: Error on deleting single response...

jotoeri (approved):
Typo see below, CS fix and the commit-history is a bit mixed up now.

But apart from that 👍

src/views/Results.vue
skjnldsv: Nice catch!

PR #635, opened 2020-10-12
Inform user if the form is anonymous

### LINKED ISSUE #530: Inform user: Is it anonymous?
- Nextcloud-Version: 19.0.0
- Forms-Version: 2.0.0

Don't know if it's anonymous.

Make it clear to the user if the form is anonymous.

Nienzu:
Fix #530

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -70,10 +70,8 @@
 				@change="autoSizeDescription"
 				@keydown="autoSizeDescription"
 				@keyup="onDescChange" />
-			<!-- Only visible if at least one question is marked as mandatory-->
-			<p v-if="mandatoryUsed" class="info-mandatory">
-				* {{ t('forms', 'Required questions') }}
-			</p>
+			<!-- Generate form information message-->
+			<p class="info-message" v-text="infoMessage" />
 		</header>
 
 		<section>
@@ -197,14 +195,21 @@ export default {
 			return this.form.questions && this.form.questions.length === 0
 		},
 
-		/**
-		 * Check if at least one question is mandatory
-		 * @returns {Boolean}
-		 */
-		mandatoryUsed() {
-			return this.form.questions.reduce(
-				(isUsed, question) => isUsed || question.mandatory
-				, false)
+		isMandatoryUsed() {
+			return this.form.questions.reduce((isUsed, question) => isUsed || question.mandatory, false)
+		},
+
+		infoMessage() {
+			let message = ''
+			if (this.form.isAnonymous) {
+				message += t('forms', 'Responses are anonymous.')
+			} else {
+				message += t('forms', 'Responses are connected to your Nextcloud account.')
+			}
+			if (this.isMandatoryUsed) {
+				message += ' ' + t('forms', ' An asterisk (*) indicates mandatory questions.')
+			}
+			return message
 		},
 	},
 
@@ -446,7 +451,7 @@ export default {
 
 		.form-title,
 		.form-desc,
-		.info-mandatory {
+		.info-message {
 			width: 100%;
 			padding: 0 16px;
 			border: none;
@@ -470,7 +475,7 @@ export default {
 			resize: none;
 		}
 
-		.info-mandatory {
+		.info-message {
 			font-size: 100%;
 			padding-bottom: 20px;
 			resize: none;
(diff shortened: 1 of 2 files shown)


Nienzu:
Implement #530

src/views/Create.vue
```diff
@@ -70,9 +70,22 @@
 				@change="autoSizeDescription"
 				@keydown="autoSizeDescription"
 				@keyup="onDescChange" />
-			<!-- Only visible if at least one question is marked as mandatory-->
+			<!-- Check the form's meta information-->
 			<p v-if="mandatoryUsed" class="info-mandatory">
-				* {{ t('forms', 'Required questions') }}
+				<template v-if="isAnonymous">
```
skjnldsv, line 75: Hey!
There is a cleaner way to do that :)
Could you create a computed like `infoMessage` that returns those strings based on the `isAnonymous`/`mandatoryUsed` directly?
That would keep the template clean :wink:

skjnldsv (changes requested):

src/views/Create.vue
Nienzu: Thanks for the suggestion!

Nienzu: I use  v-text to keep the template clean.

src/views/Create.vue
```diff
@@ -197,14 +195,23 @@ export default {
...
-		 * Check if at least one question is mandatory
-		 * @returns {Boolean}
-		 */
-		mandatoryUsed() {
-			return this.form.questions.reduce(
+		infoMessage() {
+			const isMandatoryUsed = this.form.questions.reduce(
 				(isUsed, question) => isUsed || question.mandatory
 				, false)
+			if (isMandatoryUsed) {
```
skjnldsv, line 202:
```suggestion
		isMandatoryUsed() {
			return this.form.questions.reduce((isUsed, question) => isUsed || question.mandatory, false)
		},
		infoMessage() {
			if (this.isMandatoryUsed) {
```

src/views/Create.vue
```diff
@@ -197,14 +195,23 @@ export default {
...
+				} else {
+					return t('forms', 'Responses are anonymous. An asterisk (*) indicates mandatory questions.')
+				}
+			} else {
+				if (this.form.isAnonymous) {
+					return t('forms', 'Responses are connected to your Nextcloud account.')
+				} else {
+					return t('forms', 'Responses are anonymous.')
+				}
+			}
```
skjnldsv, line 214: You can use string combination to avoid making translators write 4 different translations :)

```suggestion
			let message = ''
			if (this.form.isAnonymous) {
				message += t('forms', 'Responses are connected to your Nextcloud account.')
			} else {
				message += t('forms', 'Responses are anonymous.')
			}
			if (this.isMandatoryUsed) {
				message += ' ' + t('forms', 'An asterisk (*) indicates mandatory questions.')
			}
			return message
```

Also, are you sure about `this.form.isAnonymous` ?
shouldn't we return `Responses are anonymous` when it's true instead? Missing a `!` ?

src/views/Create.vue
```diff
@@ -70,10 +70,8 @@
 				@change="autoSizeDescription"
 				@keydown="autoSizeDescription"
 				@keyup="onDescChange" />
-			<!-- Only visible if at least one question is marked as mandatory-->
-			<p v-if="mandatoryUsed" class="info-mandatory">
-				* {{ t('forms', 'Required questions') }}
-			</p>
+			<!-- Generate form information message-->
+			<p class="info-mandatory" v-text="infoMessage" />
```
jotoeri, line 74: Maybe rename also to info-message? Just for consistency...

jotoeri (commented):
Nice! 💪 
Just one small addition to previous comments. ;)

src/views/Create.vue
Nienzu: It's my mistake. Thanks!!

src/views/Create.vue
Nienzu: That sounds good. Let's do it!

jotoeri (approved):
Looks good to me 👍 🚀

skjnldsv (approved):

skjnldsv:
:rocket: :tada:

jancborchardt:
By the way @Nienzu we have a Nextcloud Talk instance for quicker communication among contributors, with channels for e.g. the Forms app, Deck app, Vue team etc. :) If you like we can add you there via the email from your commits?

Nienzu:
@jancborchardt 
Yes, it sounds cool~Thanks for your invitation : )

PR #746, opened 2021-01-21
Add csv export and prevents CSV formula injection

skjnldsv:
1. Fill a form and add `=A1` into one answer
2. Export the form results
3. Open the csv into office and see the formula outputting another cell as data

This PR fixes and prevent this from happening.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -30,6 +30,8 @@ namespace OCA\Forms\AppInfo;
 
 use OCP\AppFramework\App;
 
+include_once __DIR__ . '/../../vendor/autoload.php';
+
 class Application extends App {
 	public const APP_ID = 'forms';
 
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -29,7 +29,7 @@ namespace OCA\Forms\Controller;
 
 use DateTimeZone;
 use Exception;
-use League\Csv\EscapeFormula;
+
 use OCA\Forms\Db\Answer;
 use OCA\Forms\Db\AnswerMapper;
 use OCA\Forms\Db\Form;
@@ -45,8 +45,8 @@ use OCA\Forms\Service\FormsService;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\IMapperException;
-use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Http\DataDownloadResponse;
+use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCS\OCSBadRequestException;
 use OCP\AppFramework\OCS\OCSForbiddenException;
 use OCP\IConfig;
@@ -59,8 +59,9 @@ use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\Security\ISecureRandom;
 
-use League\Csv\Writer;
+use League\Csv\EscapeFormula;
 use League\Csv\Reader;
+use League\Csv\Writer;
 
 class ApiController extends Controller {
 	protected $appName;
@@ -143,6 +144,7 @@ class ApiController extends Controller {
 	 * @NoAdminRequired
 	 *
 	 * Read Form-List only with necessary information for Listing.
+	 * @return DataResponse
 	 */
 	public function getForms(): DataResponse {
 		$forms = $this->formMapper->findAllByOwnerId($this->currentUser->getUID());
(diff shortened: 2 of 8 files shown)


lib/AppInfo/Application.php
```diff
@@ -40,4 +44,13 @@ class Application extends App {
 	public function __construct(array $urlParams = []) {
 		parent::__construct(self::APP_ID, $urlParams);
 	}
+
+	public function register(IRegistrationContext $context): void {
```
skjnldsv, line 48: @ChristophWurst what is the min version for this? 20?

skjnldsv: Shall I go back and put this in the __construct instead?
`<nextcloud min-version="18" max-version="21" />`

skjnldsv: Resolved

nickvergessen: its 20 only, yeah

lib/Controller/ApiController.php
```diff
@@ -896,4 +914,108 @@ public function deleteAllSubmissions(int $formId): DataResponse {
...
+		}
+
+		$fileName = $form->getTitle() . ' (' . $this->l10n->t('responses') . ').csv';
+		return new DataDownloadResponse($this->array2csv($header, $data), $fileName, 'text/csv');
+	}
+
+	/**
+	 * Convert an array to a csv string
+	 * @param array $array
+	 * @return string|false
```
ChristophWurst, line 1001: no

lib/Controller/ApiController.php
```diff
@@ -896,4 +914,108 @@ public function deleteAllSubmissions(int $formId): DataResponse {
...
+	 */
+	private function array2csv(array $header, array $records): string {
+		if (count($header) === 0 && count($records) === 0) {
```
ChristophWurst, line 1004: `empty`?

ChristophWurst: also you're not allowed to return anything but a string, this will cause a TypeError

nickvergessen (approved):

jotoeri (approved):
Oh, nice! 👍 

- I think i would have produced the csv directly, instead of having the extra data array & function inbetween. It is anyways stored row by row, just as it gets read again.
- The timestamp output is quite long, probably use a shorter format there?

But both are just thoughts, nothing breaking.... 🤷‍♂️

lib/Controller/ApiController.php
```diff
@@ -896,4 +932,109 @@ public function deleteAllSubmissions(int $formId): DataResponse {
 
 		return new DataResponse($formId);
 	}
+
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
nickvergessen, line 938: Should not be needed?

jotoeri: When using an OCSController to use it as external API, then it will be necessary, no? (PR for the Controller is incoming. ;)) Or is there another way when using it public, too?

skjnldsv: Indeed

jotoeri:
Just recognized, there are errors in the log, when a question didn't get answered.

grafik

lib/Controller/ApiController.php
nickvergessen: OCSController never needs this

skjnldsv: It's needed for now. Otherwise GET fails.
We'll see once we switched to OCSController

skjnldsv:
> Just recognized, there are errors in the log, when a question didn't get answered.

Fixed

lib/Controller/ApiController.php
skjnldsv: See #747

jotoeri:
One more thing @skjnldsv 
Currently only data is escaped, where necessary. Do you know about another/additional formatter on this library to escape everything? That seems a bit more reasonable to me. Google Forms does too.

Current download:
```
"User display name",Timestamp,"Question 1","Question 2"
jonas,"Friday, January 22, 2021 at 12:47:29 AM GMT+0:00","Option 2",Answer
jonas,"Friday, January 22, 2021 at 12:45:57 AM GMT+0:00","Option 3",NextAnswer
```

What Google Forms does:
```
"Zeitstempel","Blabla","Name","E-Mail-Adresse","Adresse","Telefonnummer","Kommentare"
"2020/02/18 9:07:53 PM MEZ","","asdf","asdf@ysrt.de","here","",""
"2020/02/18 9:08:11 PM MEZ","","me","here@web.de","unde","",""
```

skjnldsv:
@jotoeri maybe it's an option in the php library we now use? :thinking:

jotoeri:
Yeah, i found a way meanwhile and tested it yesterday. I just put it on top of #780 as i moved the csv export to a service there.

PR #747, opened 2021-01-21
Fix api to be ocs

jotoeri:
Now getting a true ocs response. Unfortunately this produces a deep object concattenation, which is why i use a small wrapper to extract the actual data.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -42,7 +42,7 @@ use OCA\Forms\Db\Submission;
 use OCA\Forms\Db\SubmissionMapper;
 use OCA\Forms\Service\FormsService;
 
-use OCP\AppFramework\Controller;
+use OCP\AppFramework\OCSController;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\IMapperException;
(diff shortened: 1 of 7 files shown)


lib/Controller/ApiController.php
```diff
@@ -123,6 +123,7 @@ public function __construct(string $appName,
 
 	/**
 	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
jotoeri, line 126: @nickvergessen , if i remove this here, i get a CSRF Error.
Is there a better way to solve this then?

jotoeri: ah... wait will check if i maybe missed a header. :(

jotoeri: Ok, sorry. My fault.
Good we talked about, thank you! 😉

nickvergessen: 🦆  Rubber duck to the rescue

jotoeri:
Splitted this, as it became too much. Now it's only what is said in the title. 😉

jotoeri:
@loelkes might want to give here a review, too? :)

loelkes:
@jotoeri Awesome! I can't look right now, maybe next week?

skjnldsv (approved):

PR #749, opened 2021-01-22
Make new option return similar to new question and form

jotoeri:
- I was wondering about this while creating the docs: newOption only returned the new id, while question and form return the full object.
This adapts the newOption api to return the full object, just like newQuestion and newForm respectively do.

- Edit: Second thing i realised during tests: Due to the different naming, we had two questionId properties after creating a new answer. But that's more or less unrelated to the title, sorry for mixing this here. 🙈


skjnldsv (approved):
Fine for me :)

Issue #202, opened 2020-02-20
Duplicate/Clone Form


laarisyko:
It would be great, if one could duplicate a form, so it could be reused. We want to use the app for internal employee surveys every month. This little feature would simplify the process a lot. 

The advanced version, would be an option for time periodic forms :)

Thank you and all the best,
Lorenz

learndyanimcs:
I was reading somewhere that this feature is already available in the Beta Version. I'm currently running Nextcloud 20.0.0 Beta 4 and installed Forms 2.0.4, but there is no option to duplicate a form. Is it possible to use the current beta?

skjnldsv:
No it is not available.
See the labels and milestone on the right :) :v:

PR #756, opened 2021-01-24
Clone Forms

jotoeri:
I was just motivated to do this, so here it is... 😉 
In fact, it is a done thing, just based upon my previous PRs, as it includes API Doc and OCS-API already. Therefore as draft for now.

Implements #202

### DIFF
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -42,12 +42,9 @@
 				@click="mobileCloseNavigation">
 				{{ t('forms', 'Responses') }}
 			</ActionRouter>
-			<!-- <ActionRouter :close-after-click="true"
-				:exact="true"
-				icon="icon-clone"
-				:to="{ name: 'clone', params: { hash: form.hash } }">
-				{{ t('forms', 'Clone form') }}
-			</ActionRouter> -->
+			<ActionButton :close-after-click="true" icon="icon-clone" @click="onCloneForm">
+				{{ t('forms', 'Copy form') }}
+			</ActionButton>
 			<ActionSeparator />
 			<ActionButton :close-after-click="true" icon="icon-delete" @click="onDeleteForm">
 				{{ t('forms', 'Delete form') }}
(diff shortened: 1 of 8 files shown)


jancborchardt:
Very nice! First design feedback would be to adjust the wording to the less technical "Copy form". Yes, "Copy share link" should then be changed to "Share link" :)

Reason for the change is that "Copy form" or "Duplicate form" are much more common words than "Clone", and copy is exactly what is done, like cut/copy/paste. :)

src/components/AppNavigationForm.vue
```diff
@@ -166,6 +163,10 @@ export default {
 			this.$emit('mobile-close-navigation')
 		},
 
+		onCloneForm() {
+			this.$emit('clone', this.form.id)
+		},
+
```
skjnldsv, line 169: Please do the cloning here. Same as the delete. No need to forward it to the parent :)

skjnldsv (changes requested):

src/components/AppNavigationForm.vue
jotoeri: Hm, i just pushed a fixup, but i'm not convinced of this. We have to insert the new Form on parent into the list, so now we have some operation in two places. If we just do everything on parent, then it's all in one place. And it seems more logical to me, if new Forms are handled on the list, instead of a form handling another form...?
So if you're ok with, i'd rather delete the fixup again... 🙈

jotoeri:
So - Wording is adapted, also adapted it on all occurences of the Share Link. Therefore, on the empty-content i included the Clippy-icon, so the button is not lost with the short text... @jancborchardt Is that good like this?
grafik
grafik
grafik

src/components/AppNavigationForm.vue
skjnldsv: Ah, I get the point.
Yeah, fair enough! :)

skjnldsv (approved):

skjnldsv:
We should really have unit tests for those stuff.

Issue #364, opened 2020-05-07
Show forms that are shared with me on the navigation


jakobroehrl:
If I use one of the yellow fields the choosen user/all users should see the form but the don't.

![grafik](https://user-images.githubusercontent.com/24530680/81281335-fed3bd80-9059-11ea-88e3-f150c0d5022f.png)

**To Reproduce**
1. create a new form
2. share it to all users or to a special user
3. log in as this user
4. navigate to forms

**Expected behavior**
The shared form should be visible

skjnldsv:
> The shared form should be visible

You mean on the left navigation ?
You can access it via the link, correct?

jakobroehrl:
OK, sorry. It's working. I got it wrong. 
I thought the user/users would see the form/form results in there instance.

skjnldsv:
> I thought the user/users would see the form/form results in there instance.

Funny, we were discussing this 20min ago :stuck_out_tongue: 
I'll leave this ticket opened to keep track of it! Because yep, we definitely need to have shared-with-me forms in the navigation for logged in users! :+1:

jakobroehrl:
That's because I find https://github.com/nextcloud/forms/issues/5 important. 
With this I can share the results of an form

jmcclelland:
Hi all - thanks for the great work on this app. I wanted to clarify this issue to see if it's the same issue i have.

As I understand it, right now "sharing" a form means indicating who should be able to _fill it out._ This ticket is asking for a new feature that would allow you to share admin control over the form with other users. So, two or more logged in users could add/edit/delete questions and also have full access to the results.

Is that what this ticket is about? Or should I open a new one?

jotoeri:
Hey @jmcclelland ,

this issue is about forms, that i don't own admin-rights to, but i am only allowed to fill. These forms should be shown on the sidebar as well, to be able to access it from there, instead of only via link.

However, i think the features you want to be implemented are tracked in #455 (for accessing only results) and #322 (for collaborative editing). Feel free to use GitHubs reactions there, if you agree.
Hope that clarifies this? :)

Greets,
Jonas

jmcclelland:
Thank you - those are definitely the issues for me to track.

2epal:
please forgive me to insist.... but i can't figure out how to share the form with other users...

i want to build a form and then (without send a link) the form to show at the other users. is it possible? or not?

jotoeri:
Hey @2epal,
[...]

Issue #539, opened 2020-08-04
Forms link shows error when not logged in and shared with users


uli-heller:
**Describe the bug**

I created a form, added some users to the form, copied the link to the form into an email which I sent to the users,
The form is not anonymous. When the users click on the link, they typically see something like this:

![image](https://user-images.githubusercontent.com/1677414/89257297-2ec53980-d626-11ea-892f-33ac55f3d385.png)

It reads "form not found".

However, if they log into nextcloud before clicking on the link, everything works ok.

**To Reproduce**
1. Create a form which is not anonymous
2. Invite some users by sending them the link to the form
3. Ask them to activate the link without being logged in within nextcloud

![image](https://user-images.githubusercontent.com/1677414/89257476-85327800-d626-11ea-9346-d83998b41640.png)

**Expected behavior**

After clicking on the link, the user should be redirected to the login page and after logging in the form should show up.

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 19.0.1
- Forms-Version: 2.0.0 and 2.0.2

**Desktop (please complete the following information):**
 - OS: Linux Ubuntu-20.04
 - Browser: Chrome
 - Version: 83

**Browser log**
```
JQMIGRATE: Migrate is installed, version 1.4.1
globals.js:61 $ is deprecated: The global jQuery is deprecated. It will be updated to v2.4 in Nextcloud 20 and v3.x in Nextcloud 21. In later versions of Nextcloud it might be removed completely. Please ship your own.
$ @ globals.js:61
get @ globals.js:92
(anonymous) @ previewplugin.js?v=ed49cdf0-0:149
globals.js:61 $ is deprecated: The global jQuery is deprecated. It will be updated to v2.4 in Nextcloud 20 and v3.x in Nextcloud 21. In later versions of Nextcloud it might be removed completely. Please ship your own.
$ @ globals.js:61
get @ globals.js:92
(anonymous) @ publicpage.js?v=ed49cdf0-0:23
globals.js:61 $ is deprecated: The global jQuery is deprecated. It will be updated to v2.4 in Nextcloud 20 and v3.x in Nextcloud 21. In later versions of Nextcloud it might be removed completely. Please ship your own.
$ @ globals.js:61
get @ globals.js:92
(anonymous) @ publicpage.js?v=ed49cdf0-0:36
[...]

PR #763, opened 2021-02-01
Show shared forms on navigation

jotoeri:
Show the shared forms on navigation. Implements #364

- [x] API Docs are still missing.
- [x] Probably better split API to have getForms & getSharedForms, which might somewhen be extended with e.g. shared Results

grafik
grafik

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -25,14 +25,11 @@
 
 return [
 	'routes' => [
-		['name' => 'page#index', 'url' => '/', 'verb' => 'GET'],
-
-		// Before /{hash} to avoid conflict
-		['name' => 'page#index', 'url' => '/{hash}/edit', 'verb' => 'GET', 'postfix' => 'edit'],
-		['name' => 'page#index', 'url' => '/{hash}/results', 'verb' => 'GET', 'postfix' => 'results'],
-		['name' => 'page#index', 'url' => '/{hash}/submit', 'verb' => 'GET', 'postfix' => 'submit'],
-
+		// Before /{hash}/{action} to avoid conflict
 		['name' => 'page#goto_form', 'url' => '/{hash}', 'verb' => 'GET'],
+
+		// As parameters have defaults, this catches all routes from '/' to '/hash/edit'
+		['name' => 'page#index', 'url' => '/{hash}/{action}', 'verb' => 'GET', 'defaults' => ['hash' => '', 'action' => '']],
 	],
 	'ocs' => [
 		// Forms
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -28,10 +28,9 @@ return [
 		['name' => 'page#index', 'url' => '/', 'verb' => 'GET'],
 
 		// Before /{hash} to avoid conflict
-		['name' => 'page#index', 'url' => '/new', 'verb' => 'GET', 'postfix' => 'create'],
 		['name' => 'page#index', 'url' => '/{hash}/edit', 'verb' => 'GET', 'postfix' => 'edit'],
-		['name' => 'page#index', 'url' => '/{hash}/clone', 'verb' => 'GET', 'postfix' => 'clone'],
 		['name' => 'page#index', 'url' => '/{hash}/results', 'verb' => 'GET', 'postfix' => 'results'],
+		['name' => 'page#index', 'url' => '/{hash}/submit', 'verb' => 'GET', 'postfix' => 'submit'],
 
 		['name' => 'page#goto_form', 'url' => '/{hash}', 'verb' => 'GET'],
 	],
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -43,6 +43,7 @@ return [
 		['name' => 'api#cloneForm', 'url' => '/api/v1/form/clone/{id}', 'verb' => 'POST'],
 		['name' => 'api#updateForm', 'url' => '/api/v1/form/update', 'verb' => 'POST'],
 		['name' => 'api#deleteForm', 'url' => '/api/v1/form/{id}', 'verb' => 'DELETE'],
+		['name' => 'api#getSharedForms', 'url' => '/api/v1/shared_forms', 'verb' => 'GET'],
 
 		// Questions
 		['name' => 'api#newQuestion', 'url' => '/api/v1/question', 'verb' => 'POST'],
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -47,7 +47,7 @@
 		</AppNavigation>
 
 		<!-- No forms & loading emptycontents -->
-		<AppContent v-if="loading || noForms || (!routeHash && $route.name !== 'create')">
+		<AppContent v-if="loading || noForms || !routeHash || !routeAllowed">
 			<EmptyContent v-if="loading" icon="icon-loading">
 				{{ t('forms', 'Loading forms …') }}
 			</EmptyContent>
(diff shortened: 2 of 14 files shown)


jotoeri:
Hm, @jancborchardt are you sure, you want it to have like this? I could imagine this becomes a bit confusing if the list becomes longer. Especially if we somewhen add further lists like for Result-Sharing or a collapsed one for kind of an archive?
The old way it was collapsable from the vue-components. Wouldn't it be better if we just use the component and probably reduce the inset/change the style there a bit (effecting of course all using apps)?
grafik

jotoeri:
And for reference the same vue-components change (half inset) also on deck and polls. 😉 I really like it like this!
grafik
grafik
grafik

skjnldsv:
Should probably be "My forms" instead. cc @jancborchardt

jotoeri:
> 
> 
> Should probably be "My forms" instead. cc @jancborchardt

Yeah, if you have a look onto the first two pictures, this is already 'Your Forms' and 'Shared with you'.

I think the most important question is for now: How do we list the forms on navigation? @jancborchardt @skjnldsv 
- Collapsable Navigation Item? Maybe with a decreased inset, as on the last picture of my previous comment? @skjnldsv Do you think we would get that through on vue-components? (This is my favourite)
- Non-Collapsable just listed with AppNavigationCaption (w/o icon) or with AppNavigationItem (w/ icon)

jancborchardt:
I think this one you posted looks most proper @jotoeri? It’s similar to the Settings with "Personal" and "Admin" headings in the nav, isn’t that in the components as well? Cause the headings themselves actually do not need to be _selectable_ entries, just separators.

> image

jotoeri:
Yes, it's also a component. I would have liked more the collapsable solution. But ok. Now its the Caption-Version.

grafik
grafik

skjnldsv:
If "your forms" is empty, we should not display an empty list, I think we should add a small message?

jotoeri:
Yeah, had that also in mind... But would need probably some AppNavigationEmptyContent on components?! Seems like the AppNavigation only takes appropriate components inside the template, no?

skjnldsv:
> But would need probably some AppNavigationEmptyContent on components?!

I don't think we'd have too much usage for it. Maybe just remove the title and spacing? Just show th shared ones?

The main content should already state that you don't have any and you can create new ones, no?

jotoeri:
Yeah, you're right. The Spacer i just removed completely as i realised, the Caption does that itself. So no need to have twice a space.
grafik
grafik

appinfo/routes.php
```diff
@@ -25,19 +25,16 @@
...
-		['name' => 'page#index', 'url' => '/new', 'verb' => 'GET', 'postfix' => 'create'],
-		['name' => 'page#index', 'url' => '/{hash}/edit', 'verb' => 'GET', 'postfix' => 'edit'],
-		['name' => 'page#index', 'url' => '/{hash}/clone', 'verb' => 'GET', 'postfix' => 'clone'],
-		['name' => 'page#index', 'url' => '/{hash}/results', 'verb' => 'GET', 'postfix' => 'results'],
-
+		// Before /{hash}/{action} to avoid conflict
 		['name' => 'page#goto_form', 'url' => '/{hash}', 'verb' => 'GET'],
+
+		// As parameters have defaults, this catches all routes from '/' to '/hash/edit'
+		['name' => 'page#index', 'url' => '/{hash}/{action}', 'verb' => 'GET', 'defaults' => ['hash' => '', 'action' => '']],
```
skjnldsv, line 32: nice!

src/Forms.vue
```diff
@@ -203,10 +246,10 @@ export default {
 		 * @param {Number} id the form id
 		 */
 		async onDeleteForm(id) {
-			const formIndex = this.forms.findIndex(form => form.id === id)
-			const deletedHash = this.forms[formIndex].hash
+			const formIndex = this.forms.owned.findIndex(form => form.id === id)
+			const deletedHash = this.forms.owned[formIndex].hash
 
-			this.forms.splice(formIndex, 1)
+			this.forms.owned.splice(formIndex, 1)
```
skjnldsv, line 252: I think I would have been more confident having two separate array for this, `forms` and `sharedForms`

src/Forms.vue
```diff
@@ -106,27 +120,45 @@ export default {
 		return {
 			loading: true,
 			sidebarOpened: true,
-			forms: [],
+			forms: {
+				owned: [],
+				shared: [],
+			},
```
skjnldsv, line 126: I think I would have been more confident having two separate array for this, `forms` and `sharedForms`

I'm really not fond of nested values in data.

skjnldsv (changes requested):

src/Forms.vue
jotoeri: 🤷‍♂️ Changed that...

jotoeri:
Just thougth about the access stuff again, to fix the redirect-thing #539. As it also clarifies/secures the work & code here, i squashed it in here, instead of a separate PR, that would change the changes of this PR afterwards.
Sorry for this change after the first review. The result here just seems to be more clear & clean to me. 🙈

src/Forms.vue
```diff
@@ -26,17 +26,28 @@
...
 					@clone="onCloneForm"
 					@delete="onDeleteForm" />
+
+				<!-- Shared Forms-->
+				<AppNavigationCaption :title="t('forms', 'Shared with you')" />
+				<AppNavigationForm v-for="form in sharedForms"
+					:key="form.id"
+					:form="form"
+					:read-only="true"
+					@mobile-close-navigation="mobileCloseNavigation" />
```
skjnldsv, line 45: Maybe also hide if there is no "shared forms" ?

jotoeri: 👍

jotoeri:
Just added and already squashed one more small fix, that i just found: https://github.com/nextcloud/forms/pull/763/commits/171d43b08f5bc88d891c135358a85cd27eef64bd

src/Forms.vue
```diff
@@ -107,26 +120,58 @@ export default {
...
+				console.error('Form not found for hash: ', this.routeHash)
+				return false
+			}
+
+			// For Routes edit or results, the form must be in owned forms.
+			if (this.$route.name === 'edit' || this.$route.name === 'results') {
+				if (this.forms.findIndex(form => form.hash === this.routeHash) < 0) {
+					return false
+				}
+			}
```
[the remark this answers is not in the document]
jotoeri: Opened an issue to keep track:  #806

src/Forms.vue
```diff
@@ -107,26 +120,58 @@ export default {
...
 		}
 	},
 
 	computed: {
 		noForms() {
-			return this.forms && this.forms.length === 0
+			return this.forms && (this.forms.length === 0) && (this.sharedForms.length === 0)
+		},
+		noOwnedForms() {
+			return this.forms && (this.forms.length === 0)
```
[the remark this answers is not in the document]
jotoeri: First line here would not work i think, but included the rest: https://github.com/nextcloud/forms/pull/763/commits/e8ec9dc5c8a34c07a66627c3dc117fa272db3b0a

src/Forms.vue
```diff
@@ -151,15 +196,26 @@ export default {
...
 			} catch (error) {
 				showError(t('forms', 'An error occurred while loading the forms list'))
 				console.error(error)
-			} finally {
-				this.loading = false
 			}
+
+			// Load shared forms
+			try {
+				const response = await axios.get(generateOcsUrl('apps/forms/api/v1', 2) + 'forms/shared')
```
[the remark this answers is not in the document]
jotoeri: Yeah, i remember struggling on this, too. I changed it. 👍

src/Forms.vue
```diff
@@ -107,26 +120,61 @@ export default {
...
 			sidebarOpened: false,
 			forms: [],
+			sharedForms: [],
 		}
 	},
 
 	computed: {
 		noForms() {
-			return this.forms && this.forms.length === 0
+			return this.noOwnedForms && this.noSharedForms
```
skjnldsv, line 129: Ah right, double negative! :grin:

skjnldsv (approved):
Good for me! Please rebase :rocket:

skjnldsv:
Ah, sorry, I thought I saw a fixup! ALl good then! :shipit:

appinfo/routes.php
jotoeri: @rullzer, @nickvergessen, @ChristophWurst does one of you know about any problems with such generic route?
We have a few users now reporting they can not open the App via Navigation. But seems a bit specific to nginx-users 🤔

Should we eventually re-include the line without any parameters again?

nickvergessen: well I guess the route `index.php/apps/forms//` might not evaluate the same with all webservers and therefor result in problems later. I would suggest to just add multiple routes instead.

PR #780, opened 2021-02-16
Export csv to cloud

jotoeri:
As i have Onlyoffice meanwhile on my club-instance, i realized how useful it would be to export to cloud directly instead of down- & upload manually.

- Further included csv-escaping now here, for not to loose it during rebase...

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -58,6 +58,7 @@ return [
 		// Submissions
 		['name' => 'api#getSubmissions', 'url' => '/api/v1/submissions/{hash}', 'verb' => 'GET'],
 		['name' => 'api#exportSubmissions', 'url' => '/api/v1/submissions/export/{hash}', 'verb' => 'GET'],
+		['name' => 'api#exportSubmissionsToCloud', 'url' => '/api/v1/submissions/export', 'verb' => 'POST'],
 		['name' => 'api#deleteAllSubmissions', 'url' => '/api/v1/submissions/{formId}', 'verb' => 'DELETE'],
 
 		['name' => 'api#insertSubmission', 'url' => '/api/v1/submission/insert', 'verb' => 'POST'],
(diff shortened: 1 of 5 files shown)


jotoeri:
@jancborchardt
- Do you have some better wording for the FilePicker Title?
- It seems to me, i can't avoid people picking files on the file-picker. How would you handle the different cases? Probably show a overwrite-warning, when a file got selected? What i currently do is:
  - If people don't select a file, i just store the file `FormName (responses).csv` into the folder. If the file exists, i just overwrite. (The data should just be an extended version, why should people change the csv-data without storing it as spreadsheet.)
  - If people select a `csv`-file, i just overwrite that file.
  - If people select a non-`csv`-file, i take the folder and write the csv `FormName (responses).csv` into that folder.

grafik
grafik
grafik

jancborchardt:
> Do you have some better wording for the FilePicker Title?

Yup, an improved wording would be "Save CSV to Files" – yes, Files with capital F as it refers to the app – and using the blue icon-folder icon for the action. :)

> It seems to me, i can't avoid people picking files on the file-picker. How would you handle the different cases? Probably show a overwrite-warning, when a file got selected? 

Yeah, good question – I think the way how you handle the different cases sound good. @skjnldsv what do you think?


And an additional point: Now that we have the new "Save CSV to Files" action, we should rename the "Export to CSV" → "Download CSV"? :)

jotoeri:
Ok - Adapted that. So from my side this is ready to go then... 🎉 

grafik
grafik

jancborchardt:
Only 1 detail: "Save" instead of "Store". :)

jotoeri:
Oups... Sorry. "Wer lesen kann...." 🙈 
Corrected that.

skjnldsv:
Rebase please :)

jotoeri:
Done 👍

skjnldsv (approved):

PR #781, opened 2021-02-17
Include TopBar Share-button

### LINKED ISSUE #581: More visibility for the "Copy share link" button
**Nextcloud:**
- Nextcloud-Version: 19.0.1
- Forms-Version: 2.0.3

Related to: Frontend, UX

Hello!

Thanks for working on this app! I've been testing Nextcloud Forms with non tech-savvy people to get UX feedback.

The overall app UX is excellent and very promising, but I got multiple remarks about the "Copy share link" button, which seems hard to find for some people.

It isn't a bother to me, and I believe it's OK for most people (since the link is both in the dropdown menu in the form list in the left sidebar **and** in the form settings menu in the right sidebar), but it seems to be a concern for some people.

This form settings menu isn't open by default (and I believe it's a good thing) but near than half of the people who were testing the application made their form **without checking the settings**, so they had trouble to find the "copy link" button.

Unlike the settings that doesn't need special attention in all use cases, copying the form link is definitely a mandatory step for anyone who uses the application, exactly like checking the form answers though the "responses" button. So what about moving it into the main view?

Something like this:
image

Thanks for reading :slightly_smiling_face:

jotoeri:
One more small thing - fixes #581 
On Result-View the new button will **not** show up, if the empty-content is shown, as the empty content already includes this button. I'd like to do this also on Create-View, if the sidebar is open which contains the same button again.
grafik
grafik

Two things to clarify:
- @skjnldsv As said above, i would like to show the button only, if the sidebar is closed. So we don't have it twice directly next to each other. Do you have a simple way to examine that on Create-View? Unfortunately i don't know how to access the sidebar-open property across the router... 🤔 
- @skjnldsv @jancborchardt Is one of you aware of the current state of `@nextcloud/dialogs`? It seems the `toasts.scss` is built twice, once on forms, once somewhere else. My browser shows me two Inline-Stylesheets with the same content, except the last change on dialogs (dark-mode icon). However the build on forms seems not to be able to resolve the path to the close-icon, so this is not visible.
grafik
grafik

### DIFF
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -60,9 +60,12 @@
 
 		<!-- No errors show router content -->
 		<template v-else>
-			<router-view :form.sync="selectedForm" />
+			<router-view
+				:form.sync="selectedForm"
+				:sidebar-opened.sync="sidebarOpened" />
 			<router-view v-if="!selectedForm.partial"
 				:form="selectedForm"
+				:opened.sync="sidebarOpened"
 				name="sidebar" />
 		</template>
 	</Content>
@@ -101,6 +104,7 @@ export default {
 	data() {
 		return {
 			loading: true,
+			sidebarOpened: true,
 			forms: [],
 		}
 	},
--- a/src/components/TopBar.vue
+++ b/src/components/TopBar.vue
@@ -26,6 +26,9 @@
 <template>
 	<div class="top-bar" role="toolbar">
 		<slot />
+		<div class="top-bar__small">
+			<slot name="small" />
+		</div>
 	</div>
 </template>
 <script>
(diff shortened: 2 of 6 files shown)


skjnldsv:
> * @skjnldsv @jancborchardt Is one of you aware of the current state of `@nextcloud/dialogs`? It seems the `toasts.scss` is built twice, once on forms, once somewhere else

It might be loaded by multiple apps.
One might be the forms, the second another app :shrug:

jotoeri:
> It might be loaded by multiple apps.

Hm, but then still the one on forms does not work. 🙈 

> @skjnldsv As said above, i would like to show the button only, if the sidebar is closed. So we don't have it twice directly next to each other. Do you have a simple way to examine that on Create-View? Unfortunately i don't know how to access the sidebar-open property across the router... 🤔

@skjnldsv This is the more important question for this PR... 😉

jotoeri:
Ok, got one way. Not that big thing, just passing the prop through parent. It's indeed not as beautiful as with the event-bus, but having both, event-bus and prop would be strange, too...

So this is the sidebar now: SharLink not shown if sidebar is opened. (Except, that wording on sidebar got adapted in #756 to "Share Link")
grafik
grafik


-> Ready to go from my side.

skjnldsv (approved):

PR #789, opened 2021-02-21
Create Activities

### LINKED ISSUE #12: Notification/Activity for new form responses
1.0.3
**Nextcloud or ownCloud?:**
Nextcloud
15.0.5

A customizable notification on form submission (either to the owner or to a defined email address) would be another great feature.

jotoeri:
Now creating some Activity:
- If a form got shared to a selected user or group
- If somebody submitted to a form

For Submissions, Activities will be merged. Settings are separated into Share & Submission, so Notifications & Emails can be set up separately.

@jancborchardt As usual - i think you have some better wording for the texts? 😉
- '{user} has shared the Form {formTitle} with you'
- '{user} has shared the Form {formTitle} with Group {group}'
- '{user} answered your Form {formTitle}'
- 'A Form has been <strong>shared</strong> with you'
- 'Someone <strong>answered</strong> a form'

grafik
- Settings are listed under "other Activities" currently. I first developed it as separate section to have it clear, but then realised this needs NC20 or higher. So as soon, as we set minversion to NC20, we can change to class `ActivitySettings`.
grafik
grafik

Fixes #12 for now.


lib/Controller/ApiController.php
```diff
@@ -66,76 +67,80 @@
...
 		$this->formsService = $formsService;
-		$this->secureRandom = $secureRandom;
-		$this->dateTimeFormatter = $dateTimeFormatter;
+
 		$this->config = $config;
+		$this->dateTimeFormatter = $dateTimeFormatter;
+		$this->l10n = $l10n;
+		$this->logger = $logger;
+		$this->userManager = $userManager;
+		$this->secureRandom = $secureRandom;
```
jotoeri, line 143: These changes are basically just the insertion of the activityManger and sorting according to the includes.

skjnldsv: lol, that diff though! :see_no_evil:

jotoeri:
Just found a nice trick on Deck, to have separate implementation depending on NC-Version. Now we have the Settings without section for NC<20 and with section for NC>=20 😍

lib/Activity/Provider.php
```diff
@@ -0,0 +1,281 @@
...
+			$formLink .= $formHash;
+			if ($route !== '') {
+				$formLink .= '/' . $route;
+			}
+		} catch (IMapperException $e) {
+			// Ignore if not found, just use stored title
+		}
+
+		return [
+			'type' => 'highlight',
```
nickvergessen, line 256: I would say define yourself for the next release, simply send a PR against https://github.com/nextcloud/server/blob/master/lib/public/RichObjectStrings/Definitions.php#L222-L246 similar to Deck

lib/Activity/Settings/FormsActivitySettings.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace OCA\Forms\Activity\Settings;
+
+/** TODO
+ * Move to only have the new version, as soon as Forms minversion is set to NC20
+ * - Just rename FormsActivitySettings20 to FormsActivitySettings
+ * - Delete this file and FormsActivitySettingsLegacy
+ */
+$version = \OCP\Util::getVersion()[0];
+if ($version >= 20) {
```
nickvergessen, line 32: They shouldn't conflict, so you can implement both interfaces in one class?

jotoeri: Hmm, so as
`abstract class FormsActivitySettings extends ActivitySettings implements ISetting`
and implementing all methods? 🤔 
Or how does that work, if NC19 does not know the `ActivitySettings` yet?

nickvergessen: Ah right, well then I guess this is okay.

jotoeri:
@skjnldsv From my side this is good to go, why developing?

lib/Activity/Provider.php
jotoeri: Ref. https://github.com/nextcloud/server/pull/25758 and #803

skjnldsv:
> @skjnldsv From my side this is good to go, why developing?

Because I thought this was not done :p

nickvergessen (approved):
Looks good, didn't test.
But note you can only use the new type after a release and as far as I know apps can't require patch versions on nextclouds. I therefor would only use it from 21 onwards

jotoeri:
> you can only use the new type after a release

Yeah, i now put the existing type 'highlighted' into the parser. So when to move to the new type will now depend on how far we can backport https://github.com/nextcloud/server/pull/25758 and when we set the corresponding (major-)minversion. I have a ticket open for this.
Concerning the required patchversion then, i don't mind of telling users to just update to the newest patch on their respective major, to make activities work then...

skjnldsv (approved):

PR #795, opened 2021-02-22
Introduce php-constants, use for predefined answerTypes

### LINKED ISSUE #524: Use php constant to list available questions
@jotoeri would be nice to have those coming from something like a const somewhere instead of being hardcoded there? :)
What do you think?

https://github.com/nextcloud/forms/blob/25bd1bd84c608ed55dde8f983a2953eb92cf983d/lib/Controller/ApiController.php#L730-L732

_Originally posted by @skjnldsv in https://github.com/nextcloud/forms/pull/520#discussion_r461344331_

jotoeri:
Fixes #524

Not ideal, to have now properties defined in two places. But i don't know a way to provide the constants from js to php, neither to provide arrow-functions as initialState.
So with this, we have everything on js now, and manually sync the necessary properties to php-constants. 🤷‍♂️

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * @copyright Copyright (c) 2021 Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @author Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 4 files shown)


skjnldsv (approved):

skjnldsv (changes requested):
I'm wondering if that wouldn't be cleaner to have every constant as 
```php
public const ANSWER_TYPE_MULTIPLE = 1;
public const ANSWER_TYPE_MULTIPLE_UNIQUE = 2;
...
public const ANSWER_TYPE_MULTIPLE_LONG = 5;
```

syntax and another dedicated array with every predefined:
```php
public const ANSWER_PREDEFINED = [ANSWER_TYPE_MULTIPLE, ANSWER_TYPE_MULTIPLE_UNIQUE...];
```

jotoeri:
Hm, that would indeed make the call a little bit simpler... However as it is now, we have the same structure as in AnswerTypes.js, which is more clear to keep the files in sync? 🤔

skjnldsv:
> Hm, that would indeed make the call a little bit simpler... However as it is now, we have the same structure as in AnswerTypes.js, which is more clear to keep the files in sync?

Not sure having the same structure make sense, both handle those types very differently. I think we should go for the cleaner approach. It doesn't make sense to nest properties in php as we never iterate over them nor do we have more than one at the moment

jotoeri:
🤷‍♂️

skjnldsv (approved):

PR #801, opened 2021-02-23
Manually set Types

### LINKED ISSUE #799: Migration not working on NC18 & NC19!
Just tested it, with the new Types `Doctrine\DBAL\Types\Types` we get an Error on NC18 & NC19.
> Exception: Class 'Doctrine\DBAL\Types\Types' not found

@skjnldsv For NC18 i would say we just put it out of support for Forms `v2.2.0`?!

@ChristophWurst In general, for apps without stable branches to support all officially supported NC-versions, it would be good to backport the DB-Types! As it is a completely new feature, this should not break anything on backwards compatibility!?

jotoeri:
I'm not really a fan of this, but seems like we need to do this to fix the support for NC19.

Fixes #799

### DIFF
--- a/lib/Migration/Version010200Date20200323141300.php
+++ b/lib/Migration/Version010200Date20200323141300.php
@@ -24,7 +24,6 @@
 
 namespace OCA\Forms\Migration;
 
-use Doctrine\DBAL\Types\Types;
 use OCP\DB\ISchemaWrapper;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IConfig;
@@ -55,6 +54,11 @@ class Version010200Date20200323141300 extends SimpleMigrationStep {
 		'dropdown' => 'multiple_unique'
 	];
 
+	private const TYPE_BOOLEAN = 'boolean';
+	private const TYPE_INTEGER = 'integer';
+	private const TYPE_JSON = 'json';
+	private const TYPE_STRING = 'string';
+
 	/**
 	 * @param IDBConnection $connection
 	 * @param IConfig $config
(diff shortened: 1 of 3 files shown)


skjnldsv:
> I'm not really a fan of this, but seems like we need to do this to fix the support for NC19.

Aaaah, damn!

nickvergessen:
https://github.com/nextcloud/3rdparty/tree/stable20/doctrine/dbal/lib/Doctrine/DBAL/Types works with 20+
but inlining is also fine until you drop 19

ChristophWurst (approved):
Please open a ticket for the cleanup once anything older than 20 was dropped so you can use the new Nextcloud constants.

lib/Migration/Version010200Date20200323141300.php
```diff
@@ -55,6 +54,11 @@ class Version010200Date20200323141300 extends SimpleMigrationStep {
 		'dropdown' => 'multiple_unique'
 	];
 
+	private const TYPE_BOOLEAN = 'boolean';
+	private const TYPE_INTEGER = 'integer';
+	private const TYPE_JSON = 'json';
```
nickvergessen, line 59: :-X does this even work on all dbs? at least not in our mappers?

jotoeri: According to doctrine-docs, doctrine will automatically fall back to text in case a DB does not support json. It worked flawless until now...
I just tested this branch with clean installations of all three dabase-types and it works without problems...

skjnldsv: @nickvergessen you're worrying about oracle?

skjnldsv (approved):

lib/Migration/Version010200Date20200323141300.php
nickvergessen: Yeah

skjnldsv: Can we easily add testing ? Do you have a github workflow file on hand we could steal?

nickvergessen: https://github.com/nextcloud/notifications/blob/master/.github/workflows/oci.yml

skjnldsv: @jotoeri can you add this test to this pr? :)

skjnldsv: Then, once green, you can merge! :rocket:

PR #802, opened 2021-02-24
Fix header-height for NC19

jotoeri:
Don't know if this is the first time somebody tests forms on NC19... 😄 
These variables were not defined on NC19 yet, so our top-bar is completely hidden and people can't open the sidebar.

### DIFF
--- a/css/forms.scss
+++ b/css/forms.scss
@@ -22,6 +22,7 @@
 
 // Various variables used by this app
 :root {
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):
> These variables were not defined on NC19 yet

Ahah, that's a good one! :see_no_evil: 
Okay, let's do this

PR #815, opened 2021-02-28
Clarify translations

### LINKED ISSUE #794: Confusing german translation for results button
@rakekniven i recently wanted to see the answers / results of a survey.
I was expected a button called `Ergebnisse` or `Auswertung` or something like that, but i wasn't able to find one.

The button turned out to be at the top right - the label says `Antworten` - i didn't notice / press it because i was afraid that it would lead to a page where i can fill out the survey as a user. I thought its meaning was `Reply`.

The translation should be adjusted to reduce the confusion :wink: 

grafik

cc @jancborchardt for information

jotoeri:
Mostly changing the Responses-Button to "Results", to have a different wording than on the Summary/Responses-Tabs. This enables Translators to have different translations for the Responses-Button and the Responses-Tab. Fixes #794.
@jancborchardt Are you ok with this change 🤔?

Further i included some comments for translators, where i think it could be useful.
@rakekniven Do you have some more points, where you struggle on translations due to a missing context?

grafik
grafik
grafik

### DIFF
--- a/lib/Activity/Provider.php
+++ b/lib/Activity/Provider.php
@@ -186,6 +186,7 @@ class Provider implements IProvider {
 			return [
 				'type' => 'highlight',
 				'id' => $userId,
+				// TRANSLATORS Shown as a users display-name
 				'name' => $this->l10n->t('Anonymous user')
 			];
 		}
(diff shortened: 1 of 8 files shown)


rakekniven:
> Do you have some more points, where you struggle on translations due to a missing context?

No, not at the moment. I need to install the app and play with it to prove translations. But not today 🌞

rakekniven (approved):

skjnldsv (approved):
Wow, very fine work!! Translators will be very happy :)

PR #816, opened 2021-03-01
Disable iPhone Auto-Zoom

### LINKED ISSUE #523: Disable Auto Zoom on iPhone
Description:
When filling a form on iPhone with Safari/Opera, input zooms inand don´t go back.

To Reproduce:
1. Create a form with text fields long and short.
2. Open the form on iPhone with Safari.
3. Tap into input some text in the field, the keyboard will appear.
4. The screen zooms into the field hiding part of the label and also the "send" button.
5. Unselect filed, the keyboard will hide now.
6. Try to zoom out back (hardly possible).

Expected behavior:
Zooms out when unselected, or not zoom in (like on other smartphones)

System:
Nextcloud 19.0.1
Forms 2.0.0-beta.4
iOS 13.5.1

Reason:
As I looked for the bug and what it cause I found out that this behavior is known on iPhones: https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone
The reason is input text smaller than 16px.

Solutions:
As suggested on StackOverflow, make text bigger for Iphones or disable all zooming on the form page.

Example:
Recorded exmaple
I would try to tweak CSS myself, but not found the right way.

jotoeri:
Including a small script to disable the auto-zooming on iphones.
- I tested it on Firefox and Edge NC20 & NC21, where zooming still works.
- I also tested on Android Firefox and LineageOS Stock-Browser, where Zooming still works more or less. Means: Just as bad as before, so this PR does not change the zooming there. (One can zoom on the sidebar and navigation, but not on the main content. Seems to me like some vue-css/config issue...)
- The idea is from https://stackoverflow.com/a/57527009/14998011 and https://stackoverflow.com/a/9039885/14998011

@idecable, @kacenka49, @auviolon, @jordan-richardo Could some of you check, if this works on different iOS devices? Ideally also, if zooming still works on some Mac Desktop devices?

Should fix #523

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -77,6 +77,9 @@ class PageController extends Controller {
 	/** @var ILogger */
 	private $logger;
 
+	/** @var IRequest */
+	protected $request;
+
 	/** @var IUrlGenerator */
 	private $urlGenerator;
 
@@ -122,6 +125,7 @@ class PageController extends Controller {
 		$this->initialStateService = $initialStateService;
 		$this->l10n = $l10n;
 		$this->logger = $logger;
+		$this->request = $request;
 		$this->urlGenerator = $urlGenerator;
 		$this->userManager = $userManager;
 		$this->userSession = $userSession;
@@ -136,6 +140,7 @@ class PageController extends Controller {
 	public function index(): TemplateResponse {
 		Util::addScript($this->appName, 'forms-main');
 		Util::addStyle($this->appName, 'forms');
+		$this->insertHeaderOnIos();
 		$this->initialStateService->provideInitialState($this->appName, 'maxStringLengths', $this->maxStringLengths);
 		return new TemplateResponse($this->appName, self::TEMPLATE_MAIN);
 	}
@@ -177,6 +182,7 @@ class PageController extends Controller {
 
 		// Main Template to fill the form
 		Util::addScript($this->appName, 'forms-submit');
+		$this->insertHeaderOnIos();
 		$this->initialStateService->provideInitialState($this->appName, 'form', $this->formsService->getPublicForm($form->getId()));
 		$this->initialStateService->provideInitialState($this->appName, 'isLoggedIn', $this->userSession->isLoggedIn());
 		$this->initialStateService->provideInitialState($this->appName, 'maxStringLengths', $this->maxStringLengths);
@@ -218,5 +224,18 @@ class PageController extends Controller {
 		}
 
 		return new TemplateResponse($this->appName, $template);
+	}
+
+	/**
+	 * Insert the extended viewport Header on iPhones to prevent automatic zooming.
+	 */
+	public function insertHeaderOnIos() {
+		$USER_AGENT_IPHONE_SAFARI = '/^Mozilla\/5\.0 \(iPhone[^)]+\) AppleWebKit\/[0-9.]+ \(KHTML, like Gecko\) Version\/[0-9.]+ Mobile\/[0-9.A-Z]+ Safari\/[0-9.A-Z]+$/';
+		if (preg_match($USER_AGENT_IPHONE_SAFARI, $this->request->getHeader('User-Agent'))) {
+			Util::addHeader('meta', [
+				'name' => 'viewport',
+				'content' => 'width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1'
+			]);
+		}
 	}
 }


skjnldsv (changes requested):
Oh no, this should most likely be done with php. :see_no_evil: 
This seems like a terrible hack to have :grin:

idecable:
> Oh no, this should most likely be done with php. 🙈
> This seems like a terrible hack to have 😁

 Can you elaborate more? This is a client side detail, not a backend problem. I don't know everything by far, always learning something new, but I have experimented. 

<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0"> adding this worked for me in the template. Have not tested on Android, but will certainly try this week.

Many talked options: changing the size of the input text fields, so that it would prevent zoom, others mentioned adding some script that would ensure the page will zoom-out when you leave the text input fields.. I'm open to all approaches, as long it does the job. I was up to a point where I wanted to attempt moving the Submit button to the left side as opposed to the right side. So many different ways. 

But yes by all means, let the best tweak win. 😁

idecable:
Wow the PageController edit is very well done. Now I want to learn more PHP

jotoeri:
Hmm, indeed, this works quite easy on php...^^
- I'm not sure, if the Regex is ideal. Maybe somebody with a bit more of Regex-Experience could double-check that. I checked it with several strings, it seems to work so far... 🤔 https://regex101.com/r/wepuAT/1 with strings from: [beamtic.com unknown?](https://beamtic.com/user-agents/?browser=Safari) and with [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#safari_ua_string)
- The viewport string is the one, that comes already with NC, just extended with the `maximum-scale`-term.

jotoeri:
@idecable did you check already on iOS if this works? 👀

idecable:
> @idecable did you check already on iOS if this works? 👀

I didn't apply your fix to my setup yet, but I like how you structured this up. 

I could create an instance with your added code in a few days, and test on Android emulator as well.

jotoeri:
> I could create an instance with your added code in a few days, and test on Android emulator as well.

That would be nice 😊 💪

skjnldsv:
> * I'm not sure, if the Regex is ideal. Maybe somebody with a bit more of Regex-Experience could double-check that. I checked it with several strings, it seems to work so far... thinking [regex101.com/r/wepuAT/1](https://regex101.com/r/wepuAT/1) with strings from: [beamtic.com unknown?](https://beamtic.com/user-agents/?browser=Safari) and with [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#safari_ua_string)

We have existing user agents constants if you want :)
https://github.com/nextcloud/server/blob/eb502c02ff7693bb36318d857985f79e7bac370c/lib/private/AppFramework/Http/Request.php#L67-L79

jotoeri:
> We have existing user agents constants if you want

Ahh, that's good, yes 👍. However the safari-constant does not match for iPhones, as the Mobile-Part is missing. I used now the syntax from the constant, but included the `iPhone` and the `Mobile`-Part.

skjnldsv (approved):

jotoeri:
Ok, so i meanwhile created a testinstance:
@idecable, @kacenka49, @auviolon, @jordan-richardo, would be cool, if some of you could check on these two forms, if the fix works:
- First one is on master branch, which should still contain the bug (for reference): http://h2643730.stratoserver.net/testcloud/index.php/apps/forms/je6MH2PEcpzP8CBS
- Second one is on this bugfix-branch, which hopefully does not show the bug anymore: http://h2643730.stratoserver.net/testcloud2/index.php/apps/forms/oqTwPeCk3xaYoS6a

jotoeri:
Ok, got one confirmation on the forum, another one by a friend of mine. This works, so let's go here. 👍 🙂

idecable:
> Ok, so i meanwhile created a testinstance:
> @idecable, @kacenka49, @auviolon, @jordan-richardo, would be cool, if some of you could check these two forms on your iPhone, for if the fix works:
> 
> * First one is on master branch, which should still contain the bug (for reference): http://h2643730.stratoserver.net/testcloud/index.php/apps/forms/je6MH2PEcpzP8CBS
> * Second one is on this bugfix-branch, which hopefully does not show the bug anymore: http://h2643730.stratoserver.net/testcloud2/index.php/apps/forms/oqTwPeCk3xaYoS6a

> 

Are these links offline? Unable to load both.

jotoeri:
Ah, yes sorry, so i was just two hours too fast. The server got deleted, as it was an old one, which i could just quickly use for a public link, before it will die. 🙈 
But it's merged now, just wait two days and you'll have it on your own instance through the app-store. 😉 🚀

PR #820, opened 2021-03-02
Use new RichObject

jotoeri:
As @nickvergessen suggested, this now uses the `forms-form` RichObject by default, but validates it and falls back to `highlight` in case our definition does not exist yet. ☺️

### DIFF
--- a/lib/Activity/Provider.php
+++ b/lib/Activity/Provider.php
@@ -35,6 +35,8 @@ use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IURLGenerator;
 use OCP\IUserManager;
+use OCP\RichObjectStrings\InvalidObjectExeption;
+use OCP\RichObjectStrings\IValidator;
 
 class Provider implements IProvider {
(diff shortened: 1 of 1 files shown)


lib/Activity/Provider.php
```diff
@@ -246,20 +253,30 @@ private function getRichFormTitle(string $formTitle, string $formHash, string $r
...
 			'id' => $formHash,
 			'name' => $formTitle,
 			'link' => $formLink
 		];
+
+		try {
+			// Validating a dummy-subject, which uses the richFormTitle.
+			$this->validator->validate('{form}', ['form' => $richFormTitle]);
+		} catch (InvalidObjectExeption $e) {
+			// In case this did not work (type does not exist for <NC21.0.1), fall back to type highlight
```
nickvergessen, line 275: Add a fixme to drop it when support for 21 is dropped

nickvergessen (approved):

skjnldsv (approved):

PR #845, opened 2021-03-11
Fix activity l10n & make codecov happy

jotoeri:
Just started working from time to time a bit on the tests... As i have no experience yet on this, probably some view also on logical side would be good here in detail...

### DIFF
--- a/tests/Unit/Activity/ProviderTest.php
+++ b/tests/Unit/Activity/ProviderTest.php
@@ -1,0 +1,452 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2021 Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @author Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OCA\Forms\Tests\Unit\Activity;
+
+use OCA\Forms\Activity\Provider;
+
+use OCA\Forms\Db\Form;
+use OCA\Forms\Db\FormMapper;
+use OCP\Activity\IEvent;
+use OCP\Activity\IEventMerger;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\IGroup;
+use OCP\IGroupManager;
+use OCP\IL10N;
+use OCP\ILogger;
+use OCP\IURLGenerator;
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\L10N\IFactory;
+use OCP\RichObjectStrings\InvalidObjectExeption;
+use OCP\RichObjectStrings\IValidator;
+use PHPUnit\Framework\MockObject\MockObject;
+use Test\TestCase;
+
+class ProviderTest extends TestCase {
+
+	/** @var Provider */
+	private $provider;
+
+	/** @var FormMapper|MockObject */
+	private $formMapper;
+
+	/** @var IEventMerger|MockObject */
+	private $eventMerger;
+
+	/** @var IGroupManager|MockObject */
+	private $groupManager;
+
+	/** @var IL10N|MockObject */
+	private $l10n;
+
+	/** @var ILogger|MockObject */
+	private $logger;
+
+	/** @var IURLGenerator|MockObject */
+	private $urlGenerator;
+
+	/** @var IUserManager|MockObject */
+	private $userManager;
+
+	/** @var IFactory|MockObject */
+	private $l10nFactory;
+
+	/** @var IValidator|MockObject */
+	private $validator;
+
+	public function setUp(): void {
+		$this->formMapper = $this->createMock(FormMapper::class);
+		$this->eventMerger = $this->createMock(IEventMerger::class);
+		$this->groupManager = $this->createMock(IGroupManager::class);
+		$this->logger = $this->createMock(ILogger::class);
+		$this->urlGenerator = $this->createMock(IURLGenerator::class);
+		$this->userManager = $this->createMock(IUserManager::class);
+		$this->l10nFactory = $this->createMock(IFactory::class);
+		$this->validator = $this->createMock(IValidator::class);
+
+		$this->urlGenerator->expects($this->any())
+			->method('linkToRouteAbsolute')
+			->with('forms.page.index')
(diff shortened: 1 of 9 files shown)


skjnldsv (approved):
And tests?! Oh my! :cupid: :rocket:

jotoeri:
@nickvergessen, as these are my first written Tests... Could you have a short look onto the basic handling? Is that good? Are there generally better approaches, that i missed? Just to have an idea, and not to produce shitty tests from now on...

jotoeri:
... just pushed a fixup, as i learned one should use use Nextclouds `Test\TestCase` instead of `PHPUnit\Framework\TestCase`

tests/Unit/Activity/Settings/FormsActivitySettingsTest.php
```diff
@@ -0,0 +1,84 @@
...
+		$this->assertEquals(true, $this->activitySettings->isDefaultEnabledNotification());
+	}
+
+	public function testCanChangeMail() {
+		$this->assertEquals(true, $this->activitySettings->canChangeMail());
+	}
+
+	public function testIsDefaultEnabledMail() {
+		$this->assertEquals(false, $this->activitySettings->isDefaultEnabledMail());
+	}
```
nickvergessen, line 83: Personal preference I developed in the meantime. tests for "empty" functions which are just `return <fixed value>;` I don't write any tests for. The methods have a return type specified that is good enough.

tests/Unit/Activity/ProviderTest.php
```diff
@@ -0,0 +1,377 @@
...
+	private $userManager;
+
+	/** @var IValidator|MockObject */
+	private $validator;
+
+	public function setUp(): void {
+		$this->formMapper = $this->createMock(FormMapper::class);
+		$this->eventMerger = $this->createMock(IEventMerger::class);
+		$this->groupManager = $this->createMock(IGroupManager::class);
+		$this->l10n = $this->createMock(IL10N::class);
```
nickvergessen, line 79: Actually this made me discover a bug.

So you are not using the `$language` variable handed in by the parse method:
https://github.com/nextcloud/forms/blob/master/lib/Activity/Provider.php#L90

That means all your strings when sent in a background job will be english.
Inject `OCP\L10N\IFactory` instead, in `parse(...)` get the requested language: `$l = $this->l10nFactory('forms', $language`);` and inject that in all methods that currently use `$this->l10n`

tests/Unit/Activity/ProviderTest.php
```diff
@@ -0,0 +1,377 @@
...
+	public function testGetSubjectString() {
+		$this->l10n->expects($this->exactly(3))
+			->method('t')
+			->will($this->returnCallback(function (string $identity) {
+				return $identity;
+			}));
+
+		$this->assertEquals('{user} has shared the form {formTitle} with you', $this->provider->getSubjectString('newshare'));
+		$this->assertEquals('{user} has shared the form {formTitle} with group {group}', $this->provider->getSubjectString('newgroupshare'));
+		$this->assertEquals('{user} answered your form {formTitle}', $this->provider->getSubjectString('newsubmission'));
```
nickvergessen, line 201: Use a data provider instead:
https://github.com/nextcloud/spreed/blob/master/tests/php/Activity/Provider/InvitationTest.php#L126-L141

```suggestion
	public function dataGetSubjectString() {
		return [
			['newshare', '{user} has shared the form {formTitle} with you'],
			...
		];
	}

	/**
	 * @dataProvider dataGetSubjectString
	 *
	 * @param string $subject
	 * @param string $expected
	 */
	public function testGetSubjectString(string $subject, string $expected) {
		$this->l10n->expects($this->exactly(3))
			->method('t')
			->will($this->returnCallback(function (string $identity) {
				return $identity;
			}));
		$this->assertEquals($expected, $this->provider->getSubjectString($subject));
```

tests/Unit/Activity/ProviderTest.php
```diff
@@ -0,0 +1,377 @@
...
+		$this->expectException(\InvalidArgumentException::class);
+		$this->provider->getSubjectString('someUnknownSubject');
```
nickvergessen, line 203: move to a new test.

tests/Unit/Activity/ProviderTest.php
```diff
@@ -0,0 +1,377 @@
...
+			'formHash' => 'abcdefg',
+			'groupId' => 'someGroup'
+		]));
+		$this->assertEquals($newSubmissionResult, $this->provider->getRichParams('newsubmission', [
+			'userId' => 'affectedUser',
+			'formTitle' => 'Some FormTitle',
+			'formHash' => 'abcdefg'
+		]));
+		$this->assertEquals([], $this->provider->getRichParams('unknownSubject', []));
+	}
```
nickvergessen, line 299: same here, data provider

tests/Unit/Activity/ProviderTest.php
```diff
@@ -0,0 +1,377 @@
...
+			'id' => 'anon-user-xyz',
+			'name' => 'Anonymous user'
+		]);
+
+		// User not found
+		$this->userManager->expects($this->once())
+			->method('get')
+			->with('someDeletedUser')
+			->willReturn(null);
+		$this->assertEquals($this->provider->getRichUser('someDeletedUser'), [
```
nickvergessen, line 324: data provider

tests/Unit/Activity/ProviderTest.php
nickvergessen: if your array ever grows big and you dont know what case 0 etc was testing for, you can use an array string key as description in new line 194, see e.g. https://github.com/nextcloud/spreed/blob/master/tests/php/Notification/NotifierTest.php#L467

tests/Unit/Activity/ProviderTest.php
```diff
@@ -0,0 +1,377 @@
...
+			'type' => 'highlight',
+			'id' => 'abcdefg',
+			'name' => 'Some Form Title',
+			'link' => 'http://localhost/apps/forms'
+		], $this->provider->getRichFormTitle('Some Form Title', 'abcdefg'));
+	}
+
+	// Test right event-icon
+	public function testGetEventIcon() {
+		$this->assertEquals('http://localhost/core/img/actions/shared.svg', $this->provider->getEventIcon('newshare'));
```
nickvergessen, line 372: data provider

tests/Unit/Activity/Settings/NewSubmissionTest.php
```diff
@@ -0,0 +1,72 @@
...
+
+	public function testGetIdentifier() {
+		$this->assertEquals('forms_newsubmission', $this->newShare->getIdentifier());
+	}
+
+	public function testGetName() {
+		$this->l10n->expects($this->once())
+			->method('t')
+			->with('Someone <strong>answered</strong> a form')
+			->willReturn('Some translated Text');
```
nickvergessen, line 65: I always use this dummy which also works with parameters in case you ever have to use some:
https://github.com/nextcloud/spreed/blob/master/tests/php/Notification/NotifierTest.php#L356-L360

nickvergessen:
Hope the comments help, if you need further clarification let me know.
Sorry if they sound rude/rough because I just dumped them in without much prosa 🙊

jotoeri:
Wow, thank you! :heart:
No Problem. I'll let you know in case i don't understand sth. 😊

jotoeri:
Okay, so fixed the translation of activity, and adapted the provider-test. Should be good to go now from my side.
Thanks again, Joas, for the Feedback! ☺️

skjnldsv (approved):

nickvergessen (approved):

PR #846, opened 2021-03-12
Fix Routing

jotoeri:
Seems like nginx appends a dash to the base-route, which always was a problem on apache. Don't know why we never ran into these problems.
This way now indeed feels a bit hacky on the page-controller, but the routes seem to be logical and it seems to work on both nginx/apache. 🤷‍♂️

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -25,11 +25,12 @@
 
 return [
 	'routes' => [
-		// Before /{hash}/{view} to avoid conflict
+		// Internal views
+		['name' => 'page#views', 'url' => '/{hash}/{view}', 'verb' => 'GET'],
+		// Share-Link & public submit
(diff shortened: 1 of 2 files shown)


appinfo/routes.php
```diff
@@ -25,11 +25,12 @@
 
 return [
 	'routes' => [
-		// Before /{hash}/{view} to avoid conflict
+		// Internal views
+		['name' => 'page#views', 'url' => '/{hash}/{view}', 'verb' => 'GET'],
```
skjnldsv, line 29: That's why I had my trick before! :wink:

skjnldsv: https://github.com/nextcloud/forms/blob/df243e4493bfecbb990f7f2355fc0391749354bb/appinfo/routes.php#L28-L33

skjnldsv (approved):

Issue #828, opened 2021-03-07
Delete a users Forms, when the user got deleted


jotoeri:
> interesting side effect, deleting as admin a user with his data from the nextcloud instance does not remove the users information from the polls tables …
Not really GDPR compliant. 

Just saw that comment on the forum. Applies to forms as to polls. I think we should delete a users forms when the user gets deleted.

Maybe one should also think about making corresponding submissions anonymous, if a user gets deleted.

skjnldsv:
There is an event you can listen too iirc.
We can also decide to add a cron task to check and scale in case there are a lot of data.

jotoeri:
Probably some combination...!? I think only using a cron would be quite some overhead to permanently check all forms. Having some listener to the event, that then tells our cron job to delete for that specifc user would be a good combination i think.
We'll see when implementing this... 😉

skjnldsv:
Yes, maybe just check if the user have a lot of forms and send that to the cron Job if too much?
Don't know the proper approach for such scaling... Maybe @rullzer or @ChristophWurst can help?

rullzer:
Don't do to much magic.

Listen to the event => inject job
Cron jobs are run often. So it should pick it up soon to delete it.

PR #856, opened 2021-03-15
Delete a deleted Users Forms

jotoeri:
As we talked about, i was interested to do this. 😄 And it was just too easy to implement. 🙈 
Basically a done PR here, i'd just still wait for a while until we don't see urgent bugs for v2.2.2 anymore. 😉 

Implements #828

### DIFF
--- a/lib/Db/FormMapper.php
+++ b/lib/Db/FormMapper.php
@@ -31,14 +31,23 @@ use OCP\IDBConnection;
 use OCP\AppFramework\Db\QBMapper;
 
 class FormMapper extends QBMapper {
+	/** @var QuestionMapper */
+	private $questionMapper;
+
+	/** @var SubmissionMapper */
+	private $submissionMapper;
 
 	/**
 	 * FormMapper constructor.
 	 *
 	 * @param IDBConnection $db
 	 */
-	public function __construct(IDBConnection $db) {
+	public function __construct(QuestionMapper $questionMapper,
+								SubmissionMapper $submissionMapper,
+								IDBConnection $db) {
 		parent::__construct($db, 'forms_v2_forms', Form::class);
+		$this->questionMapper = $questionMapper;
+		$this->submissionMapper = $submissionMapper;
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


skjnldsv:
Is this ready for reviews?

jotoeri:
It is, yes.
I just didn't want to merge until we know, we don't need a `v2.2.3`. And i didn't know how to mark that ideally.^^

lib/Db/FormMapper.php
```diff
@@ -107,4 +116,15 @@ public function findAllByOwnerId(string $ownerId): array {
...
+
+	/**
+	 * Delete a Form including connected Questions and Submissions
+	 * @param Form $form The form instance to delete
+	 */
+	public function deleteForm(Form $form): void {
+		// Delete Submissions(incl. Answers), Questions(incl. Options) and Form.
+		$this->submissionMapper->deleteByForm($form->getId());
+		$this->questionMapper->deleteByForm($form->getId());
+		$this->delete($form);
```
skjnldsv, line 128: Could there be an error here? If so do we need to catch and log?
Maybe also just log debug by default as well?

jotoeri: Logging is done one level higher on API and BackgroundJob respectively.
Regarding errors, this should be ok, as we only delete, what we first got from db. So as long as there is no kind of 'read-only'-Db that should work. 😉

skjnldsv (approved):

skjnldsv:
Rebase and squash?

skjnldsv (approved):

skjnldsv:
php cs lint fails

jotoeri:
That's the php-cs-fixer bug... So wie can either ignore it for now, or i'll rebase on the weekend onto #864 😉

jotoeri:
Rebase is done meanwhile, but please don't merge, until #871 is merged and released. 😞

skjnldsv (changes requested):
Blocking then

rullzer:
So good to go now?

jotoeri:
Yeah, should be good... 👍  I just hope we don't need a 2.2.5 🙈

skjnldsv (approved):

jotoeri:
Just rebased, so i'd say lets go here. 😉

PR #871, opened 2021-03-27
Fix export again :(

jotoeri:
Just created a new error on #860 😭 
The export works, but though throws an error on each single submission&question, if the index on `$carry[$questionId] .= $an...` does not exist yet. Looks really scary on the log, but creates no problem on functionality. 😄 

Also including a test now. 💪


skjnldsv (approved):

skjnldsv:
Code looks good, phpunit says no :)

jotoeri:
Yeah, as the test was not done at all, yet. Should be good now 👍

PR #882, opened 2021-04-04
Refactor Mandatory

### LINKED ISSUE #521: Rename  `mandatory` prop to `isMandatory`
https://github.com/nextcloud/forms/blob/53b6851a79984ba24a873e3423bb15b3fb386002/src/mixins/QuestionMixin.js#L48-L54

To be consistent with other variables, because it looks weird here
https://github.com/nextcloud/forms/blob/d9ff7a180a045404b8eddf1fc5a9631f6bf0527e/src/components/Questions/QuestionDropdown.vue#L192-L203

Chartman123:
Fixes #521 

- [x] Rename all mandatory to isRequired
- [x] New DB-Migration, renaming the Column
- [x] Compatibility of API on getQuestion() and updateQuestion()
- [x] Versioning of API to v1.1
- [x] Documentation of API & DataStructure
- [x] Changelog with Deprecation-Warning


src/components/Questions/QuestionLong.vue
```diff
@@ -24,21 +24,21 @@
...
-		@update:mandatory="onMandatoryChange"
+		@update:isRequired="onRequiredChange"
 		@delete="onDelete">
 		<div class="question__content">
 			<textarea ref="textarea"
 				:aria-label="t('forms', 'A long answer for the question “{text}”', { text })"
 				:placeholder="submissionInputPlaceholder"
 				:disabled="!readOnly"
-				:required="mandatory"
+				:isRequired="isRequired"
```
jotoeri, line 41: The `:required=` on the old code here is the HTML-Property `required` of the text-field. Onto this we apply our `isRequired` data.
So we need in the new way a
```
:required="isRequired"
```

src/components/Questions/Question.vue
```diff
@@ -43,16 +43,16 @@
 				type="text"
 				minlength="1"
 				:maxlength="maxQuestionLength"
-				required
+				isRequired
```
jotoeri, line 46: This is equivalent to a fixed `required=true`. So this must remain as `required`, which here sets the question-title as required. Basically - a question always needs a title.

src/components/Questions/QuestionDropdown.vue
```diff
@@ -33,13 +33,13 @@
...
 		:shift-drag-handle="shiftDragHandle"
 		@update:text="onTitleChange"
-		@update:mandatory="onMandatoryChange"
+		@update:isRequired="onRequiredChange"
 		@delete="onDelete">
 		<select v-if="!edit"
 			:id="text"
 			:name="text"
 			:multiple="isMultiple"
-			:required="mandatory"
```
jotoeri, line 42: Same here 😉

src/components/Questions/QuestionMultiple.vue
```diff
@@ -49,7 +49,7 @@
 							'checkbox question__checkbox': !isUnique,
 						}"
 						:name="`${id}-answer`"
-						:required="isRequired(answer.id)"
```
jotoeri, line 52: dito 😉

src/components/Questions/QuestionShort.vue
```diff
@@ -24,21 +24,21 @@
...
-		@update:mandatory="onMandatoryChange"
+		@update:isRequired="onRequiredChange"
 		@delete="onDelete">
 		<div class="question__content">
 			<input ref="input"
 				:aria-label="t('forms', 'A short answer for the question “{text}”', { text })"
 				:placeholder="submissionInputPlaceholder"
 				:disabled="!readOnly"
-				:required="mandatory"
+				:isRequired="isRequired"
```
jotoeri, line 41: 🔼

src/components/Questions/QuestionDate.vue
```diff
@@ -103,7 +103,7 @@ export default {
 		 */
 		inputAttr() {
 			return {
-				required: this.mandatory,
+				isRequired: this.isRequired,
```
jotoeri, line 106: This is some kind of a special case, but basically the same... The returned object {} will be passed to the datePicker Component, which will then apply all of these attributes directly onto the HTML-Input. Thus we also need the `required:this.isRequired` here.

jotoeri (commented):
So what i saw already:
We have to differentiate between the HTML-Property `required`, which finally applies the setting onto the textfield/input-element, and our own property within the data-model of a question. We can only rename our own property, the HTML-Property is fixed to `required`. 

So our question data-object stores some data as json, containing also the current `mandatory` boolean value. If we then want to apply this property (the specific question-instance is mandatory or not) onto the respective input-field, we pass the stored mandatory-value to the HTML-Property `required`
```
{
  id: 1,
  text: 'Some nice question-title',
  mandatory: false,
  ....
}
```
```
<input :required="mandatory" />
```
which will for this example then render as:
```
<input required="false" />
```

src/components/Questions/QuestionLong.vue
Chartman123: I reverted the changes in question :-) Thanks for clearing that up!

src/components/Questions/Question.vue
Chartman123: I reverted the changes in question :-) Thanks for clearing that up!

Chartman123:
Ok, I reverted the changes regarding the HTML properties now... Thank you for helping me understand what the code does.

However, I still don't know what needs to be done to also change the code of the API or the database. You proposed this:

> However - what just came into my mind: We could rename it everywhere on our code, and copy it on public API to a second prop of the old naming just before the API-Response is sent. It's a bit hacky, but such we don't break the API, can release the change with the next minor and just state the `mandatory` prop as deprecated and to be removed with v3.0.

So by now, I tried not to touch the code that communicates with the database and keep the word `mandatory` there.

Chartman123:
I also get some Lint warnings: Attribute ':isRequired' must be hyphenated

Should I change that to `is-required` or `required`?

jotoeri:
> 
> 
> I also get some Lint warnings: Attribute ':isRequired' must be hyphenated
> 
> Should I change that to `is-required` or `required`?

Yep, that needs a `is-required="isRequired"`, while the `isRequired` then is the property on e.g. the QuestionShort-Component and is passed via the `is-required` to the Question-Component, where it then is again written as `isRequired`. 🤷‍♂️ Honestly i also don't fully understand why this rule is given on lint. 😄

jotoeri:
> So by now, I tried not to touch the code that communicates with the database and keep the word mandatory there.

So my idea was to rename basically everything from mandatory to isRequired. Also the DB-Column, which will need a migration.
Then on API-Controller, `getForm()`, we just copy the property isRequired to a property mandatory `$form['mandatory']=$form['isRequired']`, so API will return an Object like the following, where mandatory/isRequired is of course redundand, but people can still use the old mandatory, while we already use the new isRequired.
```
{
  id: 5,
  title: 'This is a form title',
  isRequired: true,
  mandatory: true,
  ...
}
```

- Same applies of course to the `updateForm()` procedure, where we have to ensure, that if somebody wants to update the mandatory-property, we will store it into our isRequired. Maybe we can even find a way to return a deprecation-warning in this case.


So - just to have an overview:
- [ ] Rename all mandatory to isRequired
- [ ] New DB-Migration, renaming the Column
- [ ] Compatibility of API on getForm() and updateForm()
- [ ] Versioning of API to v1.1
- [ ] Documentation of API & DataStructure
- [ ] Changelog with Deprecation-Warning

Chartman123:
Ok, fine. I think we could split that to a second PR. This one should be working now. I changed everything to `isRequired` or `is-required` where possible without messing with the API or the database. I also tested it locally in a Nextcloud test instance. No more errors in the browser console so far... 👍🏻

jotoeri:
Hm, sorry, but honestly i'm no fan of splitting this. I think we should do it on this branch until it is fully done, stable and documented. Otherwise we just insert even more inconsistency and probably nobody will care for to fix this. And this is one of the key features of Git: to develop in parallel until a feature is fully implemented (reviewed) and stable, and then to merge.

Chartman123:
Ok 🤣  So, do I need to add my changed line as a second line and keep the original line in Db/Question.php so that the response will have both props?

I'm currently trying to create a new migration step. Is there a special way to create the new file regarding the file name?

Chartman123:
> So my idea was to rename basically everything from mandatory to isRequired. Also the DB-Column, which will need a migration.
> Then on API-Controller, `getForm()`, we just copy the property isRequired to a property mandatory `$form['mandatory']=$form['isRequired']`, so API will return an Object like the following, where mandatory/isRequired is of course redundand, but people can still use the old mandatory, while we already use the new isRequired.

Don't you mean getQuestion() and updateQuestion()?

jotoeri:
> add my changed line as a second line

That would probably work afaik, but just as long as we do nothing else with that property within our code (which is indeed unlikely for that specific property). However, to have it clear, i would still do it only right before the API-Response is sent (so just before the return in getQuestion()), so we only have the doubling on API and won't have to take care of the doubled information within our own code.

> create a new migration step

Yes, there is an OCC-Command to create migration-steps... Just call it with version 020300 then...
And if i remember correctly, we will need a migration step to create the new column, then copy the data in the postProcessing and in a second step we can then delete the old column. Unfortunately there is no renameColumn procedure...

> Don't you mean getQuestion() and updateQuestion()?

Ahh, yes you're right. 🙈

Chartman123:
> Just call it with version 020300 then...

Is it ok to have two migration steps with the same version number and only a different date? I saw that you generated a migration step with this version, too.

jotoeri:
That works, yes. Afaik they could theoretially(!) even be named anyways, they will just be executed once in alphabetical order...
For testing purposes you can btw. just remove the resp. migration from oc_migrations table on DB and reset the forms version on oc_appconfig (or increase the version on forms info.xml). Then it should just run again...

Chartman123:
I faced another problem now while testing the API changes:

Doctrine\DBAL\Driver\PDO\Exception: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'is_required' in 'field list'

So `this.saveQuestionProperty('isRequired', isRequiredValue)` seems to try to save the value to the column "is_required" instead of "isRequired". Is it possible to use our column name or should we rename the DB column to is_required?

jotoeri:
Yeah, that's all a bit strange tbh. 
We need a **sub**-hyphenated (don't know the correct english term?!) version `is_required` on DB, while on the DB-Entity it is already the upper/lower version `isRequired`. You can compare also e.g. to the `is_anonymous`/`isAnonymous` property for forms.
And then it is through all the code the upper-lower version, except then the already discussed Property-passes on vue again. ^^ https://github.com/nextcloud/forms/blob/c8f371da5d2f4fdd1383eb9df3704ed1e99df5d5/lib/Migration/Version010200Date20200323141300.php#L116 https://github.com/nextcloud/forms/blob/c8f371da5d2f4fdd1383eb9df3704ed1e99df5d5/lib/Db/Form.php#L60

lib/Migration/Version020300Date20210406114130.php
```diff
@@ -0,0 +1,74 @@
...
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('forms_v2_questions');
+
+		$table->addColumn('isRequired', self::TYPE_BOOLEAN, [
+			'notnull' => false,
+			'default' => 0,
+		]);
```
jotoeri, line 62: Now i would surround that with an `if(!$table->hasColumn('is_required')) `, so it will only get executed, if the column does not exist yet.
That makes it a bit more failsafe and will enable us to also change the inital table-creation.

lib/Migration/Version020300Date20210406133704.php
```diff
@@ -0,0 +1,50 @@
...
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$schema->getTable('forms_v2_questions')
+			->dropColumn('mandatory');
```
jotoeri, line 46: Same here, just with an `if(hasTable())` without `!`

An example for this can be found on spreed. This example additionally uses different return-statements, which seems to be a nice way for not to change the schema. (I also just recently learned that and don't fully know it yet.) We don't necessarily need the different returns, it also works like you did it with the `return $schema`. But what you can well see from there is the surrounding by the if-statement.
https://github.com/nextcloud/spreed/blob/77b26ae723f7dd4dc6faad19ae322a3017f59ee1/lib/Migration/Version11001Date20210212105406.php#L49

jotoeri (commented):
Looks good already! 🙂 
Except the renaming to the sub-hypenated column, and if that works then, i'd change it to the conditional schemaChange and adapt the table-creation on Version010200. Such, we will for new installations directly create the new table, while existing installations will get migrated.

Chartman123:
> That would probably work afaik, but just as long as we do nothing else with that property within our code (which is indeed unlikely for that specific property). However, to have it clear, i would still do it only right before the API-Response is sent (so just before the return in getQuestion()), so we only have the doubling on API and won't have to take care of the doubled information within our own code.

Ok, I didn't know how to do that... So I will now push a commit with the way I think it also works. Perhaps we can then discuss the way you thought of.

Chartman123:
I think I messed up a little bit when I switched to my other computer and did a sync... How can I remove the last commit 036e52d?

jotoeri:
> 
> 
> I think I messed up a little bit when I switched to my other computer and did a sync... How can I remove the last commit [036e52d](https://github.com/nextcloud/forms/commit/036e52dc5272e524b938d2ee94e594cb57c8164d)?

Uhm, by just doing a rebase you should only get all commits without the merge commit.

To check that before force-pushing you can then also do a `git diff origin/refactor-mandatory`, where `origin` is the name of your remote (In case you initially cloned your fork and added this repo as remote you might have to adapt the command.) This will show you a diff to the branch on the server, so if you do not see any diff on your terminal, then the code is equal...^^
Also quite useful to monitor that a bit are imo some tools that show you a history/tree. Either git itself `git log --oneline --graph` or some GUIs: https://git-scm.com/downloads/guis

lib/Db/Question.php
```diff
@@ -48,6 +48,7 @@ class Question extends Entity {
 	protected $order;
 	protected $type;
 	protected $mandatory;
+	protected $isRequired;
```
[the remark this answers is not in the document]
jotoeri: Honestly i also just realised, it is not the single line i thought on API, as we have multiple places, where we send question-objects.
However, in total i still think it is clearest for us, if we do those copies only on API-Controller, right before sending. Basically that is only what we need, as we don't want to disturb our API, and everything on our own code will then only handle the isRequired while mandatory does not exist....
So afais where we send questions is:
- newQuestion()
- getSubmissions()
- getForm()

On newQuestion it should just be a simple property-copy, while on the latter two we would have to loop through the questions-array and set it for each question. But that is three lines of code each.

@skjnldsv Any opinion on this? Would you do the mapping somewhere else?

jotoeri (commented):
So - two long comments. 🙈 

To have a look onto the API, i typically use an add-on on firefox called 'RESTED'. Such you can call the API and see the returned data directly. I'm currently not on my dev-machine, so i can only provide you a screenshot probably this evening, but using the API-docs you could try to get that tool running with the right url & Header... 😉

lib/Db/Question.php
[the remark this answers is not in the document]
skjnldsv: Sorry, I'm too out of context from forms.
We are not forced to keep a retro-compatibility with the api.
We can release a major and tell that this changed.

Only rename the entry in the db and replace the update/binds of `mandatory` to `is-required`.
Trying to keep everything is just going to complicate the code. Let's see that as a migration to a major?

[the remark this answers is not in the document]
jotoeri: What? 👀 I thought we discussed that already on the issue?
To have a proper versioning we would have to do a API v2 while keeping v1 alive for a while, no? That's not less complicated.... 🤔

lib/Controller/ApiController.php
```diff
@@ -634,6 +634,16 @@ public function updateQuestion(int $id, array $keyValuePairs): DataResponse {
...
 		}
 
+		// Rename deprecated mandatory key to isRequired
+		if (key_exists('mandatory', $keyValuePairs)) {
+			$this->logger->info('Key \'mandatory\' is deprecated, please use \'isRequired\'.');
+			foreach ($keyValuePairs as $key => $value) {
+				if ($key === 'mandatory') {
+					$key = 'isRequired';
+				}
+			}
```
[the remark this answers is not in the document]
Chartman123: Ok, I already did some research on how to do it but obviously I chose the wrong solution. Thank you!

lib/Db/Question.php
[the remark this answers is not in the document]
skjnldsv: Depends, you can decide to not keep the v1 and do a hard break with a major :shrug: 
Not ideal, but this whole discussion doesn't have any good solution anyway ahah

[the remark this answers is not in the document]
skjnldsv: We can also just drop it all and focus on something else ^^'

Chartman123:
I don't know why, but when I'm trying the app with my latest code, I get a request like this to the update endpoint:

`{"id":7,"keyValuePairs":{"isrequired":true}}`

The response from the server is a 404:

```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>404</statuscode>
  <message>Invalid query, please check the syntax. API specifications are here: http://www.freedesktop.org/wiki/Specifications/open-collaboration-services.
</message>
 </meta>
 <data/>
</ocs>
```

In the Nextcloud log I get the following error:

BadFunctionCallException: isrequired is not a valid attribute
```
{"reqId":"FoOfQY2oHcmib8qUQ5rz","level":3,"time":"2021-04-07T19:45:47+00:00","remoteAddr":"192.168.1.201","user":"admin","app":"no app in context","method":"POST","url":"/nextcloud/ocs/v2.php/apps/forms/api/v1/question/update","message":{"Exception":"BadFunctionCallException","Message":"isrequired is not a valid attribute","Code":0,"Trace":[{"file":"/srv/www/htdocs/nextcloud/lib/public/AppFramework/Db/Entity.php","line":151,"function":"setter","class":"OCP\\AppFramework\\Db\\Entity","type":"->"},{"file":"/srv/www/htdocs/nextcloud/lib/public/AppFramework/Db/Entity.php","line":55,"function":"__call","class":"OCP\\AppFramework\\Db\\Entity","type":"->"},{"file":"/srv/www/htdocs/nextcloud/apps/forms/lib/Controller/ApiController.php","line":645,"function":"fromParams","class":"OCP\\AppFramework\\Db\\Entity","type":"::"},{"file":"/srv/www/htdocs/nextcloud/lib/private/AppFramework/Http/Dispatcher.php","line":218,"function":"updateQuestion","class":"OCA\\Forms\\Controller\\ApiController","type":"->","args":["*** sensitive parameters replaced ***"]},{"file":"/srv/www/htdocs/nextcloud/lib/private/AppFramework/Http/Dispatcher.php","line":127,"function":"executeController","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->"},{"file":"/srv/www/htdocs/nextcloud/lib/private/AppFramework/App.php","line":157,"function":"dispatch","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->"},{"file":"/srv/www/htdocs/nextcloud/lib/private/Route/Router.php","line":302,"function":"main","class":"OC\\AppFramework\\App","type":"::"},{"file":"/srv/www/htdocs/nextcloud/ocs/v1.php","line":63,"function":"match","class":"OC\\Route\\Router","type":"->"},{"file":"/srv/www/htdocs/nextcloud/ocs/v2.php","line":24,"args":["/srv/www/htdocs/nextcloud/ocs/v1.php"],"function":"require_once"}],"File":"/srv/www/htdocs/nextcloud/lib/public/AppFramework/Db/Entity.php","Line":122,"CustomMessage":"--"},"userAgent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:87.0) Gecko/20100101 Firefox/87.0","version":"21.0.0.18","id":"606e0c212330c"}
```

Why do I lose the capital R here?

jotoeri:
> Why do I lose the capital R here?

Hm, that's strange... for me it works with the current code here on github... 🤔 
Did you build the vue-code inbetween? Maybe the built js is a bit outdated?

Chartman123:
> Hm, that's strange... for me it works with the current code here on github... 🤔
> Did you build the vue-code inbetween? Maybe the built js is a bit outdated?

Seemed like a caching problem... I rebuilt the js a few times and it started working again only after clearing my browser cache... That's why I deleted the comment.

jotoeri:
😆 I still saw it, as i didn't update the page... 🙈 

To be sure for such caching problems i meanwhile use another Add-on on Firefox called 'Clear Cache'. This clears the cache and then autmomatically reloads the page and can be placed as shortcut on the taskbar. So i in fact always use this to reload the page when developing. But it's firefox, don't know if you work with sth. else... 🤷‍♂️

Chartman123:
Thank you for the recommendations, Firefox is fine 🙃

lib/Db/Question.php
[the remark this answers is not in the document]
Chartman123: > On newQuestion it should just be a simple property-copy, while on the latter two we would have to loop through the questions-array and set it for each question. But that is three lines of code each.

If we kept `mandatory` for compatibility reasons, should we do the copy for getSubmissions() and getForm() in the respective services or also in the API controller?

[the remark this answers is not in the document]
jotoeri: I'd do it only on API and right before sending the response... Such wie have the doubled data exclusively on API and not within our own code...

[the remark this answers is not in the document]
Chartman123: Ok, the deprecated mandatory properties have been added to the respective API responses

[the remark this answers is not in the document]
skjnldsv: @Chartman123 I'll let @jotoeri be the judge of that. I'm too out of context for forms unfortunately. I don't feel confident enough to help taking a proper decision there :)

Chartman123:
@jotoeri what are the next steps to create the new API version? Do we need a second ApiController and the duplicated entries in the routes.php?

jotoeri:
To version the api we can just parametrize the apiVersion on the router. E.g. on spreed it's this:
https://github.com/nextcloud/spreed/blob/24bcefa5cd52df002386acaded3df501d5e0f846/appinfo/routes.php#L62

As the apiVersion on requirements is afaik a regex then, we need a `v(1|1\.1)` there.

And then i'd just leave it like this. Both versions will just call the same methods, and such this will of course return the isRequired also on v1. But as this is nothing breaking or disturbing i wouldn't care about.
Honestly i'm also a bit struggling currently on this versioning. But in total i think this is the least disturbing way...
-> People can use v1.1 with isRequired, while if they want to use v1 due to backwards-complatiblity to old app-versions, they will have to use mandatory.

( -> A bit simpler would be to have only 'v1' on the routes and versioning v1.1 only on the docs. But as this would produce problems, if people want to use the route v1 with isRequired on old versions, this is also not ideal 🙁 
-> And the way of just the v2 is imo not ideal, as we would disturb the usage between app-versions completely. )

docs/DataStructure.md
```diff
@@ -1,7 +1,7 @@
 # Forms Data Structure
 **State: Forms v2.2.0 - 22.01.2021**
```
jotoeri, line 2: Don't forget to update the App-Version... 😉

lib/Controller/ApiController.php
```diff
@@ -215,6 +215,14 @@ public function getForm(int $id): DataResponse {
 			throw new OCSForbiddenException();
 		}
 
+		// Backward compatibility for mandatory
```
jotoeri, line 218: Can you add a TODO-Comment here on the four places, to remove the blocks when removing API v1.1.?

lib/Controller/ApiController.php
```diff
@@ -215,6 +215,14 @@ public function getForm(int $id): DataResponse {
...
+		$questions = $form['questions'];
+		foreach ($questions as &$question) {
+			$question['mandatory'] = $question['isRequired'];
+		}
+		unset($question);
+		$form['questions'] = $questions;
```
jotoeri, line 224: Hmm, also an interesting version and i think this also works now, no? 👍 

However, there is a bit simpler and cleaner way:
```
foreach ($form['questions'] as $index => $question) {     // same as the $key => $value construct, just with different variable-names. ;)
  $form['questions'][$index]['mandatory'] = $question['isRequired'];
}
```

So you can use the array-access with square brackets also as nested model and by accessing $form['questions'] directly, you can even change the values therein directly from the foreach. The $question['isRequired'] could also be done like this, but as the copied variable already exists, one can even use it...^^

lib/Migration/Version020300Date20210406133704.php
```diff
@@ -0,0 +1,54 @@
...
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->getTable('forms_v2_questions')->hasColumn('mandatory') && $schema->getTable('forms_v2_questions')->hasColumn('is_required')) {
+			$schema->getTable('forms_v2_questions')
```
jotoeri, line 46: Can you store the table here in a variable and use it from there? Just as it's done on your other migration... That's a bit more lucid....

jotoeri (commented):
Aaand just a few comments again 😉

lib/Controller/ApiController.php
Chartman123: Yes it worked very well, but you're right: the lesser code, the better 🙂

lib/Controller/ApiController.php
Chartman123: Done

lib/Migration/Version020300Date20210406133704.php
Chartman123: Done

docs/DataStructure.md
Chartman123: I added v2.3.0 and the current date.

Chartman123:
> To version the api we can just parametrize the apiVersion on the router.

Ok, did that... Where should in the docs should we point out, that mandatory is now deprecated?

jotoeri:
Hm, so i think you did already quite well by marking the individual occurences. Same you could still do on the json in DataStructure.md. Also there, i'd set the whole mandatory line on the table as italic, to mark it a bit more. And probably add there on the table a short 'replaced by'? Should suffice if it's only on the table...

And then adding a h2 on api-docs before or after General as 'Deprecation Warning', with a short text, that mandatory on Question-Props has been replaced by isRequired?
And on the changelog in fact the same text, just in an h3 as 'Deprecated' ?

In fact it should just be clear and visible... 🤔

Chartman123:
Ok, so I think we're good for a real review 😄

jotoeri:
Nice! 🚀 
A rebase is good, yes! You don't have to necessarily squash them all into a single commit, but grouping a bit would for sure be good. Especially commits, that fix another one should be squashed, you can also change the order of commits during interactive rebase for this.
And then i'd say just group it a bit by tasks or by connected changes if possible and give the commits the corresponding naming. In fact, we should just have a readable and relatively short history in the end, so its overseeable.
If you first do a rebase without squashing, then you will see a diff before pushing, and afterwards you can just play around with the rebasing, as long as you take care to force-push only, if you have no diff to the server-branch (and/or have a backup^^).

Chartman123:
Just noticed that I commited some changes together that would have been better in seperate commits for having a clean history... Is it possible to split some of the commits now? Otherwise this would be my rebase-todo:

```
reword ba3eee4 Change mandatory to isRequired
reword fa28af1 Add schema migration from mandatory to isRequired
f da4b111 Fix lint error
f 5513854 Fix lint error
pick 10ebcad Adjust API to isRequired
f 23053c3 Fix array substitution
f 8e3fd5d Remove mandatory from Db\Questions.php
f 6357cb9 Add mandatory for deprecated API calls
f e305b46 Cleaner code and TODO comments added  // Can this be split into two commits?
f ce53f11 Remove trailing whitespace
f 43ef88e Move internal API calls to v1.1
reword 1a993d4 Change documentation to isRequired
f e03ec5d Fix typo
f 16048ef Add deprecation info to docs and changelog
pick ce5a6fe api v1.1 // Can this be split into two commits?
```

Could you please give me some more details about how to get the diffs? I'm not that much used to git...

jotoeri:
Want to have a quick call? Might be easier to explain that a bit.... 🙈 
https://cloud.nextcloud.com/call/1818210369

jotoeri:
> Add schema migration from mandatory to isRequired
- Seems like you changed the commit-message here already? Just fyi, that won't work, but will just open the commit-message, where you can change it then. I also ran into that already quite some times, but you'll see then...^^

- Yes, it is possible to split commits. To have it clear, i typically do this as separate rebase before squashing, to have git not doing too much magic in parallel.
  - Use an 'edit' on the rebase list. This will interrupt the rebase after using the respective commit.
  - Reset to the previous commit: `git reset HEAD^`
  -> Explanation: `HEAD` refers to the current commit, so `HEAD^` refers to the commit before the current one. (Can be continued with `HEAD~2`, `HEAD~3`,...). This will reset your branch to the previous commit, while keeping the changes as unstaged changes.
  - Then add the part, that you want to commit. `git add -p` (patch) might help, in case you want to partially add a file.
  - Repeat, until you have all changes comitted again. `git status` might help to keep an overview.
  - The continue the rebase-list: `git rebase --continue`
  - Ideally do a diff towards the server-branch to be sure, nothing got lost.

- The diff thing basically just compares the current code:
  - `git diff` compares the code to the current branch, such shows you the current changes, that you did not commit yet.
  - `git diff some-branch` compares the current code (i.e. current branch, if you don't have uncommitted changes) to the `some-branch` branch.
  - `git diff origin/some-branch` where origin is the name of a remote, just compares the current code to the code on the remote-branch (if up-to-date). So if your `git remote -v` shows the github repo as `origin`, then you compare to the branch on github.

jotoeri (commented):
So, looked through all again, some small points remain.

One more thing, that rankles me a bit:
I personally don't like that on fresh installations we create the column as mandatory, but then directly change and delete it again. So i'd change the Version0102 Migration, too. However, i also know of people who don't like changing old migrations, that got released already.... So just decide yourself, what you like 🙈 
In case you would change it, we would also need to add the `if(hasColumn)` part to the first Version0203-Migration within the ensureColumnIsNullable-Procedure (but keep the mandatory there!)...

Chartman123:
> However, i also know of people who don't like changing old migrations, that got released already....

That's what the dev docs recommend for migrations, so I think I'll stick with that for now :)

src/components/Questions/QuestionDropdown.vue
```diff
@@ -189,8 +189,8 @@ export default {
 		 * @returns {boolean}
 		 */
 		isRequired(id) {
```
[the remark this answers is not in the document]
Chartman123: Yes, that was due to some conflicts... As I didn't face any other problem I didn't change this line. I will update it now :)

docs/API.md
```diff
@@ -20,10 +20,13 @@ This file contains the API-Documentation. For more information on the returned D
...
 ```
 
+## Deprecation warning
+Starting with API version 1.1, the `mandatory` property of questions is deprecated and replaced by `isRequired`. It will be removed in API version 2.
+
 ## Form Endpoints
 ### List owned Forms
 Returns condensed objects of all Forms beeing owned by the authenticated user.
-- Endpoint: `/api/v1/forms`
+- Endpoint: `/api/v1.1/forms`
```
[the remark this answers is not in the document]
Chartman123: Oops... :-/ Will adjust it there, too :-)

jotoeri (approved):
LGTM 🚀

Chartman123:
Oh, was I too fast merging it?? 😳

jotoeri:
Ahh, yes a bit. 🙈 Two approvals for larger changes are typically the goal...
Anyways. We'll fix it separately if there's still stuff. 👍 
Thanks for digging into and staying on this topic. And congrats to a quasi-full-stack PR without much previous experience... 💪 😉

Chartman123:
Thank you again for your help 😇

OK, the next time I'll wait for the second approval.

PR #894, opened 2021-04-11
Bump version

jotoeri:
So let's go ahead, bump minversion & appversion. :wink:

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -11,7 +11,7 @@
 - **🔒 Data under your control!** Unlike in Google Forms, Typeform, Doodle and others, the survey info and responses are kept private on your instance.
 - **🙋 Get involved!** We have lots of stuff planned like more question types, collaboration on forms, [and much more](https://github.com/nextcloud/forms/milestones)!
     ]]></description>
-    <version>2.2.4</version>
+    <version>2.3.0</version>
     <licence>agpl</licence>
 
     <author>Affan Hussain</author>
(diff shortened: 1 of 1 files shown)


appinfo/info.xml
```diff
@@ -11,7 +11,7 @@
 - **🔒 Data under your control!** Unlike in Google Forms, Typeform, Doodle and others, the survey info and responses are kept private on your instance.
 - **🙋 Get involved!** We have lots of stuff planned like more question types, collaboration on forms, [and much more](https://github.com/nextcloud/forms/milestones)!
     ]]></description>
-    <version>2.2.4</version>
+    <version>2.3.0</version>
```
skjnldsv, line 14: Missing package.json version bump :)

jotoeri: Hm, i didn't plan to release yet... This is just to make the migrations run...
Would you anyways bump the package? Or some.other way? Manual running of migrations? 🤔

skjnldsv: I would to keep consistent, :)
But not a big deal indeed ;)

skjnldsv (approved):

appinfo/info.xml
jotoeri: I'll add it. 👍 
Just had another thought: How about using an alpha-version? That even has the advantage of beeing able to do such step again, but not having shot the version 2.3 already. 🎉

skjnldsv: Sure, I'm fine as well!
I'm using beta/alpha on other places. 

Syntax is `2.3.0-alpha.0`. Npm supports it with `npm version prerelease --preid alpha` (and various preXXXXX options)

jotoeri:
Just pushed it to master now, including the tag... https://github.com/nextcloud/forms/commit/600ce2397afb2f162f7e9da0ae0a2c84bbb8b2ea

Issue #736, opened 2021-01-11
Date format not consistent


KvoiceOxford:
The form's date format recorded in responses is dependent on the submitting browser, not the form-app. Hence different browsers produce different date-format outcomes. For example: '19 March 2021' using my Edge or my Firefox; whereas 'March 19, 2021' using my Chrome. This data-formatting outcome does not provide a consistent data format held within the NC database, hence potentially can lead to unpredictable results when analysing form responses.

Chartman123:
@jotoeri I started looking into this and I'm already storing the date/datetime values as Unix timestamps into the db and can also get it back and display it in moment 'LL' and 'LLL' formats. However, older values from the db won't be displayed correctly. Should we do a db migration to convert date/datetime answers to Unix timestamps or should we do some more checks in the front-end?

PR #895, opened 2021-04-13
Add server-side validation of submissions

### LINKED ISSUE #758: Server-side validation of Responses
**Describe the bug**
Requirements such as making a specific question mandatory seem to only be enforced client-side (e.g. in the user's browser).  If a user intercepts the HTTP request with their answer and changes the drop-down index to a non-existent one, the form response is accepted by the server and the question is marked as "No Response" (when that should not be possible as the question was marked as mandatory)

**To Reproduce**
1. Create a Form.
2. Add a drop-down question, marked as 'Required'
2. Go to the form link and select one of the answers in the drop-down.
3. Intercept the request to the server (e.g. using a browser extension such as TamperData or an intercepting proxy such as ZAP or Burp)
4. Replace the index of the answer by a non-existent one (e.g. if the answer had 3 options in the drop-down, change the selected index to '5', for example).
5. The response will be accepted.
6. When the responses are viewed by the Form creator, the drop-down answers that were tampered with non-existent indices will appear as "No response".

**Expected behavior**
The server should reject the response informing the user that a non-valid answer to the question was chosen/provided.

**Nextcloud:**
- Nextcloud-Version: 20.0.6
- Forms-Version: 2.1.0

Chartman123:
Fixes #758 by checking answers before creating a submission and sending an error "400 - Bad Request" if at least one question has no valid answer. 

- [x] required questions have at least one answer
- [x] answers are in options
- [x] questions don't have multiple answers if only one answer is expected
- [x] ~~valid date / datetime values~~ (wait for #736)
- [x] all answered questions correspond to the current form

### DIFF
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -26,6 +26,7 @@ namespace OCA\Forms\Service;
 
 use DateTimeZone;
 
+use OCA\Forms\Constants;
 use OCA\Forms\Db\FormMapper;
 use OCA\Forms\Db\QuestionMapper;
 use OCA\Forms\Db\SubmissionMapper;
@@ -239,5 +240,55 @@ class SubmissionService {
 		$csv->insertAll($records);
 
 		return $csv->getContent();
+	}
+
+	/**
+	 * Validate all answers against the questions
+	 * @param array $questions Array of the questions of the form
+	 * @param array $answers Array of the submitted answers
+	 * @return boolean If the submission is valid
+	 */
+	public function validateSubmission(array $questions, array $answers): bool {
+		
+		// Check by questions
+		foreach ($questions as $question) {
+			$questionId = $question['id'];
+			$questionAnswered = array_key_exists($questionId, $answers);
+
+			// Check if all required questions have an answer
+			if ($question['isRequired'] && (!$questionAnswered || !array_filter($answers[$questionId], 'strlen'))) {
+				return false;
+			}
+
+			// Perform further checks only for answered questions
+			// TODO Check if date questions have valid answers
+			if ($questionAnswered) {
+				// Check if non multiple questions have not more than one answer
+				if ($question['type'] !== Constants::ANSWER_TYPE_MULTIPLE && count($answers[$questionId]) > 1) {
+					return false;
+				}
+	
+				// Check if all answers are within the possible options
+				if (in_array($question['type'], Constants::ANSWER_PREDEFINED)) {
+					foreach ($answers[$questionId] as $answer) {
+						// Search corresponding option, return false if non-existent
+						if (array_search($answer, array_column($question['options'], 'id')) === false) {
+							return false;
+						}
+					}
+				}
+			}
+		}
+
+		// Check for excess answers
+		foreach ($answers as $id => $answerArray) {
+			// Search corresponding question, return false if not found
+			$questionIndex = array_search($id, array_column($questions, 'id'));
+			if ($questionIndex === false) {
+				return false;
+			}
+		}
+
+		return true;
 	}
 }
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -973,6 +973,11 @@ class ApiController extends OCSController {
 			throw new OCSForbiddenException('Already submitted');
 		}
 
+		// Is the submission valid
+		if (!$this->submissionService->validateSubmission($questions, $answers)) {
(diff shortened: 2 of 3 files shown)


Chartman123:
@jotoeri Any other cases that should get validated? If so, just add the items to the list in the first comment 👍🏻

jotoeri:
I added one more thing, that currently came to my mind... 👍

Chartman123:
I think that this will already be filtered in the insertion itself, but we can of course add this to the validation, too :)

jotoeri:
Currently yes, but we can change that... 😉 
So just from logical side - if we now have a validation in front of the insertion, that checks if the options are valid, then the logical way is to also check the questions there, having the same handling if invalid. We can even simplify the actual insertion then, as we know already, that when called it's valid.

Chartman123:
Are you aware of an easy way to parse the moment.js date/datetime strings in PHP? As far as I've seen, the values are transmitted localized as a string and are also stored like that inside the DB. Will it work to feed these strings to the date_parse() function?

jotoeri:
Hmm, honestly no. But we might also have to change that format somhow again. (See e.g. #736 ) So don't put too much effort into checking this. The problem here is especially, that it might have to fit all localized formats currently. Including especially all languages.

Chartman123:
Ok, I came up with an solution that might probably work... But I think that this might need some testing for several date formats... Otherwise I think that I have covered all validations now :)

Chartman123:
Ok, seems like I still have some errors in my code. I tried to send a submission from within the app (not with RESTED) and got a 400 from my validation 😦

Chartman123:
> Ok, seems like I still have some errors in my code. I tried to send a submission from within the app (not with RESTED) and got a 400 from my validation 😦

Should be fixed by my latest commit 👍🏻

jotoeri:
Hmm, so probably we should already add the tests before merging then... Especially on different languages for dates, so we don't start blocking any users now...
Do you want to have a look into? Or should i give it a go?

Chartman123:
You mean automated unit tests? I have never done that before. So I wouldn't mind if you did it :)

jotoeri:
> We could also just leave the date validation out of the PR for now and wait for a standardized date format for the storage.

That's probably a better way.... It's just quite error-prone, so even if we test 10 Languages, there might be another one, where it fails...

jotoeri:
Ahh, now i was just a second too slow... 😄 
Can you rebase onto current master? So just git pull on master and then rebase the branch... This just as a general thing for a clean history... 😉

Chartman123:
> Can you rebase onto current master? So just git pull on master and then rebase the branch...

Better now?

jotoeri:
Yep 👍 
grafik

jotoeri:
Just added some tests... This one was unexpectedly easy, as it does not call anything else than default php-functions 😄 🙈 

So if you still have ideas, what to test, you can just append them to the dataProvider...

Chartman123:
That's basically what I tested manually during development, so I think we're good to go now 👍🏻

lib/Service/SubmissionService.php
```diff
@@ -240,4 +241,58 @@ private function array2csv(array $header, array $records): string {
...
+						if (array_search($answer, array_column($question['options'], 'id')) === false) {
+							return false;
+						}
+					}
+				}
+	
+				// TODO Check if date questions have valid answers
+				if (in_array($question['type'], [Constants::ANSWER_TYPE_DATE, Constants::ANSWER_TYPE_DATETIME])) {
+					continue;
+				}
```
jotoeri, line 283: Hmm, forgot to add before commiting or so? 😉

Chartman123: I wanted to leave it here as a TODO for the future. Should I remove it completely?

jotoeri: Ahh, now i see, you just removed the return...^^
Yes, better remove, it's just dead code like this....

Chartman123: Exactly... So remove or put it in comments?

jotoeri (approved):

jancborchardt:
Awesome contribution @Chartman123! We have a Nextcloud Talk instance for contributors, with e.g. a general "Community chat" and also a channel for the Forms app for quicker coordination. If you are interested, I can add you there with a guest account via your email address. :)

Chartman123:
@jancborchardt thanks 😇 yes, that would be nice. 👍🏻

PR #924, opened 2021-05-07
Make datetime answers consistent

### LINKED ISSUE #736: Date format not consistent

Chartman123:
Fixes #736 

- [x] Store timestamps to database
- [x] Parse timestamps in Submission view
- [x] Parse timestamps in Summary view
- [ ] Add database migration?
- [ ] Add validation
- [ ] Adjust exports


jotoeri:
Uhm, take care a bit here, initially that were timestamps, hovewer that caused problems as the timestamp does not reflect the timezone! So if we select a time within our Timezone MESZ, the unix-timestamp stores the time of two hours ago... That was why i moved to store just the text, that the user actually selected on the form.

So there are two possiblities, i think: Either storing the timezone together with the unix-timestamp, which would be more complex for us, but a bit more beautiful on the form-results. Or we just format the timestamp ourselves, making it at least independent of browser and keep storing the text. That might however not reflect different locales between users or between users and form-admin on the results...
Any thoughts? 🤔

Chartman123:
The way I implemented it right now should store the localized value that has been selected as UTC timestamp in the DB and will be retrieved as UTC and converted back to the localized value of the browser...

So do you want to display the time across timezones as the same value? So if someone in GMT selects 12:00 pm and someone in CET selects 12:00 pm it will show two answers with 12:00 pm?

I think it's always best to store everything as UTC and display the values localized.

> This article basically outline the strategy of keep it simple, and just work with UTC time at all times until right before display to the user.

https://www.moesif.com/blog/technical/timestamp/manage-datetime-timestamp-timezones-in-api/

jotoeri:
Hmm, the article is more about how to store dates on Db and represent on API, no? While for us, the relevant point is how to handle it or behave towards the user... 🤔 
The problematic part is imo especially when handling dates, while for times one might be more aware of timezones. So if a user fills the form at 00:30, it might happen, that the user in just the next timezone sees a different date, than the one that has been filled...
Let's ask some researchers what they think, see on the chat... 😉

jotoeri:
😢

Chartman123:
Don't worry, I just wanted to start again clean and thought it would be easier to just delete the branch and create a new one 🙂

PR #976, opened 2021-06-10
Fix Activity Link

jotoeri:
Signed-off-by: Jonas Rittershofer <jotoeri@users.noreply.github.com>

### DIFF
--- a/lib/Activity/Provider.php
+++ b/lib/Activity/Provider.php
@@ -255,7 +255,7 @@ class Provider implements IProvider {
 			$formTitle = $this->formMapper->findbyHash($formHash)->getTitle();
 
 			// Append hash and route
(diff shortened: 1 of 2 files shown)


Chartman123 (approved):

jancborchardt:
Looks good regarding the "/", would that also fix the issue with the "index.php" being there or not? (Don’t have email setup on my local instance so can’t test that.)

skjnldsv (approved):

jotoeri:
> Looks good regarding the "/", would that also fix the issue with the "index.php" being there or not? (Don’t have email setup on my local instance so can’t test that.)

Nope, that's a sever thing.... Wie only use, what urlGenerator produces upto the 'apps/forms'-path...

PR #1001, opened 2021-07-02
Make date/time answers consistent

### LINKED ISSUE #736: Date format not consistent

Chartman123:
Make date/time answers consistent, this fixes #736 

- [x] Add time answer type
- [x] Store date as 'YYYY-MM-DD'
- [x] Store time as 'HH:mm'
- [x] Remove datetime (backward compatibility?)
- [x] Adjust result views
- [x] Adjust exports?
- [x] Add validation
- [x] Block creation of new datetime-questions on API.

### DIFF
--- a/docs/DataStructure.md
+++ b/docs/DataStructure.md
@@ -172,4 +172,5 @@ Currently supported Question-Types are:
 | `short`         | A short text answer. Single text line |
 | `long`          | A long text answer. Multi-line supported |
 | `date`          | Showing a dropdown calendar to select a date. |
-| `datetime`      | Showing a dropdown calendar to select a date **and** a time. |
+| _`datetime`_      | _deprecated: No longer available for new questions. Showing a dropdown calendar to select a date **and** a time._ |
+| `time`          | Showing a dropdown menu to select a time. |
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -36,6 +36,7 @@ class Constants {
 	public const ANSWER_TYPE_LONG = 'long';
 	public const ANSWER_TYPE_DATE = 'date';
 	public const ANSWER_TYPE_DATETIME = 'datetime';
+	public const ANSWER_TYPE_TIME = 'time';
 
 	// AnswerTypes, that need/have predefined Options
 	public const ANSWER_PREDEFINED = [self::ANSWER_TYPE_MULTIPLE, self::ANSWER_TYPE_MULTIPLEUNIQUE, self::ANSWER_TYPE_DROPDOWN];
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -40,4 +40,10 @@ class Constants {
 
 	// AnswerTypes, that need/have predefined Options
 	public const ANSWER_PREDEFINED = [self::ANSWER_TYPE_MULTIPLE, self::ANSWER_TYPE_MULTIPLEUNIQUE, self::ANSWER_TYPE_DROPDOWN];
+
+	// AnswerTypes for date/time questions
+	public const ANSWER_DATETIME = [self::ANSWER_TYPE_DATE, self::ANSWER_TYPE_DATETIME, self::ANSWER_TYPE_TIME];
+
+	// Formats for AnswerTypes date/datetime/time
+	public const ANSWER_PHPDATETIME_FORMAT = [self::ANSWER_TYPE_DATE => 'Y-m-d', self::ANSWER_TYPE_DATETIME => 'Y-m-d H:i', self::ANSWER_TYPE_TIME => 'H:i'];
 }
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -25,6 +25,7 @@
 namespace OCA\Forms\Service;
 
 use DateTimeZone;
+use DateTime;
 
 use OCA\Forms\Constants;
 use OCA\Forms\Db\FormMapper;
@@ -261,10 +262,18 @@ class SubmissionService {
 			}
 
 			// Perform further checks only for answered questions
-			// TODO Check if date questions have valid answers
 			if ($questionAnswered) {
 				// Check if non multiple questions have not more than one answer
 				if ($question['type'] !== Constants::ANSWER_TYPE_MULTIPLE && count($answers[$questionId]) > 1) {
+					return false;
+				}
+
+				/*
+				 * Check if date questions have valid answers
+				 * $answers[$questionId][0] -> date/time questions can only have one answer
+				 */
+				if (in_array($question['type'], Constants::ANSWER_DATETIME) &&
+					!$this->validateDateTime($answers[$questionId][0], Constants::ANSWER_PHPDATETIME_FORMAT[$question['type']])) {
 					return false;
 				}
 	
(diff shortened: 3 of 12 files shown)


Chartman123:
@jotoeri Is it possible to use the datetime picker vue component to just select a time?

jotoeri:
> @jotoeri Is it possible to use the datetime picker vue component to just select a time?

Should work, yes...:
https://github.com/mengxiong10/vue2-datepicker#props

jotoeri:
Just realised, we should also:
- [ ] Block creation of new datetime-questions on API.

:wink:

Chartman123:
@jotoeri Another missing point is the translation for the "new" time picker related strings. Where do I have to add these strings so that the can be localized by the translators?

jotoeri:
> @jotoeri Another missing point is the translation for the "new" time picker related strings. Where do I have to add these strings so that the can be localized by the translators?

That comes automatically once your PR is merged. There is a process running somewhere by NC-GmbH (once a day or so?), that will extract the translations and push it to Transifex. So nothing you need to care about. :wink:

lib/Constants.php
```diff
@@ -36,7 +36,14 @@ class Constants {
...
+	public const ANSWER_TYPE_TIME = 'time';
 
 	// AnswerTypes, that need/have predefined Options
 	public const ANSWER_PREDEFINED = [self::ANSWER_TYPE_MULTIPLE, self::ANSWER_TYPE_MULTIPLEUNIQUE, self::ANSWER_TYPE_DROPDOWN];
+
+	// AnswerTypes for date/time questions
+	public const ANSWER_DATETIME = [self::ANSWER_TYPE_DATE, self::ANSWER_TYPE_DATETIME, self::ANSWER_TYPE_TIME];
+
+	// Formats for AnswerTypes date/datetime/time
+	public const ANSWER_DATETIME_FORMAT = ['date' => 'Y-m-d', 'datetime' => 'Y-m-d H:i', 'time' => 'H:i'];
```
jotoeri, line 48: Hmm, the naming here is a bit confusing... I would have expected with the ANSWER_DATETIME & ANSWER_DATETIME_FORMAT , that the storageFormat is meant. 🤔 Maybe add a comment on this?

jotoeri, line 48: Also use the types here? `self::ANSWER_TYPE_DATE => 'Y-m-d'`

lib/Service/SubmissionService.php
```diff
@@ -291,4 +299,15 @@ public function validateSubmission(array $questions, array $answers): bool {
...
+
+	/**
+	 * Validate correct date/time formats
+	 * @param string $dateStr String with date from answer
+	 * @param string $format String with the format to validate
+	 * @return boolean If the submitted date/time is valid
+	 */
+	private function validateDateTime(string $dateStr, string $format) {
+		$d = DateTime::createFromFormat($format, $dateStr);
+		return $d && $d->format($format) === $dateStr;
```
jotoeri, line 311: `!! DateTime::createFrom....` would probably suffice... 🤔 But also ok like this for me. 🤷‍♂️

docs/DataStructure.md
```diff
@@ -173,3 +173,4 @@ Currently supported Question-Types are:
 | `long`          | A long text answer. Multi-line supported |
 | `date`          | Showing a dropdown calendar to select a date. |
 | `datetime`      | Showing a dropdown calendar to select a date **and** a time. |
```
jotoeri, line 175: Do we have anywhere a deprecation-warning for this already?

jotoeri (commented):
Some small thoughts... ;)

lib/Service/SubmissionService.php
```diff
@@ -261,12 +262,19 @@ public function validateSubmission(array $questions, array $answers): bool {
...
 				// Check if non multiple questions have not more than one answer
 				if ($question['type'] !== Constants::ANSWER_TYPE_MULTIPLE && count($answers[$questionId]) > 1) {
 					return false;
 				}
+
+				/*
+				 * Check if date questions have valid answers 
+				 * $answers[$questionId][0] -> date/time questions can only have one answer
+				 */ 
+				if (in_array($question['type'], Constants::ANSWER_DATETIME) && !$this->validateDateTime($answers[$questionId][0], Constants::ANSWER_DATETIME_FORMAT[$question['type']])) {
```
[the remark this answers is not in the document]
Chartman123: I've also thought about moving the two parts to two variables... But perhaps the linebreak would be enough

lib/Service/SubmissionService.php
Chartman123: I found this code in the official PHP docs in a comment and just copied it... So I'd say let's just keep it for the moment

src/models/AnswerTypes.js
```diff
@@ -141,5 +146,26 @@ export default {
 
 		// Using the same vue-component as date, this specifies that the component renders as datetime.
 		includeTime: true,
```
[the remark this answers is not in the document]
Chartman123: I'm not quite sure... I mean we still have existing forms with that type.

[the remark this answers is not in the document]
Chartman123: Or do you mean because of the new pickerType?

lib/Constants.php
Chartman123: Yes, will do that :)

Chartman123: I tried to make it consistent with ANSWER_PREDEFINED as opposed to the ANSWER_TPYE_... above

docs/DataStructure.md
Chartman123: No, I don't think so... also the 'time' type is still missing there.

src/models/AnswerTypes.js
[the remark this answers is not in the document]
jotoeri: These types were just used on our own code. Nothing of this is stored in DB (except the name). So as you now changed the code, the properties `includeTime` and `onlyTime` are not used anymore and can be removed. So yes - basically this property is replaced with the pickerType/storageFormat/momentFormat properties...

docs/DataStructure.md
Chartman123: I added the correct description, but the first column included 'date' for both lines 😉 So this will be another fixup commit, and also the deprecation notice.

lib/Constants.php
Chartman123: ok, now I see what you mean... I didn't think about just having the PHP DateTime format but just a default format for the answer types as we have them in the answerTypes.js, but here of course in the way the PHP formatting is done as opposed to the moment format.

jotoeri: Well our default format for answerTypes is still the storageFormat, even on php... The stored data (& its format) is still the same on php and js. And the here denoted format definitely differs from our original datetime format, that is valid around the whole app. So thus i'd still rename it to make that clear.... Or at least adding a comment describing, why this format now differs, resp. what it is for. ;)

Chartman123: Ok, it differs in the string that marks the format, but in the end it all results in YYYY-MM-DD and HH:mm ;) But no problem, I can rename them and then get the PR merged :)

jotoeri (approved):
Fine by me now. 👍 
Nice, @Chartman123 ! 🎉

PR #1056, opened 2021-10-02
Fixes to Russian translation

funbotan:
Fixed some inconsistencies, translation mistakes, and name collision with the Polls app.


jotoeri:
Hey Alexander,

first of all: thank you for engaging and contributing! :)
However, as on the Nextcloud Project nearly all translations are managed on Transifex, so is forms. So if we'd merge this here, it would soon get overwritten again from transifex.
This also means, if you want your translations to be continuous, you have to insert them on transifex and the Pull-Request to the repository will then happen automatically:

https://www.transifex.com/nextcloud/nextcloud/forms/

You might need to start with a first login on transifex and with adding Russian as one of your languages. That is all a bit confusing for the first moment iirc. For the login you can also use the SSO with Github. But once you got this, you can just start applying or discussing your corrections there. Just ask if you have questions! :)

Greets,
Jonas

funbotan:
> 
> 
> Hey Alexander,
> 
> first of all: thank you for engaging and contributing! :) However, as on the Nextcloud Project nearly all translations are managed on Transifex, so is forms. So if we'd merge this here, it would soon get overwritten again from transifex. This also means, if you want your translations to be continuous, you have to insert them on transifex and the Pull-Request to the repository will then happen automatically:
> 
> https://www.transifex.com/nextcloud/nextcloud/forms/
> 
> You might need to start with a first login on transifex and with adding Russian as one of your languages. That is all a bit confusing for the first moment iirc. For the login you can also use the SSO with Github. But once you got this, you can just start applying or discussing your corrections there. Just ask if you have questions! :)
> 
> Greets, Jonas

Thank you for the comprehensive explanation!

Issue #1012, opened 2021-07-20
Share with Circle


dominik34:
Sharing with circles is not possible.
Unlike in the sharing section of Documents, eg. there are no circles found to share with.

JOpsDev:
Would sharing with circles enable editing the form by multiples persons make possible? We don't need to edit the form at the same time but adjusting some things would be really helpful.

Chartman123:
No, it's primarily just for sharing the form with regard to submitting...

punkyard:
hi
our teams never share with groups, they share with circles, which they can create and modify as needed.
so far we can't use forms for this reason.

I'm pushing the request!!
thanks for your concern :)

punkyard:
hi @Chartman123 
could you please enlighten me with the future of this issue?
'merged' does it mean it's going to be taken into consideration for a future update?

Chartman123:
Hi @punkyard

yes, merged means that the changes in the code are in the main branch now and that they will be part of the next release.

Issue #717, opened 2020-12-15
Hard to find settings sidebar


konfiv:
Thanks for the Form!

**Describe the bug**
When logged in in nextcloud and a form link is visited, not the form will show, but the files of the NextCloud.
If you are not logged in the form is visible and working.

**To Reproduce**
1. Limit the Form app to admin group
2. create a form
3. Login with a normal user
4. Paste the link provided by the form 
5. you will not see the form, but the files 

**Expected behavior**
When go to the form link, that the form will show, not the files of NextCloud.

**Screenshots**

**Nextcloud (please complete the following information):**
- Nextcloud-Version:  19.0.4
- Forms-Version: v2.0.4 (2020-09-01)

**Desktop (please complete the following information):**
 - OS: ubuntu 18.4
 - Browser firefox 
 - Version 68.10.0esr (64-Bit)

**Smartphone (please complete the following information):**

**Browser log**
```
Open your console, reload your page and/or do the action leading to this issue and copy/paste the log in this thread.
```

<details>
<summary>How to access your browser console (Click to expand)</summary>

# Chrome
- Press either CTRL + SHIFT + J to open the “console” tab of the Developer Tools.
- Alternative method:
    1. Press either CTRL + SHIFT + I or F12 to open the Developer Tools.
    2. Click the “console” tab.

# Safari
- Press CMD + ALT + I to open the Web Inspector.
- See Chrome’s step 2. (Chrome and Safari have pretty much identical dev tools.)

# IE9
1. Press F12 to open the developer tools.
2. Click the “console” tab.

# Firefox
- Press CTRL + SHIFT + K to open the Web console (COMMAND + SHIFT + K on Macs).
- or, if Firebug is installed (recommended):
    1. Press F12 to open Firebug.
    2. Click on the “console” tab.

# Opera
1. Press CTRL + SHIFT + I to open Dragonfly.
2. Click on the “console” tab.
</details>

**Additional context**

skjnldsv:
Are you talking about filling the form? Or editing the form?
If the latest (editing), please check https://github.com/nextcloud/forms/issues/322
If the first (filling), then please double check your form sharing option and make sure the form is shared with your "normal user"

konfiv:
Hi, thanks. I meant:
1) filling a form by people.
I did not want to share the edit form option, only filling a form. So, restricting to only fill a form is not possible?
In the App options, now I enabled share with my "normal user group" than the share link is working well, when users are logged in, and when not logged in.

[...]

PR #1087, opened 2021-12-09
New Sharing

### LINKED ISSUE #425: Improve shared user component view
Show sharings as in Server

jotoeri:
Slow, but steady, i'm working from time to time on the sharing. I got a UI now, of which i think that should work.
So before i continue to work that all down to DB, @jancborchardt @nimishavijay @skjnldsv @RenataGegaj would you mind to have a first look onto?
It's just in case we need to change some of the stored data behind. The exact wording is not important yet, it's just for not to work down through all the code and later changing the whole data-structure.

grafik
grafik

grafik

grafik

Will someday fix #425 😉


jotoeri:
Hmm, just found another solution, which unfortunately looks not that clear on first, but is probably a bit less confusing:
- Like this, NO sharing is set by default. Nobody has access, only by selecting a specific option, people get access. This is more like how it is done in server.
- Two different ways of the Show-option:
grafik
grafik

RenataGegaj:
@jotoeri Sorry for the late reply, I just saw this. I think this is a better approach overall!+1 Splitting the sections into tabs is a good way to simplify the UI especially since the two sections are not quite related to each other.

I do have some usability concerns regarding the"Sharing" section. I understand some of them might be because of wording but there are also some regarding design patterns.
- The first thing a user will see is the input box(User or group name...), there is no explanation on what the feature does or a clear instruction on how it's used. "Why do I need to write a group or a username, what will happen next"
- The wording on the "Internal link" it's not very clear. Who are the users that already have access and do you mean access to the form or access to the instance or smth else? 
- The external ling option needs a description as well to inform the user. Also, I notice that there are two different icons for sharing the links, one is a clipboard and the other is a plus, curious to know why and what's the difference.
- For the last option "Show to users on the sidebar", I'm not sure I understand this feature. How are those users picked? If this is an option to allow users to share with specific people, isn't that what the first option does? 
I recently described some pain points users had with the current implementation of this feature while we did the user testings, I'll leave it here for reference https://github.com/nextcloud/forms/issues/717#issuecomment-1018030989

jotoeri:
Hey Renata,
thank you for the feedback - here some explanations, 

> The first thing a user will see is the input box(User or group name...), there is no explanation on what the feature does or a clear instruction on how it's used. "Why do I need to write a group or a username, what will happen next"

Yeah, from that view, it is just a wording thing, sth. like 'search for ...' would be better. However, the text without any addition is similar to what is done in server for sharing.... 🤔 We'll keep it to check again in the end for the wording.
> The wording on the "Internal link" it's not very clear. Who are the users that already have access and do you mean access to the form or access to the instance or smth else?

Hm ok, also intersting. It basically has the same functionality as the internal link on server. So users having access by either a direct (group-) share or if all users have access, they can use this link. Users who do not have access this way, can't use it. Do you have another proposal for the wording?
> The external ling option needs a description as well to inform the user. Also, I notice that there are two different icons for sharing the links, one is a clipboard and the other is a plus, curious to know why and what's the difference.

Same here - handling is like in server, where you see the plus, to create the first link, and afterwards you will have a menu, where you can create more links. Maybe it gets a bit more clear, if you can use it - then there are also tooltips to describe it. 😉 

> For the last option "Show to users on the sidebar", I'm not sure I understand this feature. How are those users picked? If this is an option to allow users to share with specific people, isn't that what the first option does?

This option is connected to the option above - ' Permit access to all users'. It only shows up, if the permit-option is active. So the permit-option is only for the access-control, while the 'show'-option then makes the form show up on all users sidebar. Same here - maybe let's check that again, when you can test it on some instance? :)

Greets! :)

Nemental:
@jotoeri Would it be possible to share the results of a form as well?

nimishavijay:
Oh wow! Looking good! 🎉 Here are my thoughts:
- I think the different tabs are a great idea! Similar to existing sidebar in other apps :)
- The general layout is also similar to the current sharing tab of the Files sidebar so it would feel more natural 
- I like the toggles in the settings, we read in an article that toggles are useful when there is no save button, so I think we should use it more 

Some questions I have, mostly regarding the wording, perhaps @RenataGegaj could also help out here:
- What does the "legacy link" section do? If it has to do with an outdated link, I don't think it needs to be so present. We could just skip that entire item and redirect any legacy links to the new links (does that make sense? Hope I understood the function of "Legacy link" correctly
- What does "Permit access to all users" do? Why would this be off?


Visual nitpicking:
- I would maybe change the wording a bit. We also use sentence case for our wording (Only the first letter of the sentence is a capital letter) so:
  - Internal Link --> Internal link
  - Public Share Link --> Share link
  - Logged-in users with access-permission can use this link --> Only works for logged in users who can access this form
- The list of users who have access does not need to be visible by default, as it could be a very long list. It could be similar to the current design where it is hidden behind a "Other users with access" which would be collapsed or expanded. If there are no users under this then it does not have to shown.
image

Really great work! :) Looks really nice!

jotoeri:
Hey @nimishavijay, thank you for the feedback! 🎉 ☺️ 
Here some answers to your questions.

> What does the "legacy link" section do? If it has to do with an outdated link, I don't think it needs to be so present. We could just skip that entire item and redirect any legacy links to the new links (does that make sense? Hope I understood the function of "Legacy link" correctly

Yeah, thats it nearly. The old sharing links were just the form-link itself (`apps/forms/hash`), which will now become the internal link. So the legacyLink allows the internalLink to be used publicly as temporary solution so to not just break existing links. Thus, i'd like people to change to the new way of the public-link (`apps/forms/s/newHashPerLink`, similar to server-sharing), which however needs their action, as they need to spread the new link to their applicants. And i'd say we also need to trigger them somehow, because one day we will finally remove the legacySupport, which would just break their links then instead of now...^^
I could make the exclamation-mark to be the black&white version to have it less obstrusive? 🤔 
(Also here the corresponding tooltip information:)
grafik


> What does "Permit access to all users" do? Why would this be off?

This allows all logged in users of the instance to access the form (by internal link). So this should of course be turned off, in case you want to share only to a few users.
The second part of this (show to all users) then additionally to allowing access, shows the form to all logged-in users on appNavigation, which means they can not only access it if they get the link, but they also actively get the form shown on the sidebar. This could be turned off, if you want users to be logged in for access via internal link, but don't want to actively show the form to all users of the instance on appNavigation.

> I would maybe change the wording a bit. We also use sentence case for our wording (Only the first letter of the sentence is a capital letter) so

Of course... 🤦‍♂️ 🙈 

>  It could be similar to the current design where it is hidden behind a "Other users with access"

But where do these 'other users' come from? 🤔 I just tested it on files-sharing, where i don't have any other users in that list, but everybody i share a file with comes up in the classical sharing list...?

Greets! ✌️

jotoeri:
**Review help**

- The OLD sharing was stored like:
```
form: {
   ...
   access: {
      type: 'selected',
      users: ['userId1', 'userId2'],
      groups: ['groupId1']
   }
}
```

- The NEW sharing changed to:
```
form: {
   ...
   access: {
      permitAllUsers: True,
      showToAllUsers: True,
      [ legacyLink: True ]
   },
   shares: [
      {
         id: 3,
         formId: 5,
         shareType: IShare::TYPE_USER,
         shareWith: 'userId1'
      },
      {
         id: 4,
         formId: 5,
         shareType: IShare::TYPE_LINK,
         shareWith: 'iAmSomeHash'
      },
   ]
}
```

Regarding the links:
- While the old (also publich) sharing link was the direct form link, `localhost/index.php/apps/forms/someHash`, this link will now only work as internal link for users already having access. This is similar to the files-sharing internal link.
- The new public sharing link is now like `localhost/index.php/apps/forms/s/someNewHash`, which now also can be multiple public links, linking to the same form.
- For temporary compatiblity, the migration of old forms introduces the `legacyLink`-property, which still allows the internal link to be used as old public link. But this also produces a warning for the form-owner, to replace the link. This functionality is a one-way ticket, so once deleted, it can not be restored, but a new public link has to be used.


Regarding direct access in the app:
- If a form is directly shared to single users or groups, then the form always shows up on their sidebar.
- While the property `permitAllUsers` allows all users to access the form (via spread internal link), there is now an additional switch, to show the form to all users on the sidebar. This enables now also to share, but not show to all users, which was a multiple FR on the issue-tracker.

Internal Data:
- There already is a permissions-property added, which tells the frontend, what uses are allowed to do on a form. This is currently used, to check if a user is allowed to e.g. edit a form, but also prepares for an easier result-sharing f.ex..
- Maybe clear, but to explain: Partial forms are just stripped form-objects, that contain only the data, that is necessary to show the form on the sidebar. Only when selecting a form for edit/submit, the rest of the non-partial data is loaded. The loading of partial forms got a small restructuring here...
- To submit as non-logged in users to the new public links, the share hash must be passed to the api. This is the way of access-control, if there is a public link used. The user needs to know the share-hash, but can not just submit to a form without a public hash.

jotoeri:
A try for review coordination: **Feel free to edit this, add comments, etc.!**

- [ ] appinfo/info.xml
- [ ] appinfo/routes.php
- [ ] css/icons.scss
- [ ] lib/Constants.php
- [ ] lib/Controller/ApiController.php
- [ ] lib/Controller/PageController.php
- [ ] lib/Controller/ShareApiController.php
- [ ] lib/Db/FormMapper.php
- [ ] lib/Db/Share.php
- [ ] lib/Db/ShareMapper.php
- [ ] lib/Migration/Version030000Date20211206213004.php
- [ ] lib/Service/FormsService.php
- [ ] package-lock.json
- [ ] package.json
- [ ] src/Forms.vue
- [ ] src/FormsSubmit.vue
- [ ] src/components/AppNavigationForm.vue
- [ ] src/components/ShareDiv.vue
- [ ] src/components/SidebarTabs/SettingsSidebarTab.vue
- [ ] src/components/SidebarTabs/SharingSearchDiv.vue
- [ ] src/components/SidebarTabs/SharingShareDiv.vue
- [ ] src/components/SidebarTabs/SharingSidebarTab.vue
- [ ] src/components/UserDiv.vue
- [ ] src/mixins/PermissionTypes.js
- [ ] src/mixins/ShareLinkMixin.js
- [ ] src/mixins/ShareTypes.js
- [ ] src/mixins/ViewsMixin.js
- [ ] src/views/Create.vue
- [ ] src/views/Results.vue
- [ ] src/views/Sidebar.vue
- [ ] src/views/Submit.vue
- [ ] tests/Unit/Controller/ShareApiControllerTest.php
- [ ] tests/Unit/Service/FormsServiceTest.php
- [ ] Testing the frontend and finding dumb bugs

lib/Controller/ShareApiController.php
```diff
@@ -0,0 +1,199 @@
...
+			$form = $this->formMapper->findById($formId);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
+			throw new OCSBadRequestException('Could not find form');
+		}
+
+		if ($form->getOwnerId() !== $this->currentUser->getUID()) {
+			$this->logger->debug('This form is not owned by the current user');
+			throw new OCSForbiddenException();
+		}
```
come-nc, line 125: Shouldn’t that be a permission check instead? How is that handled for files?
I’m fine if this is left as is for now, just wanted to point that out as accessing results is seen as a permission, I think resharing should also.

lib/Controller/ShareApiController.php
```diff
@@ -0,0 +1,199 @@
...
+		// Only accept usable shareTypes
+		if (array_search($shareType, Constants::SHARE_TYPES_USED) === false) {
+			$this->logger->debug('Invalid shareType');
+			throw new OCSBadRequestException('Invalid shareType');
+		}
+
+		try {
+			$form = $this->formMapper->findById($formId);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
```
come-nc, line 118:
```suggestion
			$this->logger->debug('Could not find form', ['exception' => $e]);
```
Most of the time if you have an exception for the context it’s good to pass it to the logger. (only putting that comment here but other try/catch are concerned)

lib/Controller/ShareApiController.php
```diff
@@ -0,0 +1,199 @@
...
+		$share->setShareWith($shareWith);
+
+		// For public-shares, check if hash already exists. (Unfortunately not possible here by unique index on db.)
+		if ($shareType === IShare::TYPE_LINK) {
+			try {
+				// Try loading a share to the hash.
+				$nonex = $this->shareMapper->findPublicShareByHash($share->getShareWith());
+
+				// If we come here, a share has been found --> The share hash already exists, thus aborting.
+				$this->logger->debug('Share Hash already exists.');
```
come-nc, line 148: Maybe loop on the generation instead?

lib/Controller/ShareApiController.php
```diff
@@ -0,0 +1,199 @@
...
+		$share->setShareType($shareType);
+
+		// Create public-share hash, if necessary.
+		if ($shareType === IShare::TYPE_LINK) {
+			$shareWith = $this->secureRandom->generate(
+				24,
+				ISecureRandom::CHAR_HUMAN_READABLE
+			);
+		}
+		$share->setShareWith($shareWith);
```
come-nc, line 139: It seems there is no check that $shareWith contains a valid user/group, not sure if expected or not.

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,166 @@
...
+				'length' => 256,
+			]);
+
+			$table->setPrimaryKey(['id'])
+				->addIndex(['form_id'], 'forms_shares_form')
+				->addIndex(['share_type'], 'forms_shares_type')
+				->addIndex(['share_with'], 'forms_shares_with');
+		}
+
+		return $schema;
```
come-nc, line 82: You should return null if no change is needed and only return $schema if the table was added.

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,166 @@
...
+	 * @param array $options
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		$qb_fetch = $this->connection->getQueryBuilder();
+		$qb_update_access = $this->connection->getQueryBuilder();
+		$qb_insert_shares = $this->connection->getQueryBuilder();
+
+		$qb_fetch->select('id', 'access_json')
+			->from('forms_v2_forms');
+		$cursor = $qb_fetch->execute();
```
come-nc, line 102: Use executeQuery and executeStatement

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,166 @@
...
+					$qb_update_access->setParameter('id', $row['id'], IQueryBuilder::PARAM_INT)
+						->setParameter('access_json', json_encode([
+							'permitAllUsers' => false,
+							'showToAllUsers' => false,
+						]), IQueryBuilder::PARAM_STR);
+					$qb_update_access->execute();
+
+					foreach ($access['users'] as $user) {
+						$qb_insert_shares->values([
+							'form_id' => $qb_insert_shares->createNamedParameter($row['id'], IQueryBuilder::PARAM_INT),
```
come-nc, line 144: You should add those 3 fields as parameters to the initial query and use setParameter, I think

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,166 @@
...
+							'share_type' => $qb_insert_shares->createNamedParameter(IShare::TYPE_GROUP, IQueryBuilder::PARAM_INT),
+							'share_with' => $qb_insert_shares->createNamedParameter($group, IQueryBuilder::PARAM_STR),
+						]);
+						$qb_insert_shares->execute();
+					}
+					break;
+
+				default:
+					$output->warning('Unknown access property on form. ID: ' . $row['id'], ', access: ' . $row['access_json']);
```
come-nc, line 161: This will output for all already migrated entries, no?

lib/Service/FormsService.php
```diff
@@ -173,19 +216,45 @@ public function getPublicForm(int $id): array {
...
+	 * @return boolean
 	 */
-	public function canSubmit($formId) {
+	public function canSubmit(int $formId): bool {
 		$form = $this->formMapper->findById($formId);
-		$access = $form->getAccess();
 
-		// We cannot control how many time users can submit in public mode
-		if ($access['type'] === 'public') {
+		// We cannot control how many time users can submit if public link / legacyLink available
```
come-nc, line 256: Shouldn’t we still forbid multiple submit for logged in users even if there is a public link?
(I mean "there is a public link" != "A public link is being used")

Is it possible for a form shared by public link to not be anonymous for logged in users? If yes, then submission limit should also apply.

come-nc (commented):
PHP code looks good!
A few minor comments, and I think improvement needed on the Migration, it should use a $qb var for the query builder and assign the other 3 from there, it looks quite messy right now compared to other Migrations from other apps.

Properties are not strong typed, but I thought this may be on purpose if you plan to support older Nextcloud/PHP versions. (24 requires PHP>=7.4 so we can start using strong typed properties there. But you can also require PHP>=7.4 even if you support older Nextcloud versions I think, given 7.3 is EOL. Not 100% sure of the rules there.)

A strong step in the right direction for form sharing, congrats.

lib/Controller/ShareApiController.php
jotoeri: Hmm, interesting idea, to use the new permissions for this check, yes. But as we until now use this check on all api-routes - in fact for the edit-permission - i think i'd leave it for now. A sharing/resharing permission is still far apart i think. There's other things to develop first... 😉

lib/Controller/ShareApiController.php
jotoeri: 👍

lib/Controller/ShareApiController.php
jotoeri: Hmm, i remember we had that discussion on creating forms somewhen. I think regarding code-complexity (which is indeed not too much of a problem) vs. probability of occurence, i'd go with the 2-lines-less-coding way... In case of failure, the user will just click a second time. 😉

lib/Controller/ShareApiController.php
jotoeri: Good point, added that. 👍

lib/Migration/Version030000Date20211206213004.php
jotoeri: 🤷‍♂️ 👍

jotoeri:
Thank you in any case already for your work, @come-nc ! 💪 🚀 
A few things i fixed already, the rest i'll do the next days... 😉

lib/Controller/ShareApiController.php
```diff
@@ -0,0 +1,232 @@
...
+		// Check for valid shareWith, needs to be done separately per shareType
+		switch ($shareType) {
+			case IShare::Type_USER:
+				if (!($this->userManger->get($shareWith) instanceof IUser)) {
+					$this->logger->debug('Invalid user to share with.');
+					throw new OCSBadRequestException('Invalid user to share with.');
+				}
+				break;
+
+			case IShare::TYPE_GROUP:
```
PVince81, line 157: for later, also add circles ? (could be a "// TODO: add CIRCLES" and a separate ticket)

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,167 @@
...
+	 * @param array $options
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		$qb_fetch = $this->connection->getQueryBuilder();
+		$qb_update_access = $this->connection->getQueryBuilder();
+		$qb_insert_shares = $this->connection->getQueryBuilder();
+
+		$qb_fetch->select('id', 'access_json')
+			->from('forms_v2_forms');
+		$cursor = $qb_fetch->execute();
```
CarlSchwan, line 103:
```suggestion
		$qbFetch->select('id', 'access_json')
			->from('forms_v2_forms');
		$cursor = $qb_fetch->executeQuery();
```

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,167 @@
...
+
+			switch ($access['type']) {
+				case 'public':
+					$qb_update_access->setParameter('id', $row['id'], IQueryBuilder::PARAM_INT)
+						->setParameter('access_json', json_encode([
+							'legacyLink' => true,
+							'permitAllUsers' => false,
+							'showToAllUsers' => false,
+						]), IQueryBuilder::PARAM_STR);
+					$qb_update_access->execute();
```
CarlSchwan, line 123:
```suggestion
					$qb_update_access->executeStatement();
```

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,167 @@
...
+					break;
+
+				case 'registered':
+					$qb_update_access->setParameter('id', $row['id'], IQueryBuilder::PARAM_INT)
+						->setParameter('access_json', json_encode([
+							'permitAllUsers' => true,
+							'showToAllUsers' => true,
+						]), IQueryBuilder::PARAM_STR);
+					$qb_update_access->execute();
```
CarlSchwan, line 132:
```suggestion
					$qb_update_access->executeStatement();
```

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,167 @@
...
+					break;
+
+				case 'selected':
+					$qb_update_access->setParameter('id', $row['id'], IQueryBuilder::PARAM_INT)
+						->setParameter('access_json', json_encode([
+							'permitAllUsers' => false,
+							'showToAllUsers' => false,
+						]), IQueryBuilder::PARAM_STR);
+					$qb_update_access->execute();
```
CarlSchwan, line 141:
```suggestion
					$qb_update_access->executeStatement();
```

lib/Migration/Version030000Date20211206213004.php
come-nc: Yes but entries which were already migrated before will not have access['type'], no?
I did not see how you avoid migrating already migrated entries.

You have to take into account that your migration will run also on already migrated databases, in which case it should do nothing.

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,167 @@
...
+
+					foreach ($access['users'] as $user) {
+						$qb_insert_shares->values([
+							'form_id' => $qb_insert_shares->createNamedParameter($row['id'], IQueryBuilder::PARAM_INT),
+							'share_type' => $qb_insert_shares->createNamedParameter(IShare::TYPE_USER, IQueryBuilder::PARAM_INT),
+							'share_with' => $qb_insert_shares->createNamedParameter($user, IQueryBuilder::PARAM_STR),
+						]);
+						$qb_insert_shares->execute();
```
CarlSchwan, line 149:
```suggestion
						$qb_insert_shares->executeStatement();
```

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,167 @@
...
+					}
+					foreach ($access['groups'] as $group) {
+						$qb_insert_shares->values([
+							'form_id' => $qb_insert_shares->createNamedParameter($row['id'], IQueryBuilder::PARAM_INT),
+							'share_type' => $qb_insert_shares->createNamedParameter(IShare::TYPE_GROUP, IQueryBuilder::PARAM_INT),
+							'share_with' => $qb_insert_shares->createNamedParameter($group, IQueryBuilder::PARAM_STR),
+						]);
+						$qb_insert_shares->execute();
```
CarlSchwan, line 157:
```suggestion
						$qb_insert_shares->executeStatement();
```

lib/Db/ShareMapper.php
```diff
@@ -0,0 +1,131 @@
...
+namespace OCA\Forms\Db;
+
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Db\MultipleObjectsReturnedException;
+use OCP\AppFramework\Db\QBMapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+use OCP\Share\IShare;
+
+class ShareMapper extends QBMapper {
```
CarlSchwan, line 36:
```suggestion

/**
 * @template-extends QBMapper<Share>
 */
class ShareMapper extends QBMapper {
```

artonge (changes requested):
Some logic interrogation, and some readability suggestions, feel free to ignore some if they concern copy-pasted code.

lib/Controller/ShareApiController.php
jotoeri: Yep, that's planned, but for the refactoring i thought it would be too much within one PR. 😉  #1012

lib/Db/ShareMapper.php
jotoeri: Any specific functionality on that? (Or what do i have to search for to understand it? 😉) Tbh. we don't have that on any of our mappers... 🤔

lib/Migration/Version030000Date20211206213004.php
jotoeri: Oh man. - Now i see, what you mean with messy. :D Thanks!

lib/Migration/Version030000Date20211206213004.php
jotoeri: Ahh, you're right, yes. 👍 
I added a check for migrated forms.

lib/Service/FormsService.php
jotoeri: Hmm, until now we have the setting automatically shown as acitvated once there is a public link. We earlier said, that we do not want to offer the submit-restriction, if we essentially can not guarantee its functionality.
I'd leave it here for now, i think maybe some way of one-time shares would be a better way to provide a functionality, that we can actually manage.

src/Forms.vue
```diff
@@ -140,20 +145,18 @@ export default {
...
-				if (this.forms.findIndex(form => form.hash === this.routeHash) < 0) {
-					return false
-				}
+			// Check if route in permissions-list
+			if (this.forms.concat(this.sharedForms).find(form => form.hash === this.routeHash).permissions.includes(this.$route.name)) {
+				return true
 			}
 
-			return true
+			return false
```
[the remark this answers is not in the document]
jotoeri: ❤️ 👍

src/Forms.vue
```diff
@@ -193,6 +196,19 @@ export default {
...
+		 * Open a form and its sidebar for sharing
+		 *
+		 * @param {string} hash the hash of the form to load
+		 */
+		openSharing(hash) {
+			if (hash !== this.routeHash || this.$route.name !== 'edit') {
+				this.$router.push({ name: 'edit', params: { hash } })
+			}
+			this.sidebarActive = 'forms-sharing'
+			this.sidebarOpened = true
```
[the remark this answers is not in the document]
jotoeri: Yes, the sidebar-state is not reflected in the url on forms. The sidebar is only shown in edit-mode, thus if another form or e.g. the submit-view are currently open, this switches to the right forms edit-mode and opens the sidebar.

package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "forms",
 	"description": "Forms app for nextcloud",
-	"version": "2.5.0",
+	"version": "3.0.0-alpha.0",
```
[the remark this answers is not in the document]
jotoeri: Uhm, this is how we handled it the last times, to run the migrations for us... 🤔 
How would you alternatively ensure, migrations are run on all dev-instances, if not bumping the version?

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -0,0 +1,212 @@
...
+			return this.hasPublicLink || this.form.access.legacyLink || this.form.isAnonymous
+		},
+		disableSubmitMultipleExplanation() {
+			if (this.disableSubmitMultiple) {
+				return t('forms', 'This can not be controlled, if the form has a public link or stores responses anonymously.')
+			}
+			return ''
+		},
+		hasPublicLink() {
+			return this.form.shares.slice()
```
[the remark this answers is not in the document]
jotoeri: This is used to create a copy of the array, so `filter()` does not change the original share-data. We do this in several places on forms. Spread operators would be an alternative nowadays... 🤔

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -0,0 +1,212 @@
...
+				.filter(share => share.shareType === this.SHARE_TYPES.SHARE_TYPE_LINK)
+				.length !== 0
+		},
+
+		// Inverting submitOnce for UI here. Adapt downto Db for V3, if this imposes for longterm.
+		submitMultiple() {
+			if (this.disableSubmitMultiple) {
+				return true
```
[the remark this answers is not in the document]
jotoeri: Short version is good, indeed. The functionality however is right. If the property `allowMultiple` is disabled for users, we assume the allowMultiple-value internally as if beeing true.

src/components/SidebarTabs/SharingSearchDiv.vue
```diff
@@ -0,0 +1,304 @@
...
+		/**
+		 * Text when there is no Results to be shown
+		 *
+		 * @return {string}
+		 */
+		noResultText() {
+			if (this.loading) {
+				return t('forms', 'Searching …')
+			}
+			return t('forms', 'No elements found.')
```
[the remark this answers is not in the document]
jotoeri: 'Coding style' we use on forms is without else... 😉

jotoeri:
> Properties are not strong typed, but I thought this may be on purpose if you plan to support older Nextcloud/PHP versions. (24 requires PHP>=7.4 so we can start using strong typed properties there. 

Yes, we officially still support NC22 and just recently removed php 7.2 with NC21. ^^ Thus i'd leave it for now. We'll have a separate PR to change all of our classes, when the time has come. 😉

jotoeri:
Ok - So thank you, @artonge @CarlSchwan @come-nc for your reviews here! 🚀 Quite a few new things and ideas. I had a look onto all of it, partially just set a comment, but also implemented quite a few things.

So - i think if there's no more issues from your side, this should be good to go, then! ☺️ 🎉

lib/Db/ShareMapper.php
come-nc: It gives more information to psalm to improve its error detection, basically.
You can see the template in the docblocks of https://github.com/nextcloud/server/blob/master/lib/public/AppFramework/Db/QBMapper.php#L43

lib/Controller/ShareApiController.php
```diff
@@ -0,0 +1,233 @@
...
+					$this->logger->debug('Share Hash already exists.');
+					throw new OCSException('Share Hash exists. Please retry.');
+				} catch (DoesNotExistException $e) {
+					// Just continue, this is what we expect to happen (share hash not existing yet).
+				}
+				break;
+			
+			default:
+				// This passed the check for used shareTypes, but has not been found here.
+				$this->logger->warning('Unknown, but used shareType: {shareType}. Please file an issue on GitHub.', [ 'shareType' => $shareType ]);
```
come-nc, line 181: Throw as well?

lib/Migration/Version030000Date20211206213004.php
```diff
@@ -0,0 +1,182 @@
...
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('forms_v2_shares')) {
```
come-nc, line 60: Invert the if and exit early please

come-nc (approved):
Good apart from nitpicking.

I did not test.

jotoeri:
@artonge Do you wanna check your requested changes? Or should i just dismiss the review as i looked into your comments?

src/components/SidebarTabs/SettingsSidebarTab.vue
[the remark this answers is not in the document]
artonge: `filter()` does not change the original array

artonge (approved):

come-nc:
:rocket: :partying_face:

PR #1126, opened 2022-02-22
API v2

jotoeri:
This is in fact a part of #1087. Splitting this off for easier review.

- Moving the whole API to v2
- I first tried to keep API v1.1, however mapping the new sharing to the old way is way too hard for its benefit. Thus - even that not good at all - I'd drop API v1.1. To have at least some 'compatibility' for outside API users, a capabilities response should hint on which API exists on the instance. #1146
- Adapt API Documentation to changes
- Add Sharing-Part to API-Doc

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -54,9 +54,11 @@ jobs:
       - name: Set up Nextcloud
         env:
           DB_PORT: 4444
+          OC_PASS: test
         run: |
           mkdir data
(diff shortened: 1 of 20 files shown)


jotoeri:
- [ ] Revert #1162 to get proper response again

jotoeri:
Ok - so @Chartman123 @skjnldsv This should now be a quite simple review, as it is only the first commit to check. The second commit has been done with #1148

Chartman123:
> Ok - so @Chartman123 @skjnldsv This should now be a quite simple review, as it is only the first commit to check. The second commit has been done with #1148 

OK, I'll try to find some time to dive into this. :)

Chartman123 (approved):
Didn't run the code, but couldn't find any obvious problems by reviewing the code :)

jotoeri:
I tested it quite a few times meanwhile, i think. 😉

PR #1139, opened 2022-03-24
make forms OCS API CORS compatible

everlanes:
When accessing the public OCS API in an external web application the browser performs a CORS check.

These changes add the functionality, that the Forms API responds with the needed information to the OPTIONS request.
It is adding the builtin ApiController.preflightedCors endpoint and extends the Allowed Header list.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -65,7 +65,17 @@ return [
 	],
 
 	'ocs' => [
-		
+		// CORS Preflight
+		[
+			'name' => 'api#preflightedCors',
+			'url' => '/api/{apiVersion}/{path}',
+			'verb' => 'OPTIONS',
+			'requirements' => [
+				'path' => '.+',
+				'apiVersion' => 'v2'
+			]
+		],
+
 		// Forms
 		[
 			'name' => 'api#getForms',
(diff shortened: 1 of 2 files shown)


jotoeri:
Hey @everlanes,

thank you for watching into this and giving some contribution back! 🙂 I'm totally with you, that this would be good if the API works with CORS.
However - i just created a simple testpage on another subdomain of my testlinux. Unfortunately, on firefox the cors-request does still not work with only your PR. On the other hand, if i add the @CORS comment to the methods as said in the dev-docs, then the CORS-Request works, but the forms app within the cloud does not work anymore. When i integrated the OCS-Stuff somewhen, i got told that the CORS stuff is not properly implemented/working in nc-server. It seems to me, this is still the case...?!
How did you test this? Can you give an example, that you could make running with your PR?

Greets! :)
Jonas

jotoeri:
Oh sorry for not coming back here before... The point is, that this works for you more - i would say - 'by luck'. 😉 It works only on the insert procedure, while on the other endpoints it does not - and who would have thought, my tests ran with 'getForms'. 🙃 

I also created a PR now on server, which helps making it work (& thereby included yours, which is also necessary 😉 ). Let's see how fast we get this done... ;)

jotoeri:
So, this works now with NC25. Wanna test it once again, @everlanes?  😉

jotoeri (approved):

Chartman123 (approved):

Issue #818, opened 2021-03-01
Minversion 21 - Todo


jotoeri:
I assume this won't be the last point, so i create this overview already to have a common structure. :wink

- [ ] Move to NC-Internal DB-Types https://github.com/nextcloud/forms/issues/737

PR #1149, opened 2022-03-29
Move to NC-Internal Db-Types

### LINKED ISSUE #737: Minversion 21 - Move to NC internal db column Types
Hi there,

I found that your app uses some classes of Nextcloud's third party libraries. Since an upgrade of those for php8 compatibility, your app will have to be adjusted as well. See https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/upgrade-guide.html#doctrine-dbal for details. Your app uses `Doctrine\DBAL\Types\Type` which lost the old constants: https://github.com/doctrine/dbal/blob/3.0.x/UPGRADE.md#deprecated-type-constants.

Cheers

jotoeri:
Fixes #737 and thus #818  :)

### DIFF
--- a/lib/Migration/Version030000Date20220402100057.php
+++ b/lib/Migration/Version030000Date20220402100057.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2022 Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @author Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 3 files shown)


jotoeri:
Now also added the String-Length fix here. 🙃 - Fix #1152

lib/Migration/Version030000Date20220402100057.php
```diff
@@ -0,0 +1,71 @@
...
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$update_necesssary = false;
+
+		// Change Type of Description from string to text. Necessary due to length restrictions.
+		if ($schema->hasTable('forms_v2_forms') && $schema->getTable('forms_v2_forms')->hasColumn('description')) {
```
nickvergessen, line 50: This should be the case, so you should not need to check for it. Quite the opposite? When someone updates and this column/table is missing you want it to error so they know something is wrong.

jotoeri: Interesting point, yes. We had until now always these hasTable() used.... 🤔

I changed it. 👍

appinfo/info.xml
```diff
@@ -45,7 +45,7 @@
     <screenshot>https://raw.githubusercontent.com/nextcloud/forms/master/screenshots/forms3.png</screenshot>
 
     <dependencies>
-        <nextcloud min-version="20" max-version="24" />
+        <nextcloud min-version="21" max-version="24" />
```
nickvergessen, line 48: 21 is EOL already, could directly jump to 22

jotoeri: Jep, but that is in #1150, as it 'needs' one more adaption... ;)

skjnldsv (approved):

jotoeri:
Fixing php8 separately.

PR #1156, opened 2022-04-02
Remove old forms tables

### LINKED ISSUE #1155: Drop old forms tables
We still have the old tables of forms < v2 on db. I think with v3 we can drop these db-cadavers.

jotoeri:
In case somebody has still the old tables of Forms < v2 on DB, i think we can do a small cleanup now. 😉 

Closes #1155


skjnldsv (approved):

skjnldsv (changes requested):
Blocked

PR #1168, opened 2022-04-13
Rework Navigation

### LINKED ISSUE #422: Show number of responses on responses button
Show number of responses on responses button like "15 responses", and in the action menu as well

### LINKED ISSUE #423: Add navigation second line status
@jancborchardt Can you give details on this? Which information did you have in mind?

Chartman123:
Closes #422 and closes #423

This will partially close the two issues: Sharing will be reworked so it won't be easy to classify. Showing the number of answers redundant in the sidebar and on the responses button won't make that much sense.

### DIFF
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -21,14 +21,22 @@
   -->
 
 <template>
-	<AppNavigationItem ref="navigationItem"
-		:icon="icon"
+	<ListItem ref="navigationItem"
 		:title="formTitle"
 		:to="{
 			name: routerTarget,
 			params: { hash: form.hash }
 		}"
+		:counter-number="form.submissionCount"
+		:active="isActive"
+		:compact="true"
 		@click="mobileCloseNavigation">
+		<template #icon>
+			<div :class="icon" />
+		</template>
+		<template v-if="hasSubtitle" #subtitle>
+			{{ formSubtitle }}
+		</template>
 		<template v-if="!loading && !readOnly" #actions>
 			<ActionButton :close-after-click="true" icon="icon-share" @click="onShareForm">
 				{{ t('forms', 'Share form') }}
@@ -48,7 +56,7 @@
 				{{ t('forms', 'Delete form') }}
 			</ActionButton>
 		</template>
-	</AppNavigationItem>
+	</ListItem>
 </template>
 
 <script>
@@ -57,7 +65,7 @@ import { showError } from '@nextcloud/dialogs'
 import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
 import ActionRouter from '@nextcloud/vue/dist/Components/ActionRouter'
 import ActionSeparator from '@nextcloud/vue/dist/Components/ActionSeparator'
-import AppNavigationItem from '@nextcloud/vue/dist/Components/AppNavigationItem'
+import ListItem from '@nextcloud/vue/dist/Components/ListItem'
 import axios from '@nextcloud/axios'
 import moment from '@nextcloud/moment'
 
@@ -67,7 +75,7 @@ export default {
 	name: 'AppNavigationForm',
 
 	components: {
-		AppNavigationItem,
+		ListItem,
 		ActionButton,
 		ActionRouter,
 		ActionSeparator,
@@ -101,6 +109,16 @@ export default {
 			return 'icon-forms'
 		},
 
+		/**
+		 * Check if form is current form and set active
+		 */
+		isActive() {
+			return this.form.hash === this.$route.params.hash
+		},
+
+		/**
+		 * Check if form is expired
+		 */
 		isExpired() {
 			return this.form.expires && moment().unix() > this.form.expires
 		},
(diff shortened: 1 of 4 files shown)


jotoeri:
So - after quite a bunch of more css-shit, i think i got a solution, that i'd be fine with. How do you think, @Chartman123 ? 😉 
grafik

Chartman123:
Looking good :) I will check it out later.

jancborchardt:
Looking very nice! :) Just for clarification, does the number counter show the responses since last viewing, or the total response count?

Chartman123:
> does the number counter show the responses since last viewing, or the total response count?

At the moment only the total. If we want to show a highlighted counter for new responses since the last visit, we would need to store the last visit timestamp and compare it to the submissions for the form.

src/components/AppNavigationForm.vue
```diff
@@ -217,3 +256,62 @@ export default {
 
 }
 </script>
+
+<style lang="scss">
+
+.list-item-form {
```
skjnldsv, line 262: I don't think that is needed, you should not need custom css as the [`ListItem`](https://nextcloud-vue-components.netlify.app/#/Components/List%20items?id=listitem) should already cover the proper design

jancborchardt:
Thanks @Chartman123, total count is fine, no need to overcomplicate it since we have notifications for new responses. :)

src/components/AppNavigationForm.vue
jotoeri: That was basically me, introducing that... 🙃 
It implements two points:
- The Listitem as is, is really large, which is especially compared to the old AppNavigationItem a massive increase of used space, which is IMO not necessary. Especially if the form list becomes long, this will be quite annoying to use. Thus the CSS decreases its height a bit. Also the mix of items with/without second line needs a little more compact design to blend in properly.
- The ListItem is not made to be used without second line but with Counter. Thus the second part of the CSS brings this in a usable view. Would of course be better to solve that in the lib on long term, but that might be breaking changes there, so I'd for now solve our problem here.

The picture in my comment above basically shows the total result including all necessary variants, that we might show.
I can give you more details and images when I'm back home next week... (Or @Chartman123 can of course, too 😉 )

jotoeri: Old way with AppNavigationItem:
grafik

ListItem with CSS as in this PR:
Animation


ListItem without CSS, same content:
Animation

src/components/AppNavigationForm.vue
```diff
@@ -110,6 +119,19 @@ export default {
...
 		},
 
+		/**
+		 * Check if form is current form and set active
+		 */
+		isActive() {
+			if (this.form.hash === this.$route.params.hash) {
+				return true
+			}
+			return false
```
jotoeri, line 129: A bit shorter & cleaner, if you directly return the result:
```suggestion
			return this.form.hash === this.$route.params.hash
```

src/components/AppNavigationForm.vue
```diff
@@ -126,6 +148,23 @@ export default {
...
+		/**
+		 * Return expiration details for subtitle
+		 */
+		formSubtitle() {
+			if (this.form.expires) {
+				const expireDate = moment(this.form.expires, 'X').fromNow()
+				if (this.isExpired) {
+					return t('forms', 'Expired {relativeDate}', { relativeDate: expireDate })
+				}
+				return t('forms', 'Expires {relativeDate}', { relativeDate: expireDate })
```
jotoeri, line 160: Also a small code-beauty: With according naming, you can directly pass the variable without separately assigning.
```suggestion
				const relativeDate = moment(this.form.expires, 'X').fromNow()
				if (this.isExpired) {
					return t('forms', 'Expired {relativeDate}', { relativeDate })
				}
				return t('forms', 'Expires {relativeDate}', { relativeDate })
```

jotoeri (approved):
Small code-beauty would be good, but apart from that i'm happy with the current state! 🎉

src/components/AppNavigationForm.vue
jancborchardt: Sounds like something that should be fixed in the Vue components instead though?

That is, the height should ideally stay the same, but it's nice that entries without a subline have their title and counter vertically centered.

@skjnldsv @marcoambrosini

skjnldsv: > Sounds like something that should be fixed in the Vue components instead though?

Exactly what i was gnna say! :wink:

skjnldsv (changes requested):
Please fix the component instead

jotoeri:
Did a rebase here, maybe i'll find some time to look into the library. 😉

Chartman123:
> * [x]  To fix on public link if not logged in

Fixed by checking for $this->currentUser in the if statements

jotoeri:
I wont get to work on it on the weekend. Hopefully end of next week then.

jotoeri:
With https://github.com/nextcloud/nextcloud-vue/pull/2856 now looks like:
new

src/components/AppNavigationForm.vue
jotoeri: Done. :muscle:

Chartman123:
@jancborchardt @skjnldsv The Vue component has been made ready for this now. Could you please check again if we should adjust something here or if we can merge it? Thanks 🙃

jotoeri (approved):
Still fine by me... 👍 🙃

PR #1170, opened 2022-04-14
Update version on master

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -45,7 +45,7 @@
 	<screenshot>https://raw.githubusercontent.com/nextcloud/forms/master/screenshots/forms3.png</screenshot>
 
 	<dependencies>
-		<nextcloud min-version="22" max-version="24" />
+		<nextcloud min-version="22" max-version="25" />
 	</dependencies>
 
 	<navigations>


jotoeri:
We'll need it for the tests of course.
@skjnldsv How does that work with your new release-workflow? Needs most probably manual adaption before, then? 🤔

skjnldsv:
Yeah, it's a good question. I'm not sure we should have 25 here. Or else we'll have to edit to 24 before each release. Not very efficient :see_no_evil: 

But thinking about it, as long as 25 is not released, that doesn't change anything, as the only people that will install this app on 25 would be devs. So we could indeed keep this as 22-25, but when we're reaching 25 release date, we'll have to release a fully tested version of Forms on 25 anyway. So that doesn't change anything in then end.

Yeah, I think we should merge it.

skjnldsv (approved):

nickvergessen:
Yeah that is also my conclusion. The problem with not bumping the requirement is that you can't test with master anymore.
I used to manually change the info.xml before packaging and then reverting it, but since we now build via ci you would need to commit it all the time.

nursoda:
Interesting…but somewhat irritating. You assume –in case NC25 does have incompatible requirements– that

- you will timely release a bugfix version that reverts info.xml to 24.x
- admins and users only update their apps from the Nextcloud app store (and not from local caches)

While I agree that these assumptions hold true for most installations, it doesn't make the uneasiness go away that the human readable semantics is misleading. As admin, I also don't expect max-version downgrades. I'd propose to exclude the max-version checks if run on non-released nightly master/CI build. This obviously should not be discussed on an app's pull request…

jotoeri:
>I'd propose to exclude the max-version checks if run on non-released nightly master/CI build

Interesting point... @nickvergessen Is there any way of using the possibliity to 'activate untested app' via occ? 🤔

nickvergessen:
`occ app:enable --force` should do

jotoeri:
Interesting... We have that already on the phpunit tests 🙈. So probably we wouldn't even need to change the serverversion for running the tests. 🤔

nickvergessen:
Yeah I was not aware of the `--force` until recently. It might be the solution for all the cases in tests of server and when other apps are needed to test, as well as in multi apps themselves.

So a potentially new way of handling them would be:
1. Always develop with info.xml being set to supported versions only
2. Master of Nextcloud can be tested/developed against with the `--force` option
3. Once nearing a new major beta/rc/final of server, the version is added to info.xml

Issue #321, opened 2020-04-30
Description for each question


Pilzinsel64:
Hello!

Sometimes it is really helpful (also for the user) to have a quick description or introduction about a question, so he/she know exactly what you want and type in the current answer.

As example, see this Form here I made using Microsoft Forms:
[SM64 ROM Manager - Bug Report](https://forms.office.com/Pages/ResponsePage.aspx?id=DQSIkWdsW0yxEjajBLZtrQAAAAAAAAAAAAN__taH74pUOFIzWjFDT0JaQVFDTDVYUTFaMlRPQ1dWWS4u)

This feature already has been discussed also here: #316

ei8fdb:
This is a helpful feature. :100: agree. For me (doing user research) this would be a priority.

Hanskes:
# 624 doesn't seem like a duplicate to me. This is about a description above a single question. # 624 is about the ability to put multiple questions in a section. Multiple sections, each with its own title and description, would be very useful.

Hanskes:
#624 doesn't seem like a duplicate to me. This is about a description above a single question. #624 is about the ability to put multiple questions in a section. Multiple sections, each with its own title and description, would be very useful.

loelkes:
> #624 doesn't seem like a duplicate to me. This is about a description above a single question. #624 is about the ability to put multiple questions in a section. Multiple sections, each with its own title and description, would be very useful.

What I meant for #624 was to have text blocks without questions. To have the questions _inside_ of sections would be nice, but for me it would be enough for it to _look like_ they're in sections with text blocks between them.

Another solution is to hide the question by creating a hidden field but with the description/title still visible. It sounds like a dirty hack but I would consider my issue "solved".

skjnldsv:
> What I meant for #624 was to have text blocks without questions. To have the questions _inside_ of sections would be nice, but for me it would be enough for it to _look like_ they're in sections with text blocks between them.

Yes, thus why I re-opened :wink:

PR #1172, opened 2022-04-14
Add Question Description

jotoeri:
Adding a simple description to questions. Could somewhen be formattable, but for now implements #321 

One thing, that i personally dont like is that the scrollHeight of the textarea seems to have a minimum, that we cannot set. Thus, for short Descriptions, the textarea is relatively large and the content around jumps a bit when enabling/disabling edit-mode. So if sbdy has a good idea to solve that, please tell me, apart from that i think we can go for now with this...

grafik

<details>
<summary>Initial image</summary>

grafik

</details>

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -25,6 +25,18 @@ namespace OCA\Forms;
 
 class Constants {
 	/**
+	 * Maximum String lengths, the database is set to store.
+	 */
+	public const MAX_STRING_LENGTHS = [
+		'formTitle' => 256,
+		'formDescription' => 8192,
+		'questionText' => 2048,
+		'questionDescription' => 4096,
+		'optionText' => 1024,
+		'answerText' => 4096,
+	];
+
+	/**
 	 * !! Keep in sync with src/models/AnswerTypes.js !!
 	 */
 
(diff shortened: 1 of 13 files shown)


jotoeri:
Note to self:
- [ ] Move maxStringLengths to Constants::
https://github.com/nextcloud/forms/blob/4b23e63a67e720d06dfcd17466f7ae2266bdb682/lib/Controller/PageController.php#L96

Chartman123:
> and the content around jumps a bit

@jotoeri I wouldn't mind. For existing questions without description the lower elements also jump around when you enter the edit mode for that question later

Chartman123:
I was able to fix it for Chrome and at least enhance it for Firefox by adding rows="1" to the textarea.

Chartman123 (approved):
If you autosquash my fixup commit, I think that we're good to go here. Perhaps you can find an even better solution for firefox if you dive deeper into the web 🙂

jancborchardt (changes requested):
Very cool enhancement! :) Just 1 detail: Instead of using italic which can be difficult to read for people, we can use `color: var(--color-text-maxcontrast)`.

skjnldsv (approved):
Looks good aside from Jan's comment :)

jotoeri:
Changed to color maxcontrast. Now i got some other issue, that the color does not apply anymore as long as i hold the mouse button. @skjnldsv @jancborchardt Any idea, which class/property could kick in here? 🤔 

Animation

PR #1189, opened 2022-04-26
Radio box choice translation (french)

Jerome-Herbinet:
Hello,
First of all, I would like to inform you that I have searched for your project at Transiflex, but without success.
Explanation of the change made and desired: 
The problem is indeed only a vocabulary problem and not a functional problem. In a form, when we talk about multiple or unique choices, we are talking about the possibilities of answers given to the people who will fill in the form. 
If the choice is multiple, it means that the person will be able to choose several answers (check boxes). 
If the choice is not multiple, and therefore unique, we are in the case of a radio box.
So, I think that the notion of "multiple" for the radio box misleads the users (whether it is the one who creates the form or the one who will fill it).
All this thinking is done in the French linguistic culture, so I don't question the translations in other languages.
Sincerely yours,
Jérôme


Chartman123:
AFAIK, all Nextcloud translations are done here: https://www.transifex.com/nextcloud/nextcloud/

You can also try to contact other translators over in the community forums: https://help.nextcloud.com/c/translations/23

Jerome-Herbinet:
Thank your for your fast answer @Chartman123 
So I finally found the translations in the global Nextcloud project, in the "forms" sub-project and it turns out that someone already made an identical suggestion. So I voted (+1) his two proposals. Since I strongly believe that the current French translations for the creation of the radio boxes can be really misleading or suggest that there is a bug in the application, do you recommend, in order to get this taken care of quickly, to leave the pull-request in place or to close it? What do you recommend?

Chartman123:
I'd propose to close this PR as otherwise your changes will be overwritten again by the bot.

Glad I could help you :)

Jerome-Herbinet:
OK, i'll close immediately the PR and i've added a comment with "high priority" in Transiflex (copy/past of my GitHub PR explanation). Thank you for your help :-)

PR #1210, opened 2022-05-24
Show deleted userid on export

jotoeri:
Once a user is not found by the userManager, it is currently shown as anonymous.

This changes the behaviour such:
- For true anonymous users, old behaviour - no uid but DisplayName 'Anonymous user' is shown
- For users whose uid has been stored, but they are currently not found on the instance (e.g. deleted users), the uid is exported, just the displayName remains empty.

### DIFF
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -177,12 +177,21 @@ class SubmissionService {
 
 			// User
 			$user = $this->userManager->get($submission->getUserId());
+
+			// If user not found...
 			if ($user === null) {
-				// Give empty userId
-				$row[] = '';
-				// TRANSLATORS Shown on export if no Display-Name is available.
-				$row[] = $this->l10n->t('Anonymous user');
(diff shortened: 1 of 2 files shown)


Chartman123:
Should we really export the uid of deleted users? Might this be a problem regarding the GDPR?

jotoeri:
:thinking: Good question. The more i think about, the less i want to open that barrell. Maybe never touch a running system. :see_no_evil:

Chartman123:
> :thinking: Good question. The more i think about, the less i want to open that barrell. Maybe never touch a running system. :see_no_evil: 

We could however just add "Deleted User" as the Display Name if you want to have a distinction between anonymous and deleted users...

jotoeri:
Nah, thats not that important to me... I thought more about having the information of the user, as a deleted account not necessarily means a gdpr-request to delete all personal connected data. And the anonymisation of the submission in fact backwards changes the data IMO.... But that differentiation would then need much more thoughts and handling, so... 🤷‍♂️

PR #1270, opened 2022-07-29
Set border color to maxcontrast

### LINKED ISSUE #1254: Separator line is barely visible due to low contrast
Input fields are shown with a thin, grey line with too low contrast to the white background.

image

Chartman123:
This fixes #1254 by changing bottom-border color from `--color-border-dark` to `--color-text-maxcontrast`

### DIFF
--- a/src/components/Questions/AnswerInput.vue
+++ b/src/components/Questions/AnswerInput.vue
@@ -235,7 +235,7 @@ export default {
 	margin: 0;
 	padding: 0 0;
 	border: 0;
-	border-bottom: 1px dotted var(--color-border-dark);
+	border-bottom: 1px dotted var(--color-text-maxcontrast);
 	border-radius: 0;
 	font-size: 14px;
 	position: relative;
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -303,7 +303,7 @@ export default {
 			}
 
 			&__text[type=text] {
-				border-bottom-color: var(--color-border-dark);
+				border-bottom-color: var(--color-text-maxcontrast);
 				margin-top: 2px;
 			}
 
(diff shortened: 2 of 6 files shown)


jancborchardt (approved):
Much better, although I’m not sure if it really counts as fixing the reported issue, as it’s still not soo visible.

We could merge this as it’s definitely an improvement, and then use the TextInput component as soon as it’s done? We probably need to see how it looks like e.g. when you have multiple fields below each other like for a multiselect or radio options, cause there could be too many borders floating around. cc @marcoambrosini to keep in mind maybe?

marcoambrosini:
I'll add this style to the textField field component in a follow-up

Chartman123:
> Much better, although I’m not sure if it really counts as fixing the reported issue, as it’s still not soo visible.

@jancborchardt We could also set the border size to 2px instead of just one. That makes the line much more visible.

src/components/Questions/AnswerInput.vue
```diff
@@ -235,7 +235,7 @@ export default {
 	margin: 0;
 	padding: 0 0;
 	border: 0;
-	border-bottom: 1px dotted var(--color-border-dark);
```
jotoeri, line 238: Is it necessary here? As @jancborchardt already said, that looks quite hard now, if there are multiple fields below each other. Also we here have the Checkboxes as indicating element in front, while the issue mentioned the text-input fields, where this line is the input-indicator, as the checkboxes are for themselves. :thinking:

marcoambrosini: Could you post a screenshot @jotoeri please?

jotoeri: | Before | After |
|---|---|
| before | after |

Chartman123: I wouldn't mind the darker lines here. 🙈 But I could also live with the lighter ones. So I think that this should also be rated from an accessibility point of view. And perhaps to make the look more unified.

jotoeri: Anyways. Maybe i just have to geht used to it. 🙈

jotoeri (approved):

Issue #354, opened 2020-05-06
Allow multiple choice or checkbox answers to be reordered


ei8fdb:
**Which Forms version are you running?**
v2.0.0 beta 1 (build 20200505)

**Nextcloud or ownCloud?:**
Nextcloud

## Feature request
Allow multiple choice or checkbox answers to be reordered

## Rational
Right now if a form admin wants to force an ordering on answers they have to make sure they enter the answers in the correct order - reordering the answers is not possible. It would be helpful to allow reordering the answers.

## UI
Providing a "grab handle" (on the left or right hand side of the answer) would allow for this.

## Priority
Medium/Low

For me as a UX researcher, this would be a helpful option. It's not critical but it would make life much easier. :smile: 

Thoughts?

PR #1271, opened 2022-07-29
Feature: Allow shuffling of answer options | Sort options

### LINKED ISSUE #1007: Order of choices is not stable

**Describe the bug**
I put answers for a multiple choice or dropdown type of question but their order is not the same when I access then the form to complete it. It doesn't happen all the time, it looks random.

**To Reproduce**
Create a multiple answers or dropdown type question (didn't try with other)

**Expected behavior**
Order of answers should be consistent and follow what from author has set.

**Screenshots**
I did a video capture as it may be easier to understand.

https://user-images.githubusercontent.com/764332/124580143-22f28080-de50-11eb-9054-6408b401851a.mp4

The video shows only the case for multiple choices but it's the same for dropdowns from what I tested (and maybe other type also)

- Nextcloud-Version: 19
- Forms-Version: 2.2.4

 - OS: Mac 10.16.6
 - Browser FF
 - Version 89.0.2

**Additional context**

Thanks for your great help and this awesome app. Can't wait to use it fully :)

### LINKED ISSUE #1067: Ability to randomise order of multiple choice answers
- Nextcloud-Version: 19.0.13
- Forms-Version: 2.2.4

For some multiple choice questions it would be very useful to provide a purposefully unordered list of choices so as not to give positional weight to any answer(s).

This would optimally be a per-question setting so some questions would have static order while others are randomised.

susnux:
* Resolves #1067 and fixes #1007

---

This introduces an question option to randomize the order of the answer options.
Shuffling the order of the answer choices reduces bias in responses. (#1067)

If no shuffling is enabled the options are sorted in the order they were
created, as currently they are shown as returned by the database which
is not necessarily sorted. (#1007)

### DIFF
--- a/lib/Db/Question.php
+++ b/lib/Db/Question.php
@@ -44,6 +44,8 @@ use OCP\AppFramework\Db\Entity;
  * @method void setText(string $value)
  * @method string getDescription()
  * @method void setDescription(string $value)
+ * @method array getExtraSettings()
+ * @method void setExtraSettings(array $value)
  */
 class Question extends Entity {
 	protected $formId;
@@ -52,6 +54,7 @@ class Question extends Entity {
 	protected $isRequired;
 	protected $text;
 	protected $description;
+	protected $extraSettingsJson;
 
 	public function __construct() {
 		$this->addType('formId', 'integer');
@@ -60,6 +63,15 @@ class Question extends Entity {
 		$this->addType('isRequired', 'bool');
 		$this->addType('text', 'string');
 		$this->addType('description', 'string');
+	}
+
+	public function getExtraSettings(): object {
+		return json_decode($this->getExtraSettingsJson() ?: '{}');
+	}
+
+	public function setExtraSettings(array $extraSettings) {
+		// Make sure to be an object (empty assoc. array)
+		$this->setExtraSettingsJson(json_encode($extraSettings, JSON_FORCE_OBJECT));
 	}
 
 	public function read(): array {
@@ -71,6 +83,7 @@ class Question extends Entity {
 			'isRequired' => $this->getIsRequired(),
 			'text' => htmlspecialchars_decode($this->getText()),
 			'description' => htmlspecialchars_decode($this->getDescription()),
+			'extraSettings' => $this->getExtraSettings(),
 		];
 	}
 }
--- a/lib/Migration/Version030000Date20220831195000.php
+++ b/lib/Migration/Version030000Date20220831195000.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2022 Ferdinand Thiessen <rpm@fthiessen.de>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OCA\Forms\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version030000Date20220831195000 extends SimpleMigrationStep {
+
+	// Types::JSON is only available starting with NC24, we still support NC22
+	private const TYPE_JSON = 'json';
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+		$table = $schema->getTable('forms_v2_questions');
+
+		if (!$table->hasColumn('extra_settings_json')) {
+			$table->addColumn('extra_settings_json', self::TYPE_JSON, [
+				'notnull' => false,
+			]);
+
+			return $schema;
+		}
+		
+		return null;
+	}
+}
--- a/src/mixins/QuestionMixin.js
+++ b/src/mixins/QuestionMixin.js
@@ -104,6 +104,17 @@ export default {
 			type: Object,
 			required: true,
 		},
+
+		/**
+		 * Extra settings
+		 * Currently only contains whether options should be shuffled
+		 */
+		extraSettings: {
+			type: Object,
+			default: () => {
+				return {}
+			},
+		},
 	},
 
 	components: {
@@ -115,6 +126,17 @@ export default {
 			// Do we display this question in edit or fill mode
 			edit: false,
 		}
+	},
+
+	computed: {
+		sortedOptions() {
+			// Only shuffle options if not in editing mode (and shuffling is enabled)
+			if (!this.edit && this.extraSettings?.shuffleOptions) {
+				return this.shuffleArray(this.options)
+			}
+			// Ensure order of options always is the same
+			return [...this.options].sort((a, b) => a.id - b.id)
+		},
 	},
 
 	methods: {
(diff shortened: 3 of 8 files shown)


src/mixins/QuestionMixin.js
```diff
@@ -116,6 +121,12 @@ export default {
 		}
 	},
 
+	computed: {
+		sortedOptions() {
+			return this.shuffleOptions ? [...this.options].sort(() => 0.5 - Math.random()) : [...this.options].sort((a, b) => a.id - b.id)
```
jotoeri, line 126: I'm not sure, if this will solve the issue or probably just shift it. Afaik Prostgresql can reuse ids if a row got deleted. Thus it might happen, that a newly created option has a smaller id than an older one.
This more as a thought. From my side we can also go with this for now.

lib/Db/Question.php
```diff
@@ -58,6 +59,7 @@ public function __construct() {
 		$this->addType('order', 'integer');
 		$this->addType('type', 'string');
 		$this->addType('isRequired', 'bool');
+		$this->addType('shuffleOptions', 'bool');
```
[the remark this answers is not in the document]
susnux: Sounds good!
Would you prefer the extra column on the question or a extra table like

<table>
<tr><th colspan=3> question_options</th></tr>
<tr><td>question</td><td>option</td><td>value</td></tr>
</table>

I only ask because the latter option is used by another project I work on and allows querying on the database for that option (e.g. allows filtering on the database layer).

src/mixins/QuestionMixin.js
susnux: The `SERIAL` data type just uses a `SEQUENCE` so I do not think it will ever reuse IDs.

But we could add a `order` property so the forms author could easily swap the entries or rearrange the order of them.
That would also fix / implement https://github.com/nextcloud/forms/issues/354

Chartman123: Yes, that would be nice to make the order adjustable while already touching the ordering 👍🏻

lib/Db/Question.php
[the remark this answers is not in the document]
jotoeri: Uhhm, i think i would for now go with a column on the questions-table, that should suffice, i think.

src/mixins/QuestionMixin.js
Chartman123: We could rely on this here: https://github.com/SortableJS/Vue.Draggable

Or are there any other Nextcloud apps that already have an UI for re-ordering records of a list?

susnux: > We could rely on this here: https://github.com/SortableJS/Vue.Draggable

Yes that is nice and already a dependency of forms, I will try to rewrite this PR to comply with the comments :)

jotoeri: ❤️

Just as an idea, that came to my mind: Having the draggable-icon as a hover on the checkbox/radio dummys i could imagine of a way without UI Overload, but sufficiently intuitive. 🤔 @jancborchardt Any idea for a good UI to drag options around?

lib/Db/Question.php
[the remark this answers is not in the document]
susnux: Changed it accordingly

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
 
 ### Deprecated
 
+### Fixed
```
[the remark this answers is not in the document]
susnux: Ah thank you, reverted that change :)

src/mixins/QuestionMixin.js
susnux: I think it is better to postpone this, as it is better suited in a following, separate, pull request. Or what do you think?

Chartman123: Fine by me 🙂

susnux: Ok than this PR is ready to be reviewed and later on I will submit an other one providing reorder support based on this

lib/Migration/Version030000Date20220831195000.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2022
```
jotoeri, line 5: You can add your own name there, if you want! 😉

lib/Migration/Version030000Date20220831195000.php
```diff
@@ -0,0 +1,57 @@
...
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version030000Date20220831195000 extends SimpleMigrationStep {
+
+	// Types::JSON is only available starting with NC24, we still support NC22
```
jotoeri, line 33: Oh, nice, didn't know that! 🚀

lib/Migration/Version030000Date20220831195000.php
susnux: If you want me to do it, I can add it but having my name somewhere is not important for me^^

jotoeri (approved):
Should be fine, i think 🤔

lib/Migration/Version030000Date20220831195000.php
jotoeri: Haha, i'm no lawyer, if i think about, if a Copyright should have somebody owning the right 🤔
But don't care, fine by me without.

jotoeri (changes requested):
Argh, sorry, i was too fast pushing the button.
Just getting this on console, when adding a new question:
grafik

susnux:
> Just getting this on console, when adding a new question

Oh yes that happens if there are no settings as the empty array would be serialized as an array instead of an object (assoc. array). This is now fixed.

lib/Migration/Version030000Date20220831195000.php
susnux: Added it

jotoeri:
Top 👍 
Just one more thing i found now:
When checking/unchecking the Shuffle-Checkbox, we should update the computed property of `sortedOptions`. Currently, this only takes effect after reloading the page (while editing the form).

susnux:
Hm while editing (at least in the editing view) options should not be shuffled (only in the submit view).
But I will test this as soon as I got some time for preparing a test env, as currently master is broken (due to `nextcloud-eventbus` 3.0.1 dependency of `nextcloud-vue` which is already fixed but 3.0.2 is not released)

jotoeri:
> Hm while editing (at least in the editing view) options should not be shuffled (only in the submit view).

When editing a single question, i'm totally with you. Then they should be ordered.
But on Form-Editing, when not editing a single question (i.e. Within Create.vue, but `edit=false` of Question.vue), the options are currently unsorted (if set). I think that is fine, as this situation should indicate how the final form looks like, but it should also be reactive, so when unchecking the shuffle, then the options are directly shown as sorted (and the other way round).

> due to nextcloud-eventbus 3.0.1 dependency of nextcloud-vue which is already fixed but 3.0.2 is not released

Do you know `npm link`? You can relatively quickly just clone eventbus master, ci, build, link and then link `@nextcloud/event-bus` on forms. Then it works with current master, so you can use it already. 😉

susnux:
> When checking/unchecking the Shuffle-Checkbox, we should update the computed property of `sortedOptions`. Currently, this only takes effect after reloading the page (while editing the form).

Fixed that.

One general thing I noticed: Currently every interaction is debounced, resulting in a slight UI latency. Would it not be smoother to just debounce the request to the server? (As the UI is currently also not checking the request for success, meaning there is no rollback if that request failed).

jotoeri (commented):
- Old Questions are not loaded/shown (i.e. Questions with `ExtraSettingsJson = NULL`)
- The way of sorting is quite biased, no? Can we get to a Fisher-Yates here? Should be of same Order, but much more effective.

- Can you please test it on your own properly? Takes also quite much time for me, to do this...

jotoeri:
> One general thing I noticed: Currently every interaction is debounced, resulting in a slight UI latency.

Hmm, i don't think the delay is much of a problem. It is anyways only visible on the two boolean switches and if we fix the actions to be inside of the Editing-Mode, then it even won't be visible that much. (#1329) I also have in mind, that chaning this would also increase the load on the browser due to permanently updating and triggering corresponding reactivities.

susnux:
@jotoeri Fixed your comments. Sorry for the inconvenience,  I will try to test better.

jotoeri (approved):
Should be fine now. :+1:

jotoeri:
@Chartman123 wanna double-check here? :wink:

lib/Db/Question.php
```diff
@@ -62,6 +65,15 @@ public function __construct() {
...
 	}
 
+	public function getExtraSettings(): array {
+		return json_decode($this->getExtraSettingsJson() ?: '{}', true);
+	}
+
+	public function setExtraSettings(array $extraSettings) {
+		// Make sure to be an object (empty assoc. array)
+		$this->setExtraSettingsJson(json_encode($extraSettings, JSON_FORCE_OBJECT));
+	}
```
Chartman123, line 75: I get a warning/error in the browser console:

```
[Vue warn]: Invalid prop: type check failed for prop "extraSettings". Expected Object, got Array 
```

Probably because of using arrays here and an object there: https://github.com/nextcloud/forms/blob/217d12a63cc7dd5f7032068794db7a9dbc296088/src/mixins/QuestionMixin.js#L112-L116

Chartman123: Basically the same as @jotoeri's comment here: https://github.com/nextcloud/forms/pull/1271#pullrequestreview-1093979213

susnux: That are associative arrays for php but objects for js (`JSON_FORCE_OBJECT` for the encode part and `associative=true` for the decode part).

Are you sure you are on hash 217d12a63cc7dd5f7032068794db7a9dbc296088 ?

Chartman123: Yes, did a double check right now: 

image

Chartman123: @susnux As far as I understand the associative=true returns an array if the json is null... That would totally explain the error in the first place: It appears for questions not having extraSettings. The frontend however expects an object...

grafik

Chartman123:
```
	public function getExtraSettings(): object {
		return json_decode($this->getExtraSettingsJson() ?: '{}');
	}
```
This way it works fine for me...

susnux: > This way it works fine for me...

Yes that is working, I now changed the code accordingly.

Chartman123 (approved):

PR #1281, opened 2022-08-04
Move to button component & first material icons

jotoeri:
Moving to NcButton instead of manually styled ones.

New Toolbar is icon-only with tooltip 🙂:

grafik
grafik

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -31,24 +31,25 @@
 
 	<NcAppContent v-else>
 		<!-- Show results & sidebar button -->
-		<TopBar>
-			<template #default>
-				<button @click="showResults">
-					<span class="icon-comment" role="img" />
-					<!-- TRANSLATORS Button to switch to the Result-View -->
-					{{ t('forms', 'Results') }}
-				</button>
-				<button v-if="!sidebarOpened" @click="onShareForm">
-					<span class="icon-share" role="img" />
-					{{ t('forms', 'Share form') }}
-				</button>
-			</template>
-			<template #small>
-				<button v-tooltip="t('forms', 'Toggle settings')"
-					@click="toggleSidebar">
-					<span class="icon-menu-sidebar" role="img" />
-				</button>
-			</template>
+		<TopBar :sidebar-opened="sidebarOpened"
+			@update:sidebarOpened="onSidebarChange">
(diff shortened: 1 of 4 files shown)


src/views/Create.vue
```diff
@@ -31,24 +31,25 @@
...
-			</template>
+		<TopBar :sidebar-opened="sidebarOpened"
+			@update:sidebarOpened="onSidebarChange">
+			<NcButton v-tooltip="t('forms', 'Results')"
+				:aria-label="t('forms', 'Results')"
+				type="tertiary"
+				@click="showResults">
+				<template #icon>
+					<IconMessageReplyText :size="20" />
+				</template>
```
Chartman123, line 42: What about making this button a 'primary' button with the text?
grafik

jotoeri: Hm, i think we have too many blue elements then. 
Primary should only be one button imo (i.e. Add question), secondary is imo too shady and already very present on the appnavigation and without outline (tertiary with text), the buttons are not separated. I think this is how i came there to keep a clean and simple design overall. 🤔 

What i thought about, to have the results-access a bit less hidden, was to add a results button to the appnavigation-actions again?

Chartman123: Ok, and if we kept the tertiary (or tertiary with background) style but just add the text?

Isn't the results button already in the action button of the navigation?

jotoeri: Tertiary with background does not exist. And without background the separation of the buttons is not given. This is exactly why i removed the text. 😉 

The results button was once in the appnavigation. I'm not sure, but i think we removed it, when adding the results-button to the main-view.

grafik

Chartman123: https://github.com/nextcloud/forms/blob/master/src/components/AppNavigationForm.vue#L44-L50

It's still there :)

So for me the icon only button is also fine.

jotoeri: Uhh, then there is another problem 🤔 
grafik

Chartman123: grafik

Yes, for me it's clearly visible ;-)

Chartman123 (approved):

skjnldsv (approved):

jancborchardt:
Hm, would still be good to use a button with text there? Secondary with icon for both "Results" and "Back to questions" would be good?

PR #1283, opened 2022-08-05
Use MutliSelect for dropdown questions

### LINKED ISSUE #498: Use the multiselect component in dropdown component
> Because the call on whether to use the multiselect component or its own component has changed around so much, I’d prefer if you can create that follow-up issue with the relevant info. :)

That would make the create preview also be the same as what the real final form would look like :rocket:

cc @jancborchardt
_Originally posted by @skjnldsv in https://github.com/nextcloud/forms/pull/461#pullrequestreview-443725230_

Chartman123:
This fixes #498 

### DIFF
--- a/src/components/Questions/QuestionDropdown.vue
+++ b/src/components/Questions/QuestionDropdown.vue
@@ -36,23 +36,17 @@
 		@update:description="onDescriptionChange"
 		@update:isRequired="onRequiredChange"
 		@delete="onDelete">
-		<select v-if="!edit"
-			:id="text"
+		<Multiselect v-if="!edit"
+			v-model="selectedOption"
 			:name="text"
+			:placeholder="selectOptionPlaceholder"
 			:multiple="isMultiple"
 			:required="isRequired"
-			class="question__content"
-			@change="onChange">
-			<option value="">
-				{{ selectOptionPlaceholder }}
-			</option>
-			<option v-for="answer in options"
-				:key="answer.id"
-				:value="answer.id"
-				:selected="isChecked(answer.id)">
-				{{ answer.text }}
-			</option>
-		</select>
+			:options="options"
+			:searchable="false"
+			label="text"
+			track-by="id"
+			@select="onSelect" />
 
 		<ol v-if="edit" class="question__content">
 			<!-- Answer text input edit -->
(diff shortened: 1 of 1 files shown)


Chartman123:
@jotoeri Do you know why the selected option doesn't show up in the dropdown? If I understand it correctly that should be the default behaviour for the `option` slot...

jotoeri:
Maybe you need to provide a variable for storing this. Thus having a variable and using v-model should work, i think :thinking: 
https://nextcloud-vue-components.netlify.app/#/Components/Multiselect

Chartman123:
> Thus having a variable and using v-model should work

@jotoeri Thanks, that did the trick :)

jancborchardt:
Ah I get it, this is only for the creation process?

Considering your question in the issue @Chartman123 
> Do you mean that the component should be used to add new options, too, or only for the "real" preview as long as you're not in the edit mode for the options? Because I couldn't find a way to add new options directly within the Multiselect component other than for tags.

Ideally yes, it could be possible to add new options there too. But @skjnldsv or other Vue component experts will have to chime in on if that’s already possible.

Chartman123:
> Or is this only to use the component and multiselection is limited?

By now the dropdown allows only one option to be selected, however, there are some code elements prepared for multiple options to be selected.

Yes would be nice if some Vue experts could join this conversation :)

src/components/Questions/QuestionDropdown.vue
```diff
@@ -36,23 +36,16 @@
...
-				{{ selectOptionPlaceholder }}
-			</option>
-			<option v-for="answer in options"
-				:key="answer.id"
-				:value="answer.id"
-				:selected="isChecked(answer.id)">
-				{{ answer.text }}
-			</option>
-		</select>
+			:options="options"
```
jotoeri, line 44: Placeholder is missing 😉

jotoeri (commented):
Just one small thing.

jotoeri:
> Ideally yes, it could be possible to add new options there too. 

So as far as i see, the MultiSelect Component does currently not provide that functionality.

> Or is this only to use the component and multiselection is limited?

Yes, i'd keep this one for now only to move the dropdown to the component. We can add the other functionality later.

And here's a screenshot of how users can submit then: 😉 
grafik

jancborchardt (approved):
Seems good to me then :)

src/components/Questions/QuestionDropdown.vue
Chartman123: Fixed it, so we're good to go 🙂

PR #1298, opened 2022-08-18
Use NC/vue beta 3

jotoeri:
So -  to go on here, lets start by moving to the beta, then fixing and improving stuff. 😉


Chartman123 (approved):
Looks good :)

PR #1310, opened 2022-08-23
Implement warning when leaving an unsubmitted form

### LINKED ISSUE #319: Alert when leaving the page with an unfinished filled form
To avoid data loss

Chartman123:
Fixes #319 by adding a warning

### DIFF
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -145,6 +145,32 @@ export default {
 
 	mixins: [ViewsMixin],
 
+	/*
+	 * This is used to confirm that the user wants to leave the page
+	 * if the form is unsubmitted.
+	 */
+	beforeRouteUpdate(to, from, next) {
+		// This navigation guard is called when the route parameters changed (e.g. form hash)
+		// continue with the navigation if there are no changes or the user confirms to leave the form
+		if (this.confirmLeaveForm()) {
+			next()
+		} else {
+			// Otherwise cancel the navigation
+			next(false)
+		}
+	},
+
+	beforeRouteLeave(to, from, next) {
+		// This navigation guard is called when the route changed and a new view should be shown
+		// continue with the navigation if there are no changes or the user confirms to leave the form
+		if (this.confirmLeaveForm()) {
+			next()
+		} else {
+			// Otherwise cancel the navigation
+			next(false)
+		}
+	},
+
 	props: {
 		isLoggedIn: {
 			type: Boolean,
@@ -164,6 +190,8 @@ export default {
 			answers: {},
 			loading: false,
 			success: false,
+			/** Submit state of the form, true if changes are currently submitted */
+			submitForm: false,
 		}
 	},
 
@@ -232,6 +260,13 @@ export default {
 		},
 	},
 
+	beforeDestroy() {
+		window.removeEventListener('beforeunload', this.beforeWindowUnload)
+	},
+	created() {
+		window.addEventListener('beforeunload', this.beforeWindowUnload)
+	},
+
 	beforeMount() {
 		// Public Views get their form by initial-state from parent. No fetch necessary.
 		if (this.publicView) {
@@ -266,11 +301,34 @@ export default {
 			this.$refs.submitButton.click()
 		},
 
+		/*
+		 * Methods for catching unwanted unload events
+		 */
+		beforeWindowUnload(e) {
+			if (!this.confirmLeaveForm()) {
+				// Cancel the window unload event
+				e.preventDefault()
+				e.returnValue = ''
+			}
+		},
+
+		/**
+		Check if the form contains unsaved changes, returns true if the the form can be leaved safely, false if the navigation should be canceled.
+		 */
+		confirmLeaveForm() {
+			return (
+				this.submitForm
+				|| Object.keys(this.answers).length === 0
+				|| confirm(t('forms', 'You have unsaved changes! Do you still want to leave?'))
+			)
+		},
+
 		/**
 		 * Submit the form after the browser validated it 🚀
 		 */
 		async onSubmit() {
 			this.loading = true
+			this.submitForm = true
 
 			try {
 				await axios.post(generateOcsUrl('apps/forms/api/v2.1/submission/insert'), {
@@ -295,6 +353,7 @@ export default {
 			this.answers = {}
 			this.loading = false
 			this.success = false
+			this.submitForm = false
 		},
 	},
 


Chartman123:
@jotoeri any ideas why the custom message isn't shown in the window.confirm() dialog?

jotoeri:
Seems like just `confirm` should work. But to be honest, i currently can't tell you exactly, why. :laughing:

https://github.com/nextcloud/forms/blob/6c257edeea2231aeb55f1b62ab85dbb49d6ecef7/src/components/AppNavigationForm.vue#L207

Chartman123:
@jotoeri Tried it with just confirm() but doesn't work either... Even with static text. Did just research and it seems that it doesn't accept custom text for the onbeforeunload event

jotoeri:
Ok, sorry @Chartman123 i checked it once before quickly, but i definitely have not time to get into detail currently. Needs probably to be postponed.

Chartman123:
No problem, I think the design parts are more important for 3.0 than this here :)

susnux:
`confirm` and `alert` are not allowed within `beforeunload`.
Also all browser dropped support for custom messages, so you only can use the default one.


    Chrome removed support for custom message in ver 51
    Opera removed support for custom message in ver 38
    Firefox removed support for custom message in ver 44.0
    Safari removed support for custom message in ver 9.1

Chartman123:
Yes, that's true. I'm currently trying to move the warning to a NcModal like in the Cookbook app... This should at least work for leaving a form without closing the window/tab.

src/views/Submit.vue
```diff
@@ -262,11 +289,32 @@ export default {
...
+				e.returnValue = ''
+			}
+		},
+		confirmLeavingPage() {
+			return confirm(t('forms', 'You have unsaved changes! Do you still want to leave?'))
+		},
+		confirmStayInEditedForm() {
+			return (
+				!this.submitForm
+                && !this.confirmLeavingPage()
```
susnux, line 308: This will even show the dialog if there are *no* unsaved changes, even if you simply in the submit view of your own form for previewing and want to switch back to edit view.
I would suggest to check if there are any answers already like this

```suggestion
				!this.submitForm
				&& Object.keys(this.answers).length > 0
				&& !this.confirmLeavingPage()
```

susnux (changes requested):
Currently every time you open the submit view a confirmation dialog is shown, see code comment.

Moreover if you click to stay on the site, an error is thrown in the console because the navigation gets canceled (as requested) by the navigation guard.
So we should add a catch block to the `router.push` calls on the `TopBar` component and ignore every navigation guard related errors.

Chartman123:
@susnux Thanks for your input :) I'm also planning to move the warning to a modal. If you want, you can help me there. Everytime I tried it ended up being even worse 😆

Chartman123:
@susnux I think I got this working as expected right now. I think we should merge this without the modal right now, so that we can take it as a basis for @hamza221's PR regarding the local storage.

src/views/Submit.vue
```diff
@@ -145,6 +145,21 @@ export default {
 
 	mixins: [ViewsMixin],
 
+	/*
+     * This is used to confirm that the user wants to leave the page
+     * if the form is unsubmitted.
+     */
```
susnux, line 151:
```suggestion
	 * This is used to confirm that the user wants to leave the page
	 * if the form is unsubmitted.
	 */
```

Mix of tab and spaces

src/views/Submit.vue
```diff
@@ -145,6 +145,21 @@ export default {
...
+	beforeRouteUpdate(to, from, next) {
+		// beforeRouteLeave is called when the static route changes.
```
susnux, line 153: Comment is about "beforeRouteLeave" but "beforeRouteUpdate" is used, I think the comment is wrong as "beforeRouteUpdate" seems to be the proper guard?

src/views/Submit.vue
```diff
@@ -145,6 +145,21 @@ export default {
...
+		// Cancel the navigation, if the form has unsaved edits and the user did not
+		// confirm leave. This prevents accidentally losing changes
+		if (this.confirmLeaveForm()) {
+			next()
+		}
+
+		return false
```
susnux, line 160: This seems to be wrong, in the unload event listener you have:
`if (confirmLeaveForm) then cancel`

But here it is reversed, I would stick with this logic here as "confirmLeave" sounds like it is confirmed to leave if true, so if true -> leave. Otherwise cancel.

Also we still use vue-router 3 so the syntax should be:
```suggestion
		if (this.confirmLeaveForm()) {
			next()
		}

		next(false)
```

src/views/Submit.vue
```diff
@@ -266,11 +289,31 @@ export default {
 			this.$refs.submitButton.click()
 		},
 
+		/*
+		 * Methods for catching unwanted unload events
+		 */
+		beforeWindowUnload(e) {
+			if (this.confirmLeaveForm()) {
```
susnux, line 296: See above, the logic seems inverted, if leaving is confirmed we should *not* cancel the event.
```suggestion
			if (!this.confirmLeaveForm()) {
```

src/views/Submit.vue
```diff
@@ -266,11 +289,31 @@ export default {
...
+				e.returnValue = ''
+			}
+		},
+
+		confirmLeaveForm() {
+			return (
+				!this.submitForm
+				&& Object.keys(this.answers).length > 0
+				&& confirm(t('forms', 'You have unsaved changes! Do you still want to leave?'))
+			)
```
susnux, line 308: See above.
It sounds like leaving is confirmed, so we should return whenever the navigation should continue and false if we want to cancel the navigation.

```suggestion
		/**
		Check if the form contains unsaved changes, returns true if the the form can be leaved safely, false if the navigation should be canceled.
		 */
		confirmLeaveForm() {
			return (
				this.submitForm
				|| Object.keys(this.answers).length === 0
				|| confirm(t('forms', 'You have unsaved changes! Do you still want to leave?'))
			)
```

src/views/Submit.vue
```diff
@@ -164,6 +179,7 @@ export default {
 			answers: {},
 			loading: false,
 			success: false,
+			submitForm: false,
```
susnux, line 182:
```suggestion
			/** Submit state of the form, true if changes are already submitted */
			submitForm: false,
```

Some kind of documentation would be good as the name is not that self explanatory.

susnux (changes requested):

src/views/Submit.vue
Chartman123: Oh, some copy paste leftovers... 🙈

src/views/Submit.vue
Chartman123: Yes, beforeRouteLeave doesn't work if you work with parameters in the router path. So I'll change the comment here

src/views/Submit.vue
Chartman123: Did you already consider reversing the logic in this suggestion? Or shouldn't it be !this.confirmLeaveForm() then?

susnux: All suggestions were based on the reversed logic, meaning confirmLeaveForm return true if you should continue with the navigation.

So here is no negation needed

src/views/Submit.vue
```diff
@@ -145,6 +145,21 @@ export default {
...
+	beforeRouteUpdate(to, from, next) {
+		// This navigation guard is called when the route changes.
+		// continue with the navigation if there are no changes or the user confirms to leave the form
+		if (this.confirmLeaveForm()) {
+			next()
+		}
+
+		// Otherwise cancel the navigation
+		next(false)
+	},
```
susnux, line 161:
```suggestion
		// This navigation guard is called when the route parameters changed (e.g. form hash)
		// continue with the navigation if there are no changes or the user confirms to leave the form
		if (this.confirmLeaveForm()) {
			next()
		} else {
			// Otherwise cancel the navigation
			next(false)
		}
	},

	beforeRouteLeave(to, from, next) {
		// This navigation guard is called when the route changed and a new view should be shown
		// continue with the navigation if there are no changes or the user confirms to leave the form
		if (this.confirmLeaveForm()) {
			next()
		} else {
			// Otherwise cancel the navigation
			next(false)
		}
	},
```

After testing this PR: We need both route update and route leave. Otherwise not all cases are handled.

susnux (approved):
Tested and works now perfectly!

Issue #1023, opened 2021-07-28
Fill out form myself


vzsze:
**Nextcloud (please complete the following information):**
- Nextcloud-Version: [e.g. 20.0.11]
- Forms-Version: [e.g. 2.3.0]

**Is your feature request related to a problem? Please describe.**
After setting up a form, I need to fill it out myself. When opening the forn in the forms tab, I get to the form editor, but can't seem to fill out the form.

**Describe the solution you'd like**
I would like to be able to fill out a form that I created myself, too.

**Describe alternatives you've considered**
Using a separate user to create and to fill out forms. But this is often not possible.

jotoeri:
Hey @vzsze,

if you copy the form-link, you can open it yourself in your browser. Alternatively just remove the "/edit"-part from the url.

Greets,
Jonas

@jancborchardt How do you think about a separate Link "Fill form" in the Actions on App-Navigation? Or does it become too much there then already? Or maybe even replace the copy-link button on app-navigation by a link to fill the form?

vzsze:
Thank you for pointing out the workaround with the form-link. An UI-Item in the Actions would be nice.

JanLikar:
Sure, it's possible to copy the link, but I think we can agree that's not really an ideal UX flow.

If the maintainers agree, I'm willing to add the button into the menu.
 
In addition to that, I would argue it makes more sense to make this view the default view when the owner clicks on the form. But that's a different story.

jotoeri:
> In addition to that, I would argue it makes more sense to make this view the default view when the owner clicks on the form. But that's a different story.

Interesting idea. And should be not too difficult to solve: Default view is fill-view, then a button linking to edit-mode in the top-bar and in the menu. @jancborchardt @nimishavijay ?

nimishavijay:
This sounds like a good idea. A button saying "View form" or something would be nice, that opens the form in a new tab. Google Forms and Typeform use just an eye icon for this, so I'm guessing it's a pattern now. 

![image](https://user-images.githubusercontent.com/52440189/141445682-d0c314fc-ca71-43b1-af16-36c6e06f91d0.png)
![image](https://user-images.githubusercontent.com/52440189/141445632-1bb707e0-0940-4411-9d85-08d37846362d.png)

I don't think replacing the share link button with a view button is a good idea since it's the biggest way people share forms now and if that button goes missing it might be confusing.

> I would argue it makes more sense to make this view the default view when the owner clicks on the form

[...]

PR #1320, opened 2022-09-01
Add UI to preview and fill out own forms

susnux:
This implements #1023 by providing a button to switch to the submit view
of an owned form like suggested in the [issue comment](https://github.com/nextcloud/forms/issues/1023#issuecomment-972117908).

So basically this buttons are added:
View form button in TopBar

View form button in forms list context menu

This also adds the `TopBar` to the submit view (for owned forms) to make the UX more consistent.

### DIFF
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -40,6 +40,15 @@
 			{{ formSubtitle }}
 		</template>
 		<template v-if="!loading && !readOnly" #actions>
+			<NcActionRouter :close-after-click="true"
+				:exact="true"
+				:to="{ name: 'edit', params: { hash: form.hash } }"
+				@click="mobileCloseNavigation">
+				<template #icon>
+					<IconPencil :size="20" />
+				</template>
+				{{ t('forms', 'Edit form') }}
+			</NcActionRouter>
 			<NcActionButton :close-after-click="true" @click="onShareForm">
 				<template #icon>
 					<IconShareVariant :size="20" />
@@ -85,6 +94,7 @@ import moment from '@nextcloud/moment'
 import IconCheck from 'vue-material-design-icons/Check'
 import IconContentCopy from 'vue-material-design-icons/ContentCopy'
 import IconDelete from 'vue-material-design-icons/Delete'
+import IconPencil from 'vue-material-design-icons/Pencil'
 import IconMessageReplyText from 'vue-material-design-icons/MessageReplyText'
 import IconShareVariant from 'vue-material-design-icons/ShareVariant'
 
@@ -100,6 +110,7 @@ export default {
 		IconCheck,
 		IconContentCopy,
 		IconDelete,
+		IconPencil,
 		IconMessageReplyText,
 		IconShareVariant,
 		NcActionButton,
--- a/src/components/TopBar.vue
+++ b/src/components/TopBar.vue
@@ -26,6 +26,42 @@
 <template>
 	<div class="top-bar" role="toolbar">
 		<slot />
+		<NcButton v-if="canSubmit && $route.name !== 'submit'"
+			v-tooltip="t('forms', 'View form')"
+			:aria-label="t('forms', 'View form')"
+			type="tertiary"
+			@click="showSubmit">
+			<template #icon>
+				<IconEye :size="20" />
+			</template>
+		</NcButton>
+		<NcButton v-if="canEdit && $route.name !== 'edit'"
+			v-tooltip="t('forms', 'Edit form')"
+			:aria-label="t('forms', 'Edit form')"
+			type="tertiary"
+			@click="showEdit">
+			<template #icon>
+				<IconPencil :size="20" />
+			</template>
+		</NcButton>
+		<NcButton v-if="canSeeResults && $route.name !== 'results'"
+			v-tooltip="t('forms', 'Results')"
+			:aria-label="t('forms', 'Results')"
+			type="tertiary"
+			@click="showResults">
+			<template #icon>
+				<IconMessageReplyText :size="20" />
+			</template>
+		</NcButton>
+		<NcButton v-if="canShare && !sidebarOpened"
+			v-tooltip="t('forms', 'Share form')"
+			:aria-label="t('forms', 'Share form')"
+			type="tertiary"
+			@click="onShareForm">
+			<template #icon>
+				<IconShareVariant :size="20" />
+			</template>
+		</NcButton>
 		<NcButton v-if="showSidebarToggle"
 			v-tooltip="t('forms', 'Toggle settings')"
 			:aria-label="t('forms', 'Toggle settings')"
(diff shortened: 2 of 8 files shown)


src/views/Create.vue
```diff
@@ -41,6 +41,14 @@
 					<IconMessageReplyText :size="20" />
 				</template>
 			</NcButton>
+			<NcButton v-tooltip="t('forms', 'Fill out form')"
+				:aria-label="t('forms', 'Fill out form')"
```
Chartman123, line 45: I'd prefer the text "View form" or "View/fill out form" which would also better suite the Eye Icon.

src/views/Results.vue
```diff
@@ -39,6 +39,14 @@
 					<IconReply :size="24" />
 				</template>
 			</NcButton>
+			<NcButton v-tooltip="t('forms', 'Fill out form')"
+				:aria-label="t('forms', 'Fill out form')"
```
Chartman123, line 43: Same here

src/views/Create.vue
susnux: You are right, changed this

src/views/Results.vue
```diff
@@ -39,6 +39,14 @@
 					<IconReply :size="24" />
```
jotoeri, line 39: You can change that then to the Pencil-Edit Icon with according wording, too 😉

src/views/Submit.vue
```diff
@@ -28,6 +28,25 @@
...
 	<NcAppContent v-else>
+		<TopBar v-if="isOwnedForm">
+			<NcButton v-tooltip="t('forms', 'Back to questions')"
+				:aria-label="t('forms', 'Back to questions')"
+				type="tertiary"
+				@click="showEdit">
+				<template #icon>
+					<IconReply :size="24" />
+				</template>
+			</NcButton>
```
jotoeri, line 39: Pencil Icon and `Edit Form` or so? 😉

jotoeri (commented):
Nice, that you tackle this, @susnux! However - I'd say let us turn things completely around, then! 🚀 
Let's make the Submit-View the default, having the known Edit-Icon (Pencil) in the Actions-Menu and in the Topbar and just linking between the three views. As Nimisha said in her comment in the issue, we need to take care of a warning on expired forms then, but i'd say that also can be a follow-up fix.

susnux:
@jotoeri I pushed a new commit implementing your comments including the warning if the form has expired

Chartman123:
Nice 👍🏻 

> warning if the form has expired
I think we should have the same appearance here as for the share link (see #1321)

susnux:
> > warning if the form has expired
> 
> I think we should have the same appearance here as for the share link (see #1321)

Adjusted it 😊

src/views/Submit.vue
```diff
@@ -46,6 +51,13 @@
 		<EmptyContent v-else-if="success || !form.canSubmit" icon="icon-checkmark">
 			{{ t('forms', 'Thank you for completing the form!') }}
 		</EmptyContent>
+		<NcEmptyContent v-else-if="isExpired"
+			:title="t('forms', 'Form expired')"
+			:description="t('forms', 'This form has expired and is no longer taking answers')">
+			<template #icon>
+				<FormsIcon />
+			</template>
+		</NcEmptyContent>
```
Chartman123, line 60: Better, but do we really need the header and the description for expired forms? @jotoeri What do you think?

src/components/TopBar.vue
```diff
@@ -26,6 +26,42 @@
...
+		<NcButton v-if="canSeeResults && $route.name !== 'results'"
+			v-tooltip="t('forms', 'Results')"
+			:aria-label="t('forms', 'Results')"
+			type="tertiary"
+			@click="showResults">
+			<template #icon>
+				<IconMessageReplyText :size="24" />
+			</template>
+		</NcButton>
+		<NcButton v-if="canShare"
```
jotoeri, line 56: Here the `v-if="sidebarOpened"` got lost. This should disappear, if the sidebar already shows the sharing-options.

jotoeri (commented):
Sorry for that bunch of comments again. 🙈

src/views/Submit.vue
jotoeri: I think at least on the internal view, the Title and description were kept on purpose. You're right, maybe it would be better to have the info-message not shown anymore, but i'd say let's not overload this PR and do this separately.

src/mixins/ViewsMixin.js
```diff
@@ -58,7 +60,23 @@ export default {
...
+	computed: {
+		canEdit() {
+			return !!this.form?.permissions.includes(this.PERMISSION_TYPES.PERMISSION_EDIT)
+		},
+		canSubmit() {
+			return !!this.form?.permissions.includes(this.PERMISSION_TYPES.PERMISSION_SUBMIT)
+		},
+		canSeeResults() {
+			return !!this.form?.permissions.includes(this.PERMISSION_TYPES.PERMISSION_RESULTS)
+		},
```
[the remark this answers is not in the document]
susnux: Yes, but the optional chain operator (`?`) might return `undefined` in which case the `!!` casts the value from `undefined` to `false`.

src/views/Submit.vue
```diff
@@ -54,6 +59,13 @@
 				<IconCheck :size="64" />
 			</template>
 		</NcEmptyContent>
+		<NcEmptyContent v-else-if="isExpired"
+			:title="t('forms', 'Form expired')"
+			:description="t('forms', 'This form has expired and is no longer taking answers')">
+			<template #icon>
+				<FormsIcon />
```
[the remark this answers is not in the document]
susnux: I changed this from `IconClockAlert` (requested https://github.com/nextcloud/forms/pull/1320#issuecomment-1236412591 ), but I was not completely sure thats the right icon, I will change this.

src/components/AppNavigationForm.vue
```diff
@@ -55,6 +55,15 @@
 				</template>
 				{{ t('forms', 'Results') }}
 			</NcActionRouter>
+			<NcActionRouter :close-after-click="true"
```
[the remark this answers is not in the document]
susnux: sounds reasonable ^^

src/mixins/ViewsMixin.js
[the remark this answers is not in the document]
jotoeri: Hm, yes, you're right. Should still work without, as undefined is a falsy value on js. But that's no blocker here. 🤷‍♂️

susnux:
> Sorry for that bunch of comments again.

@jotoeri no problem, I think thats a good thing (being so attentive increases the trustability of a project I think :grinning:  )

src/mixins/ViewsMixin.js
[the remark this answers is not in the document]
susnux: Yes, but it will trigger a vue error (wrong prop type). But changed this now anyway.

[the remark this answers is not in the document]
jotoeri: Ahh, ok, yes. 👍

src/views/Submit.vue
```diff
@@ -88,10 +97,11 @@ import { loadState } from '@nextcloud/initial-state'
 import { generateOcsUrl } from '@nextcloud/router'
 import { showError } from '@nextcloud/dialogs'
 import axios from '@nextcloud/axios'
+import moment from '@nextcloud/moment'
 import NcAppContent from '@nextcloud/vue/dist/Components/NcAppContent'
+import NcButton from '@nextcloud/vue/dist/Components/NcButton'
```
jotoeri, line 102: Not necessary anymore. Same on Create.vue.

jotoeri (approved):
One small nitpick, but all in all fine by me now. ☺️ 

Very nice step Ux-wise! 🚀

src/views/Submit.vue
susnux: fixed it :)

Chartman123 (approved):

jotoeri:
Ah and can you squash the two commits? The first one still shows the old approach. 😉 
But then you can also merge if you want.

PR #1321, opened 2022-09-01
Empty Content to Vue

jotoeri:
Moving the empty-content templates to a small vue-script, due to missing icon-classes in NC25.

grafik
grafik


Chartman123:
@jotoeri: I just tried to see the second screenshot (epired form) in action and went directly to the submit-route of one of my expired forms. However, I could see the form and the submit button and then get a 403 when I wanted to submit the form. Do we have a bug there?

Chartman123:
On the public share links they both seemed to work. Should we make them also working for the internal links?

jotoeri:
> @jotoeri: I just tried to see the second screenshot (epired form) in action and went directly to the submit-route of one of my expired forms. However, I could see the form and the submit button and then get a 403 when I wanted to submit the form. Do we have a bug there?

Hmm, yes, probably we missed that. But seems like this will be solved with #1320 then. This PR is more the technical part to move the templates to a vue-component. So only on public view.

Chartman123 (approved):

PR #1334, opened 2022-09-06
Fix migration 20220414 not working on SQLite

### LINKED ISSUE #1333: Migrate from 2.5.1 to current master fails with SQLite database
**Describe the bug**
I noticed that if you do not use MySQL / MariaDB, or at least this happens for SQLite, you can not migrate from latest stable to current master because the migration `Version030000Date20220414203511` does not work:

```
Exception: Database error when running migration 030000Date20220414203511 for app forms An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL
```

This happens because the migration adds the `description` column as `not null` without providing a default value.
Which is not allowed for SQLite databases on modification (only if you create a table).

**Nextcloud**
- Nextcloud-Version: 25 beta (current git)
- Forms-Version: current master

susnux:
* Fixed #1333 

SQLite does not allow to add columns `notnull` without default value the default value is also not allowed to be an empty string.

`notnull` without a default is only allowed on table creation, not on modification later on.

### DIFF
--- a/lib/Migration/Version030000Date20220414203511.php
+++ b/lib/Migration/Version030000Date20220414203511.php
@@ -44,7 +44,7 @@ class Version030000Date20220414203511 extends SimpleMigrationStep {
 
 		if (!$table->hasColumn('description')) {
 			$table->addColumn('description', Types::TEXT, [
-				'notnull' => true,
+				'notnull' => false,
 				'length' => 4096,
 			]);
 


susnux:
Tested this to work with SQLite.

But not sure if this is the best solution as afterwards you can not rely on the column properties:
You can not rely on `NOT NULL` but must assume it is a SQLite instance meaning the column allows null values.

jotoeri:
I'd say lets make the `notnull=>false` and `default=>''` just for all databases? Or do you have anything speaking against?

Chartman123:
The question description isn't required for all questions, is it? So wouldn't it make sense to make the column `notnull=>false` without a default value in the first place?

jotoeri:
> The question description isn't required for all questions, is it? So wouldn't it make sense to make the column `notnull=>false` without a default value in the first place?

Also fine :+1:

jotoeri:
Sorry, just took over here, to have it done. :wink:

jotoeri (approved):

Chartman123 (approved):

susnux:
> Sorry, just took over here, to have it done. wink

No problem, last few days my time was very limited, so thanks for caring :)

PR #1338, opened 2022-09-14
Adjust styling for NC25 & Update `@nextcloud/vue`

susnux:
Currently forms is somewhat broken for the current NC master (or 25 beta), as the topbar is rendered inside the navigation menu.
This fixes the compatibility with the upcoming NC, but probably breaks the compatibility with anything below 25.

* Update `@nextcloud/vue` to version 7.0.0-beta.3 to support the latest changes of the nc frontend in `server` (`appcontent` as scroll container)
* Adjusted the styling of the TopBar and views accordingly

Without this it looks like this on NC25 (especially the toolbar on the top right):
toolbar is overlapping the navigation

### DIFF
--- a/src/components/TopBar.vue
+++ b/src/components/TopBar.vue
@@ -171,19 +171,15 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-$top-bar-height: 60px;
-
 .top-bar {
-	position: sticky;
+	top: 0;
 	z-index: 100;
-	top: var(--header-height);
 	display: flex;
+	position: sticky;
 	align-items: center;
 	align-self: flex-end;
 	justify-content: flex-end;
-	height: $top-bar-height;
-	margin-top: calc($top-bar-height * -1);
-	padding: 0 6px;
+	padding: calc(var(--default-grid-baseline, 4px) * 2);
 }
 
 .icon--flipped {
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
 				"@nextcloud/logger": "^2.3.0",
 				"@nextcloud/moment": "^1.2.1",
 				"@nextcloud/router": "^2.0.0",
-				"@nextcloud/vue": "^6.0.0-beta.7",
+				"@nextcloud/vue": "^7.0.0-beta.3",
 				"core-js": "^3.25.1",
 				"crypto-js": "^4.1.1",
 				"debounce": "^1.2.1",
@@ -34,8 +34,8 @@
 			},
 			"devDependencies": {
 				"@nextcloud/babel-config": "^1.0.0",
-				"@nextcloud/browserslist-config": "^2.2.0",
-				"@nextcloud/eslint-config": "^8.0.0",
+				"@nextcloud/browserslist-config": "^2.3.0",
+				"@nextcloud/eslint-config": "8.0.0",
 				"@nextcloud/stylelint-config": "^2.2.0",
 				"@nextcloud/webpack-vue-config": "^5.3.0"
 			},
@@ -2037,10 +2037,14 @@
 			}
 		},
 		"node_modules/@nextcloud/browserslist-config": {
-			"version": "2.2.0",
-			"resolved": "https://registry.npmjs.org/@nextcloud/browserslist-config/-/browserslist-config-2.2.0.tgz",
-			"integrity": "sha512-kC42RQW5rZjZZsRaEjVlIQpp6aW/yxm+zZdETnrRQnUzcPwBgF4wO4makfGT63Ckd+LkgUW+geesPiPRqxFVew==",
-			"dev": true
+			"version": "2.3.0",
+			"resolved": "https://registry.npmjs.org/@nextcloud/browserslist-config/-/browserslist-config-2.3.0.tgz",
+			"integrity": "sha512-1Tpkof2e9Q0UicHWahQnXXrubJoqyiaqsH9G52v3cjGeVeH3BCfa1FOa41eBwBSFe2/Jxj/wCH2YVLgIXpWbBg==",
+			"dev": true,
+			"engines": {
+				"node": "^16.0.0",
+				"npm": "^7.0.0 || ^8.0.0"
+			}
 		},
 		"node_modules/@nextcloud/calendar-js": {
 			"version": "3.0.0",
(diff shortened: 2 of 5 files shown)


src/components/TopBar.vue
```diff
@@ -171,18 +171,15 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-$top-bar-height: 60px;
-
 .top-bar {
 	position: sticky;
 	z-index: 100;
-	top: var(--header-height);
+	top: 0;
```
jotoeri, line 177: Can be removed, has no effect.

src/components/TopBar.vue
```diff
@@ -171,18 +171,15 @@ export default {
...
 	display: flex;
 	align-items: center;
 	align-self: flex-end;
 	justify-content: flex-end;
-	height: $top-bar-height;
-	margin-top: calc($top-bar-height * -1);
+	height: 44px;
 	padding: 0 6px;
```
jotoeri, line 183: Can you replace the padding with a `margin: 8px 12px;`, please?

src/components/TopBar.vue
```diff
@@ -171,18 +171,15 @@ export default {
...
 	z-index: 100;
-	top: var(--header-height);
+	top: 0;
 	display: flex;
 	align-items: center;
 	align-self: flex-end;
 	justify-content: flex-end;
-	height: $top-bar-height;
-	margin-top: calc($top-bar-height * -1);
+	height: 44px;
```
jotoeri, line 182: Can be removed, has no effect.

jotoeri (commented):
That was intended until now, to keep the support for NC<25. But with some information from today about NC wanting to promote forms further, seems like this is necessary. Also if i'm not really happy with it tbh.

Few small things on the css, then i think we can go with this.

Chartman123:
Any problems with this regarding the compatibility with NC24?

src/components/TopBar.vue
susnux: No it is required for the TopBar to stick on the top (see attached screen recording):


https://user-images.githubusercontent.com/1855448/190260026-668a7fa5-0995-46c9-9b6e-e96508571e10.mp4

jotoeri: Uuuhps, sorry. Didn't test it with scrolling. 👍

src/components/TopBar.vue
susnux: How about `margin: 4px 4px 12px;`? This would adjust the margin with the "app navigation" button on the left side.

jotoeri: Can you do a Screenshot? I think i tested it with 4px before writing this comment and it felt too close still.

susnux: Or maybe `4px 12px`, as only the different vertical spacing is a bit irritating. 

`4px 8px`:
top bar 4px 8px margin

`8px 12px`:
top bar 8px 12px margin

src/components/TopBar.vue
susnux: That's right

susnux:
> keep the support for NC<25

Maybe do it like the *text* app is maintaining different branches for old versions? Like branching before this and then only doing fixes for old stable and features into current stable.
But not sure how complex that would be and if this is a possible option for *forms*.

> Any problems with this regarding the compatibility with NC24?

The top bar is off center:

https://user-images.githubusercontent.com/1855448/190265035-bcbb1050-72e8-481c-84e5-9382f84c46bb.mp4

src/components/TopBar.vue
jotoeri: I think i'd still go with the 8/12. Imo it looks better and these icons belong to the content, while the other one belongs to the sidebar. So no need to have them perfectly aligned, maybe even good to have them a little bit separate?
But thats just my opinion.
@Chartman123 @jancborchardt @nimishavijay ?

Chartman123: I also like the bigger margin better... On the other hand, there is a variable `topbar-margin` with 4px

susnux: > On the other hand, there is a variable `topbar-margin` with 4px

I adjusted the code to use the `topbar-margin` for the y-axis but stick with the 12px x-axis margin, I also like that better.
Even though that variable is currently broken ( https://github.com/nextcloud/nextcloud-vue/pull/3261 )

jotoeri: My opinion was more referring to the top-margin than the right one. 🙈

jancborchardt: We should do the same as for the navigation entries on the left, which is:
`padding: calc(var(--default-grid-baseline, 4px)*2);` (so 8px, but ideally use that variable, it has a fallback set there :)
Then it is nicely aligned on the top and sides (important is that padding is the same for both there), and also aligned with the left navigation toggle.
image

Btw, the colors of the buttons in the screenshot are only to illustrate the margins, right? Cause they should not all be primary buttons. (And as said elsewhere, we need at least 1 button there to have text as well, most likely the "Results" button.)

susnux: > We should do the same as for the navigation entries on the left

Ok! Using that variable for the margin looks good.

> and also aligned with the left navigation toggle.

BTW that navigation toggle is 4px shifted compared to the entries (spaced 4px from the top, while the entries using that variable are spaced 8px from the top).

> Btw, the colors of the buttons in the screenshot are only to illustrate the margins, right?

Exactly, just for illustration purpose :)

jotoeri (approved):
Small thought, but no blocker anymore. :+1:

src/components/TopBar.vue
Chartman123: > (And as said elsewhere, we need at least 1 button there to have text as well, most likely the "Results" button.)

@jancborchardt I will add this to the other PR regarding the design review

jotoeri: > > (And as said elsewhere, we need at least 1 button there to have text as well, most likely the "Results" button.)
> 
> @jancborchardt I will add this to the other PR regarding the design review

Take care - that will be quite a mess! We have different buttons on different views! And on the submit-view, definitely the edit button is the more important one than the results. I'm not sure, if that's a good idea to have changing buttons highlighted like that or to sometimes have a highlighted one at all, but sometimes not.

@jancborchardt Any specific reason, why you would like to have some text?

Chartman123 (approved):

PR #1347, opened 2022-09-20
Use NcAppNavigationItem instead of NcListItem for app navigation

susnux:
This fixes icons not being placed correctly within the list items, see following example (focus on the left spacing of the icons).

**Currently:**
icons are placed without any spacing

**With this PR**
icons are centered within the 44px

### DIFF
--- a/src/components/AppNavigationForm.vue
+++ b/src/components/AppNavigationForm.vue
@@ -21,7 +21,7 @@
   -->
 
 <template>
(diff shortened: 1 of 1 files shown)


Chartman123:
We went just the other way round in #1168 for being able to use a second line...

jotoeri:
> So I guess the styling should be fixed in NcListItem...

That's it basically. Sorry. 🤷‍♂️

PR #1353, opened 2022-09-21
Allow embedding forms within other websites

### LINKED ISSUE #324: Ability to embed form into website
Similar to how we do with the Calendar, it would be great to be able to embed a shared form into a website, as sort of a permanent survey or contact form. Google Forms allows this and it’s often used on websites.

@rullzer said it needs some special view (like Calendar has) and also needs some CSP etc magic set.

susnux:
* Resolves #324 

* Add embedded endpoint for page controller and allow inserting submissions without CSFR as anonymous submissions for public shares.
* Added sub-menu entry for copying the embedding code to the clipboard and added documentation on how to use the embedded view.
* Switched to `vue-clipboard2` to allow copying to clipboard from sub-menu entry (allows setting a container for the copy action).


jotoeri:
Hmm, i'm not sure, if tweaking the csp is a good idea here. I would need to dig into that to properly judge it. Maybe @ChristophWurst or @nickvergessen could give a comment on that, if they find some time.

Probably also to allow the cors-requests now could allow to make it work without the csp stuff. CORS should work now with latest server 25. #1139
I personally have unfortunately no time for this until at least november. Sorry for that.

susnux:
> i'm not sure, if tweaking the csp is a good idea here

At least that is what appointments and calendar do (I draw inspiration from them).

With this changes CSFR protection is disabled for public shares, but at least you can not assign responses to an specific user, because if the CSFR token does not match (which happens if you embed the form) the responses are assigned to anonymous.


> I personally have unfortunately no time for this until at least november. Sorry for that.

No problem :)

nickvergessen:
Sorry, that's not my area of expertise

docs/Embedding.md
```diff
@@ -0,0 +1,34 @@
...
+```html
+<iframe src="EMBEDDABLE_URL" width="750" height="900"></iframe>
+```
+The size parameters are based on our default forms styling.
+
+## Custom styling
+To apply custom styles on the embedded forms the [Custom CSS App](https://apps.nextcloud.com/apps/theming_customcss) can be used.
+
+The embedded form provides the `app-forms-embedded` class, so you can apply you styles.\
+For example if want the form to fill out the frame you can use this:
```
Chartman123, line 26:
```suggestion
The embedded form provides the `app-forms-embedded` class, so you can apply your styles.\
For example if you want the form to fill out the frame you can use this:
```

Chartman123 (changes requested):

Chartman123:
I think that you should try to further improve the styling of the embedded form. I opened the embedded route in a maximized browser window and got the following result:

grafik

For the embedded view, I'd suggest to remove the background and the rounded corners and only show the app content itself. Furthermore I'd suggest to keep the width of the form content the way we have it inside Nextcloud or for the public view 🙂

susnux:
> For the embedded view, I'd suggest to remove the background and the rounded corners and only show the app content itself.

That is already documented in the `Embedding.md":
```
#content-vue.app-forms-embedded {
    width: 100%;
    height: 100%;
    border-radius: 0;
    margin: 0;
}
``` 

But I could change it to be the default.

> Furthermore I'd suggest to keep the width of the form content the way we have it inside Nextcloud or for the public view

The only difference to the public view is: No width limit (so taking 100%), that is also why it is not centered anymore.
I did this so the view is bound to the `iframe` size, as you would control the size and alignment normally by the `iframe` when embedding.

susnux:
@Chartman123 Fixed that by removing the background on the embedded view by default but still using our styling on the form.

screenshot

Chartman123:
Nice work :) I think this way it can be embedded much nicer into other websites

css/embedded.css
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/* Remove background for embedded view */
+body {
+	background-color: var(--color-main-background) !important;
+	background-image: none!important;
+}
```
Chartman123, line 25:
```suggestion
body {
	background-color: var(--color-main-background) !important;
	background-image: none !important;
}

```

Some small improvements here: missing space and missing new line at the end of the file :)

Chartman123 (approved):
Some small changes 🙂  And you could try to add some more tests for the PHP additions 👍🏻

susnux:
For the failing unit test I need some feedback @jotoeri :

I "fixed" this so it works exactly as on the main branch, so no security changes, but there is a different underlying problem:

Currently the `insertSubmission` endpoint is vulnerable to CSRF attacks.

For `OCSController` the CSRF checks are disabled by default[¹](https://github.com/nextcloud/server/blob/master/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L208).
Be we have set `@CORS` which enforces basic auth, to prevent CSRF attacks, but this is disabled when setting `@PublicPage`[²](https://github.com/nextcloud/server/blob/master/lib/private/AppFramework/Middleware/Security/CORSMiddleware.php#L86) which is set on `insertSubmission`.

So if I am right this means currently it is possible to insert submissions using session authentication which is vulnerable for CSRF. I do not see a simple solution besides using two endpoints: One for users (with `@CORS` without `@PublicPage`) and one for anonymous (with both). But maybe I am missing something?

This is unrelated to this PR, but I think this should be discussed.

docs/Embedding.md
```diff
@@ -19,6 +19,39 @@ Using the copy-embedding-code button on the *sharing sidebar* will automatically
 ```
 The size parameters are based on our default forms styling.
 
+## Auto resizing the `iframe`
+The embedded view provides a `MessageEvent` to communicate its size with its parent window.
+This is done as accessing the document within an `iframe` is no possible if not on the same domain.
```
Chartman123, line 24:
```suggestion
This is done as accessing the document within an `iframe` is not possible if not on the same domain.
```

susnux:
Added a message event to allow parent windows to resize the `iframe` to match its content (as direct access is forbidden if embedded from a different domain).

(from user feedback)

Chartman123:
@susnux where did the removal of the clipboard related stuff from?

susnux:
> @susnux where did the removal of the clipboard related stuff come from?

I remember copying the embedding code was not possible, because of some bugs, but it seems line v-clipboard got some updates after 4 year (there is a release from December), I will try if it works now.

Chartman123:
Ah ok, then it should also be OK, if we don't need this code anywhere else 👍🏻

susnux:
I split the copying related change to this PR https://github.com/nextcloud/forms/pull/1478
I will then rebase this onto that PR, I think this is much cleaner.

susnux:
> I split the copying related change to this PR #1478 I will then rebase this onto that PR, I think this is much cleaner.

Rebased, much cleaner now

docs/Embedding.md
```diff
@@ -0,0 +1,78 @@
...
+#content-vue.app-forms-embedded {
+    width: 100%;
+    height: 100%;
+    border-radius: 0;
+    margin: 0;
+}
+```
+
+Or if you want the form to fill the screen:
+```scss
```
jotoeri, line 70: Did you test that? I'm not sure, if scss insertion is working with the customcss app...

lib/Controller/PageController.php
```diff
@@ -287,4 +289,32 @@ public function insertHeaderOnIos() {
 			]);
 		}
 	}
+
+	/**
+	 * @NoAdminRequired
+	 * @PublicPage
+	 * @NoCSRFRequired
+	 * @NoSameSiteCookieRequired
```
jotoeri, line 297: Not necessary.

jotoeri (commented):
So -  now i got to this one.
- A few comments are in the code.
- Generally, i think we shouldnt have the embedding active by default on share links. So some kind of setting to allow embedding would be good. Maybe UI-wise it could also be a nice way to have a button like "Convert to embedding link" or sth. like that.
- Submit currently doesn't work, since csrf check fails, even that a token is provided. Needs further investigation, but i can also have a look when i find some time 🤔

src/components/SidebarTabs/SharingSidebarTab.vue
```diff
@@ -167,6 +173,7 @@ import NcActionButton from '@nextcloud/vue/dist/Components/NcActionButton.js'
 import NcCheckboxRadioSwitch from '@nextcloud/vue/dist/Components/NcCheckboxRadioSwitch.js'
 import IconAccountMultiple from 'vue-material-design-icons/AccountMultiple.vue'
 import IconAlertCircleOutline from 'vue-material-design-icons/AlertCircleOutline.vue'
+import IconCodeBrackets from 'vue-material-design-icons/CodeBrackets.vue'
```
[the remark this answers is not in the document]
susnux: This is the icon other apps already use for embedding code, e.g. the Calendar app:
https://github.com/nextcloud/calendar/blob/483f45d537806a9ee82db6ba1f9e94a0db58b299/src/components/AppNavigation/EditCalendarModal/PublishCalendar.vue#L93

susnux:
Thank you for your review, I fixed most of the code comments.

>     * Generally, i think we shouldnt have the embedding active by default on share links. So some kind of setting to allow embedding would be good. Maybe UI-wise it could also be a nice way to have a button like "Convert to embedding link" or sth. like that.

Do you mean admin settings or per form settings for embedding?

>     * Submit currently doesn't work, since csrf check fails, even that a token is provided. Needs further investigation, but i can also have a look when i find some time thinking

I can not confirm this, we already use this as a custom "preview" version and submitting works as expected (anonymous & logged in).

jotoeri:
> Do you mean admin settings or per form settings for embedding?

Admin Setting would be a nice thing, but that's separate then.
What i meant was more like per link. So that not just every link is by default usable for embedding, but you have to activate that. Could be just a checkbox to allow embedding per share link.
UI-wise it would probably be a bit nicer to have instead of the checkbox a convert-button, that converts the whole share-link line into an embedding-link line, probobly with the resp. icon and then copying the embedding code by default on the copy-button.

But as said - the main though was, that not just every link is by default usable by everybody to embed it somewhere.

src/components/SidebarTabs/SharingSidebarTab.vue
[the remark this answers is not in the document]
jotoeri: 😕 👍 🤷‍♂️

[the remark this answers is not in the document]
Chartman123: Perhaps something to talk about in the c.nc.c Design team?

susnux:
> I can not confirm this, we already use this feature and submitting works as expected (anonymous & logged in).

I have to revoke my statement: I now can reproduce this:
It works because cloud and embedding website are on the same site, but if you try to embed on a different site you will get CORS issues.
Also if you are currently logged in the embedding view does not work either because of same site cookie.
I currently think the CSRF issues happen because you get a csrf token from the API, but the further API requests do not include the session (because the session is stored in a same site cookie which only allows same site or safe HTTP requests (`GET`)).

susnux:
Reset to "developing", I will have to figure out how to fix submissions from embedded pages.

src/FormsSubmit.vue
```diff
@@ -46,8 +46,45 @@ export default {
...
+		if (this.isEmbedded) {
+			// Communicate window size to parent window in iframes
+			const resizeObserver = new ResizeObserver(entries => {
+				this.emitResizeMessage(entries[0].target)
+			})
+			this.$nextTick(() => resizeObserver.observe(document.querySelector('.app-forms-embedded form')))
+		}
+	},
+
+	methods: {
```
vitormattos, line 64:
```suggestion
	destroyed() {
		unsubscribe('forms:last-updated:set')
	},

	methods: {
```

Unsubscribe the event.

src/FormsSubmit.vue
```diff
@@ -46,8 +46,45 @@ export default {
 		return {
 			form: loadState('forms', 'form'),
 			isLoggedIn: loadState('forms', 'isLoggedIn'),
+			isEmbedded: loadState('forms', 'isEmbedded', false),
 			shareHash: loadState('forms', 'shareHash'),
 		}
 	},
+
+	mounted() {
+		if (this.isEmbedded) {
```
vitormattos, line 55:
```suggestion
		if (this.isEmbedded) {
			subscribe('forms:last-updated:set', (id) => {
				window.parent?.postMessage({
					type: 'form-saved',
					payload: {
						id,
					},
				}, '*')
			})
```
To be possible trigger the saved event at window parent

docs/Embedding.md
```diff
@@ -0,0 +1,78 @@
...
+		"width": 750,
+		"height": 900,
+	},
+}
+```
+
+To receive this information on your parent site:
+```js
+window.addEventListener("message", (event) => {
+	if (event.origin !== "http://your-nextcloud-server.com") {
```
vitormattos, line 40: To be more generic:

```suggestion
	if (event.origin !== window.location.protocol + '//' + window.location.host) {
```

src/FormsSubmit.vue
```diff
@@ -21,7 +21,7 @@
   -->
 
 <template>
-	<NcContent app-name="forms">
+	<NcContent app-name="forms" :class="{'app-forms-embedded': isEmbedded}">
 		<Submit :form="form"
 			:public-view="true"
 			:share-hash="shareHash"
```
vitormattos, line 27:
```suggestion
			:share-hash="shareHash"
			:sidebar-opened="!isEmbedded"
```
To prevent the follow error:

Screenshot_20230728_140444

docs/Embedding.md
susnux: I do not think this is good here, as it implies your nextcloud instance is running on the same domain as your website.
So having the "fixed" URL there will make it clearer you have to adjust it for you setup (like it will otherwise break if your nextcloud instance is running on `cloud.example.com` but your website is running on `www.example.com`).

src/FormsSubmit.vue
susnux: Nice catch!

docs/Embedding.md
vitormattos: Is only an example, can mark as resolved without changes at this point :-)

src/FormsSubmit.vue
```diff
@@ -46,8 +46,37 @@ export default {
...
 		}
 	},
+
+	mounted() {
+		if (this.isEmbedded) {
+			// Communicate window size to parent window in iframes
+			const resizeObserver = new ResizeObserver(entries => {
+				this.emitResizeMessage(entries[0].target)
+			})
+			this.$nextTick(() => resizeObserver.observe(document.querySelector('.app-forms-embedded form')))
```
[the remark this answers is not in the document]
susnux: will not work as we do not need the container width but the inner form width

susnux:
One change to note: The submission insert no longer sets the CSRF annotation, because this is already done by the CORS annotation:
- If a user session is found (logged in) on of the following must succeed:
  - the CSRF check is performed
  - check that basic auth is used

src/components/SidebarTabs/SharingSidebarTab.vue
```diff
@@ -45,20 +45,34 @@
 		<TransitionGroup v-else tag="div">
 			<div v-for="share in publicLinkShares"
 				:key="'share-' + share.shareType + '-' + share.shareWith"
-				class="share-div share-div--link">
+				:set="void(isEmbeddable = isEmbeddingAllowed(share))"
+				class="share-div share-div--link"
+				:class="{ 'share-div--embeddable': isEmbeddingAllowed(share) }">
```
Chartman123, line 50: Why don't you use isEmbeddable here?

src/components/SidebarTabs/SharingSidebarTab.vue
```diff
@@ -279,6 +301,14 @@ export default {
 			}
 		},
 
+		/**
+		 * Make a share embeddable into websites (sets the internal permission)
+		 * @param {{ permissions: string[] }} share
```
Chartman123, line 306: Please add missing JSDoc param

src/mixins/ShareLinkMixin.js
```diff
@@ -37,11 +38,25 @@ export default {
...
+				url = generateUrl(`/apps/forms/embed/${share.shareWith}`)
+			} else {
+				url = generateUrl(`/apps/forms/s/${share.shareWith}`)
+			}
+			return (new URL(url, window.location)).href
+		},
+
+		/**
+		 * Check if a share can be used for embedding
+		 * @param {{ shareType: number, permissions: string[] }} share
```
Chartman123, line 56: Please add missing JSDoc param

src/mixins/ShareLinkMixin.js
```diff
@@ -57,9 +72,24 @@ export default {
...
+			const code = `<iframe src="${this.getPublicShareLink(share)}" width="750" height="900"></iframe>`
+			try {
+				await navigator.clipboard.writeText(code)
+				showSuccess(t('forms', 'Embedding code copied'))
+			} catch (error) {
+				showError(t('forms', 'Cannot copy the code'))
+				logger.error('Copy embedding code failed', { error })
 			}
-			// Set back focus as clipboard removes focus
-			event.target.focus()
```
Chartman123, line 62: Is this no longer needed?

tests/Unit/Controller/ShareApiControllerTest.php
```diff
@@ -657,6 +672,7 @@ public function dataUpdateShare() {
 			],
 			'invalid-permission' => [
 				'share' => [
+
```
Chartman123, line 675: why a new line here?

src/components/SidebarTabs/SharingSidebarTab.vue
susnux: It does not work on the element were you "set" it, it will still use the value from the loop before

Chartman123: Thanks for the explanation :)

lib/Constants.php
```diff
@@ -147,12 +147,15 @@ class Constants {
 	public const PERMISSION_RESULTS = 'results';
 	public const PERMISSION_RESULTS_DELETE = 'results_delete';
 	public const PERMISSION_SUBMIT = 'submit';
+	/** Special internal permissions to allow embedding a form (share) into external websites */
+	public const PERMISSION_EMBEDD = 'embedd';
```
Chartman123, line 151:
```suggestion
	public const PERMISSION_EMBED = 'embed';
```

lib/Constants.php
```diff
@@ -147,12 +147,15 @@ class Constants {
...
 
 	public const PERMISSION_ALL = [
 		self::PERMISSION_EDIT,
 		self::PERMISSION_RESULTS,
 		self::PERMISSION_RESULTS_DELETE,
 		self::PERMISSION_SUBMIT,
+		self::PERMISSION_EMBEDD,
```
Chartman123, line 158:
```suggestion
		self::PERMISSION_EMBED,
```

lib/Controller/PageController.php
```diff
@@ -152,46 +153,87 @@ public function internalLinkView(string $hash): Response {
...
+	 * @NoCSRFRequired
+	 *
+	 * @param string $hash
+	 * @return Response
+	 */
+	public function embeddedFormView(string $hash): Response {
 		try {
 			$share = $this->shareMapper->findPublicShareByHash($hash);
+			// Check if the form is allwed to be embedded
+			if (!in_array(Constants::PERMISSION_EMBEDD, $share->getPermissions())) {
```
Chartman123, line 178:
```suggestion
			if (!in_array(Constants::PERMISSION_EMBED, $share->getPermissions())) {
```

src/components/SidebarTabs/SharingSidebarTab.vue
```diff
@@ -279,6 +301,14 @@ export default {
 			}
 		},
 
+		/**
+		 * Make a share embeddable into websites (sets the internal permission)
+		 * @param {{ permissions: string[] }} share The public link share to make embeddable
+		 */
+		makeEmbeddable(share) {
+			this.updateShare({ ...share, permissions: [...share.permissions, this.PERMISSION_TYPES.PERMISSION_EMBEDD] })
```
Chartman123, line 309:
```suggestion
			this.updateShare({ ...share, permissions: [...share.permissions, this.PERMISSION_TYPES.PERMISSION_EMBED] })
```

src/mixins/PermissionTypes.js
```diff
@@ -31,6 +31,8 @@ export default {
 				PERMISSION_RESULTS: 'results',
 				PERMISSION_RESULTS_DELETE: 'results_delete',
 				PERMISSION_SUBMIT: 'submit',
+				/** Internal permission to mark public link shares as embeddable */
+				PERMISSION_EMBEDD: 'embedd',
```
Chartman123, line 35:
```suggestion
				PERMISSION_EMBED: 'embed',
```

src/mixins/ShareLinkMixin.js
```diff
@@ -37,11 +38,25 @@ export default {
...
+			}
+			return (new URL(url, window.location)).href
+		},
+
+		/**
+		 * Check if a share can be used for embedding
+		 * @param {{ shareType: number, permissions: string[] }} share The share to check
+		 */
+		isEmbeddingAllowed(share) {
+			return share.shareType === this.SHARE_TYPES.SHARE_TYPE_LINK && share.permissions?.includes(this.PERMISSION_TYPES.PERMISSION_EMBEDD)
```
Chartman123, line 59:
```suggestion
			return share.shareType === this.SHARE_TYPES.SHARE_TYPE_LINK && share.permissions?.includes(this.PERMISSION_TYPES.PERMISSION_EMBED)
```

tests/Unit/Controller/ShareApiControllerTest.php
```diff
@@ -609,6 +609,21 @@ public function dataUpdateShare() {
...
+				'share' => [
+					'id' => 1,
+					'formId' => 5,
+					'shareType' => IShare::TYPE_LINK,
+					'shareWith' => 'somehash',
+					'permissions' => [Constants::PERMISSION_SUBMIT]
+				],
+				'formOwner' => 'currentUser',
+				'keyValuePairs' => [
+					'permissions' => [Constants::PERMISSION_SUBMIT, Constants::PERMISSION_EMBEDD]
```
Chartman123, line 622:
```suggestion
					'permissions' => [Constants::PERMISSION_SUBMIT, Constants::PERMISSION_EMBED]
```

Chartman123 (commented):
I fixed a few typos regarding the new `embed` permission. :)

Chartman123 (approved):

Chartman123:
@susnux After rebasing it here via Github, we lost the DCO check ;) Could you please do another rebase and push it again? After that I think we're good to go :)

whoamiafterall:
Nice - i'm looking forward to this. When will be the next release?

susnux:
Thank you :)

> When will be the next release?

Somewhen before Nextcloud 29 release, we do not have a strict schedule for the next release.

PR #1355, opened 2022-09-22
Add Nextcloud 25 support

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -46,7 +46,7 @@
 	<screenshot>https://raw.githubusercontent.com/nextcloud/forms/master/screenshots/forms3.png</screenshot>
 
 	<dependencies>
-		<nextcloud min-version="25" max-version="25" />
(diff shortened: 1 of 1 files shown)


jotoeri:
As we have the force-enable on the tests, we don't need this.
https://github.com/nextcloud/forms/blob/ddadc349c8bf3452af9866db6ac68b19758a7d99/.github/workflows/phpunit.yml#L62

Ref. Discussion in #1170

nickvergessen:
Yeah, sorry about that. As I'm running a script over ~70 apps, it's kinda hard to remember this.
I can remove the forms app or you simply have to close it every time 😎

jotoeri:
Yeah, no problem. I thought that already, that you don't do this by hand.^^
If you can remove forms, would probably be nice, i see some risk to forget it myself and merge then. 🙈

PR #1356, opened 2022-09-22
Use `--gradient-primary-background` for summary

### LINKED ISSUE #1350: Fix Results Bar
Look changed somehow with nc25. Was discussed on the chat somewhen, should ideally look like the old version.

grafik

Chartman123:
Fixes #1350 by using `--gradient-primary-background` variable for the results for now.

grafik


jotoeri (approved):
Hmm, nearly no gradient at all now, no?
But better than before. 👍

Chartman123:
Yes, the old styling would indeed be better, but there's currently no variable that has this value inside the theming app. Should I add a TODO comment to switch it to another variable when it becomes available?

PR #1362, opened 2022-09-27
Fix question movement

jotoeri:
@Chartman123 How about this one to reduce the text jumping? :blush:
(Also i found that the label uses Material Icons meanwhile. So using these instead of the pseudoInput.)

image


Chartman123:
@jotoeri I like :)

jotoeri:
Should be sufficient for now.^^

Chartman123 (approved):

PR #1364, opened 2022-09-28
Adjust look of long text questions

Chartman123:
Before:
grafik

After:
grafik

### DIFF
--- a/src/components/Questions/QuestionLong.vue
+++ b/src/components/Questions/QuestionLong.vue
@@ -87,7 +87,7 @@ export default {
 		autoSizeText() {
 			const textarea = this.$refs.textarea
 			textarea.style.cssText = 'height:auto; padding:0'
-			textarea.style.cssText = `height: ${textarea.scrollHeight + 20}px`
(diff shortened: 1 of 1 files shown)


src/components/Questions/QuestionLong.vue
```diff
@@ -99,18 +99,8 @@ export default {
 <style lang="scss" scoped>
 .question__text {
 	// make sure height calculations are correct
```
jotoeri, line 101: Comment to remove. Was meant to the box-sizing css.

jotoeri (approved):
Na gut. :see_no_evil: :wink:

src/components/Questions/QuestionLong.vue
Chartman123: I've kept it because of the needed adjustments in autoSizeText(), but I can also remove it :)

jotoeri: Uhm, it was not meant like 'make sure, if you change anything', but was meant like 'with the content-box instead of the default border-box we make sure that the calculations are correct'. So with removing the content-box, it doesn't have any connection anymore at this position.

PR #1365, opened 2022-09-29
Fix public View

jotoeri:
Two more small fixes:

- Header margin on submit-view got lost on alignment adjustments
- Compensate the missing top-spacing, when the topbar is disabled on public view.

|Before|After|
|---|---|
|Screenshot from 2022-09-29 10-07-37|Screenshot from 2022-09-29 10-26-54|


src/views/Submit.vue
```diff
@@ -297,6 +297,11 @@ export default {
 	align-items: center;
 	flex-direction: column;
 
+	&--public {
+		// Compensate top-padding for missing topbar
+		padding-top: 60px;
```
jotoeri, line 302: 60px are the exact same height of the topBar. But maybe 50px would suffice and look a bit better here? :thinking:

Chartman123: I think I'd go for the 50 px...

Chartman123 (approved):

PR #1381, opened 2022-10-15
More usable and accessible navigation between View/Edit/Results

jancborchardt:
The icon-only navigation on the top right is not super visible and not understandable, so I worked on some improvements sort-of inspired by our mockups for the new tabs (which we can also eventually use once it’s updated in the component) https://github.com/nextcloud/nextcloud-vue/issues/603

## Before
- Lots of icons
- Some icons show only in some context

https://user-images.githubusercontent.com/925062/195996592-ec09cdde-ca2d-4c1f-95fc-3f1d7be9d979.mp4

## After
- Everything has text
- View/Edit/Results is grouped
- All buttons show always

https://user-images.githubusercontent.com/925062/195996589-2614b829-cd5c-4efd-aedc-ef93ff4d97e7.mp4

## To do
Issues which still need fixing and some of which I need help with @jotoeri @Chartman123 :)
- [x] Sidebar toggle only works in "Edit" mode ("Share" button always switches to "Edit" mode too). This should be possible from everywhere.
- [x] Mobile responsiveness, buttons could be icon-only on mobile
- [x] Test if it works in all views like sharing
- [ ] Loading overlay makes the buttons flash (not essential)

### DIFF
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -49,6 +49,7 @@
 					:key="form.id"
 					:form="form"
 					:read-only="true"
+					@open-sharing="openSharing"
 					@mobile-close-navigation="mobileCloseNavigation" />
 			</template>
 		</NcAppNavigation>
@@ -92,9 +93,9 @@
 			<router-view :form.sync="selectedForm"
 				:sidebar-opened.sync="sidebarOpened"
 				@open-sharing="openSharing" />
-			<router-view v-if="!selectedForm.partial"
+			<router-view v-if="!selectedForm.partial && canEdit"
 				:form="selectedForm"
-				:opened.sync="sidebarOpened"
+				:sidebar-opened.sync="sidebarOpened"
 				:active.sync="sidebarActive"
 				name="sidebar" />
 		</template>
@@ -158,6 +159,9 @@ export default {
 	},
 
 	computed: {
+		canEdit() {
+			return this.selectedForm.permissions.includes(this.PERMISSION_TYPES.PERMISSION_EDIT)
+		},
 		hasForms() {
 			return !this.noOwnedForms || !this.noSharedForms
 		},
@@ -236,7 +240,7 @@ export default {
 		 * @param {string} hash the hash of the form to load
 		 */
 		openSharing(hash) {
-			if (hash !== this.routeHash || this.$route.name !== 'edit') {
+			if (hash !== this.routeHash) {
 				this.$router.push({ name: 'edit', params: { hash } })
 			}
 			this.sidebarActive = 'forms-sharing'
--- a/src/components/TopBar.vue
+++ b/src/components/TopBar.vue
@@ -25,41 +25,50 @@
   -->
 <template>
 	<div class="top-bar" role="toolbar">
-		<slot />
-		<NcButton v-if="canSubmit && $route.name !== 'submit'"
-			v-tooltip="t('forms', 'View form')"
-			:aria-label="t('forms', 'View form')"
-			type="tertiary"
-			@click="showSubmit">
-			<template #icon>
-				<IconEye :size="20" />
-			</template>
-		</NcButton>
-		<NcButton v-if="canEdit && $route.name !== 'edit'"
-			v-tooltip="t('forms', 'Edit form')"
-			:aria-label="t('forms', 'Edit form')"
-			type="tertiary"
-			@click="showEdit">
-			<template #icon>
-				<IconPencil :size="20" />
-			</template>
-		</NcButton>
-		<NcButton v-if="canSeeResults && $route.name !== 'results'"
-			v-tooltip="t('forms', 'Results')"
-			:aria-label="t('forms', 'Results')"
-			type="tertiary"
-			@click="showResults">
-			<template #icon>
-				<IconPoll :size="20" />
-			</template>
-		</NcButton>
+		<div v-if="!canOnlySubmit" class="top-bar__view-select">
+			<NcButton v-if="canSubmit"
+				:aria-label="isMobile ? t('forms', 'View form') : null"
+				:type="$route.name === 'submit' ? 'secondary' : 'tertiary'"
+				@click="showSubmit">
+				<template #icon>
+					<IconEye :size="20" />
+				</template>
+				<template v-if="!isMobile">
+					{{ t('forms', 'View') }}
+				</template>
+			</NcButton>
+			<NcButton v-if="canEdit"
+				:aria-label="isMobile ? t('forms', 'Edit form') : null"
+				:type="$route.name === 'edit' ? 'secondary' : 'tertiary'"
+				@click="showEdit">
+				<template #icon>
+					<IconPencil :size="20" />
+				</template>
+				<template v-if="!isMobile">
+					{{ t('forms', 'Edit') }}
+				</template>
+			</NcButton>
+			<NcButton v-if="canSeeResults"
+				:aria-label="isMobile ? t('forms', 'Show results') : null"
+				:type="$route.name === 'results' ? 'secondary' : 'tertiary'"
+				@click="showResults">
+				<template #icon>
+					<IconPoll :size="20" />
+				</template>
+				<template v-if="!isMobile">
+					{{ t('forms', 'Results') }}
+				</template>
+			</NcButton>
+		</div>
 		<NcButton v-if="canShare && !sidebarOpened"
-			v-tooltip="t('forms', 'Share form')"
-			:aria-label="t('forms', 'Share form')"
+			:aria-label="isMobile ? t('forms', 'Share form') : null"
 			type="tertiary"
 			@click="onShareForm">
 			<template #icon>
 				<IconShareVariant :size="20" />
+			</template>
+			<template v-if="!isMobile">
+				{{ t('forms', 'Share') }}
 			</template>
 		</NcButton>
 		<NcButton v-if="showSidebarToggle"
@@ -77,6 +86,7 @@
 
 <script>
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
+import isMobile from '@nextcloud/vue/dist/Mixins/isMobile.js'
 import IconEye from 'vue-material-design-icons/Eye.vue'
 import IconMenuOpen from 'vue-material-design-icons/MenuOpen.vue'
 import IconPencil from 'vue-material-design-icons/Pencil.vue'
@@ -96,7 +106,7 @@ export default {
 		NcButton,
 	},
 
-	mixins: [PermissionTypes],
+	mixins: [isMobile, PermissionTypes],
 
 	props: {
 		sidebarOpened: {
@@ -123,8 +133,11 @@ export default {
 			// This probably can get a permission of itself
 			return this.canEdit || this.canSeeResults
 		},
+		canOnlySubmit() {
+			return this.permissions.length === 1 && this.permissions.includes(this.PERMISSION_TYPES.PERMISSION_SUBMIT)
+		},
 		showSidebarToggle() {
-			return this.sidebarOpened !== null
+			return this.canEdit && this.sidebarOpened !== null
 		},
 	},
 
@@ -137,30 +150,36 @@ export default {
 		 * Router methods
 		 */
 		showEdit() {
-			this.$router.push({
-				name: 'edit',
-				params: {
-					hash: this.$route.params.hash,
-				},
-			})
+			if (this.$route.name !== 'edit') {
+				this.$router.push({
+					name: 'edit',
+					params: {
+						hash: this.$route.params.hash,
+					},
+				})
+			}
 		},
 
 		showResults() {
-			this.$router.push({
-				name: 'results',
-				params: {
-					hash: this.$route.params.hash,
-				},
-			})
+			if (this.$route.name !== 'results') {
+				this.$router.push({
+					name: 'results',
+					params: {
+						hash: this.$route.params.hash,
+					},
+				})
+			}
 		},
 
 		showSubmit() {
-			this.$router.push({
-				name: 'submit',
-				params: {
-					hash: this.$route.params.hash,
-				},
-			})
+			if (this.$route.name !== 'submit') {
+				this.$router.push({
+					name: 'submit',
+					params: {
+						hash: this.$route.params.hash,
+					},
+				})
+			}
 		},
 
 		onShareForm() {
@@ -180,6 +199,17 @@ export default {
 	align-self: flex-end;
 	justify-content: flex-end;
 	padding: calc(var(--default-grid-baseline, 4px) * 2);
+
+	&__view-select {
+		display: flex;
+		height: 44px;
+		align-items: center;
+		align-self: flex-end;
+		justify-content: flex-end;
+		background: var(--color-main-background);
+		border: 2px solid var(--color-border);
+		border-radius: var(--border-radius-pill);
+	}
 }
 
 .icon--flipped {
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -32,7 +32,10 @@
 	</NcAppContent>
 
 	<NcAppContent v-else>
-		<TopBar :permissions="form?.permissions" @share-form="onShareForm" />
+		<TopBar :permissions="form?.permissions"
+			:sidebar-opened="sidebarOpened"
+			@update:sidebarOpened="onSidebarChange"
+			@share-form="onShareForm" />
 		<header v-if="!noSubmissions">
 			<h2>{{ formTitle }}</h2>
 			<p>{{ t('forms', '{amount} responses', { amount: form.submissions.length }) }}</p>
--- a/src/mixins/ViewsMixin.js
+++ b/src/mixins/ViewsMixin.js
@@ -46,6 +46,10 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		sidebarOpened: {
+			type: Boolean,
+			required: true,
+		},
 	},
 
 	data() {
@@ -61,6 +65,10 @@ export default {
 	methods: {
 		onShareForm() {
 			this.$emit('open-sharing', this.form.hash)
+		},
+
+		onSidebarChange(newState) {
+			this.$emit('update:sidebarOpened', newState)
 		},
 
 		/**
--- a/src/router.js
+++ b/src/router.js
@@ -63,15 +63,21 @@ export default new Router({
 		},
 		{
 			path: '/:hash/results',
-			component: Results,
+			components: {
+				default: Results,
+				sidebar: Sidebar,
+			},
 			name: 'results',
-			props: true,
+			props: { default: true },
 		},
 		{
 			path: '/:hash/submit',
-			component: Submit,
+			components: {
+				default: Submit,
+				sidebar: Sidebar,
+			},
 			name: 'submit',
-			props: true,
+			props: { default: true },
 		},
 	],
 })
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -161,13 +161,6 @@ export default {
 
 	mixins: [ViewsMixin],
 
-	props: {
-		sidebarOpened: {
-			type: Boolean,
-			required: true,
-		},
-	},
-
 	data() {
 		return {
 			maxStringLengths: loadState('forms', 'maxStringLengths'),
@@ -258,9 +251,6 @@ export default {
 		onDescChange() {
 			this.autoSizeDescription()
 			this.saveDescription()
-		},
-		onSidebarChange(newState) {
-			this.$emit('update:sidebarOpened', newState)
 		},
 
 		/**
--- a/src/views/Sidebar.vue
+++ b/src/views/Sidebar.vue
@@ -21,7 +21,7 @@
  -->
 
 <template>
-	<NcAppSidebar v-show="opened"
+	<NcAppSidebar v-show="sidebarOpened"
 		:active="active"
 		:title="t('forms', 'Form settings')"
 		@close="onClose"
@@ -71,16 +71,13 @@ export default {
 		SharingSidebarTab,
 		SettingsSidebarTab,
 	},
+
 	mixins: [ViewsMixin],
 
 	props: {
 		active: {
 			type: String,
 			default: 'forms-sharing',
-		},
-		opened: {
-			type: Boolean,
-			required: true,
 		},
 	},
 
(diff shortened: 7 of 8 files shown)


src/components/TopBar.vue
```diff
@@ -26,43 +26,41 @@
...
-			</template>
-		</NcButton>
+		<div class="top-bar-buttons">
+			<NcButton v-if="canSubmit"
+				:type="$route.name === 'submit' ? 'secondary' : 'tertiary'"
+				@click="showSubmit">
+				<template #icon>
+					<IconEye :size="20" />
+				</template>
+				<template>{{ t('forms', 'View') }}</template>
```
Chartman123, line 36:
```suggestion
				{{ t('forms', 'View') }}
```

src/components/TopBar.vue
```diff
@@ -26,43 +26,41 @@
...
+			</NcButton>
+			<NcButton v-if="canEdit"
+				:type="$route.name === 'edit' ? 'secondary' : 'tertiary'"
+				@click="showEdit">
+				<template #icon>
+					<IconPencil :size="20" />
+				</template>
+				<template>{{ t('forms', 'Edit') }}</template>
```
Chartman123, line 44:
```suggestion
				{{ t('forms', 'Edit') }}
```

src/components/TopBar.vue
```diff
@@ -26,43 +26,41 @@
...
+			</NcButton>
+			<NcButton v-if="canSeeResults"
+				:type="$route.name === 'results' ? 'secondary' : 'tertiary'"
+				@click="showResults">
+				<template #icon>
+					<IconPoll :size="20" />
+				</template>
+				<template>{{ t('forms', 'Results') }}</template>
```
Chartman123, line 52:
```suggestion
				{{ t('forms', 'Results') }}
```

src/components/TopBar.vue
```diff
@@ -26,43 +26,41 @@
...
+		</div>
 		<NcButton v-if="canShare && !sidebarOpened"
-			v-tooltip="t('forms', 'Share form')"
-			:aria-label="t('forms', 'Share form')"
 			type="tertiary"
 			@click="onShareForm">
 			<template #icon>
 				<IconShareVariant :size="20" />
 			</template>
+			<template>{{ t('forms', 'Share form') }}</template>
```
Chartman123, line 61:
```suggestion
			{{ t('forms', 'Share form') }}
```

src/components/TopBar.vue
```diff
@@ -26,43 +26,41 @@
...
 		</NcButton>
-		<NcButton v-if="showSidebarToggle"
+		<NcButton
 			v-tooltip="t('forms', 'Toggle settings')"
```
Chartman123, line 64:
```suggestion
		<NcButton v-tooltip="t('forms', 'Toggle settings')"
```

Chartman123 (changes requested):
Some smaller changes to fix the Lint warnings and error. Also I'm not quite sure if we really need the border around the first three buttons

jancborchardt:
> Some smaller changes to fix the Lint warnings and error. Also I'm not quite sure if we really need the border around the first three buttons

Thank you for the help, I will look into it! :)
The border is intended to group them, as those 3 belong together, like a "tab bar" of the ones which change the content. Both "Share form" and the sidebar icon are different things, so it’s good to visually separate.

jotoeri:
> The border is intended to group them, as those 3 belong together, like a "tab bar" of the ones which change the content. Both "Share form" and the sidebar icon are different things, so it’s good to visually separate.

I agree with that. But i also think we should go with less border radius. Similar to the design you linked in vue. On the results we have it completely without inner radius, but thats indeed probably strange with the inner one of the three.

Chartman123:
> On the results we have it completely without inner radius, but thats indeed probably strange with the inner one of the three.

@jancborchardt I know that this would be some bigger change also to vue component, but would it make sense to switch to the NcCheckboxRadioSwitch think with the alternate button layout and horizontal grouping? At the moment there is no icon slot and I also don't know if that's planned or possible... But would make it a lot more easier for single applications to have such a grouped button element... I already did some experiments with this for the results view switching, but currently it's still very ugly. 😄

jotoeri:
@Chartman123 Hmm, be careful not to project anything onto the same component, just because it has a similar base... Components are in fact a reusable version of specialisations of the underlying HTML with specialised behaviour and design.
That does neither forbid us, to use the HTML directly, nor does a single component need to cover the (kind of) generalisation to support all possible use-cases.

Chartman123:
@jotoeri Yes that's true. But if there are any plans at Nextcloud to have these "button groups" standardized for other apps as well, it might be a good thing to place it inside nextcloud-vue and have a re-usable design. I didn't neccessarily mean to "abuse" the NcCheckboxRadioSwitch, but this was my first starting point as the alternate horizontal layout kind of looks like the thing we'd need here.

jancborchardt:
> I agree with that. But i also think we should go with less border radius. Similar to the design you linked in vue. On the results we have it completely without inner radius, but thats indeed probably strange with the inner one of the three.

Right, so for now I didn’t go with the design like in Vue because in our case it’s a horizontal button, not a vertical one. I would propose we first go with this design as it’s the least invasive and we use standard buttons there anyway – and then when the tab component is updated, we can use that.

I would also say that NcCheckboxRadioSwitch is not correct semantically here.

What do you think?

jancborchardt:
Cool, then I need help with at least that 1 point in the todo :)
> Sidebar toggle only works in "Edit" mode ("Share" button always switches to "Edit" mode too). This should be possible from everywhere.

The sidebar and share button should show in every view as otherwise the tab bar is moving positions.

jotoeri:
> > I agree with that. But i also think we should go with less border radius. Similar to the design you linked in vue. On the results we have it completely without inner radius, but thats indeed probably strange with the inner one of the three.
> 
> Right, so for now I didn’t go with the design like in Vue because in our case it’s a horizontal button, not a vertical one. I would propose we first go with this design as it’s the least invasive and we use standard buttons there anyway – and then when the tab component is updated, we can use that.
> 
> I would also say that NcCheckboxRadioSwitch is not correct semantically here.
> 
> What do you think?

Tbh. I'm not fine with that. Why doing half things? It should be just a bit of more CSS (just border-radius?) to get a much better design here.

Chartman123:
@jancborchardt while having a look at the sidebar thing I found a little regression: as you show all the buttons all the time you can also click on a button for the route you're currently at (e.g. click edit button twice). There is an error message then in the browser console...

jancborchardt:
> Tbh. I'm not fine with that. Why doing half things? It should be just a bit of more CSS (just border-radius?) to get a much better design here.

@jotoeri I tried it out, and actually the pill radius looks much better and is more fitting to our updated circular design (and the buttons nearby):

## border-radius-pill (like buttons)
https://user-images.githubusercontent.com/925062/195996589-2614b829-cd5c-4efd-aedc-ef93ff4d97e7.mp4

## border-radius-large (like tab bar mockup)
Sorry no video here as my screen recording app seems broken.
image

@nimishavijay @marcoambrosini what do you think? Ref the tab bar mockups at https://github.com/nextcloud/nextcloud-vue/issues/603

jancborchardt:
> @jancborchardt while having a look at the sidebar thing I found a little regression: as you show all the buttons all the time you can also click on a button for the route you're currently at (e.g. click edit button twice). There is an error message then in the browser console...

@Chartman123 Fixed, thanks! :) Still need help with the following if you can:

> Sidebar toggle only works in "Edit" mode ("Share" button always switches to "Edit" mode too). This should be possible from everywhere.

jotoeri:
> @jotoeri I tried it out, and actually the pill radius looks much better and is more fitting to our updated circular design (and the buttons nearby). I did commit it so feel free to try it out. :)


Hm, ok. Then i'm with you as a long term design here. 👍

src/components/TopBar.vue
```diff
@@ -26,44 +26,44 @@
...
-		<NcButton v-if="canSeeResults && $route.name !== 'results'"
-			v-tooltip="t('forms', 'Results')"
-			:aria-label="t('forms', 'Results')"
-			type="tertiary"
-			@click="showResults">
-			<template #icon>
-				<IconPoll :size="20" />
-			</template>
-		</NcButton>
+		<div class="top-bar-buttons">
```
marcoambrosini, line 29:
```suggestion
		<div class="top-bar__buttons">
```

src/components/TopBar.vue
```diff
@@ -180,6 +180,26 @@ export default {
 	align-self: flex-end;
 	justify-content: flex-end;
 	padding: calc(var(--default-grid-baseline, 4px) * 2);
+
+	&-buttons {
```
marcoambrosini, line 184:
```suggestion
	&__buttons {
```

marcoambrosini (approved):

marcoambrosini:
tbh though I prefer the borderless version, by far

Screenshot 2022-10-26 at 11 14 41

jancborchardt:
@marcoambrosini borderless just looks not connected at all, it’s not obvious what kind of element it is. What do you think regarding pill style vs rounded borders?

Apart from that, others also do connected, pill-style tab bars (iOS does connected but rounded borders sometimes but I can’t find a proper pattern link):

MS Office on iPad:
MS office iPad toolbar

Material design docs website:
Material tab bar icons

src/components/TopBar.vue
jancborchardt: Fixed :)

src/components/TopBar.vue
jancborchardt: Fixed :)

jancborchardt:
Addressed your issues @Chartman123 @marcoambrosini, just as said need help with the fact that the sidebar is only available in "Edit" mode. Possibly could be a follow-up pull request, you tell me. :)

Chartman123:
I'd prefer to have the working sidebar button already in this PR :)

Chartman123:
@jancborchardt I think I fixed the sidebar now. 🙂

jancborchardt:
Nice @Chartman123, tested and works! :) Can we then have another review @susnux @hamza221 @jotoeri?

susnux (commented):
First of all I like this very much! Nice work :+1: 
I tested it and it works for me :smile: 

To me using three different elements in the top bar looks a bit inconsistent or messy (meaning it does not look clean like the other UI).
So there is the 3-button array, then there is the "share form" button and next to it is an other icon-only sidebar-toggle.
Maybe one could integrate that share button with the 3-button array or maybe even better with the "form settings"-sidebar-toggle.

But that is only my personal preference, you might ignore that :wink: 


A more important thing is the second ToDo, on mobile it is unusable, this is how it looks for me:
Menubar overlows on the left side, it is unusable

Chartman123:
> on mobile it is unusable,

Should be fixed now, I added the isMobile mixin 🙂

src/components/TopBar.vue
```diff
@@ -25,42 +25,50 @@
...
+			<NcButton v-if="canSubmit"
+				:disabled="$route.name === 'submit'"
+				:type="$route.name === 'submit' ? 'secondary' : 'tertiary'"
+				@click="showSubmit">
+				<template #icon>
+					<IconEye :size="20" />
+				</template>
+				<template v-if="!isMobile">
+					{{ t('forms', 'View') }}
+				</template>
```
Chartman123, line 38: @jancborchardt do we need the aria-labels again now that the text is shown only if !isMobile?

src/components/TopBar.vue
```diff
@@ -25,42 +25,50 @@
...
+			<NcButton v-if="canEdit"
+				:disabled="$route.name === 'edit'"
+				:type="$route.name === 'edit' ? 'secondary' : 'tertiary'"
+				@click="showEdit">
+				<template #icon>
+					<IconPencil :size="20" />
+				</template>
+				<template v-if="!isMobile">
+					{{ t('forms', 'Edit') }}
+				</template>
```
Chartman123, line 49: Same here

src/components/TopBar.vue
```diff
@@ -25,42 +25,50 @@
...
+			<NcButton v-if="canSeeResults"
+				:disabled="$route.name === 'results'"
+				:type="$route.name === 'results' ? 'secondary' : 'tertiary'"
+				@click="showResults">
+				<template #icon>
+					<IconPoll :size="20" />
+				</template>
+				<template v-if="!isMobile">
+					{{ t('forms', 'Results') }}
+				</template>
```
Chartman123, line 60: Same here

src/components/TopBar.vue
```diff
@@ -25,42 +25,50 @@
...
-			v-tooltip="t('forms', 'Share form')"
-			:aria-label="t('forms', 'Share form')"
 			type="tertiary"
 			@click="onShareForm">
 			<template #icon>
 				<IconShareVariant :size="20" />
 			</template>
+			<template v-if="!isMobile">
+				{{ t('forms', 'Share form') }}
+			</template>
```
Chartman123, line 71: Same here

Chartman123 (commented):
Question regarding aria-labels

src/components/TopBar.vue
jancborchardt: Ah yes you are absolutely right! On isMobile we should use aria-label then.

src/components/TopBar.vue
jancborchardt: Yup, agree. Also for simplicity, the visual text here can be called just "Share".

For aria-label, "Share form" is better.

src/Forms.vue
```diff
@@ -49,6 +49,7 @@
 					:key="form.id"
 					:form="form"
 					:read-only="true"
+					@open-sharing="openSharing"
```
jotoeri, line 52: Seems to be a rest of some testing. That functionality will not be available currently if i see that right. Not best practice for a PR, but i'm ok with leaving it here, since we will most probably use it somewhen.

src/Forms.vue
```diff
@@ -92,7 +93,7 @@
 			<router-view :form.sync="selectedForm"
 				:sidebar-opened.sync="sidebarOpened"
 				@open-sharing="openSharing" />
-			<router-view v-if="!selectedForm.partial"
+			<router-view v-if="!selectedForm.partial && selectedForm.permissions.includes(PERMISSION_TYPES.PERMISSION_EDIT)"
```
jotoeri, line 96: computed `canEdit` for nice code?

src/Forms.vue
```diff
@@ -236,7 +237,8 @@ export default {
 		 * @param {string} hash the hash of the form to load
 		 */
 		openSharing(hash) {
-			if (hash !== this.routeHash || this.$route.name !== 'edit') {
+			// if (hash !== this.routeHash || this.$route.name !== 'edit') {
```
jotoeri, line 240: To remove

src/components/TopBar.vue
```diff
@@ -25,42 +25,50 @@
...
-		<NcButton v-if="canSeeResults && $route.name !== 'results'"
-			v-tooltip="t('forms', 'Results')"
-			:aria-label="t('forms', 'Results')"
-			type="tertiary"
-			@click="showResults">
-			<template #icon>
-				<IconPoll :size="20" />
-			</template>
-		</NcButton>
+		<div class="top-bar__buttons">
```
jotoeri, line 28: Just naming wise - There are more buttons in the toolbar. Probably more a `top-bar__view-select` or sth like that?

src/components/TopBar.vue
```diff
@@ -180,6 +189,24 @@ export default {
...
+		border: 2px solid var(--color-border);
+		border-radius: var(--border-radius-pill);
+
+		.button-vue {
+
+			&:disabled {
+				opacity: initial;
+				filter: initial;
+			}
+		}
```
jotoeri, line 208: Reasonably-wise it is strange to disable the buttons but overriding the look to not-disabled.
Maybe it would be more reasonable to conditionally stop the problematic router-push (or whatever it is) instead of disabling the buttons?

src/components/TopBar.vue
Chartman123: @jancborchardt Do you want to add this to the PR? I'm currently quite busy...

src/components/TopBar.vue
Chartman123: "Share" would be fine for me

src/views/Results.vue
```diff
@@ -183,6 +186,13 @@ export default {
 
 	mixins: [ViewsMixin],
 
+	props: {
+		sidebarOpened: {
+			type: Boolean,
+			required: true,
+		},
+	},
+
```
jotoeri, line 195: This is now used on all three views. So it should better go to the ViewsMixin?

src/views/Results.vue
```diff
@@ -324,6 +334,10 @@ export default {
 
 			return submissions
 		},
+
+		onSidebarChange(newState) {
+			this.$emit('update:sidebarOpened', newState)
+		},
```
jotoeri, line 340: Same here - ViewsMixin?

jotoeri (commented):
Didn't test, just had a short look into the code. :relaxed:

src/Forms.vue
Chartman123: Ah yes, I had that on my list, but forgot to remove it yesterday when I squashed the fixup commits...

src/components/TopBar.vue
jancborchardt: Yep, I will try with these changes and also the ones from @jotoeri that I can manage – I might have to ask you (or @hamza221 ;) how to do the stuff that go beyond my frontend skills. :D

jancborchardt: So I would use `:aria-label="isMobile ? t('forms', 'View form') : ''"` but then we end up with an empty `aria-label=""` on desktop, which is dirty. Is there a way to get around this?
(Tbh instead of the isMobile mixin I would normally do this via CSS media queries?)

Chartman123: @jancborchardt I'm also not a vuejs pro 😆 I just tried the following and it seems to do the job: `:aria-label="isMobile ? t('forms', 'View form'):null"`

src/Forms.vue
jancborchardt: Removed :)

src/components/TopBar.vue
jancborchardt: You’re right, fixed!

src/components/TopBar.vue
jancborchardt: You’re right. I removed disabling the controls and the styling as this is also not done in other solutions.

There is a Vue router issue (`NavigationDuplicated Navigating to current location`) which is out of my expertise though. I found this for further reference, maybe @Chartman123 @jotoeri you can have a look?
- https://github.com/vuejs/vue-router/issues/2881#issuecomment-520554378
- https://stackoverflow.com/questions/57837758/navigationduplicated-navigating-to-current-location-search-is-not-allowed

src/views/Results.vue
jancborchardt: Fixed - I needed to remove the `required` attribute as otherwise Sidebar.vue had an error of a missing required prop – let me know if it should be done otherwise. (@hamza221 said we could also pass `:sidebarOpened` to the view in Forms.vue, but with what value then – it’s the sidebar itself. ;)

src/components/TopBar.vue
jancborchardt: Niiice, thanks a lot @Chartman123 that totally worked indeed. :) Fixed!

jancborchardt:
@chartman123 @jotoeri fixed all the issues except the `Duplicated Navigation` thing at https://github.com/nextcloud/forms/pull/1381#discussion_r1037452929 – I either need some pointers, or you are welcome to fix it, or we accept the error (not sure if that should be done). Thanks! :)

src/views/Results.vue
jotoeri: Uhm, the sidebar anyways takes a required value `opened`. If you replace that one and its usage with `sidebarOpened` it should be good. :)

src/components/TopBar.vue
jotoeri: I'd just surround the router push in `showEdit`... with the if-statement, that you basically had before to disable the button. That's also what the second part of the accepted answer on your linked stackoverflow thread wants to do. That cathes the error before it gets produced. The other mentioned solution would produce the error and catch it afterwards. 🤷‍♂️

Chartman123: Should be fixed with my latest commit 👍🏻

Chartman123:
@jancborchardt Nice 🎉  I think that we should clean up the commits a little bit. In my opinion we can squash all commits except the last one into a big one. What do you think?

src/components/TopBar.vue
jotoeri: Uhhhm, i'm sorry, but i don't think this is a good way:
- Normally one would try more to get from sync calls to async calls, since the latter does not block the whole process. This is also most probably, why the Router process changed from sync to async as mentioned in the issue. The now implemented way does just the oposite. 🙈 
- Quite complex code, that one will wonder in 2 years, why this is done
- Calls a method, of which we exactly know, that it will throw an error, just to catch the error then.

I see, that it was for sure nice to implement that and see, that it works, but i'd rather go with the simple if-version i mentioned above. 😉 Sorry 🤷‍♂️

Chartman123: I didn't read the second link as I tried to remove the error after reading the first link ;) But yes, makes more sense. So I'll revert my commit then 👍🏻

susnux (approved):

src/views/Results.vue
jotoeri: Unfortunately i couldn't test it now, but i think the added commit should be it already.

jotoeri:
Could just test it, the Sidebar thing works. Just two more things:
- We shouldn't show the View Switch, if only 'view' is available. Having a shared form with no permissions than submitting, it currently shows the View-Topbar Part.
- @jancborchardt Can we get the border somewhat inside of the view-select? Currently we have a height of 44px+2*2px border.  Can we just fix the height of the view-select to 44px?

Chartman123:
> * We shouldn't show the View Switch, if only 'view' is available. Having a shared form with no permissions than submitting, it currently shows the View-Topbar Part.

@jotoeri fixed it with my last commit

jotoeri:
> @jotoeri fixed it with my last commit

I shifted it once more, since the viewselect border was still visible.
Also added the fixed height now for the view-select.

https://github.com/nextcloud/forms/pull/1381/commits/71a6837d55676269e739c2a89d8e3b628432cdf5

|before|after|
|---|---|
|grafik|grafik|

jotoeri (approved):
Fine by me now

Chartman123 (approved):

PR #1382, opened 2022-10-15
Remember input in LocalStorage

### LINKED ISSUE #456: Ability to come back to a partially filled form
Currently when you fill out a Form partly, close the tab (accidentally or willingly) and open it again, all your responses are gone. It would be very nice if they are kept in the browser local storage so they are available even if your browser or computer crashes.

hamza221:
closes #456

### DIFF
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -98,7 +98,8 @@
 					v-bind="question"
 					:values.sync="answers[question.id]"
 					@keydown.enter="onKeydownEnter"
-					@keydown.ctrl.enter="onKeydownCtrlEnter" />
+					@keydown.ctrl.enter="onKeydownCtrlEnter"
+					@update:values="addFormFieldToLocalStorage(question)" />
 			</ul>
 			<input ref="submitButton"
 				class="primary"
@@ -201,6 +202,13 @@ export default {
 	},
 
 	computed: {
+		formValuesForLocalStorage() {
+			const fromLocalStorage = localStorage.getItem(`nextcloud_forms_${this.publicView ? this.shareHash : this.hash}`)
+			if (fromLocalStorage) {
+				return JSON.parse(fromLocalStorage)
+			}
+			return {}
+		},
 		validQuestions() {
 			return this.form.questions.filter(question => {
 				// All questions must have a valid title
@@ -271,6 +279,7 @@ export default {
 			this.resetData()
 			// Fetch full form on change
 			this.fetchFullForm(this.form.id)
+			this.initFromLocalHost()
 			SetWindowTitle(this.formTitle)
 		},
 	},
@@ -290,9 +299,31 @@ export default {
 			this.fetchFullForm(this.form.id)
 		}
 		SetWindowTitle(this.formTitle)
+		if (this.isLoggedIn) {
+			this.initFromLocalHost()
+		}
 	},
 
 	methods: {
+		initFromLocalHost() {
+			if (localStorage.getItem(`nextcloud_forms_${this.publicView ? this.shareHash : this.hash}`)) {
+				for (const key in this.formValuesForLocalStorage) {
+					const answer = this.formValuesForLocalStorage[key]
+					const answers = []
+					switch (answer?.type) {
+					case 'QuestionMultiple':
+						answer.value.forEach(num => {
+							answers.push(num.toString())
+						})
+						this.answers[key] = answers
+						break
+					default:
+						this.answers[key] = answer.value
+						break
+					}
+				}
+			}
+		},
 		/**
 		 * On Enter, focus next form-element
 		 * Last form element is the submit button, the form submits on enter then
@@ -314,6 +345,20 @@ export default {
 		onKeydownCtrlEnter() {
 			// Using button-click event to not bypass validity-checks and use our specified behaviour
 			this.$refs.submitButton.click()
+		},
+		addFormFieldToLocalStorage(question) {
+			if (!this.isLoggedIn) {
+				return
+			}
+			this.formValuesForLocalStorage[`${question.id}`] = { value: this.answers[question.id], type: answerTypes[question.type].component.name }
+			const parsed = JSON.stringify(this.formValuesForLocalStorage)
+			localStorage.setItem(`nextcloud_forms_${this.publicView ? this.shareHash : this.hash}`, parsed)
+		},
+		deleteFormFieldFromLocalStorage() {
+			if (!this.isLoggedIn) {
+				return
+			}
+			localStorage.removeItem(`nextcloud_forms_${this.publicView ? this.shareHash : this.hash}`)
 		},
 
 		/*
@@ -352,6 +397,7 @@ export default {
 					shareHash: this.shareHash,
 				})
 				this.success = true
+				this.deleteFormFieldFromLocalStorage()
 				emit('forms:last-updated:set', this.form.id)
 			} catch (error) {
 				logger.error('Error while submitting the form', { error })


hamza221:
@jotoeri @Chartman123 Can you please help figuring out how to change QuestionDate State on mounted() :)

jotoeri:
Nice start but probably still a bit to do.
However, i need to write my Masters Thesis currently, so i'm sorry, but i'm unfortunately not able to dig into until mid/end of November. Just aftwerwards i could help you...

jancborchardt:
@Chartman123 @susnux @skjnldsv maybe you can help? :)

@jotoeri all the best for the thesis! :)

Chartman123:
I will try to find some time after the weekend :)

susnux:
Nice start on this one!

>  how to change QuestionDate State on mounted?

Currently this is not possible, as `Dropdown` and `Date` are not initialized from any property but start blank (whereas the other types are already initialized by the `values` property).

I filed a PR for initializing those question types from the values property as well ( #1383 ), which should make this more easy.

---

I did not have a in depth view on your changes, but one thing I was thinking about is:
What happens if you have multiple forms with *saved states*?

hamza221:
@susnux Hey good catch with the multiple forms issue, I didn't pay attention to this scenario. I'll update the PR and Fix the failing Checks.

src/views/Submit.vue
```diff
@@ -261,11 +283,23 @@ export default {
 			// Using button-click event to not bypass validity-checks and use our specified behaviour
 			this.$refs.submitButton.click()
 		},
+		addFormFieldToLocalStorage(question) {
+
+			this.formValuesForLocalStorage[question.id] = { id: question.id, value: this.answers[question.id], type: answerTypes[question.type].component.name }
```
Chartman123, line 288: Unneeded extra new line:

```suggestion
		addFormFieldToLocalStorage(question) {
			this.formValuesForLocalStorage[question.id] = { id: question.id, value: this.answers[question.id], type: answerTypes[question.type].component.name }
```

src/views/Submit.vue
```diff
@@ -236,6 +238,26 @@ export default {
 			this.fetchFullForm(this.form.id)
 		}
 		SetWindowTitle(this.formTitle)
+		if (localStorage.getItem(`${this.shareHash}`)) {
+			this.formValuesForLocalStorage = JSON.parse(localStorage.getItem(`${this.shareHash}`))
```
Chartman123, line 242:
```suggestion
		if (localStorage.getItem(`${this.hash}`)) {
			this.formValuesForLocalStorage = JSON.parse(localStorage.getItem(`${this.hash}`))
```
Should be just "hash". Otherwise the values are stored with an empyt key ""

src/views/Submit.vue
```diff
@@ -261,11 +283,23 @@ export default {
 			// Using button-click event to not bypass validity-checks and use our specified behaviour
 			this.$refs.submitButton.click()
 		},
+		addFormFieldToLocalStorage(question) {
+
+			this.formValuesForLocalStorage[question.id] = { id: question.id, value: this.answers[question.id], type: answerTypes[question.type].component.name }
+			const parsed = JSON.stringify(this.formValuesForLocalStorage)
+			localStorage.setItem(`${this.shareHash}`, parsed)
```
Chartman123, line 290:
```suggestion
			localStorage.setItem(`${this.hash}`, parsed)
```

src/views/Submit.vue
```diff
@@ -261,11 +283,23 @@ export default {
...
+		},
+		deleteFormFieldFromLocalStorage() {
+			this.formValuesForLocalStorage = []
+			const parsed = JSON.stringify(this.formValuesForLocalStorage)
+			localStorage.setItem(`${this.shareHash}`, parsed)
```
Chartman123, line 295:
```suggestion
			localStorage.setItem(`${this.hash}`, parsed)
```

Chartman123 (commented):
Hi @hamza221! Nice to see you back here 🙂 

I don't know much about LocalStorage. However, I think that this is not working as it's intended. I can only see the stored values as long as I navigate within the same form (e.g. between submit - edit - submit). As soon as I leave the form and go to another one and then come back all fields are empty again. As soon as I then do a reload on the page, the values are loaded.

Also could you please do a rebase on current main as we have quite some newer versions in composer.lock 👍🏻

src/views/Submit.vue
Chartman123: One more thing: I think that we should also add some prefix to the storage key like `nextcloud_forms_`

hamza221:
Thank you @Chartman123 
> As soon as I leave the form and go to another one and then come back all fields are empty again
The problem here is that the init from localhost is done in beforeMount hook 
and  apparently all the forms are mounted to the Dom from the beginning we can either change the mounting/unmounting logic or find a work around. what do you suggest

Chartman123:
On a public share link `this.hash` and `this.shareHash` seem to be empty, so the storage key is then `nextcloud_forms_`. I don't know how to add the hash of the form in this case.

Furthermore the array still has lots of `null` values. Probably because you add the values to the local storage with `this.formValuesForLocalStorage[question.id]`. But I don't know if it's possible to do this another way. It just makes the LocalStorage hard to read in the browser console, but as long as it works correctly I wouldn't mind too much.

@jancborchardt Should we also add a "Reset Form" button to clear the LocalStorage and start fresh?

hamza221:
The null, and empty hash are solved, @jancborchardt  should we add a Reset Form button?

Chartman123:
@hamza221 Looks way better now :) However, when you switch the form within one session you end up with multiple records with duplicated contents:
grafik

Looks like we need to reset the object between opening different forms so that the old contents don't get written to the content of another form. I think we need to move the init function away from "beforeMount".

Also, I think that we can get rid of the "id" key inside the object as you're already using it as a key in the top object.

And please rebase on current main 🙂

hamza221:
same job as #1416 failing :/

Chartman123:
The failed jobs must be something general with the images... I also tried to use the same as in the mail repo, but that didn't help either.

Chartman123 (approved):
Looks good now to me. 👍🏻

susnux:
Basically this is good now :+1: 

But there is one thing I am unsure about: Do we want to store values on public shares for anonymous users?
As there might be places where multiple users use the same device, in this case on could see the input of the previous one. (just came to my mind, what do you think?)

Chartman123:
Yes, that's a valid point.

What about combining this here with my started PR regarding the warning when you leave an unsubmitted form? So that we only store the values to the local storage when the user is leaving an unsubmitted form and chooses to save them?

Chartman123:
@hamza221 Submit.vue now handles leaving the form with unsubmitted changes. You could try to update this PR so that data will only be stored in local storage when the user leaves the form without a submission and answers the dialog accordingly

hamza221:
@Chartman123  I'm not sure I understand how this will fix @susnux 's concern. So if I understand correctly ->

1. User attempts to leave unsubmitted form
2. He gets a warning 
3. clicks on Leave page 
4. we save his answers in local storage

Or do you mean we save it when he chooses to stay? If so I'm not sure how is that useful. 

What about saving the answers only if the user is connected and showing the warning otherwise? 

I'm sorry I'm just trying to figure out what's the most useful path we should take :))

Chartman123:
@hamza221 I meant something like this:

1. User has unsubmitted data and leaves the form
2. Popup message/modal: "You have unsaved data. Do you want to save before you leave?"
3. Yes -> trigger save, no -> leave without saving

> What about saving the answers only if the user is connected and showing the warning otherwise?
> [...]
> Google Forms only saves for logged in users and only shows Warning for anonymous users

In this case we could also store the data to the database instead of local storage (see your second update) with the benefit of being independent of the browser. However, this is of course more work as we will have to adjust the API/backend as well.

@susnux @jancborchardt @jotoeri What do you think?

susnux:
I would prefer to go with you suggested solution but only for logged-in users

>    User has unsubmitted data and leaves the form
>    Popup message/modal: "You have unsaved data. Do you want to save before you leave?"
>    Yes -> trigger save, no -> leave without saving

jancborchardt:
Yeah, it's better to have localStorage saving now for error prevention than work longer to get it into the database. The main case is accidentally closing tabs or briwser crashes. :)

hamza221:
| Chrome | Safari |
|--------|--------|
| image | image
 | 

`confirm(t('forms', 'You have unsaved changes! Do you still want to leave?'))`
The confirmation text is not matching, moreover as far as I understand even if we manage to fix the text and make it 
> You have unsaved data. Do you want to save before you leave?

The buttons' text [can't be changed](
https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) to confirm , leave for eg 
I'm working on a workaround using ncModal if that's okay with everyone

Chartman123:
@hamza221 So as far as I can see you already implemented that we will only store and restore the answers to/from the LocalStorage if a user is logged in. This would fix the original issue and so for me it's fine. We can try to combine it with the leave message in a follow-up PR.

> confirm(t('forms', 'You have unsaved changes! Do you still want to leave?'))
> The confirmation text is not matching, [...]

This could be implemented in `beforeRouteLeave()` and `beforeRouteUpdate()`

I will do the code review and testing today :) Thank you very much!

src/views/Submit.vue
```diff
@@ -315,6 +346,22 @@ export default {
...
+			const parsed = JSON.stringify(this.formValuesForLocalStorage)
+			localStorage.setItem(`nextcloud_forms_${this.publicView ? this.shareHash : this.hash}`, parsed)
+		},
+		deleteFormFieldFromLocalStorage() {
+			if (!this.isLoggedIn) {
+				return
+			}
+			this.formValuesForLocalStorage = {}
+			const parsed = JSON.stringify(this.formValuesForLocalStorage)
+			localStorage.setItem(`nextcloud_forms_${this.publicView ? this.shareHash : this.hash}`, parsed)
```
Chartman123, line 363:
```suggestion
			localStorage.removeItem(`nextcloud_forms_${this.publicView ? this.shareHash : this.hash}`, parsed)
```
Is there a reason why you just set the item to an empty object and don't remove the item completely?

hamza221:
> This could be implemented in `beforeRouteLeave()` and `beforeRouteUpdate()`

I'm not sure how . Which modal would you use. I don't think confirm() or alert () are adequate for this use case and I'm not sure how to make a custom modal blocking.

Chartman123:
> I'm not sure how . Which modal would you use. I don't think confirm() or alert () are adequate for this use case and I'm not sure how to make a custom modal blocking.

All fine :) Let's move to a follow-up 👍🏻

src/views/Submit.vue
hamza221: I can't think of a reason why I did that haha, fixed :))

src/views/Submit.vue
```diff
@@ -342,6 +387,7 @@ export default {
 		 * Submit the form after the browser validated it 🚀
 		 */
 		async onSubmit() {
+			this.deleteFormFieldFromLocalStorage()
```
Chartman123, line 390: I think we should move this line to after a successful submission. So if the submission fails for some reason we still have the values in LocalStorage

hamza221: good point, on it

Chartman123:
@hamza221 Could you please have another look at the code: I missed a problem when you reload a form with saved values: the UI doesn't update when you choose an option of a checkbox/radio question...

Chartman123:
@susnux @hamza221 The problem looks a lot like the problem that we fixed here for "other" answer questions... https://github.com/nextcloud/forms/pull/1764

Might this also come from having `:values.sync...` and `@update:values...` in Submit.vue for the Questions?

Chartman123:
Created a new issue for the problem: #1780

PR #1383, opened 2022-10-16
Init all answers from props

susnux:
Before this only the text answers were initialized with the `values` property.
This now also initializes `Dropdown` and `Date` from the `values` property, which is required for loading a saved state of a form like in #1382

### DIFF
--- a/src/components/Questions/QuestionDate.vue
+++ b/src/components/Questions/QuestionDate.vue
@@ -92,6 +92,13 @@ export default {
 		},
 	},
 
+	mounted() {
+		// Init time from values prop
+		if (this.values) {
+			this.time = this.parse(this.values[0])
+		}
+	},
+
 	methods: {
 		/**
 		 * DateTimepicker show text in picker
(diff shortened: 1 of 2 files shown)


Chartman123 (approved):

hamza221 (approved):

Chartman123:
@susnux I talked to @jotoeri and he wants to have a closer look on this when he'll find some more time for it. So we should wait with the merge 🙂

jotoeri:
Nah, sorry, @Chartman123 it was not meant that hard... I'm not sure, if this is good or if we should more have a look in a common way between the questions but that I would need to check properly. This is why I didn't react here.

Chartman123:
Yes, that's what I understood, so perhaps I didn't find the right words here 😇

susnux:
No problem take your time to review!

hamza221:
Any updates on merging this PR?

jotoeri:
Finally found some time. Sorry for that. Seems to be ok currently. 🙈

jotoeri (approved):

jotoeri:
Argh. Now i was too fast. Rebase would have been good. 🤦‍♂️

PR #1390, opened 2022-10-24
Save Submission automatically in files on submit

hamza221:
On Submission answers are saved to a csv file /forms/formName,csv

Question/ help needed?

- [ ]  What Should I change the Save CSV to files to?  @jancborchardt suggested redirecting to the file in the question how would you recommend doing that from a UX pov

Signed-off-by: hamza mahjoubi <hamzamahjoubi221@gmail.com>

### DIFF
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -20,7 +20,7 @@
  -
  -->
 
-<template>
+ <template>
 	<NcAppContent v-if="isLoadingForm">
 		<NcEmptyContent :title="t('forms', 'Loading {title} …', { title: form.title })">
 			<template #icon>
@@ -96,18 +96,16 @@
 
 <script>
 import { loadState } from '@nextcloud/initial-state'
-import { generateOcsUrl } from '@nextcloud/router'
-import { showError } from '@nextcloud/dialogs'
+import { generateOcsUrl, generateRemoteUrl } from '@nextcloud/router'
 import axios from '@nextcloud/axios'
 import moment from '@nextcloud/moment'
 import NcAppContent from '@nextcloud/vue/dist/Components/NcAppContent.js'
 import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
 import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 import IconCheck from 'vue-material-design-icons/Check.vue'
-
+import { showSuccess, showError } from '@nextcloud/dialogs'
 import answerTypes from '../models/AnswerTypes.js'
 import logger from '../utils/Logger.js'
-
 import Question from '../components/Questions/Question.vue'
 import QuestionLong from '../components/Questions/QuestionLong.vue'
 import QuestionShort from '../components/Questions/QuestionShort.vue'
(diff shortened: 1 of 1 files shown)


Chartman123:
Hi @hamza221 

Just a quick note before you put more effort into this: you shouldn't do this in the vue files as they are only used when a user actually uses the web interface. As soon as a submission will be made via the API it will not get written into the csv file. So you should put this into the ApiController.php and/or the SubmissionService.php :)

Chartman123:
> On Submission answers are saved to a csv file /forms/formName,csv

I don't think it's a good idea to use a hard coded path for storing the file. IMHO it would be better to make this configurable in the settings tab in the sidebar of the form.

> * What Should I change the Save CSV to files to?  @jancborchardt suggested redirecting to the file in the question how would you recommend doing that from a UX pov

I don't fully understand what you mean, could you please explain it a little bit more?

jancborchardt:
> I don't think it's a good idea to use a hard coded path for storing the file. IMHO it would be better to make this configurable in the settings tab in the sidebar of the form.

@Chartman123 Just FYI on that, having a "Forms" folder (uppercase first character just like the app name, and translated) is similar to how we do it for other apps like Talk (shared images), Notes, Collectives, and possibly more. Some of them might have it configurable, but I think it’s fine to have it "hard coded" in the first iteration.

That’s why I suggested doing it like that to @hamza221 :)

----

> What Should I change the Save CSV to files to?
> @jancborchardt suggested redirecting to the file in the question how would you recommend doing that from a UX pov

@hamza221 Yeah, I would say renaming the button "Save CSV to Files" to "Open spreadsheet in Files" would be good.

Chartman123:
> @Chartman123 Just FYI on that, having a "Forms" folder (uppercase first character just like the app name, and translated) is similar to how we do it for other apps like Talk (shared images), Notes, Collectives, and possibly more. Some of them might have it configurable, but I think it’s fine to have it "hard coded" in the first iteration.

@jancborchardt yes I understand that. So this could be the default location. But I think for larger organizations this wouldn't bring that much of benefit. So making it configurable could also enable the users to store the responses to e. g. a group folder :)

jancborchardt:
@Chartman123 yep, just saying it's probably better to do that in a follow-up.

Because putting it all in one folder completely allows for your use-case by simply sharing in Files, either fully or on a file basis.

Putting it in Forms itself then raises the question which level it should be on, and if your suggested use-case should be covered it probably has to be on a file/form basis, which introduces a lot of complexity.

It's ok for the first iteration to be simple, we always do stuff step by step. :)

Chartman123:
> it probably has to be on a file/form basis

@jancborchardt Yes, but this should be a "must have". Or do you want to automatically export every submission for every form into a csv file?

jancborchardt:
> Or do you want to automatically export every submission for every form into a csv file?

Ah ok, so I was under the impression that Google does this automatically for any new form, but you have to first enable it, here is the full flow @hamza221 for reference:

You can "Create spreadsheet" for responses
image

I’d say for our first vesion it’s ok to skip this step and go with reasonable naming `Forms/Formname (responses).csv`.
An issue in the current version is that if you rename or move the file it does not update that file but creates a new one, so instead it should act on the file id?
image

And from then on it’s linked:
image

And you can also unlink the form again:
image


What do you think @hamza221 @Chartman123?

Chartman123:
@jancborchardt Yes that sounds good to me 🙂

hamza221:
I agree , the way google handles it is way better

Issue #1386, opened 2022-10-24
Can't clone form


tzugen:
Server error 500:
`
[...]

PR #1391, opened 2022-10-25
Fix `setExtraSettings` to accept same type as `getExtraSettings`

susnux:
* Resolves: #1386

## Summary
This fixes the extra settings setter to accept the same parameter type as the getter emits.
So this resolves an issue where a form can not be cloned.


Chartman123 (approved):
Did a quick check with copy form, edit questions, switch shuffle checkbox, submit... Looks good for me

jotoeri:
Nice, so let's get it out then! ❤️ 🚀

Issue #489, opened 2020-06-28
Allow formatting with Markdown


EricKerby:
- Nextcloud-Version: 19.0.0
- Forms-Version: 2.0.0-beta4

It would be helpful to allow formatting the text in the description and field names with markdown or HTML. For recent forms I created, I found myself wanted to add emphasis to some text as well as a link or two.

skjnldsv:
@juliushaertl the dedicated component is not yet available, right? :)

juliushaertl:
No but work in progress https://github.com/juliushaertl/nextcloud-text/ 

It already is in kind of a working state except for some node types like images, but needs some more polishing for a first release. Unfortunately I didn't have much time lately.

skjnldsv:
It's definitely not a priority! Take your time :)

EricKerby:
Thanks, @juliushaertl ! Providing a visual text editor component which Forms and other apps can then utilize certainly sounds like a solid approach to satisfy this enhancement request.

bernd-wechner:
Not sure why the thumbs down from @skjnldsv  because, given formatting is not available I put the description of the form onto a PicoCMS page, no problem, and it just links to the form for completion, but I wanted badly to put in a link back from the form to the intro page as well, measly little link. Turns out it's easy enough to do with a small tweak:

```
diff --git a/src/views/Submit.vue b/src/views/Submit.vue
index 7fd055b..52981ee 100644
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -30,7 +30,7 @@
                                </h2>
                                <!-- Do not wrap the following line between tags! `white-space:pre-line` respects `\n` but would produce additional empty first line -->
                                <!-- eslint-disable-next-line -->
-                               <p v-if="!loading && !success" class="form-desc">{{ form.description }}</p>
+                               <p v-if="!loading && !success" class="form-desc" v-html="form.description"></p>
                                <!-- Only visible if at least one question is marked as mandatory-->
                                <p v-if="mandatoryUsed && !loading && !success" class="info-mandatory">
                                        * {{ t('forms', 'Required questions') }}
```

[...]

Issue #441, opened 2020-06-05
Adding images to questions


jancborchardt:
Google Forms has the ability to add images to questions, e.g. here to multiple choice, which is a nice way to visualize it:
![image](https://user-images.githubusercontent.com/925062/83910138-5143ef00-a76a-11ea-803e-a90868f88bc3.png)

maxolasersquad:
I'd like to add, images in both questions and answers would be preferable. This may be resolved automatically by implementing #489.

Andreas-Kainz:
If add image is possible to the answer would it be possible to make an image with the mobile and it will be uploaded?

Cause than I can use form for site inspections.

fwsmit:
You can add images using markdown:

! [] (URL)

This is a bit finicky though and doesn't work entirely as you expect.

alauzon:
> You can add images using markdown:
> 
> ! [] (URL)
> 
> This is a bit finicky though and doesn't work entirely as you expect.

Forms is not using markdown so this is useless.

Chartman123:
In the latest release, you can use markdown for the form description and the question descriptions. See #489

elephantastyczny:
Maybe it's worth to extend it to allow users to upload a picture to a form?
The we could use Forms as a "ticketing solution".

AtiusAmy:
I hope this will be possible soon :sob:

LAfricain:
How do I use an image uploaded to nextcloud in Markdown? I have only the share link.

solitonmedic:
> How do I use an image uploaded to nextcloud in Markdown? I have only the share link. I can answer the question: select the picture, in the settings of the picture, right click, copy the download link.

Thanks for this, this is exactly what I was looking for.
I don't understand why can't we use hosted images elsewhere, must we upload them to our cloud first? Such a convoluted process to get a single image to show up at the end of a survey.

chrismey-71:
You can already use markdown in the description of forms. However, the markdown command for inserting an image doesn't work. Only the alternative text appears. Is this a bug or is markdown not yet fully implemented? NC 29.0.7 // forms 4.2.4

Chartman123:
@chrismey-71 afaik only some basic markdown features are currently supported

maptal:
This would be awesome to have!

Jacques-Olivier-Farcy:
I was trying some markdown syntaxes in the form description, in vain.
So yes, it would be a real awesome feature!

mrjive:
[...]

PR #1394, opened 2022-10-26
Allow formatting question and form descriptions using markdown

### LINKED ISSUE #1253: Allow links to be embedded into description
**Nextcloud:**
- Nextcloud-Version: 24.0.2
- Forms-Version: 2.5.1

When creating forms, the possibility to embed links can be helpful to clear up things, include example images and more. Forms currently does not allow embedding of links.

Automatically wrap URLs in description within a `<a href='...'></a>` tag to make them clickable.

Placing links inside description does display them, however them not being interactable reduces user experience a lot and might cause issues for some people.

susnux:
* Resolves: #489 and resolves #1253

This will allow users to use markdown (CommonMark) for formatting descriptions (both form and question).
See also this recording for example usage:

https://user-images.githubusercontent.com/1855448/198033394-49cbb74b-3ba0-4180-8660-5c17f4a76235.mp4


Chartman123:
@susnux just a quick question: is there a reason why you moved the title handling to the mixin?

susnux:
> @susnux just a quick question: is there a reason why you moved the title handling to the mixin?

Because it is duplicated code / shared code between the create and submit view.
So I moved it to the mixin instead of copying for both views.

> I think that the css styles need some more love... at least currently for me the layout moves when I switch between edit and view mode...

What kind of move? If the input is just plain text there should be no difference, but of cause markdown and the rendered output might take different space and so there might be movement.

Chartman123:
Yes that makes sense... But you also did a switch from textarea to input and textarea to NcRichContenteditable. So there's probably some other CSS applied to those elements

Chartman123:
@susnux I think I found the source of the jumping: 

https://github.com/nextcloud/forms/blob/458bea1cc27b2c7e27e5baf86a77ac09df835e86/src/components/Questions/Question.vue#L373

I disabled this rule in the dev tools and the jumping between edit and view is gone... Could you please verify this (for one line descriptions) and put it into the PR?

gohrner:
This may also fix https://github.com/nextcloud/forms/issues/441 ?

Chartman123:
> This may also fix #441 ?
> 
> Not sure if this PR also supports markdown in the individual answer options (checkmark / radio button descriptions)?

I think it makes more sense to keep the focus on the descriptions here.

susnux:
>Could you please verify this (for one line descriptions) and put it into the PR?

@Chartman123 I can verify this was the cause, I removed that rule and pushed it to this PR. No jumping here anymore :)

Chartman123 (approved):
So for me it's fine now :)

Perhaps @jotoeri will find some time to review as well

src/views/Create.vue
```diff
@@ -38,30 +38,37 @@
...
 					:placeholder="t('forms', 'Form title')"
 					:required="true"
 					autofocus
 					type="text"
-					@input="onTitleChange" />
+					@input="onTitleChange">
+				<div v-else
+					id="form-title"
+					class="form-title"
+					v-text="formTitle" />
```
jotoeri, line 59: Please use the same element as in submit.
https://github.com/nextcloud/forms/blob/1a9445e00b78af3e3f59bd9ffe92ca8c56df1dd6/src/views/Submit.vue#L39-L41

src/views/Create.vue
```diff
@@ -38,30 +38,37 @@
...
 			</h2>
 			<label class="hidden-visually" for="form-desc">{{ t('forms', 'Description') }}</label>
-			<textarea ref="description"
-				v-model="form.description"
+			<NcRichContenteditable v-if="edit || !form.title"
+				ref="description"
```
jotoeri, line 63: Should be unused now?

src/views/Create.vue
```diff
@@ -38,30 +38,37 @@
 			@update:sidebarOpened="onSidebarChange"
 			@share-form="onShareForm" />
 		<!-- Forms title & description-->
-		<header>
+		<header v-click-outside="disableEdit" @click="enableEdit">
 			<h2>
 				<label class="hidden-visually" for="form-title">{{ t('forms', 'Form title') }}</label>
-				<textarea id="form-title"
+				<input v-if="edit || !form.title"
```
jotoeri, line 44: This kills the multiline input of Form titles. The textarea was somewhen added on purpose.

src/views/Create.vue
jotoeri: And please just insert the text with the brackets instead of the vue-directive, so we have common code across the app (and across most of nextcloud btw.)

src/components/Questions/Question.vue
```diff
@@ -79,18 +79,16 @@
...
 					:value="description"
-					:maxlength="maxStringLengths.questionDescription"
 					:placeholder="t('forms', 'Description')"
+					:maxlength="maxStringLengths.questionDescription"
 					class="question__header__description__input"
-					@input="onDescriptionChange" />
+					@update:value="onDescriptionChange" />
 
-				<!-- Do not wrap the following line between tags! `white-space:pre-line` respects `\n` but would produce additional empty first line -->
 				<!-- eslint-disable-next-line -->
```
jotoeri, line 90: Can you please replace that with the specific version? 
`<!-- eslint-disable-next-line vue/no-v-html -->`

src/components/Questions/Question.vue
```diff
@@ -212,9 +218,9 @@ export default {
 			this.$emit('update:text', target.value)
 		},
 
-		onDescriptionChange({ target }) {
+		onDescriptionChange(value) {
 			this.autoSizeDescription()
```
jotoeri, line 222: autosize has no effect now.

src/views/Create.vue
```diff
@@ -38,30 +38,37 @@
...
 				class="form-desc"
-				rows="1"
-				:maxlength="maxStringLengths.formDescription"
+				:multiline="true"
+				:value.sync="form.description"
 				:placeholder="t('forms', 'Description')"
-				@input="onDescChange" />
+				:maxlength="maxStringLengths.formDescription"
+				@input="saveDescription" />
+			<!-- eslint-disable-next-line -->
```
jotoeri, line 70: `<!-- eslint-disable-next-line vue/no-v-html -->`

src/views/Create.vue
```diff
@@ -170,30 +184,19 @@ export default {
 
 	data() {
 		return {
-			maxStringLengths: loadState('forms', 'maxStringLengths'),
-
-			questionMenuOpened: false,
```
jotoeri, line 175: Why?

src/views/Submit.vue
```diff
@@ -39,9 +39,8 @@
 			<h2 ref="title" class="form-title">
 				{{ formTitle }}
 			</h2>
-			<!-- Do not wrap the following line between tags! `white-space:pre-line` respects `\n` but would produce additional empty first line -->
 			<!-- eslint-disable-next-line -->
```
jotoeri, line 42: `<!-- eslint-disable-next-line vue/no-v-html -->`

src/views/Create.vue
```diff
@@ -438,11 +432,13 @@ export default {
...
-			font-size: 100%;
-			line-height: 150%;
-			padding-bottom: 20px;
 			margin: 0px;
 			resize: none;
+			font-size: 100%;
+			min-height: unset;
+			line-height: 1.5em;
+			padding-bottom: 20px;
+			color: var(--color-text-maxcontrast) !important;
```
jotoeri, line 441: Christmas tree sorting??

src/mixins/ViewsMixin.js
```diff
@@ -58,6 +59,25 @@ export default {
...
+		 * Return form title, or placeholder if not set
+		 *
+		 * @return {string}
+		 */
+		formTitle() {
+			if (this.form.title) {
+				return this.form.title
+			}
+			return t('forms', 'New form')
+		},
```
jotoeri, line 73: Can you remove it also from Results.vue then?

jotoeri (changes requested):
- Newlines in the description get lost
- FormHead still jumps around between edit/view
- Having no border on the question description, the same should apply to the form description. Or the other way round, i fear we have to move to borders on all, since one otherwise does not see the mode change within the header.

susnux:
I will work through this points tomorrow.

> * Newlines in the description get lost

This is because simple newlines are soft breaks in markdown so this is expected behavior you will have to use double newlines or hardbreaks (backslash+newline or two space + newline).

But I think we will have to migrate old descriptions to markdown for this, I will also have a look at this topic tomorrow.

By the way I want to thank you for your work, your reviews are very in depth, sadly a rare thing on open source projects :)

susnux:
Sorry for the delay, had some busy weeks, but I have some time for this now :)

src/components/Questions/Question.vue
```diff
@@ -189,6 +190,11 @@ export default {
 			return this.text
 		},
 
+		computedDescription() {
+			const md = MarkdownIt()
```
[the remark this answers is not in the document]
susnux: better: per view

src/views/Create.vue
```diff
@@ -422,9 +417,8 @@ export default {
 		.info-message {
 			width: 100%;
 			padding: 0 16px;
-			border: none;
 		}
-		.form-title {
```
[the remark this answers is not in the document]
susnux: reverted

src/mixins/ViewsMixin.js
susnux: done

susnux:
@jotoeri I think I resolved all review notes :rocket:

src/views/Create.vue
```diff
@@ -403,41 +367,40 @@ export default {
...
+		}
+
+		.form-desc {
+			color: var(--color-text-maxcontrast) !important;
+			line-height: 1.5em;
 			margin: 0px;
-			resize: none;
+
+			&.readonly {
+				padding: 0px 15px 17px; // compensate border of editable content
```
Chartman123, line 399: Still some moving of the elements between view and edit mode for the form description. If you add the 1px padding to the top at least for me it stays in place 🙂 

```suggestion
				padding: 1px 15px 17px; // compensate border of editable content
```

susnux: Thank you! Works :)

susnux:
I think this ready to go now :)

jotoeri (commented):
@jancborchardt Could you also comment design-wise? With some distancing i'm not sure, what would be best.

Generally, @susnux, please try to keep the PR's simple. Don't mix changes, that are not necessarily related. Also just shifting around of Code without obvious reason is just annoying and hard to review, especially if there is much more and reasonable code on the PR, that needs to be distinguished. Better create separate small PR's, that can be quickly reviewed.

jotoeri (commented):
Ah -  and empty form-title gets lost on sidebar and submit currently.

grafik

src/views/Results.vue
```diff
@@ -198,15 +198,12 @@ export default {
...
 		 *
 		 * @return {string}
 		 */
-		formTitle() {
-			if (this.form.title) {
-				return this.form.title
-			}
-			return t('forms', 'New form')
+		downloadUrl() {
+			return generateOcsUrl('apps/forms/api/v2/submissions/export/{hash}', { hash: this.form.hash })
```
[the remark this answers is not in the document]
susnux: Oh sorry this seems to be a artifact from rebasing

src/components/Questions/Question.vue
```diff
@@ -100,16 +97,19 @@
 </template>
 
 <script>
-import { directive as ClickOutside } from 'v-click-outside'
```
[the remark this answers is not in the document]
susnux: Ok, good to known, fixed that :)

src/components/Questions/Question.vue
```diff
@@ -100,16 +97,19 @@
...
 
 import IconAlertCircleOutline from 'vue-material-design-icons/AlertCircleOutline.vue'
 import IconDelete from 'vue-material-design-icons/Delete.vue'
 import IconDragHorizontalVariant from 'vue-material-design-icons/DragHorizontalVariant.vue'
 
-export default {
+import { directive as ClickOutside } from 'v-click-outside'
+import { defineComponent } from 'vue'
+
+export default defineComponent({
```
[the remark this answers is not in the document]
susnux: Sure :)
BTW it is not needed if you do not use typescript, even with vue3, but it is quite useful for IDE like for IntelliSense.

src/views/Create.vue
Chartman123: Did this disappear again? I can see a small movement when switching between the modes... This might also be related to the NcRichContenteditable as this component seems to have a thinner border than the other input fields. I created an issue for the lib: https://github.com/nextcloud/nextcloud-vue/issues/3635

src/views/Create.vue
```diff
@@ -403,41 +367,40 @@ export default {
 		margin-bottom: 24px;
 		margin-left: 56px;
 
-		h2 {
-			margin-bottom: 0; // because the input field has enough padding
-		}
-
-		.form-title,
-		.form-desc,
```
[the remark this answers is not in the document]
Chartman123: Just saw this comment after hitting the send button on my other comment... For reference here's the issue from the vue lib again https://github.com/nextcloud/nextcloud-vue/issues/3635

src/views/Create.vue
```diff
@@ -38,30 +38,29 @@
...
 				<label class="hidden-visually" for="form-title">{{ t('forms', 'Form title') }}</label>
-				<textarea id="form-title"
-					ref="title"
-					v-model="form.title"
-					class="form-title"
-					rows="1"
-					:minlength="0"
+				<NcRichContenteditable id="form-title"
+					:value.sync="form.title"
+					:multiline="true"
```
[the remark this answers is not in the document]
susnux: Reverted the switch to `NcRichContent` for the title, the also solved the issue with empty form names on the app navigation (`NcRichContent` always adds a newline to the value resulting in a not empty string but `'\n'`).

susnux:
> Ah - and empty form-title gets lost on sidebar and submit currently.

Fixed that, this was cause by `NcRichContent` which always added a newline to the value.

jotoeri (changes requested):
Did you test this? Did you check the diff here?
- Title looks completely different on create.vue
- Distance between Desc. & Title different between Create and Submit

susnux:
> Did you test this? Did you check the diff here?
> 
>     * Title looks completely different on create.vue
>     * Distance between Desc. & Title different between Create and Submit

Of cause I tested this, I even took screenshots to pixel compare them (checking if there is any jumping between edit edit-view and submit).
But I did not spot any difference:

| Submit | Create read | Create edit |
|-------|---------|--------|
| submit view | create view readonly | create view edit |

src/views/Create.vue
susnux: I think I do not understand the issue here, as your said the title must not be multi-line https://github.com/nextcloud/forms/pull/1394#discussion_r1060832519

susnux: Oh... Did you mean that `textarea` allows the text to break, even without newline, if the text is too long?
Because the forms title seems to support two "lines" until it overflows.

My intention behind `<input>` was to unified the it with the title input of `Question.vue` which was always `<input>`.

But... I am sorry, simply misunderstood you comment :/

jotoeri:
🤔 I now checked a second time and i think this is how it looks, if i check out the current PR... Can you double-check, if you pushed all commits or so?

| Submit | Create Read | Create Edit
|---|---|---|
|grafik|grafik|grafik|

src/views/Create.vue
jotoeri: > Oh... Did you mean that `textarea` allows the text to break, even without newline, if the text is too long?

Yes, thats it. 👍 The title should break automatically if it is too long, but people should not be able to create a title with `\n` newlines.

- Most probably it would be worth to adapt the Question-titles to also use textarea (Or the RichContent as you started with). The current state is more of a historical reason, that we earlier had inputs everywhere, but changed the title somewhen because of complaints. - But please, if you're motivated to change it, do it in a separate PR. 😉

susnux:
> thinking I now checked a second time and i think this is how it looks, if i check out the current PR... Can you double-check, if you pushed all commits or so?

Everything up-to-date, but that looks like it uses a different font size within the input, which is not changed by the CSS, so this might be browser related (I use Firefox 108) or maybe caching?
Nevertheless as I revert the change to `input` this should be "fixed" then.

susnux:
I rebased this branch and reverted the textarea change to fix the issues mentioned above.
I checked also there is no jumping or other visual issues anymore.

| Submit | Create | Create edit | Create edit questions |
|------------|-----------|------------------|---------------------------------|
| screenshot submit view | screenshot create view | screenshot create view editing header | screenshot create view editing questions |

Chartman123:
@susnux: Just a little thing that I saw right now: In your question edit screenshot the textboxes for the descriptions/options are a little bit wider than the title. I don't know if this was also before, but perhaps you can have a look at this :)

susnux:
> @susnux: Just a little thing that I saw right now: In your question edit screenshot the textboxes for the descriptions/options are a little bit wider than the title. I don't know if this was also before, but perhaps you can have a look at this

Sure! Fixed the border alignment, it now looks like this:
screenshot question borders aligned

src/views/Create.vue
Chartman123: @jotoeri Btw currently you can add newlines to the form title on main branch. However, they will not be shown in the submit view.

jotoeri: Oh, really? 👀 Probably one more good reason to get to the Rich-Content Component in a next PR?!

Chartman123:
@susnux @jotoeri Did some review of the application in the browser. Couldn't find any unwanted differences to the current main branch. So from my side everything should be ok. If there is nothing more codewise, we can give it a go :)

package-lock.json
```diff
@@ -2626,15 +2627,15 @@
 			}
 		},
 		"node_modules/@typescript-eslint/eslint-plugin": {
-			"version": "5.48.1",
-			"resolved": "https://registry.npmjs.org/@typescript-eslint/eslint-plugin/-/eslint-plugin-5.48.1.tgz",
-			"integrity": "sha512-9nY5K1Rp2ppmpb9s9S2aBiF3xo5uExCehMDmYmmFqqyxgenbHJ3qbarcLt4ITgaD6r/2ypdlcFRdcuVPnks+fQ==",
+			"version": "5.48.2",
+			"resolved": "https://registry.npmjs.org/@typescript-eslint/eslint-plugin/-/eslint-plugin-5.48.2.tgz",
+			"integrity": "sha512-sR0Gja9Ky1teIq4qJOl0nC+Tk64/uYdX+mi+5iB//MH8gwyx8e3SOyhEzeLZEFEEfCaLf8KJq+Bd/6je1t+CAg==",
```
Chartman123, line 2632: Are these changes needed here? Also valid for the other related changes in here...

Chartman123 (approved):
One more comment on the package-lock.json. Rest seems fine

jotoeri:
- Now its on the left side and even worse?! 
grafik
- From empty Title, when entering the first char, the title input changes size for me. @Chartman123 can you confirm this?
- I think Form title and description should **always** change edit-mode together. Currently, if the form title is empty, the description disappears. That looks strange imo.
- If there's no form-description, the Title and the info-message are now quite far apart. In fact 20px more, than without this PR. Can we get that down again? I think this doesn't necessarily be pixel-perfect between create.vue and submit.vue, but more important it should look good on submit.vue.

@Chartman123 @susnux  any opinions?

package-lock.json
jotoeri: Ah, and i personally dont mind too much on this, as its only on the lockfile here... Its indeed not related, but Dependabot would anyways change it with the next run...

Chartman123:
> * Now its on the left side and even worse?!

That's basically the same in main and there since Forms 3 iirc



> * From empty Title, when entering the first char, the title input changes size for me. @Chartman123 can you confirm this?

No, can't reproduce here.

> * I think Form title and description should **always** change edit-mode together. Currently, if the form title is empty, the description disappears. That looks strange imo.

Yes, same here... didn't try this before, but yes: should be linked

> * If there's no form-description, the Title and the info-message are now quite far apart. In fact 20px more, than without this PR. Can we get that down again? I think this doesn't necessarily be pixel-perfect between create.vue and submit.vue, but more important it should look good on submit.vue.

Confirm, I think the description element should be condtionally shown only if there is some text

susnux:
I fall into line :)

> * Now its on the left side and even worse?!

I can remove this but I think this would be better placed in a follow up PR as it is unrelated, but I do not think this is bad as it makes clear what is the question header and what the body.

> * I think Form title and description should **always** change edit-mode together. Currently, if the form title is empty, the description disappears. That looks strange imo.

I agree, fixed that.

> * If there's no form-description, the Title and the info-message are now quite far apart. In fact 20px more, than without this PR. Can we get that down again? I think this doesn't necessarily be pixel-perfect between create.vue and submit.vue, but more important it should look good on submit.vue.

Same, pushed a commit which fixes this :)

jotoeri:
> Now its on the left side and even worse?!
Well, 'unrelated' depends - you increased the offset from 2px to 4px in this PR, instead of aligning the description to the existing field...
I'm fine with having a clear difference between question and options, but currently it looks more like a bug imo. So i'd say, @susnux  please clarify with @jancborchardt how it should like and eventually fix it in a separate PR.

> From empty Title, when entering the first char, the title input changes size for me.
That's strange... I still see it on my Firefox 109 but not on edge. 🤔

jotoeri (approved):
🤷‍♂️

jotoeri:
@susnux wanna rebase & squash the commits here? Then we can merge it.

susnux:
> wanna rebase & squash the commits here? Then we can merge it.

Nice you released the hotfix update :)

hanserasmus:
Are these changes included in version 3.0.4? I have 3.0.4 installed, and none of the above Markdown stuff works for me. Trying to add a URL to a question and/or description and it displays the raw text, not MD formatted.

susnux:
No this will be included with upcoming 3.1.0 release.

hanserasmus:
Thank you @susnux. Is there a place where I can watch the project timeline? I need this urgently for a project, so if needed be, I will build the app from source code to have this capability. But don't want to build from source if 3.1.0 is imminent.

susnux:
I do not think there is a fixed schedule, but I am not into that topic, maybe @jotoeri can help.

jotoeri:
Yes, there generally is no fixed release schedule for forms. But we anyways have to release for NC26 within the next 3 days. Maybe i'll find some time later, otherwise tomorrow.

hanserasmus:
Thank you @jotoeri . I saw the release, updated, and it broke. :-( Bug report filed: https://github.com/nextcloud/forms/issues/1528
Thank you still for the quick release! This will definitely help me once I get it working again.

PR #1403, opened 2022-11-08
Submission Files can be linked and unlinked to form #5

hamza221:
Signed-off-by: hamza mahjoubi <hamzamahjoubi221@gmail.com>

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -105,7 +105,7 @@ class ApiController extends OCSController {
 
 	/** @var IUserManager */
 	private $userManager;
-
+	
 	public function __construct(string $appName,
 								ActivityManager $activityManager,
 								AnswerMapper $answerMapper,
@@ -134,7 +134,7 @@ class ApiController extends OCSController {
 		$this->configService = $configService;
 		$this->formsService = $formsService;
 		$this->submissionService = $submissionService;
-
+		
 		$this->l10n = $l10n;
 		$this->logger = $logger;
 		$this->userManager = $userManager;
@@ -348,7 +348,7 @@ class ApiController extends OCSController {
 	 * @param array $keyValuePairs Array of key=>value pairs to update.
 	 * @return DataResponse
 	 * @throws OCSBadRequestException
-	 * @throws OCSForbiddenException
+	 * @throws OCSForbiddenException 
 	 */
 	public function updateForm(int $id, array $keyValuePairs): DataResponse {
 		$this->logger->debug('Updating form: FormId: {id}, values: {keyValuePairs}', [
@@ -992,9 +992,17 @@ class ApiController extends OCSController {
 			}
 		}
 
+
 		//Create Activity
 		$this->activityManager->publishNewSubmission($form, $submission->getUserId());
 
+		$filePath = $form->getFilePath();
+		$hash = $form->getHash();
+		$ownerId = $form->getOwnerId();
+
+		if($form->getFileId() !=null){
+			$this->submissionService->updateOnSave($hash, $filePath,$ownerId );
+		}
 		return new DataResponse();
 	}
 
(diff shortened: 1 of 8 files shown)


hamza221:
I'm aware of the failing checks 🥲, I'll acknowledge them once I get feedback. I'll work on other features in the meantime

Chartman123:
@hamza221 Had a quick look at the code and think this is the way to go here. However, I didn't have the time to have a closer look. One thing I saw: I think that you can stay with the API v2 here, as we're not breaking compatibility with this change.

hamza221:
Yes, I was actually not sure about that, I'll change it in my next commit

hamza221:
FYI: Before testing this branch migration version.  3000Date20221101205756 should be executed 
Navigate to server folder and run ` php occ migrations:execute forms 3000Date20221101205756`

jancborchardt (changes requested):
Very nice! :) Some changes as discussed:
- [x] Move the "Link to a spreadsheet" entry out of the menu into a button saying "Create spreadsheet", using the "table" icon
- [x] When it says "Open spreadsheet", it can also use the "table" icon instead of folder. Also note fixed capitalization from "spreadSheet" ;)
- [x] The "Unlink spreadsheet" should be the first item in the menu.
- [ ] Check if the file_id can just use VARCHAR(64) instead of 255

Chartman123:
What came to my mind regarding the wording: The files are basically csv files and no spreadsheets, aren't they?

jancborchardt:
@Chartman123 yes they are, but we were thinking that "spreadsheet" is a more human term in this case cause it will then be open-able via Nextcloud Office. :) 

Also, Google Forms calls this functionality "Create spreadsheet".

Chartman123:
> Also, Google Forms calls this functionality "Create spreadsheet".

@jancborchardt Yes, because in their case, they really create a Google Spreadsheet document and not only a CSV ;-) I'm thinking about the users that don't have installed a office solution in their nextcloud...

jancborchardt:
> > Also, Google Forms calls this functionality "Create spreadsheet".
> 
> @jancborchardt Yes, because in their case, they really create a Google Spreadsheet document and not only a CSV ;-) I'm thinking about the users that don't have installed a office solution in their nextcloud...

Yeah understood – even in that case, it will still work as a spreadsheet. :) It’s just that it’s a more accessible name than "CSV" for a function that will be now very visible in the interface as a button.

lib/Controller/ApiController.php
```diff
@@ -992,9 +992,17 @@ public function insertSubmission(int $formId, array $answers, string $shareHash
...
 		//Create Activity
 		$this->activityManager->publishNewSubmission($form, $submission->getUserId());
 
+		$filePath = $form->getFilePath();
+		$hash = $form->getHash();
+		$ownerId = $form->getOwnerId();
+
+		if ($form->getFileId() != null) {
+			$this->submissionService->updateOnSave($hash, $filePath, $ownerId);
+		}
```
Chartman123, line 1005: This should only be executed if there is a file linked

lib/Migration/Version3000Date20221101205756.php
```diff
@@ -0,0 +1,64 @@
...
+		$table->addColumn('file_id', Types::STRING, [
+			'notnull' => false,
+			'default' => null,
+			'length' => 64,
+		]);
+
+		$table->addColumn('file_path', Types::STRING, [
+			'notnull' => false,
+			'default' => null,
+		]);
```
Chartman123, line 61: @jotoeri should we move the file path and the file id to a JSON field "extra_settings_json" just like for the questions?

And @hamza221 could you please rename the migration file to match the names of the other migrations? So probably Version030100...

lib/Controller/ApiController.php
```diff
@@ -1136,4 +1144,57 @@ public function exportSubmissionsToCloud(string $hash, string $path) {
...
+			$this->logger->debug('This form is not owned by the current user');
+			throw new OCSForbiddenException();
+		}
+		$this->exportSubmissionsToCloud($hash, $path);
+		$fileId = $this->submissionService->getFileId($path);
+		$form->setFileId($fileId);
+		$form->setFilePath($path);
+		$this->formMapper->update($form);
+		return new DataResponse($fileId);
+	}
```
Chartman123, line 1199: I'm not quite sure how the file operations are handled within Nextcloud, but I think that we must make sure that the given path is accessible and writeable by the user.

Chartman123 (changes requested):

lib/Controller/ApiController.php
hamza221: It is set from the file picker, So it is already being set by the user. If I got your comment correctly

Chartman123: Yes, but you can also directly call the API and then you have to make sure that the path exists and the user can access it.

jdaviescoates:
> they really create a Google Spreadsheet

That's what #5 is about too " Write responses into spreadsheet (like Google Forms) #5 "

That's the feature that is needed imho.

AIlkiv:
Hello!

1. What is blocking this pull request?
2. I believe it would be more practical to save it in ODT format, as CSV cannot edited in Office.
3. It would be beneficial to append to the file instead of overwriting it. For instance, the administrator could mark the answers as processed/not processed in a separate column.

Chartman123:
Hi @AIlkiv
> 1. What is blocking this pull request?
Probably time for working on it 
> 2. I believe it would be more practical to save it in ODT format, as CSV cannot edited in Office.
CSV can be imported in other spreadsheet applications and is the easiest to achieve as it's purely text based.
> 3. It would be beneficial to append to the file instead of overwriting it. For instance, the administrator could mark the answers as processed/not processed in a separate column.
I fully agree on this 👍🏻

PR #1407, opened 2022-11-12
add multiple options with one paste ##337

hamza221:


src/components/Questions/AnswerInput.vue
```diff
@@ -3,7 +3,7 @@
 		<div :is="pseudoIcon"
 			v-if="!isDropdown"
 			class="question__item__pseudoInput" />
-		<input ref="input"
+		<textarea ref="input"
```
Chartman123, line 6: The switch from input to textarea makes the input fields much higher and so the layout jumps a lot when you switch between edit and view.

Chartman123 (changes requested):
Besides the specific layout problem, I also get some errors when I paste a blank line or remove the text from an option sot that it's empty

src/components/Questions/AnswerInput.vue
hamza221: Would you suggest tinkering with css on the edit side? or you're against the use of textArea in the first place?
I'll check the blank line scenario and fix it

Chartman123: I'm not sure if pasting "new lines" to an input instead of to a textarea would allow the same result. Regarding the styling you wouldn't have to adjust it again to match the output... So yes I'd prefer the input over the textarea, but I'm not completely against the textarea here

hamza221: I don't think it's feasible with an input but I believe in the latest commit it's indistinguishable or at least seems to be.

jotoeri: Still looks completely different once clicking inside of the textarea.

- How about a modal here, too? Just keep everything as it is with simple inputs and add a modal(-component) to the actions-menu, that has one large textarea? Or even better the NcRichContenteditable then...

hamza221:
I pushed a quick POC of @jotoeri 's idea. If everyone agrees this is the way to go, I would appreciate some design advices.  from @jancborchardt  or anyone who has an input on the matter. 
Happy holidays everyone  🎄

Chartman123:
@hamza221 I like the modal 🎉  Especially the representation of the options below the input field. I think it might be useful to show a text that makes it clear that you have to use new lines for new options.

hamza221:
I only rebased and fixed conflicts, still not ready for review

Chartman123:
@susnux @hamza221 I did some rebasing and refactoring of the code. The modal/dialog itself works like expected. However the functionality and design still needs some more love. I think we could get this ready without too much effort for 4.3.

Chartman123 (dismissed):

jancborchardt (changes requested):
The bubbles below with the recognized options (that's what it is, right?) could be using the regular tag style like in Mail. :)

Chartman123:
> The bubbles below with the recognized options (that's what it is, right?) could be using the regular tag style like in Mail. :)

Yes, that's what these are for :) good idea 👍🏻

Chartman123:
@jancborchardt I think the NcSelect component fits our needs here as well and is less code to implement. I also added a loading indicator as it might take some time until all options get created on the server.

src/components/OptionInputModal.vue
```diff
@@ -0,0 +1,134 @@
...
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
```
susnux, line 23: Can we please have this file renamed to `...Dialog`? I think it reflects the used component better :)

src/components/OptionInputModal.vue
```diff
@@ -0,0 +1,134 @@
...
+			required: true,
+		},
+	},
+
+	emits: ['update:open'],
+
+	data() {
+		return {
+			enteredOptions: '',
+			buttons: [
```
susnux, line 75: The buttons are not changed, as such they do not need to be reactive. So instead of `data` move it to `computed`
(slightly performance improvement and better code style with Vue)

src/components/OptionInputModal.vue
```diff
@@ -0,0 +1,134 @@
...
+				},
+			],
+		}
+	},
+
+	computed: {
+		isDialogShown() {
+			return this.open
+		},
+
```
susnux, line 95:
```suggestion
```

src/components/OptionInputModal.vue
```diff
@@ -0,0 +1,134 @@
...
+	emits: ['update:open'],
+
+	data() {
+		return {
+			enteredOptions: '',
+			buttons: [
+				{
+					label: t('forms', 'Cancel'),
+					icon: IconCancel,
+					callback: () => { this.isDialogShown = false },
```
susnux, line 79:
```suggestion
					callback: () => { this.$emit('update:open', false) },
```

src/components/OptionInputModal.vue
```diff
@@ -0,0 +1,134 @@
...
+			return allOptions.filter(answer => { return answer.trim().length > 0 })
+		},
+	},
+
+	methods: {
+		t,
+
+		onMultipleOptions() {
+			this.isDialogShown = false
+			this.$nextTick(() => {
```
susnux, line 107: why next tick?

src/components/OptionInputModal.vue
```diff
@@ -0,0 +1,134 @@
...
+			const allOptions = this.enteredOptions.split(/\r?\n/g)
+			return allOptions.filter(answer => { return answer.trim().length > 0 })
+		},
+	},
+
+	methods: {
+		t,
+
+		onMultipleOptions() {
+			this.isDialogShown = false
```
susnux, line 106:
```suggestion
			this.$emit('update:open', false)
```

src/components/OptionInputModal.vue
```diff
@@ -0,0 +1,134 @@
...
+	},
+})
+</script>
+
+<style scoped>
+:deep(.options-modal) {
+	padding: 20px;
+}
+
+:deep(.v-select) {
```
susnux, line 127: why this custom styles for the select?

src/components/Questions/QuestionDropdown.vue
```diff
@@ -32,6 +32,13 @@
 				@update:checked="onShuffleOptionsChange">
 				{{ t('forms', 'Shuffle options') }}
 			</NcActionCheckbox>
+			<NcActionButton close-after-click
+				@click="isOptionModalOpen = true">
+				<template #icon>
+					<IconContentPaste :size="20" />
+				</template>
+				{{ t('forms', 'Multiple Options') }}
```
susnux, line 40:
```suggestion
				{{ t('forms', 'Add multiple options') }}
```

src/components/Questions/QuestionDropdown.vue
```diff
@@ -44,33 +51,46 @@
...
+							:aria-label="t('forms', 'Add a new answer')"
+							:placeholder="t('forms', 'Add a new answer')"
+							class="question__input"
+							:maxlength="maxStringLengths.optionText"
+							minlength="1"
+							type="text"
+							@input="addNewEntry">
+					</li>
+				</ol>
+			</template>
```
susnux, line 88:
```suggestion
			<div v-if="isLoading">
				<NcLoadingIcon :size="64" />
			</div>
			<ol v-else class="question__content">
				<!-- Answer text input edit -->
				<AnswerInput v-for="(answer, index) in options"
					:key="index /* using index to keep the same vnode after new answer creation */"
					ref="input"
					:answer="answer"
					:index="index"
					:is-unique="!isMultiple"
					:is-dropdown="true"
					:max-option-length="maxStringLengths.optionText"
					@delete="deleteOption"
					@update:answer="updateAnswer"
					@focus-next="focusNextInput"
					@tabbed-out="checkValidOption" />

				<li v-if="!isLastEmpty || hasNoAnswer" class="question__item">
					<input ref="pseudoInput"
						v-model="inputValue"
						:aria-label="t('forms', 'Add a new answer')"
						:placeholder="t('forms', 'Add a new answer')"
						class="question__input"
						:maxlength="maxStringLengths.optionText"
						minlength="1"
						type="text"
						@input="addNewEntry">
				</li>
			</ol>
```

src/components/OptionInputModal.vue
susnux: I think dialog matches better, but I am fine with both. But we should align all of them then (also #1606 )

src/components/OptionInputModal.vue
Chartman123: I think it was already in Hamza's original code

src/components/OptionInputModal.vue
Chartman123: To have it the whole width of the dialog and a slightly bigger margin on top of it.

jancborchardt (changes requested):
Looks very nice, just some details :)
- [x] "Multiple options" should have a lowercase "o" as per "Sentence case"
- [x] "Entered options" wording is a bit strange, maybe better "Options summary"
- [x] The "Cancel" button should not have an icon as it’s not primary.
- [x] The loading time on adding the options is quite long, can that be optimized somehow?

Chartman123:
> - [ ] The loading time on adding the options is quite long, can that be optimized somehow?

We already looked into it. The problem is that we can only submit one option per API request at the moment. So we could do three things:

- send all requests at basically the same time and await all responses to then store the server ID/order of the options to the front end. -> Options won't keep the original order from the input. (optional with checkbox?) 

- keep it that way and adjust API in the future to post several options in one request

- keep it that way and adjust API in the future to add the order to the post request

susnux:
> What about just "Options"?

:+1:

susnux:
> keep it that way and adjust API in the future to add the order to the post request

I think this is hard as setting the order requires to know the total amount of options which is hard to validate on the backend, e.g. if option 5 arrives before option 4.

> send all requests at basically the same time and await all responses to then store the server ID/order of the options to the front end. -> Options won't keep the original order from the input. (optional with checkbox?)

For the mentioned reason this is bad UX

> keep it that way and adjust API in the future to post several options in one request

I think this is the way to go

Chartman123:
@susnux ok, so we merge this and do a follow up for the new API version?

jancborchardt (approved):
Sounds good to me then, thanks for the explanation. :)

src/components/OptionInputModal.vue
Chartman123: @susnux Can we resolve this conversation or do you have a better solution for the styling?

src/components/OptionInputDialog.vue
```diff
@@ -0,0 +1,138 @@
...
+			// in case of only one option, just show an error message because it is probably missuse of the feature
+			showError(t('forms', 'Options should be seperated by new line!'))
+		},
+	},
+})
+</script>
+
+<style scoped>
+:deep(.options-modal) {
+	padding: 20px;
```
susnux, line 128: This makes it not centered (left will be more space then right.
I would prefer:

```suggestion
	padding-block: 0px 12px;
	padding-inline: 8px 20px;
```

susnux (approved):

PR #1408, opened 2022-11-13
Save a Single Question as csv

hamza221:


lib/Controller/ApiController.php
```diff
@@ -1096,6 +1096,39 @@ public function exportSubmissions(string $hash): DataDownloadResponse {
...
+	 * Export submissions of a specified Question
+	 *
+	 * @param int $id the question
+	 * @return DataDownloadResponse
+	 * @throws OCSBadRequestException
+	 * @throws OCSForbiddenException
+	 */
+	public function exportQuestion(int $questionId): DataDownloadResponse {
+		$this->logger->debug('Export submissions for form: {hash}', [
+			'hash' => $hash,
```
come-nc, line 1112: Where is $hash defined?

lib/Controller/ApiController.php
```diff
@@ -1096,6 +1096,39 @@ public function exportSubmissions(string $hash): DataDownloadResponse {
 		$csv = $this->submissionService->getSubmissionsCsv($hash);
 		return new DataDownloadResponse($csv['data'], $csv['fileName'], 'text/csv');
 	}
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 *
+	 * Export submissions of a specified Question
+	 *
+	 * @param int $id the question
```
come-nc, line 1105:
```suggestion
	 * @param int $questionId the question
```
This needs to match signature

lib/Controller/ApiController.php
```diff
@@ -1096,6 +1096,39 @@ public function exportSubmissions(string $hash): DataDownloadResponse {
...
+		try {
+			$question = $this->questionMapper->findById($questionId);
+			$formId = $question->getFormId();
+			$form = $this->formMapper->findById($formId);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find question');
+			throw new OCSBadRequestException();
+		}
+
+		if ($form->getOwnerId() !== $this->currentUser->getUID()) {
```
come-nc, line 1124: This should be some kind of permission check instead, no? Is that really only allowed for owner?

lib/Db/AnswerMapper.php
```diff
@@ -56,6 +56,28 @@ public function findBySubmission(int $submissionId): array {
...
 		return $this->findEntities($qb);
 	}
+	/**
+	 * @param int $submissionId
+	 * @param int $questionId
+	 * @throws \OCP\AppFramework\Db\DoesNotExistException if not found
+	 * @return Answer[]
+	 */
+
+	public function findBySubmissionAndQuestionId(int $submissionId, int $questionId): array {
```
come-nc, line 66:
```suggestion

	/**
	 * @param int $submissionId
	 * @param int $questionId
	 * @throws \OCP\AppFramework\Db\DoesNotExistException if not found
	 * @return Answer[]
	 */
	public function findBySubmissionAndQuestionId(int $submissionId, int $questionId): array {
```
nitpick

lib/Db/QuestionMapper.php
```diff
@@ -46,7 +46,7 @@ public function __construct(IDBConnection $db, OptionMapper $optionMapper) {
 	 * @return Question[]
 	 */
 
-	public function findByForm(int $formId, bool $loadDeleted = false): array {
+	public function findByForm(int $formId, bool $loadDeleted = false, int $questionId = -1): array|Question {
```
come-nc, line 49: Union types are not available in PHP 7.4. Is the application forcing PHP 8 already?

come-nc: I would also feel better if getting a question by id was its own method with a restrictive return type.

lib/Service/SubmissionService.php
```diff
@@ -264,6 +264,91 @@ public function getSubmissionsCsv(string $hash): array {
 			'data' => $this->array2csv($header, $data),
 		];
 	}
+	/**
+	 * Create Question specific CSV from Submissions to form
+	 * @param string $formId id of the form
+	 * @param string $questionId id of the question
+	 * @return array Array with 'fileName' and 'data'
```
come-nc, line 271:
```suggestion
	 * @return array{fileName:string,data:string} Array with 'fileName' and 'data'
```

lib/Db/QuestionMapper.php
hamza221: Mb, I agree creating a new method for getting question by Id would be better

lib/Controller/ApiController.php
hamza221: That was a mistake 😅, fixed

lib/Controller/ApiController.php
hamza221: done

hamza221: done

lib/Db/AnswerMapper.php
hamza221: ack

lib/Service/SubmissionService.php
hamza221: ack

lib/Db/QuestionMapper.php
hamza221: I didn't pay attention while making the first commit, a method already exists. I'm using it instead

jancborchardt:
And also as discussed, we should probably remove the (responses) part from the export file cause it can mess with import and translations. cc @Chartman123  @jotoeri

lib/Controller/ApiController.php
come-nc: But there is already a permission `PERMISSION_RESULTS` ?
But looking at the code I do not understand where/how these permissions are used :thinking:

come-nc:
Can you also add tests? Codecov is unhappy.

jotoeri:
Ok - sorry for kicking in here with a bad comment, but that should probably be discussed before putting too much more work into this:
Where is the benefit of having a single question exported as csv? Csv is a pure data representation, not being intended for direct reading by a user. As such, it will in 90% of the cases anyways be used with a certain data-handling / table-calculation software or alike, so the effort of deleting unused questions from the data is not very much.
So overall i currently don't see the benefit of having only one question on a pure data-export, while coding-wise and UI-wise it just adds more complexity. As i always have the 'Simple forms app'(!) in mind, i would currently vote against adding a feature with such less really new functionality. But i'm open to get convinced if there is a real need for this. :wink: 

(An export of the single questions summary-graphics would e.g. be a different thing! Then one could see the benefit of quickly having a human-readable representation of the summary.)

gerazo:
I fully agree from the technical perspective with you @jotoeri to keep things simple. However, NC was recently started to be used extensively as a complete backoffice solution for no-code office process management. This is very good news, because this means that it can completely replace the other, proprietary online offices, but it also means that some functionality cannot be simple unfortunately. The controlled data collection, surveying inside a company via Forms is also a crucial feature in othe office solutions. So unfortunately, this contradicts with you "simple forms" vision.
What I can tell from my experience is that it is quite common that the form creator monitors the incoming answers as they come and continuously manages partial data in her/his own calc sheet. This is definitely made much easier if getting an answer and pasting it is just two clicks instead of copying several fields by hand. This feature would help this case.

jancborchardt:
@gerazo note that right now this pull request implements the export of a single **Question** and its answers. But I think what you are referring to is the exporting of a single **Response**, correct?

@hamza221 and I talked about it and yeah, it does make sense. @jotoeri @Chartman123 what do you think?

Exporting question like via this PR | Where exporting individual responses could go
-|-
image|image

appinfo/routes.php
```diff
@@ -245,6 +245,22 @@
 				'apiVersion' => 'v2(\.1)?'
 			]
 		],
+		[
+			'name' => 'api#exportQuestion',
+			'url' => '/api/{apiVersion}/submissions/exportQuestion/{questionId}',
+			'verb' => 'GET',
+			'requirements' => [
+				'apiVersion' => 'v2'
```
Chartman123, line 253:
```suggestion
				'apiVersion' => 'v2.2'
```
Should only be available on 2.2 :)

appinfo/routes.php
```diff
@@ -245,6 +245,22 @@
...
+			]
+		],
+		[
+			'name' => 'api#exportSubmission',
+			'url' => '/api/{apiVersion}/submissions/exportSubmission/{submissionId}',
+			'verb' => 'GET',
+			'requirements' => [
+				'apiVersion' => 'v2'
```
Chartman123, line 261:
```suggestion
				'apiVersion' => 'v2.2'
```
Same here

lib/Controller/ApiController.php
```diff
@@ -1143,7 +1143,72 @@ public function exportSubmissions(string $hash): DataDownloadResponse {
...
```
Chartman123, line 1171: Shouldn't this be checked against the results permission, too?

Chartman123 (changes requested):
I've added some comments. Please add the new API endpoints to the documentation, too :)

And currently one can only download the CSV file. I think it would be nice to be able to save it to the cloud just like for all submissions.

PR #1416, opened 2022-11-26
transfer ownership of a form

hamza221:
image

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -128,6 +128,14 @@ return [
 			]
 		],
 		[
+			'name' => 'api#transferOwner',
+			'url' => '/api/{apiVersion}/form/transfer',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2.2'
+			]
+		],
+		[
 			'name' => 'api#deleteForm',
 			'url' => '/api/{apiVersion}/form/{id}',
 			'verb' => 'DELETE',
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -51,6 +51,7 @@ use OCP\AppFramework\OCS\OCSBadRequestException;
 use OCP\AppFramework\OCS\OCSException;
 use OCP\AppFramework\OCS\OCSForbiddenException;
 use OCP\AppFramework\OCSController;
+use OCP\Files\NotFoundException;
 use OCP\Files\NotPermittedException;
 use OCP\IL10N;
 use OCP\IRequest;
@@ -349,6 +350,50 @@ class ApiController extends OCSController {
 		$this->formsService->setLastUpdatedTimestamp($id);
 
 		return new DataResponse($form->getId());
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * Transfer ownership of a form to another user
+	 *
+	 * @param int $formId id of the form to update
+	 * @param string $uid id of the new owner
+	 * @return DataResponse
+	 * @throws OCSBadRequestException
+	 * @throws OCSForbiddenException
+	 */
+	public function transferOwner(int $formId, string $uid): DataResponse {
+		$this->logger->debug('Updating owner: formId: {formId}, userId: {uid}', [
+			'formId' => $formId,
+			'uid' => $uid
+		]);
+
+		try {
+			$form = $this->formMapper->findById($formId);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
+			throw new NotFoundException('Could not find form');
+		}
+
+		$user = $this->userManager->get($uid);
+		if($user == null) {
+			$this->logger->debug('Could not find new form owner');
+			throw new OCSBadRequestException('Could not find new form owner');
+		}
+
+		if ($form->getOwnerId() !== $this->currentUser->getUID()) {
+			$this->logger->debug('This form is not owned by the current user');
+			throw new OCSForbiddenException('This form is not owned by the current user');
+		}
+
+		// update form owner
+		$form->setOwnerId($uid);
+
+		// Update changed Columns in Db.
+		$this->formMapper->update($form);
+		
+		return new DataResponse($form->getOwnerId());
 	}
 
 	/**
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -230,10 +230,12 @@ export default {
 
 	mounted() {
 		subscribe('forms:last-updated:set', (id) => this.onLastUpdatedByEventBus(id))
+		subscribe('forms:ownership-transfered', (id) => this.onDeleteForm(id))
 	},
 
 	unmounted() {
 		unsubscribe('forms:last-updated:set', (id) => this.onLastUpdatedByEventBus(id))
+		unsubscribe('forms:ownership-transfered', (id) => this.onDeleteForm(id))
 	},
 
 	methods: {
--- a/src/components/SidebarTabs/TransferOwnership.vue
+++ b/src/components/SidebarTabs/TransferOwnership.vue
@@ -1,0 +1,182 @@
+<!--
+  - @copyright Copyright (c) 2022 Hamza Mahjoubi <hamzamahjoubi221@gmail.com>
+  -
+  - @author Hamza Mahjoubi <hamzamahjoubi221@gmail.com>
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div>
+		<NcButton class="transfer-button"
+			alignment="start"
+			type="tertiary"
+			:wide="true"
+			@click="openModal">
+			<span class="transfer-button__text">{{ t('forms', 'Transfer ownership') }}</span>
+		</NcButton>
+
+		<NcDialog :open.sync="showModal"
+			content-classes="modal-content"
+			:name="t('forms', 'Transfer ownership')"
+			:out-transition="true"
+			@close="closeModal">
+			<template #default>
+				<!-- eslint-disable-next-line vue/no-v-html -->
+				<p v-html="t('forms', 'You\'re going to transfer the ownership of {name} to another account. Please select the account to which you want to transfer ownership.', { name: `<strong>${form.title}</strong>` }, undefined, { escape: false })" />
+				<NcSelect v-model="selected"
+					class="modal-content__select"
+					:reset-focus-on-options-change="false"
+					:clear-search-on-select="true"
+					:close-on-select="true"
+					:loading="loading"
+					:get-option-key="(option) => option.key"
+					:options="options"
+					:placeholder="t('forms', 'Search for a user')"
+					:user-select="true"
+					label="displayName"
+					@search="(query)=>asyncSearch(query,true)">
+					<template #no-options>
+						{{ noResultText }}
+					</template>
+				</NcSelect>
+
+				<br>
+
+				<!-- eslint-disable-next-line vue/no-v-html -->
+				<p v-html="t('forms', 'Type {text} to confirm.', { text: `<strong>${confirmationString}</strong>` }, undefined, { escape: false })" />
+				<NcTextField :label="t('forms', 'Confirmation text')"
+					:value.sync="confirmationInput"
+					:success="confirmationInput === confirmationString" />
+
+				<br>
+
+				<p><strong>{{ t('forms','This can not be undone.') }}</strong></p>
+			</template>
+			<template #actions>
+				<NcButton :disabled="!canTransfer" type="error" @click="onOwnershipTransfer">
+					{{ t('forms', 'I understand, transfer this form') }}
+				</NcButton>
+			</template>
+		</NcDialog>
+	</div>
+</template>
+
+<script>
+import { showSuccess, showError } from '@nextcloud/dialogs'
+import { emit } from '@nextcloud/event-bus'
+import { generateOcsUrl } from '@nextcloud/router'
+import axios from '@nextcloud/axios'
+import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
+import NcDialog from '@nextcloud/vue/dist/Components/NcDialog.js'
+import NcTextField from '@nextcloud/vue/dist/Components/NcTextField.js'
+import NcSelect from '@nextcloud/vue/dist/Components/NcSelect.js'
+import UserSearchMixin from '../../mixins/UserSearchMixin.js'
+import logger from '../../utils/Logger.js'
+
+export default {
+	components: {
+		NcButton,
+		NcDialog,
+		NcTextField,
+		NcSelect,
+	},
+	mixins: [UserSearchMixin],
+
+	props: {
+		form: {
+			type: Object,
+			required: true,
+		},
+	},
+
+	data() {
+		return {
+			selected: null,
+			showModal: false,
+			confirmationInput: '',
+			loading: false,
+		}
+	},
+	computed: {
+		canTransfer() {
+			return this.confirmationInput === this.confirmationString && !!this.selected
+		},
+		confirmationString() {
+			return `${this.form.ownerId}/${this.form.title}`
+		},
+		options() {
+			if (this.isValidQuery) {
+				// Suggestions without existing shares
+				return this.suggestions
+			}
+			// Recommendations without existing shares
+			return this.recommendations
+		},
+	},
+	methods: {
+		clearText() {
+			this.confirmationInput = ''
+		},
+		closeModal() {
+			this.showModal = false
+		},
+		openModal() {
+			this.showModal = true
+		},
+		async onOwnershipTransfer() {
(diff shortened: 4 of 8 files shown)


src/Forms.vue
```diff
@@ -96,17 +96,37 @@
 				:form="selectedForm"
 				:opened.sync="sidebarOpened"
 				:active.sync="sidebarActive"
-				name="sidebar" />
+				name="sidebar"
+				@transfer:ownership="openModal" />
 		</template>
+		<NcModal v-if="modal"
+			size="normal"
+			title="Transfer Ownership"
```
Chartman123, line 104: ownership should not be capitalized here

src/Forms.vue
```diff
@@ -96,17 +96,37 @@
...
+			name="NcModal"
+			:outTransition="true"
+			@close="closeModal">
+			<div class="modal__content">
+				{{ t('forms', 'Are you sure you want to tranfer Ownership of this Form to ') }} {{ tranferData.displayName }}?
```
Chartman123, line 109: ownership should not be capitalized here

Chartman123 (changes requested):
I get an error when I actually transfer the form to another user. Probably because you're staying in the same route after the transfer and therefore on a form for which the current user doesn't have edit permissions anymore.

Second, a little loading indicator would be nice between selecting a user and showing the modal. It takes some time (around 3 to 5 seconds for me) between choosing the user and the modal being shown.

src/Forms.vue
hamza221: Can you please the console output when opening the modal, there's a vue warn that I couldn't solve. Maybe you've seen it before and can help me figure out the cause

jancborchardt (changes requested):
Nice! Some design details:
- [x] This setting should be all the way on the bottom of the settings, since it’s the least common one
- [x] It would be better as a Button than a switch
- [x] The label "Transfer ownership" is clearer
- [x] I’d expect the modal to open on button press also for the name input directly, like GitHub does on repository transfers. Then we also only have one modal
   image
- [x] The modal needs a heading, like "Transfer [formname]"
- [x] The buttons are wrongly colored and labeled – "Yes/No" is too inconcrete. Similar to Github, we only need a confirmation button, which is red, and labeled "Transfer [formname] to [username]". For canceling we have the x in the top right
- [ ] The words "ownership", "form" and "cancelled" need to be lowercase when part of a sentence

hamza221:
This last commit addresses all the mentioned issues but loading because I didn't find a straightforward way to address it since the event is triggered in one component "SettingsSidebarTab" and handled in another one "Forms".

I'll get back to it. But currently writing my thesis ( I Have to submit it on the 10th) so I'm trying to gain time by getting all PRs to a state where I can write documentation about them.  and I'll get all of them to a merging state after the 10th.

jotoeri:
Can you create a separate component for the transfer? Should be good, if it just contains the Button and the Modal, including all corresponding procedures. Then you can just load the new component on the sidebar.

> But currently writing my thesis

Good luck! 🤞 💪

hamza221:
Per @jotoeri  recommendation I created a separate component for ownership transfer. I'll add the missing tests later this week

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,176 @@
...
+	methods: {
+		onChangeOwner() {
+			this.openModal()
+		},
+		clearText() {
+			this.confirmation = ''
+		},
+		closeModal() {
+			this.modal = false
+			showError(t('forms', 'Ownership transfer was Cancelled'))
```
Chartman123, line 117:
```suggestion
			showError(t('forms', 'Ownership transfer was cancelled'))
```
Lower case "c" :)

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,176 @@
...
+					<div v-if="transferData.displayName.length > 0" class="selected_user">
+						<p>{{ transferData.displayName }}</p>
+						<NcButton type="tertiary-no-background" @click="clearSelected">
+							X
+						</NcButton>
+					</div>
+				</div>
+				<div class="modal_section">
+					<p class="modal_text">
+						{{ t('forms', 'Type ') }} {{ confirmationString }} {{ t('forms', ' to confirm') }}
```
Chartman123, line 53:
```suggestion
						{{ t('forms', 'Type ') }} <strong>{{ confirmationString }}</strong> {{ t('forms', ' to confirm') }}
```
I think we should highlight the text that the user has to type more.

Chartman123:
@hamza221 The current solution works much smoother than before. Nice job 👍🏻

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,176 @@
...
+}
+
+.modal_section {
+    margin-bottom: 20px;
+}
+
+.selected_user {
+    display: flex;
+    align-items: center;
+}
```
jotoeri, line 175: Just CSS-wise, you could maybe try to get an idea how it is done in the form-header within submit.vue? There it works without sections.

src/views/Sidebar.vue
```diff
@@ -94,6 +95,9 @@ export default {
 		onUpdateActive(active) {
 			this.$emit('update:active', active)
 		},
+		onOwnershipTransfer(share) {
+			this.$emit('transfer:ownership', share, this.form.id)
+		},
```
jotoeri, line 100: Dead code?

jotoeri (commented):
Nice @hamza221 :) I still have a bunch of comments here, but that's more since i now commented many small things. Design-wise and css-wise there is also a bit to do still, but maybe @jancborchardt can also help you on this. But in general, it feels like this is on a good way now. :)

A rebase (& squash) would also be good, when you're on it.

appinfo/routes.php
```diff
@@ -131,6 +131,14 @@
 				'apiVersion' => 'v2'
 			]
 		],
+		[
+			'name' => 'api#ownerTransfer',
+			'url' => '/api/{apiVersion}/form/transfer',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2'
```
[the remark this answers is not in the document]
jotoeri: Ahh, and adding it to the documentation would be good then. But that can be a very last thing. 😉

src/components/SidebarTabs/TransferOwnership.vue
jotoeri: And if possible, make it able to copy it. 😋

jospoortvliet:
Hi, just curious - what's still on the todo here? Does it 'just' need another review & approval?

Chartman123:
@jospoortvliet I'll have another look at it :)

Chartman123:
@hamza221 @jospoortvliet I think there is still a little work to do. On loading the modal the input box has a little glitch:
grafik

When you closse the recommendation by clicking somewhere else in the modal and then click back into the user selection, everything looks fine.

Another thing that I noticed: When you select a user, the input field should lose the focus so that you don't have to click anywhere else to close the lower part.

And the button to start the transfer only reacts to the "security" input of the form name. It doesn't matter if you have selected a user or not and then of course throws an error in the console when you didn't select a user first.

jospoortvliet:
awesome! Now, with those fixed, what's missing? The reviews are all implemented, correct? So a final thumbs up and merge?

lib/Controller/ApiController.php
```diff
@@ -405,6 +405,44 @@ public function updateForm(int $id, array $keyValuePairs): DataResponse {
...
+	 *
+	 * Transfer ownership of a form to another user
+	 *
+	 * @param int $formId id of the form to update
+	 * @param string $uid id of the new owner
+	 * @return DataResponse
+	 * @throws OCSBadRequestException
+	 * @throws OCSForbiddenException
+	 */
+	public function ownerTransfer(int $formId, string $uid): DataResponse {
```
Chartman123, line 419:
```suggestion
	public function transferOwner(int $formId, string $uid): DataResponse {
```
I think we should change the name here

appinfo/routes.php
```diff
@@ -117,6 +117,14 @@
 				'apiVersion' => 'v2(\.1)?'
 			]
 		],
+		[
+			'name' => 'api#ownerTransfer',
```
Chartman123, line 121:
```suggestion
			'name' => 'api#transferOwner',
```

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,176 @@
...
+		form: {
+			type: Object,
+			required: true,
+		},
+	},
+
+	data() {
+		return {
+			modal: false,
+			transferData: { formId: null, userId: null, displayName: '' },
```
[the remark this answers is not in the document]
Chartman123: This is still open

Chartman123 (changes requested):
There are still some Lint errors. I would also like to just have another look at the code (mostly naming related).

hamza221:
still working on the changes will push my changes in a bit the previous commit was just a rebase

hamza221:
I couldn't fix the input glitch, @Chartman123  do you think it might be a bug in NcSelect itself?

Chartman123:
It might be related to the order of the rendering. Like if the input field were rendered before the padding of the modal was applied and then the dropdown part were not re-rendered.

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,193 @@
...
+						<template #no-options>
+							{{ noResultText }}
+						</template>
+					</NcSelect>
+					<div v-if="!!selected" class="modal-content__user">
+						<p>{{ selected.displayName }}</p>
+						<NcButton type="tertiary-no-background" @click="clearSelected">
+							X
+						</NcButton>
+					</div>
```
Chartman123, line 61: Collectives has a nice implementation for the selected users:
https://github.com/nextcloud/collectives/blob/5a69647aef5c537ae3056c1225fa0cd588591918/src/components/Member/MemberPicker.vue#L19-L33

https://nextcloud-vue-components.netlify.app/#/Components/NcUserBubble

appinfo/routes.php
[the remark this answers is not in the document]
Chartman123: Documentation is still missing :)

src/components/SidebarTabs/TransferOwnership.vue
Chartman123: @hamza221 just another thought about it: do we really need the NcUserBubble? We could also just use NcSelect like in the example here: https://nextcloud-vue-components.netlify.app/#/Components/NcSelect?id=ncselect-1

hamza221: It's already working like the example, I can see how NcUserBubble is unnecessary in this context, I'll just remove it

hamza221:
are the failing PHPUnit  related? it show failed 
One or more containers failed to start.

Chartman123:
> are the failing PHPUnit  related? it shows: 
> One or more containers failed to start.

At least the OCI job fails more often. For the postgres job, it's the first time I see it failing. I'll try a re-run of the failed jobs

Chartman123 (changes requested):
I've found a few more things that you could please work on. But overall I think that this is now close to being ready. Please also add the documentation of the API.

The "glitch" with the NcSelect list is now much better but sometimes I still see that the lower part is a little wider than the input field itself.

Chartman123 (commented):
A few more comments. The animation problem seems to be fixed now, too. But the modal changes size when you have selected a user and click back into the field. The cursor looks misplaced, too.

[forms.webm](https://github.com/nextcloud/forms/assets/16020496/039a026a-d66a-4916-a3f2-976790b978b0)

hamza221:
> A few more comments. The animation problem seems to be fixed now, too. But the modal changes size when you have selected a user and click back into the field. The cursor looks misplaced, too.
I can't reproduce after the nextcloud/vue bump , @Chartman123  does it still happen on your end?

Chartman123:
> I can't reproduce after the nextcloud/vue bump , @Chartman123 does it still happen on your end?

@hamza221 seems fixed upstream, yes 👍🏻  no problems here anymore

Chartman123:
`could not invoke event listener TypeError: this.sharedForms[sharedFormIndex] is undefined`

I get the above error after clicking the transfer button in the modal.

src/views/Results.vue
```diff
@@ -168,7 +168,6 @@ import PermissionTypes from '../mixins/PermissionTypes.js'
 
 const picker = getFilePickerBuilder(t('forms', 'Save CSV to Files'))
 	.setMultiSelect(false)
-	.setModal(true)
```
hamza221, line 171: @susnux  [This is deprecated](https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/911), And I can't build the project with it set.
Should we remove it in this Pr?

hamza221: split in a new Pr https://github.com/nextcloud/forms/pull/1826

Chartman123 (approved):

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,200 @@
...
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div>
+		<NcButton @click="openModal">
+			{{ t('forms', 'Transfer ownership') }}
+		</NcButton>
+		<NcModal :show.sync="showModal"
```
susnux, line 28: I would really recommend `NcDialog`.

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,200 @@
...
+			:title="'Transfer ' + form.title"
+			:out-transition="true"
+			@close="closeModal">
+			<div class="modal-content">
+				<h2 class="modal-content__heading">
+					{{ t('forms', 'Transfer ') }} {{ form.title }}
+				</h2>
+				<div class="modal-content__section">
+					<p class="modal-content__text">
+						{{ t('forms', 'Please select the account to which you want to transfer ownership"') }}
```
susnux, line 38:
```suggestion
						{{ t('forms', 'Please select the account to which you want to transfer ownership') }}
```

Extra quote

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,200 @@
...
+				</div>
+				<div class="modal-content__section">
+					<p class="modal-content__text">
+						{{ t('forms', 'Type ') }} <strong> {{ confirmationString }} </strong>{{ t('forms', ' to confirm') }}
+					</p>
+					<NcTextField :value.sync="confirmationInput"
+						:success="confirmationInput === confirmationString" />
+				</div>
+				<NcButton :disabled="!canTransfer" type="error" @click="onOwnershipTransfer">
+					{{ t('forms', 'I understand, transfer this form') }}
```
susnux, line 65: Not sure about this sentence here, for github there is a warning before so that you confirm the warning and notes.
I think it might make sense to add text before this like "This can not be undone" or similar.

susnux (changes requested):
Nice works! But a few comments.

hamza221:
image

src/components/SidebarTabs/TransferOwnership.vue
Chartman123: > "This can not be undone"

I'm not sure if we really need this here. The action can be undone when the new owner transfers ownership back to the current owner.

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,200 @@
...
+			{{ t('forms', 'Transfer ownership') }}
+		</NcButton>
+		<NcModal :show.sync="showModal"
+			:title="'Transfer ' + form.title"
+			:out-transition="true"
+			@close="closeModal">
+			<div class="modal-content">
+				<h2 class="modal-content__heading">
+					{{ t('forms', 'Transfer ') }} {{ form.title }}
+				</h2>
```
Chartman123, line 35: Now that the NcDialog already has this in the title, I would omit it here.

src/components/SidebarTabs/TransferOwnership.vue
```diff
@@ -0,0 +1,200 @@
...
+				<div class="modal-content__section">
+					<p class="modal-content__text">
+						{{ t('forms', 'Type ') }} <strong> {{ confirmationString }} </strong>{{ t('forms', ' to confirm') }}
+					</p>
+					<NcTextField :value.sync="confirmationInput"
+						:success="confirmationInput === confirmationString" />
+				</div>
+				<NcButton :disabled="!canTransfer" type="error" @click="onOwnershipTransfer">
+					{{ t('forms', 'I understand, transfer this form') }}
+				</NcButton>
```
Chartman123, line 69: Could you please move this button to the #actions slot?

hamza221: It's already there

susnux (approved):

appinfo/routes.php
```diff
@@ -127,6 +127,14 @@
 				'apiVersion' => 'v2.2'
 			]
 		],
+		[
+			'name' => 'api#transferOwner',
```
susnux, line 131: The only thing I do not know if this is needed and we not simply can use the update endpoint and change the owner?

Chartman123 (approved):

appinfo/routes.php
Chartman123: I'd keep it separate for now. And I'll create a issue as follow up that we need to add a test for this

hbiering:
I tried to transfer ownership from myself (as admin) to a group. The Group-name was found when I started typing it in the upper field - and the transfer option was enabled when I entered the required (bold) phrase in the second field. But when continuing, I got the error message: "An error occurred while transfering ownership"

Is the transfer option intended to support transfer to User Groups (as the group shows up when you start to type) or is it limited to individual users. Although you can share final results in a spreadsheet the latter is a problem for collaboration in teams.

susnux:
Ownership is bound to users so the bug is that the search does not limit to users.

susnux:
Ownership is bound to users so the bug is that the search does not limit to users.

PR #1417, opened 2022-11-30
Added the possibility to collaborate on forms

hamza221:
This is the first iteration, real-time collaboration is not yet set 

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -131,6 +131,14 @@ return [
 				'apiVersion' => 'v2'
 			]
 		],
+		[
+			'name' => 'api#getCollaborationForms',
+			'url' => '/api/{apiVersion}/collaboration_forms',
+			'verb' => 'GET',
+			'requirements' => [
+				'apiVersion' => 'v2'
+			]
+		],
 
 		// Questions
 		[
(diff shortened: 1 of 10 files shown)


hamza221:
Version3000Date20221127191108 This migration should be run before testing these changes
Navigate to /server  and run `php occ  migrations:migrate 3000Date20221127191108`

susnux:
I am not really sure here. First I agree that concurrent editing is not on the schedule and out of scope.

But I rather think we should change the editing a bit:
1. check user has edit permissions
2. acquire edit lock on server (needed to prevent collisions)
3. submit changes
4. release lock

This would required two columns I think (`locked_by` and `locked_since`).
The `locked_by` defines the current user holding the lock (`null` if not locked) and `locked_since` is used to release the lock after a defined IDLE time to not deadlock editing (e.g. 5 minutes without activity).

What do you think? Also cc @Chartman123

Chartman123:
Yes, some kind of locking would be necessary. I'd just use `locked_until` instead of `locked_since` but in the end it's the same. Instead of using a timeout for locking we could also implement a method to manually unlock a form (perhaps only for the owner?)

susnux:
> Instead of using a timeout for locking we could also implement a method to manually unlock a form (perhaps only for the owner?)

I think we need the timeout, but could do both. Otherwise we might end up in a lot of situations where the owner is needed to unlock.

gwnmikedus:
Hi! This commit is getting quite old. It’s a rly wanted feature to be able to edit the form without having to transfer ownership to every person who needs to make an edit. Is there any ETA for this PR to be be merged?

Chartman123:
No, as you can see there's no progress. Other than that, please don't just comment asking for ETAs on PRs. If at all, use the corresponding issues.

PR #1428, opened 2022-12-23
Update workflows

jotoeri:
- Remove php 7.3
- Update checkout action
- Update mariadb version
- Update Node-Versions

Signed-off-by: Jonas Rittershofer <jotoeri@users.noreply.github.com>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,12 +13,12 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ["7.3", "7.4", "8.0", "8.1"]
+        php-versions: ["7.4", "8.0", "8.1"]
 
-    name: php${{ matrix.php-versions }}
+    name: php-lint (${{ matrix.php-versions }})
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up php ${{ matrix.php-versions }}
         uses: shivammathur/setup-php@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -13,12 +13,12 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ["7.3", "7.4", "8.0", "8.1"]
+        php-versions: ["7.4", "8.0", "8.1"]
```
susnux, line 16: Maybe add 8.2 to the matrix? So we can test for https://github.com/nextcloud/server/issues/32595

jotoeri: I had that at first... But currently it still fails, as long as the PR is not merged. The interesting question is, how long it will still take, until that gets merged...? I didn't dig deeper, but currently i don't see a clear timeframe to get this in on server.

Chartman123 (approved):

.github/workflows/lint.yml
susnux: NC26 should support 8.2 but there is still this version check in master, so we have to wait :/
https://github.com/nextcloud/server/blob/d635d58d19d5ab65c0be754fc32fce99672c249f/lib/versioncheck.php#L37

susnux (approved):

.github/workflows/lint.yml
ChristophWurst: Unless … https://github.com/nextcloud/mail/blob/ea03dc172ca9a8006a368e09e25c666925a9d597/.github/workflows/test.yml#L130-L132 :smiling_imp:

Issue #936, opened 2021-05-12
Lock form 🔒


sbernard31:
- Nextcloud-Version: 19.0.4
- Forms-Version: 2.2.2

When I finished a form, I would like to avoid to modify it unintentionally.

A solution could be to lock it to forbid modification. (We should be able to unlock it later)

An alternative solution, could be to be able to tag the state and have a kind of versioning.  (but probably more expensive than just a simple lock)

susnux:
I think this is required for #322

Issue #596, opened 2020-09-08
Participation limitation


Supergamerx3000:
**Nextcloud
- Nextcloud-Version: 18.0.8
- Forms-Version: 2.0.4

**Is your feature request related to a problem?
I would like to create a registration form that should expire automatically on a certain date (that works) but I would also like to set a number of participants, e.g. 80 participations and then the form should be closed automatically. That no one else can register.

skleber:
+1

The same thing would be interesting for me too.

mstoss:
> but I would also like to set a number of participants, e.g. 80 participations and then the form should be closed automatically. That no one else can register.

It would be a great enhancement !
Thank you

Finki008:
Is there an update on this? Can I somehow support?

Chartman123:
> Is there an update on this? Can I somehow support?

@Finki008 it's not currently on our list of features we're working on. But if you like you can give it a try and start coding :)

cedricsi35:
Hi, 

> > Is there an update on this? Can I somehow support?
> 
> [@Finki008](https://github.com/Finki008) it's not currently on our list of features we're working on. But if you like you can give it a try and start coding :)

I am also interested in this feature for registering participants in workshops with limited places.

PR #1432, opened 2022-12-30
Show relative expiration date inside form

### LINKED ISSUE #1431: End Date visible to participants
- Nextcloud-Version: 25
- Forms-Version: 3.02

We use forms for sign up to events and though we communicate the last possible sign up date, people often ask because the set enddate is only visible to the form creator.

Show the Enddate (and time) to all users who open the form (probably best to make this optional).

A simple workaround is to write it in the description, but you can forget it, either from the start or when you change the enddate.

It's by no means anything important but it would be a nice quality of life improvement.

Thx and best regards.

Chartman123:
Fixes #1431

- [x] Show expiration date in Submit.vue
- [x] Show expiration date in Create.vue
- [x] Make it configurable in the settings of the form
- [x] Adjust documentation

grafik

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -266,6 +266,7 @@ class ApiController extends OCSController {
 			'showToAllUsers' => false,
 		]);
 		$form->setSubmitMultiple(false);
+		$form->setShowExpiration(false);
 
 		$this->formMapper->insert($form);
 
--- a/lib/Db/Form.php
+++ b/lib/Db/Form.php
@@ -48,6 +48,8 @@ use OCP\AppFramework\Db\Entity;
  * @method void setIsAnonymous(bool $value)
  * @method integer getSubmitMultiple()
  * @method void setSubmitMultiple(bool $value)
+ * @method integer getShowExpiration()
+ * @method void setShowExpiration(bool $value)
  */
 class Form extends Entity {
 	protected $hash;
@@ -59,6 +61,7 @@ class Form extends Entity {
 	protected $expires;
 	protected $isAnonymous;
 	protected $submitMultiple;
+	protected $showExpiration;
 
 	/**
 	 * Form constructor.
@@ -68,6 +71,7 @@ class Form extends Entity {
 		$this->addType('expires', 'integer');
 		$this->addType('isAnonymous', 'bool');
 		$this->addType('submitMultiple', 'bool');
+		$this->addType('showExpiration', 'bool');
 	}
 
 	// JSON-Decoding of access-column.
@@ -92,7 +96,8 @@ class Form extends Entity {
 			'access' => $this->getAccess(),
 			'expires' => $this->getExpires(),
 			'isAnonymous' => $this->getIsAnonymous(),
-			'submitMultiple' => $this->getSubmitMultiple()
+			'submitMultiple' => $this->getSubmitMultiple(),
+			'showExpiration' => $this->getShowExpiration()
 		];
 	}
 }
--- a/src/components/SidebarTabs/SettingsSidebarTab.vue
+++ b/src/components/SidebarTabs/SettingsSidebarTab.vue
@@ -36,14 +36,13 @@
 			@update:checked="onSubmitMultipleChange">
 			{{ t('forms', 'Allow multiple responses per person') }}
 		</NcCheckboxRadioSwitch>
-		<div>
-			<NcCheckboxRadioSwitch :checked="formExpires"
-				type="switch"
-				@update:checked="onFormExpiresChange">
-				{{ t('forms', 'Set expiration date') }}
-			</NcCheckboxRadioSwitch>
-			<NcDatetimePicker v-show="formExpires"
-				id="expiresDatetimePicker"
+		<NcCheckboxRadioSwitch :checked="formExpires"
+			type="switch"
+			@update:checked="onFormExpiresChange">
+			{{ t('forms', 'Set expiration date') }}
+		</NcCheckboxRadioSwitch>
+		<div v-show="formExpires" class="settings-div--indent">
+			<NcDatetimePicker id="expiresDatetimePicker"
 				:clearable="false"
 				:disabled-date="notBeforeToday"
 				:disabled-time="notBeforeNow"
@@ -54,6 +53,11 @@
 				:value="expirationDate"
 				type="datetime"
 				@change="onExpirationDateChange" />
+			<NcCheckboxRadioSwitch :checked="form.showExpiration"
+				type="switch"
+				@update:checked="onShowExpirationChange">
+				{{ t('forms', 'Show expiration date on form') }}
+			</NcCheckboxRadioSwitch>
 		</div>
 	</div>
 </template>
@@ -142,6 +146,9 @@ export default {
 				this.$emit('update:formProp', 'expires', 0)
 			}
 		},
+		onShowExpirationChange(checked) {
+			this.$emit('update:formProp', 'showExpiration', checked)
+		},
 
 		/**
 		 * On date picker change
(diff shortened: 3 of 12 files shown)


jancborchardt:
@Chartman123 can you include a screenshot for quick design review? :)

I would say we do not need a setting for this – if there is an expiration date, potential respondents should always be made aware of it.

Chartman123:
> screenshot for quick design review? :)

@jancborchardt Yes, for sure 👍🏻  Added it to the first post.

Chartman123:
@jotoeri I added the text for expired forms as well as this will be shown for the owner. Now I have the same method for AppNavigationForms.vue and Submit.vue. Should this be moved to a new mixin to avoid the duplicated code?

susnux:
> > I would say we do not need a setting for this – if there is an expiration date, potential respondents should always be made aware of it.
> 
> If the form is only internally shared to the instance/other users, they can see the expiration date already in the AppNavigation. So in these cases I'd prefer to have it configurable to avoid redundance...

I agree with that, and maybe there are use cases where you do not want users to see the limit, e.g. later if we implement #1432 you might not want users to see the limit / current number of "registrations".

lib/Db/Form.php
```diff
@@ -59,6 +61,7 @@ class Form extends Entity {
 	protected $expires;
 	protected $isAnonymous;
 	protected $submitMultiple;
+	protected $showExpiration;
```
susnux, line 64: Maybe we should introduce `extraSettings` as for `Db/Questions` instead?

lib/Controller/ApiController.php
```diff
@@ -266,6 +266,7 @@ public function newForm(): DataResponse {
 			'showToAllUsers' => false,
 		]);
 		$form->setSubmitMultiple(false);
+		$form->setShowExpiration(false);
```
susnux, line 269: Can we add this to the API test?

lib/Db/Form.php
Chartman123: I had a discussion about this with @jotoeri and we agreed on having a column in this case :)

susnux: Ok because I can think of at least this two other settings we might add in the (near?) future:
* `submissionLimit` (for #596 )
* `locked` (for #936 )

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -55,6 +55,13 @@
 				type="datetime"
 				@change="onExpirationDateChange" />
 		</div>
+		<NcCheckboxRadioSwitch v-show="form.expires"
```
jotoeri, line 58: Better inside of the div, no?

jotoeri, line 58: Using `formExpires`, as for the Picker?

jotoeri: Or maybe even the expires-switch out of the div and the v-show and the position/margin applied to the whole div?

src/views/Submit.vue
```diff
@@ -44,6 +44,8 @@
 			<!-- Do not wrap the following line between tags! `white-space:pre-line` respects `\n` but would produce additional empty first line -->
 			<!-- eslint-disable-next-line -->
 			<p v-if="!loading && !success" class="form-desc">{{ form.description }}</p>
+			<!-- Show expiration message-->
+			<p v-if="form.expires && form.showExpiration" class="info-message" v-text="expirationMessage" />
```
jotoeri, line 48: Also necessary in Create.vue

jotoeri, line 48: Lets keep the bracket syntax instead of the v-text? You can also change it on the infoMessage then, iirc i didnt review that that once since i had already many other comments 🙈 .

lib/Controller/ApiController.php
Chartman123: @susnux What do you mean with that comment? If I'm correct this is already tested with https://github.com/nextcloud/forms/blob/b80928c474dbd95af3330dd85ddb3ada7fa21d01/tests/Integration/Api/ApiV2Test.php#L409-L461

src/views/Submit.vue
```diff
@@ -210,6 +212,14 @@ export default {
 
 			return message
 		},
+
+		expirationMessage() {
+			const relativeDate = moment(this.form.expires, 'X').fromNow()
+			if (this.isExpired) {
+				return t('forms', 'Expired {relativeDate}', { relativeDate })
```
jotoeri, line 219: Message with a trailing dot here?!

jotoeri:
> @jotoeri I added the text for expired forms as well. Now I have the same method for AppNavigationForms.vue and Submit.vue (and it should probably also be included into Create.vue). Should this be moved to a new mixin to avoid the duplicated code?

So - with my last comment, the message would differ by a dot?! 😆 
I think generally you could of course put it into a mixin (or probably better a util with just that function) - but it's not that critical imo. So just do as you want i'd say. 😉

src/views/Submit.vue
Chartman123: Yes, but in this case we would need a second translation for the otherwise exact same text. If thats no problem, I'll add it with the `.`

src/views/Submit.vue
Chartman123: I used the v-text syntax as it wouldn't trigger the linter ;) For the form description we also have the white-space thing in the scss part... Do we need this here, too?

Chartman123: See my comment with the question regarding the mixin...

src/components/SidebarTabs/SettingsSidebarTab.vue
Chartman123: Or should we also abandon the computed property and rely totally on the original value from the database?

lib/Db/Form.php
susnux: Thank you for the clarification! :)

src/components/SidebarTabs/SettingsSidebarTab.vue
jotoeri: Hmm, you can try. But i think the computed prop was taken, since the switch doesn't accept numerical values.

src/views/Submit.vue
jotoeri: Hmm. Indeed, a completely separate translation has the danger of getting different results. But still, it just looks bad in the header without dot, no? Maybe adding the dot manually here in the procedure (or by second param in the util, then)? 🤔

src/views/Submit.vue
Chartman123: Thanks for clarifying :)

src/components/SidebarTabs/SettingsSidebarTab.vue
Chartman123: > since the switch doesn't accept numerical values

Yes, that's it :)

src/views/Submit.vue
Chartman123: A hard coded dot is not the best solution in my opinion. I'm not sure if all languages use a `.` for ending a sentence. And without it, it's also not looking that good...

Chartman123:
@jotoeri All comments (except for the punctuation) tackled successfully 👍🏻

jotoeri (approved):
Fine by me 👍

susnux (approved):

Chartman123:
> > screenshot for quick design review? :)
> 
> @jancborchardt Yes, for sure 👍🏻 Added it to the first post.
> 
> > I would say we do not need a setting for this – if there is an expiration date, potential respondents should always be made aware of it.
> 
> If the form is only internally shared to the instance/other users, they can see the expiration date already in the AppNavigation. So in these cases I'd prefer to have it configurable to avoid redundance...

@jancborchardt Just a quick ping if you're ok with this :)

lib/Controller/ApiController.php
Chartman123: @susnux @jotoeri I found the reason for the Codecov message: the integration tests are currently disabled. I tried to enable them, however they are not successful at the moment. So I think that we should move this to a separate PR and fix the integration tests in general.

PR #1440, opened 2023-01-02
Import form question extra settings

### LINKED ISSUE #1439: Questions extra settings missing in FormMigrator

**Describe the bug**
In the import function of FormsMigrator.php, the extraSettings JSON for the questions is missing.

https://github.com/nextcloud/forms/blob/main/lib/FormsMigrator.php#L205-L212

This is also true for the corresponding tests and the documentation

Copephobia:
Fixes #1439 

### Bug Reproduction Steps

1. Have environment on NC25, with "User Migration" and "Forms" apps enabled
2. Create a form that has a question with extra settings, such as a multiple choice question with "Shuffle options" extra setting selected
3. Perform an export of the Forms app data
4. Verify the export data JSON contains the extra setting value
5. Delete the form created in step 2
6. Import the exported data from step 3
7. Verify the multiple choice question does NOT have "Shuffle options" selected (as the extra settings did not import)

### Bug Fix Validation

1. Apply the code change in this PR to the "Forms" app
2. Perform steps 2-6 in the reproduction steps above
3. Verify the form created from the import has the "Shuffle options" selected (the extra settings have imported successfully)

### DIFF
--- a/tests/Unit/FormsMigratorTest.php
+++ b/tests/Unit/FormsMigratorTest.php
@@ -114,7 +114,7 @@ class FormsMigratorTest extends TestCase {
 	public function dataExport() {
 		return [
 			'exactlyOneOfEach' => [
-				'expectedJson' => '[{"title":"Link","description":"","created":1646251830,"access":{"permitAllUsers":false,"showToAllUsers":false},"expires":0,"isAnonymous":false,"submitMultiple":false,"questions":[{"id":14,"order":2,"type":"multiple","isRequired":false,"text":"checkbox","description":"huhu","options":[{"text":"ans1"}]}],"submissions":[{"userId":"anyUser@localhost","timestamp":1651354059,"answers":[{"questionId":14,"text":"ans1"}]}]}]'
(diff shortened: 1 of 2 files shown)


jotoeri:
Looks good to me in General, just needs to fix the test now. :)

tests/Unit/FormsMigratorTest.php
```diff
@@ -162,6 +162,7 @@ public function testExport(string $expectedJson) {
 					"isRequired" => false,
 					"text" => "checkbox",
 					"description" => "huhu",
+					"extraSettings" => (new \stdClass()),
```
jotoeri, line 165: The class looks a bit strange - what we need is an object. I think i'd use the `(object)[]` as in https://github.com/nextcloud/forms/blob/bd1d901d390ab90ff6f3f066f26f688761f508c3/tests/Unit/Service/FormsServiceTest.php#L189

jotoeri:
And then two more small things:
- Ideally squash the commits to have only one
- If you manage to fix the DCO-Test would be a nice gimmick ;)
> Commit sha: [208eddf](https://github.com/nextcloud/forms/pull/1440/commits/208eddf5c2ca8b67ccc90c13af51dc3df1bd3cc0), Author: Andrew Cope, Committer: Andrew Cope; Expected "Andrew Cope [*@rit.edu](mailto:*@rit.edu)", but got "Andrew Cope [*@gmail.com](mailto:*@gmail.com)".

tests/Unit/FormsMigratorTest.php
Copephobia: Thanks! My brain could not think of the simpler way to create the empty object. Clearly I need to brush up on my PHP skills!

Chartman123 (approved):
Looks good to me 👍🏻 

I think we should add the extraSettings to the docs in a separate PR as this isn't really about the import but just missing documentation of the field itself.

jotoeri (approved):
👍 🎉 
Just wait for a second review before merging. 😉  - Edit: Chartman was faster. 😆

PR #1441, opened 2023-01-03
Show info message only if available

susnux:
Removes dead space between description and questions, the space gets even worse if we add new messages like with #1432, this only affects the public view if the current user is not logged in.

For comparison, this is how it looks (with or without this) logged in:
submit form, logged in view

And this is how it looks like if the user is not logged in:
| :framed_picture: Before | :framed_picture: :sparkles:  After |
| ---------------------------------- | ------------------------------------------------ |
| submit form, current, lot dead space | submit form, changed, only the needed margin is used |

### DIFF
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -45,7 +45,7 @@
 			<!-- eslint-disable-next-line -->
 			<p v-if="!loading && !success" class="form-desc">{{ form.description }}</p>
 			<!-- Generate form information message-->
-			<p class="info-message" v-text="infoMessage" />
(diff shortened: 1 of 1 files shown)


jotoeri (approved):

Chartman123 (approved):

jancborchardt:
Btw @susnux, I see you already have an account on our Nextcloud instance – you can join the Talk room "Forms team public" if you like. :)

susnux:
@jancborchardt 
> I see you already have an account on our Nextcloud instance – you can join the Talk room "Forms team public" if you like. :)

I would love to, I can see every "public team" room except the forms room :sweat_smile:

PR #1447, opened 2023-01-09
Scope SCSS in Create.vue and adjust Results.vue

### LINKED ISSUE #1442: Scope CSS in Create.vue
CSS in Create.vue is not scoped and thus spitting anywhere. Most probably needs CSS adaptions in other places to correct for visual changes.

Chartman123:
Fixes #1442 

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -327,20 +327,39 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-.app-content header {
-	h2 {
-		font-size: 2em;
-		font-weight: bold;
-		margin-top: 32px;
-		padding-left: 14px;
-		padding-bottom: 8px;
-		overflow: hidden;
-		text-overflow: ellipsis;
-		white-space: nowrap;
+.app-content {
+	display: flex;
+	align-items: center;
+	flex-direction: column;
+
+	header,
+	section {
+		width: 100%;
+		max-width: 750px;
 	}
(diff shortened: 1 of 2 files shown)


src/views/Results.vue
```diff
@@ -327,20 +327,39 @@ export default {
...
+		h2 {
+			margin-bottom: 0; // because the input field has enough padding
+			font-size: 28px;
+			font-weight: bold;
+			margin-top: 32px;
+			padding-left: 14px;
+			padding-bottom: 8px;
+			overflow: hidden;
+			text-overflow: ellipsis;
+			white-space: nowrap;
```
jotoeri, line 357: Not necessary? 🤔

jotoeri (approved):
Thanks for tackling directly! :)
- One small thing commented, but that was already like that before as it seems. 🤷‍♂️

src/views/Results.vue
Chartman123: With those rules:
grafik
Without:
grafik

jotoeri: Ah, sorry - I just clicked and meant the `white-space` line.
Regarding the title, i think i'd leave it here as is. There's no need to show the full title here again, no?

Chartman123: For me the shortened version is fine. :)

susnux (approved):

PR #1449, opened 2023-01-10
Enable PHPUnit integration tests and fix them

### LINKED ISSUE #1448: Enable integration tests and fix them
https://github.com/nextcloud/forms/blob/336ae8913b8ac43571d6e9f963e16a2e9049cbc7/tests/phpunit.integration.xml#L6-L9

Integration tests are currently disabled. Enabling them leads to many test failures and some errors. Those have to be fixed, too.

Chartman123:
Fixes #1448 

- [x] enable integration tests
- [x] fix failing tests / errors

### DIFF
--- a/lib/Db/OptionMapper.php
+++ b/lib/Db/OptionMapper.php
@@ -54,7 +54,8 @@ class OptionMapper extends QBMapper {
 			->from($this->getTableName())
 			->where(
 				$qb->expr()->eq('question_id', $qb->createNamedParameter($questionId))
-			);
+			)
+			->orderBy('id');
 
 		return $this->findEntities($qb);
 	}
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -57,8 +57,9 @@ jobs:
           OC_PASS: test
         run: |
           mkdir data
-          ./occ maintenance:install --verbose --database=${{ matrix.databases }} --database-name=nextcloud --database-host=127.0.0.1 --database-port=$DB_PORT --database-user=root --database-pass=rootpassword --admin-user admin --admin-pass password 
+          ./occ maintenance:install --verbose --database=${{ matrix.databases }} --database-name=nextcloud --database-host=127.0.0.1 --database-port=$DB_PORT --database-user=root --database-pass=rootpassword --admin-user admin --admin-pass password
           ./occ user:add --password-from-env --display-name="Test Displayname" test
+          ./occ user:add --password-from-env --display-name="User No. 1" user1
           ./occ app:enable --force ${{ env.APP_NAME }}
           php -S localhost:8080 &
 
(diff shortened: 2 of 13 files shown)


Chartman123:
@jotoeri Could you please have a look at the three remaining failing test scenarios? I can't see why they are failing...

Chartman123:
@jotoeri I had a look at the response from the API of my dev instance:

```
    "questions": [
        {
          "id": 38,
          "formId": 12,
          "order": 1,
          "type": "time",
          "isRequired": false,
          "text": "test",
          "description": "",
          "extraSettings": {},
          "options": []
        },
        {
          "id": 44,
          "formId": 12,
          "order": 2,
          "type": "short",
          "isRequired": false,
          "text": "Test2",
          "description": "Dies ist der Beschreibungstext zur Frage",
          "extraSettings": {},
          "options": []
        },
        {
          "id": 45,
          "formId": 12,
          "order": 3,
          "type": "time",
          "isRequired": false,
          "text": "a",
          "description": "a",
          "extraSettings": {},
          "options": []
        }
      ],
```

The questions array seems to contain some objects. Our expectation however is an array of question arrays. Might this be the problem? However, I don't know why this would only fail for MySQL and OCI then.

jotoeri:
So - Problem was, that the description was not explicitely set on the questions. This resulted in a NULL value on DB, which the `htmlspecialchars_decode()` within `Question.php` didn't like and thus made the `read()` fail. On the API we set the Description explicitely on each question creation, thus it was no problem there.
Seems like the htmlspecialchars earlier just returned the empty string, but with php8.1 throws an error or so...

jotoeri:
@nickvergessen Do you have an idea, why the Oracle setup is doing different stuff here?
https://github.com/nextcloud/forms/actions/runs/3915897580/jobs/6694496713

Do you maybe know about a docker container, that i can easily spin up to run a server master with oracle?

jotoeri:
Seems like the problematic `htmlspecialchars_decode()` is just an artifact from Forms <v2. Since we actually never do the corresponding encoding `htmlspecialchars()`, i can't imagine currently, why this should be necessary, so i think we can just remove it.
Any thoughts, @Chartman123 @nickvergessen @susnux ?
https://github.com/nextcloud/forms/pull/1449/commits/05c4fe88da76d9ddf8883d0ef001e29cf077a488

jotoeri:
Seems like we never were compatible with oracle before.^^ It essentially makes the NotNull Property of columns unusable. :(
Just as a self reminder building ontop of 2269025f5863d4382ea029da9662ceb77d1bd8f9 :
- [ ] Ensure Changed Notnull columns to be nullable
- [ ] Add Tests for these cases.

nickvergessen:
Tldr: you can have nullable columns, but empty strings are null on Oracle. So if you add a nullable columns to an existing table you have to give it a different default value than empty string.
When you create a new table you can have a nullable with default empty string, but it basically means you have to specify it on every insert as the default can not be saved.

susnux:
> i can't imagine currently, why this should be necessary, so i think we can just remove it. Any thoughts

I agree, I think this this is a leftover from the very first version where every text was encoded, but the last encode call was removed with 612af52f50dbda7a77f9b118cd74f68e9c3021a1

Chartman123:
@jotoeri I did a quick rebase to resolve the merge conflicts after merging the "show expiration" PR.

jotoeri:
> @jotoeri I did a quick rebase to resolve the merge conflicts after merging the "show expiration" PR.

👍 Should also be good to review and go now.

nickvergessen:
We can add Chartman123 to the Nextcloud org, so you can request it

lib/Db/OptionMapper.php
```diff
@@ -54,7 +54,8 @@ public function findByQuestion(int $questionId): array {
 			->from($this->getTableName())
 			->where(
 				$qb->expr()->eq('question_id', $qb->createNamedParameter($questionId))
-			);
+			)
+			->orderBy('id');
```
Chartman123, line 58: @jotoeri what is this for?

Chartman123 (commented):
@jotoeri I can't approve the PR because I'm the author... I did a double check and think that everything should be fine by now.

lib/Db/OptionMapper.php
jotoeri: Oracle somehow returned the Options in reverse order. The orderBy solved it.

jotoeri (approved):

susnux (approved):

Issue #455, opened 2020-06-12
Ability to share responses / results


jancborchardt:
It would be great if not only the form creator could see the responses, but they could also be shareable via link, or to a group of other Nextcloud accounts.

(Splitting this off from the "Collaborative editing" https://github.com/nextcloud/forms/issues/322 )

skjnldsv:
Seems a bit hard and lots of work no?
Maybe 3.0 ?

jancborchardt:
Let’s say 2.2 for now – it would be quite useful, and it’s better to properly assess it before. It was also said that making the summary stats page is hard. ;)

tigexplorer:
It would be very nice, to add the optional ability to show the user after sending the form, the summary of the responses. 

Also very helpful could be, to share a link to the responses of a form for non nextcloud users.

dosch:
or, if this proves hard; to share a form with a co-worker. 
In this scenario I just help a colleague to make the form, then transfer ownership of the form to them.
This way they are also the owner of the answers.

ostasevych:
Let me also add opposite requirement: to limit sharing responses with limited number of people.

pebkacs:
We are using forms for signups for a Minecraft server, and it works great.  It would be ideal to be able to share the responses with the rest of the MC server admins without my intervention each time there is a response.

42l-neil-tmp:
Hi,

This feature has been asked during UX tests on my end, involving a situation where:
- There's only one person in a team with a Nextcloud account ;
- A public form is created and shared, meant to be filled by outsiders ;
- The person wants to share the answers with their team without giving out their login credentials, nor requiring all members of their team to create a Nextcloud account.

This situation sounds similar to @pebkacs's as described in the comment above.

A sharing link for the responses view should answer those needs. Another solution would be to share the whole form (including the "questions" view) in read-only mode.

biva:
Maybe a workaround to make it easier for a first step: create a link to download the latest version of the CSV. As a first step, this link can be public, and as a second step we could decide if it's limited to Nextcloud users or for a Nextcloud group.

pebkacs:
> Maybe a workaround to make it easier for a first step: create a link to download the latest version of the CSV. As a first step, this link can be public, and as a second step we could decide if it's limited to Nextcloud users or for a Nextcloud group.

[...]

PR #1461, opened 2023-01-22
Allow sharees to see results | Implement `PERMISSION_RESULTS`

susnux:
* Resolves: #455 

### Summary

I currently really need the ability to share responses / have multiple users to be able to see the results of a form.
So this implements this feature by adding the available but currently unused permissions to shares and allowing users with permission `results` to see the submissions.

### Screenshot
Not much visually added except for the permission checkbox (later an other checkbox for the `edit` permission could be added):
share sidebar with permission checkbox per share

### Todo
- [x] Test: `ShareApiController`
- [x] Test: `ApiController`
- [x] Test: `FormsService`
- [x] Integration Tests
- [x] Frontend


susnux:
Ready for review :rocket: 

@jotoeri one thing is still missing because I need some feedback: Should users with `RESULTS` permission also be able to delete all submissions? If yes I need to change the permission check for this API endpoint, if not I need to disable the `delete all submissions` button for users with shared access.

Chartman123:
> If yes I need to change the permission check for this API endpoint, if not I need to disable the `delete all submissions` button for users with shared access.

@susnux An UI-only solution should never be used for permission related stuff... The actions should always be checked against the permissions in the backend

susnux:
@Chartman123 
> An UI-only solution should never be used for permission related stuff... The actions should always be checked against the permissions in the backend

I know that is also how it is implemented, but this is only about following two options:
Should users, not the owner, be able to delete all submissions?
1. Yes -> then I need to change the backend as it currently forbids this
2. No -> then I need to disable that button on the frontend as it currently shows that option but it will currently fail (see 1)

Chartman123:
@susnux @jotoeri Should we then make the permissions even more granular? Do users (not owners) have the permission to delete single submissions? Then they could also delete all submissions one by one... I'd prefer having to options: read results and delete results.

jancborchardt:
> I'd prefer having to options: read results and delete results.

Yep, agree. The wording should be "View responses" and "Delete responses".

And we could also for now just with the simple "View responses", as that will cover what you said @susnux (emphasis mine):

> I currently really need the ability to share responses / have multiple users to be able to **see** the results of a form.

susnux:
> Should we then make the permissions even more granular? Do users (not owners) have the permission to delete single submissions? Then they could also delete all submissions one by one... I'd prefer having to options: read results and delete results.

I agree with that, I think the permissions should be extended to (bold == new):
* *submit* (implicitly grated with every share)
* **share**
* *edit*
* **results_read** (currently *results*)
* **results_delete**

*(`share` would be an other topic so we can postpone that, would also require to discuss what permissions are allowed to share)*

jotoeri:
> 
    submit (implicitly grated with every share)
    share
    edit
    results_read (currently results)
    results_delete

I'm fine with splitting the permissions to delete/see results. I'd just say, ideally keep the results permission as is and just add the `results_delete` permission. Otherwise its a breaking change on the API.
Share is of course a separate thing for when being implemented...

lib/Service/FormsService.php
```diff
@@ -495,4 +482,49 @@ public function notifyNewShares(Form $form, Share $share): void {
...
+	 *
+	 * @param Share $share
+	 */
+	protected function getSharePermissions(Share $share): array {
+		$permissions = $share->getPermissions();
+		if (count($permissions) > 0) {
+			return $permissions;
+		}
+		return [ Constants::PERMISSION_SUBMIT ];
+	}
```
jotoeri, line 529: How about doing this in Db/Share.php? There you currently return an empty array, just to fill the array here then.

src/components/SidebarTabs/SharingShareDiv.vue
```diff
@@ -41,22 +46,31 @@
...
-import NcAvatar from '@nextcloud/vue/dist/Components/NcAvatar.js'
-import IconClose from 'vue-material-design-icons/Close.vue'
+import {
+	NcActions,
+	NcActionButton,
+	NcActionCaption,
+	NcActionCheckbox,
+	NcActionSeparator,
+	NcAvatar,
+} from '@nextcloud/vue'
```
jotoeri, line 56: This unnecessarily increases the bundle size by 1.8MiB, since webpack tree-shaking is not really working. So please revert that.

We could try in a separate PR, if it meanwhile is better to use the whole package like this, but that would need to change all occurences of nc-vue imports

src/components/SidebarTabs/SharingShareDiv.vue
```diff
@@ -81,6 +98,19 @@ export default {
 		removeShare() {
 			this.$emit('remove-share', this.share)
 		},
+
+		/**
+		 * @param {boolean} resultsPermission If the results permission should be granted
+		 */
+		updateResultsPermission(resultsPermission) {
+			const share = Object.assign({}, this.share, {
+				permissions: [
```
jotoeri, line 107: Since you just override the permissions, this won't work when we have other permissions.
Also i'd rather go with a simple code - If true, push, else filter/find.

tests/Unit/Controller/ApiControllerTest.php
```diff
@@ -0,0 +1,286 @@
...
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\IUserSession;
+
+use PHPUnit\Framework\MockObject\MockObject;
+use Test\TestCase;
+
+use Psr\Log\LoggerInterface;
+
+class ApiControllerTest extends TestCase {
```
jotoeri, line 55: Nice Start, thanks! 🎉 🚀
Was long overdue to do this... 🙈

lib/Controller/ShareApiController.php
```diff
@@ -244,4 +248,79 @@ public function deleteShare(int $id): DataResponse {
 
 		return new DataResponse($id);
 	}
+
+	/**
+	 * @NoAdminRequired
```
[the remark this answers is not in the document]
susnux: Used the other functions as a template but you already submitted a that PR to fix them as well :sweat_smile:

lib/Service/FormsService.php
susnux: I was unsure about the separation of model and logic, but if you prefer it in the model I am fine with in :)

lib/Controller/ShareApiController.php
```diff
@@ -244,4 +248,79 @@ public function deleteShare(int $id): DataResponse {
...
+		}
+
+		$validatedPermissions = array_intersect(Constants::PERMISSION_ALL, $permissions);
+		if (count($validatedPermissions) < count($permissions)) {
+			$this->logger->debug('Invalid permission given', ['invalid_permissions' => array_diff($permissions, $validatedPermissions)]);
+			throw new OCSBadRequestException('Invalid permission given');
+		}
+
+		// Make sure indices are fixed
+		$validatedPermissions = array_values($validatedPermissions);
```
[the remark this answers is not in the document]
susnux: No this is not an error, e.g. if the permissions are provided in a different order the resulting `validatedPermissions` could look like `[1 => 'submit']`. Then the following `$validatedPermissions[0] !== Constants::PERMISSION_SUBMIT` would fail.
So we just make sure `[1 => 'submit']` is corrected to `[0 => 'submit']`

lib/Controller/ShareApiController.php
```diff
@@ -244,4 +248,79 @@ public function deleteShare(int $id): DataResponse {
...
+	}
+
+	/**
+	 * Validate user given permission array
+	 *
+	 * @param array $permissions User given permissions
+	 * @return array Sanitized array of permissions
+	 * @throws OCSBadRequestException If invalid permission was given
+	 */
+	protected function validatePermissions(array $permissions, int $shareType): array {
```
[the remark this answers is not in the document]
susnux: This is sanitizing of user input, should I move this from controller to service?

lib/Controller/ShareApiController.php
```diff
@@ -151,6 +152,8 @@ public function newShare(int $formId, int $shareType, string $shareWith = ''): D
 			throw new OCSForbiddenException();
 		}
 
+		$validatedPermissions = $this->validatePermissions($permissions, $shareType);
```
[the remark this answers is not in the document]
susnux: My idea was to remove duplicated code, but yes I will change this to be consistent with the other checks.

lib/Controller/ShareApiController.php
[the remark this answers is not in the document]
susnux: I rewrote that function to return only a boolean so this was dropped anyways.

susnux:
@jotoeri thank you for your review, I resolved all of your comments.

I decided to stub implement `results_delete` by adding it to the constants and documentation but only use it currently to hide the *delete all submissions* and *delete submission* buttons on the frontend.
Meaning only the forms owner, as before this PR, can delete submissions.

I did so to keep this PR "small", I will submit another PR properly implementing the deleting results permission after this one is merged.

tests/Integration/Api/ApiV2Test.php
```diff
@@ -24,13 +24,13 @@
...
 namespace OCA\Forms\Tests\Integration\Api;
 
-use OCA\Forms\Db\Form;
 use OCA\Forms\Db\FormMapper;
 
 use OCP\DB\QueryBuilder\IQueryBuilder;
 
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\ClientException;
+use OCA\Forms\Constants;
```
Chartman123, line 33: Move up to the other use OCA\...?

Chartman123:
On a shared form with the results permission the TopBar shows the "Share" button on first load. Clicking it does nothing and the button disappears:
grafik

susnux:
> On a shared form with the results permission the TopBar shows the "Share" button on first load. Clicking it does nothing and the button disappears: :framed_picture: 

This was due to wrong "share permissions" on the frontend, fixed that

appinfo/routes.php
```diff
@@ -219,6 +219,14 @@
 				'apiVersion' => 'v2'
 			]
 		],
+		[
+			'name' => 'shareApi#updateShare',
+			'url' => '/api/{apiVersion}/share/update',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2'
```
Chartman123, line 227: Should we move to v2.1 here as well as it is an addition to the v2 API?

docs/API.md
```diff
@@ -350,18 +350,20 @@ Contains only manipulative question-endpoints. To retrieve options, request the
...
-  |-------------|---------|-------------|
-  | _formId_    | Integer | Id of the form to share |
-  | _shareType_ | String  | NC-shareType, out of the used shareTypes. |
-  | _shareWith_ | String  | User/Group for the share. Not used for link-shares. |
+  | Parameter    | Type     | Description |
+  |--------------|----------|-------------|
+  | _formId_     | Integer  | Id of the form to share |
+  | _shareType_  | String   | NC-shareType, out of the used shareTypes. |
+  | _shareWith_  | String   | User/Group for the share. Not used for link-shares. |
+  | _permission_ | String[] | Permissions of the sharees, see [DataStructure](DataStructure.md#Permissions). |
```
Chartman123, line 358:
```suggestion
  | _permissions_ | String[] | Permissions of the sharees, see [DataStructure](DataStructure.md#Permissions). |
```

appinfo/routes.php
susnux: @jotoeri ?

jotoeri: Definitely yes here... 👍

susnux: Should all other routes support 2.1 as well? (`v2(.1)?`)

jotoeri: Yes, but as separate PR, please.

susnux: > Definitely yes here... +1

:heavy_check_mark:  Done

susnux:
I resolved the comments and added an additionally integration test for updating share permissions.

Chartman123 (approved):
@susnux From my point of view this is ready. Just try to cleanup your commits by squashing the fix commits into the original commits. 🙂 

Then let's wait for @jotoeri to approve this and get this merged afterwards 👍🏻

jotoeri (commented):
Just one small doc remaining, code should be ok.

- But let's first get #1465 merged and released. Then we can adapt the API calmly afterwards.

docs/API.md
```diff
@@ -378,6 +380,18 @@ Contains only manipulative question-endpoints. To retrieve options, request the
 "data": 5
 ```
 
+### Update a Share permissions
+- Endpoint: `/api/v2.1/share/update`
+- Url-Parameter:
+  | Parameter     | Type     | Description |
+  |---------------|----------|-------------|
+  | _id_          | Integer  | ID of the share to update |
+  | _permissions_ | String[] | Array of permissions |
```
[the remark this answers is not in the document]
susnux: :see_no_evil:

susnux:
> Just one small doc remaining, code should be ok.
> But let's first get #1465 merged and released. Then we can adapt the API calmly afterwards.

Fixed that :)
I would have waited for that PR to get merged before anyways, so we could release a fix release quickly before implementing new features :)

jotoeri (approved):

PR #1465, opened 2023-01-26
Fix insert

jotoeri:
@susnux I think you're right, that the insert is open currently, since the combination of PublicPage and CORS is not good.
However, i think we can remove the PublicPage condition in server, since that seems to me has been the fix when the CSRF-token was not allowed on CORS routes yet. As a quick-fix, i'd just execute the CORS-Routine here for now. We can just revert the commit, once server got fixed accordingly.

### DIFF
--- a/lib/Middleware/PublicCorsMiddleware.php
+++ b/lib/Middleware/PublicCorsMiddleware.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ * @copyright Copyright (c) 2023 Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @author Jonas Rittershofer <jotoeri@users.noreply.github.com>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Forms\Middleware;
+
+use OC\AppFramework\Middleware\Security\Exceptions\SecurityException;
+
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\JSONResponse;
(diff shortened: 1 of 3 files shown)


susnux:
Looks like a valid solution, I thought about something similar: Adding a custom Middleware for our API controller with a similar validation function. This might be cleaner?

jotoeri:
Yeah, changed it to a middleware now. It's a slightly modified procedure to mainly use the public interfaces, but basically the same as in the CORSMiddleware.

lib/Middleware/PublicCorsMiddleware.php
```diff
@@ -0,0 +1,108 @@
...
+	 *
+	 * @param Controller $controller the controller that is being called
+	 * @param string $methodName the name of the method that will be called on
+	 *                           the controller
+	 * @throws SecurityException
+	 */
+	public function beforeController($controller, $methodName) {
+		// ensure that @CORS annotated API routes are not used in conjunction
+		// with session authentication since this enables CSRF attack vectors
+		if ($this->reflector->hasAnnotation('PublicCORSFix')) {
```
susnux, line 70: If I am correct, this will check every user and even throw if anonymous submit as the user is not set.
In this case we could simply drop `@PublicPage`.

If we want to support anonymous submission we should change this to:
```php
if ($this->reflector->hasAnnotation('PublicCORSFix') && $this->userSession->isLoggedIn())
```

jotoeri: Public Pages pass the CSRF Check. Thus it works to submit anonymously.

susnux: So `request->passesCSRFCheck()` returns true for anonymous requests without csrf token?
Because if not, anonymous requests would be blocked by line 79.
That's why I suggested the change to only check logged in users, so that anonymous responses are still possible.

jotoeri: Nah, csrfcheck result should of course be independent of the user/anonymous. But our own pages pass the csrf check. Didn't test now, but even with the embedding we should still have a valid csrf token.

Indeed, what does not work are anonymous requests from a foreign source. But do we really want to open that? Or shouldn't all external requests be authenticated? Feels a bit like a hole on its own to open that... 😕

susnux: <delete>I would not change the behavior here, as currently anonymous submissions from extern are allowed, so this would break such usage cases. Not sure on how high the threat of CSRF of anonymous submissions is, you could submit with the IP of someone else or send spam with the IP of someone else.

(But having a breaking change is the only real reason here for me not disallowing anonymous submissions without session)</delete>

While writing this I noticed this is the only endpoint working anonymously, so it does not make much sense having this available from external pages, as you can not even request a public share from external without authentication. (only *no admin required* and *cors* is set so the user must be authenticated).

susnux (approved):
Tested and works as expected :+1:

Chartman123 (approved):

PR #1470, opened 2023-01-30
feat: Add slot for additional actions to the question component

susnux:
Added a slot for additional, question specific, settings and moved the shuffle option to only the required questions.
Did this to allow further custom question settings without adding more, mostly useless, properties to the question component. E.g. such a feature could be the validation of a short question.

I am not 100% sure about the position of the slot, I think `delete the question` should always be the last option, but not sure if all other common options should be above the slot or also below the slot.

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -65,11 +65,7 @@
 						<!-- TRANSLATORS Making this question necessary to be answered when submitting to a form -->
 						{{ t('forms', 'Required') }}
 					</NcActionCheckbox>
-					<NcActionCheckbox v-if="shuffleOptions !== undefined"
-						:checked="shuffleOptions"
-						@update:checked="onShuffleOptionsChange">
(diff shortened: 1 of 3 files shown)


Chartman123:
@susnux on the other hand this means duplicated code if an extra setting is valid for multiple question types... I'm not sure what's better here.

susnux:
> on the other hand this means duplicated code if an extra setting is valid for multiple question types... I'm not sure what's better here.

True. I moved the shuffle option to show how this could be used, but if you think we should keep it in question (shared for two question types) I am fine with it and revert that.
I just need the slot for options specific for one question type (short answer type).

jotoeri:
As it is now, the shuffleOptions feels a bit more logically bound to the question-type it belongs to. So for me, both is fine here...

Chartman123:
Ok, then let's move it to the question types :)

Chartman123 (approved):

jotoeri (approved):

PR #1471, opened 2023-01-30
Replace deprecated NcMultiselect with recommended NcSelect

susnux:
`NcMultiselect` is [deprecated](https://nextcloud-vue-components.netlify.app/#/Components/NcMultiselect) the recommended alternative is `NcSelect` (mostly a drop-in replacement but few properties and slots changed or removed).

### DIFF
--- a/src/components/Questions/QuestionDropdown.vue
+++ b/src/components/Questions/QuestionDropdown.vue
@@ -42,16 +42,16 @@
 				{{ t('forms', 'Shuffle options') }}
 			</NcActionCheckbox>
 		</template>
-		<NcMultiselect v-if="!edit"
+		<NcSelect v-if="!edit"
 			v-model="selectedOption"
 			:name="text"
 			:placeholder="selectOptionPlaceholder"
 			:multiple="isMultiple"
 			:required="isRequired"
 			:options="sortedOptions"
+			:searchable="false"
 			label="text"
-			track-by="id"
-			@select="onSelect" />
+			@input="onInput" />
 
 		<ol v-if="edit" class="question__content">
 			<!-- Answer text input edit -->
@@ -87,7 +87,7 @@ import { emit } from '@nextcloud/event-bus'
 import { generateOcsUrl } from '@nextcloud/router'
 import axios from '@nextcloud/axios'
 import NcActionCheckbox from '@nextcloud/vue/dist/Components/NcActionCheckbox.js'
-import NcMultiselect from '@nextcloud/vue/dist/Components/NcMultiselect.js'
+import NcSelect from '@nextcloud/vue/dist/Components/NcSelect.js'
 
 import AnswerInput from './AnswerInput.vue'
 import QuestionMixin from '../../mixins/QuestionMixin.js'
@@ -100,7 +100,7 @@ export default {
 	components: {
 		AnswerInput,
 		NcActionCheckbox,
-		NcMultiselect,
+		NcSelect,
 	},
 
 	mixins: [QuestionMixin],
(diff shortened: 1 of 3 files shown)


Chartman123:
@susnux Looks good, but the styling does not fully fit the other components (border color and probably also the text color). I think that this should be fixed in the vue-lib.

Dark mode:
grafik

Light mode:
grafik

susnux:
@Chartman123 Do you emptied your cache? Exactly the opposite should be the case.
As `NcMultiselect` uses `--color-border-dark` while every other component, including `NcSelect` uses `--color-border-maxcontrast`

So it should now look like this:
darkmode, same borders but placeholder looks different
bright mode, same borders but placeholder looks different

---

But there are two differences:
1. The `NcSelect` does not have primary color borders on hover and when selected, I already filed a PR to fix this: https://github.com/nextcloud/nextcloud-vue/pull/3701
2. The placeholder looks differently because it uses `--color-text-maxcontrast` while "normal" inputs do not have a placeholder color set, this should be either fixed on the server side (global `input::placeholder`) or we could set this only for forms.

susnux:
Or maybe are you using NC 25.0.2? As the border was changed with this commit: https://github.com/nextcloud/server/commit/20da75f5ab66af28f43c3bebe0e562e382c75820 on master (I am testing with current server master) which was backported to 25.0.3

src/components/Questions/QuestionDropdown.vue
```diff
@@ -172,7 +172,7 @@ export default {
 	},
 
 	methods: {
-		onSelect(option) {
+		onInput(option) {
 			// Simple select
 			if (!this.isMultiple) {
 				this.$emit('update:values', [option.id])
```
jotoeri, line 178: clearing results in a console error currently.

jotoeri (commented):
Two small things here, apart from that, the theming does also not work properly....

(Indeed, quite specific, since System theme is dark, but NC Theme is set to light here...)
grafik

Chartman123:
> (Indeed, quite specific, since System theme is dark, but NC Theme is set to light here...)\n

That's how it looked for me yesterday at first. I then ran another `npm run dev` and then it changed to the way on my Screenshots... So probably really some caching problems here.

@susnux i will check the version later...

susnux:
> (Indeed, quite specific, since System theme is dark, but NC Theme is set to light here...) :framed_picture: 

This reminds me of https://github.com/nextcloud/text/issues/3369

src/components/Questions/QuestionDropdown.vue
susnux: Fixed that :)

jotoeri (approved):

Chartman123:
@susnux: So should we merge this before the server fix?

susnux:
> @susnux: So should we merge this before the server fix?

There is some discussion whether this is a server bug or should be fixed in the apps / libs. So I will file a PR against `@nextcloud/vue` so we get this fixed with the next release of it. I think we should wait for this to be fixed, so if we release before that, users might be annoyed by this bug.

Chartman123:
@susnux The border color is now correct, but the text color still differs...

grafik

grafik

Chartman123:
@susnux with the latest update of nc-vue the placeholder text is no longer shown

susnux:
> with the latest update of nc-vue the placeholder text is no longer shown

A regression if the select is configured to be not searchable: https://github.com/nextcloud/nextcloud-vue/pull/3833
(Probably fixed with 7.8.0)

susnux:
Rebased onto `main`.
@Chartman123  as `main` is now using version 7.8.0 this issue is fixed.

Chartman123 (approved):
@susnux the text color is still different to the other fields, but I think that we can merge this now :)

Chartman123:
@susnux the text color is still different to the other fields, but I think that we can merge this now :)

PR #1473, opened 2023-01-31
Add missing computed property to edit view

Chartman123:
This fixes the wrong text (present instead of past tense) being shown in the edit view when the form is expired.

Signed-off-by: Chartman123 <chris-hartmann@gmx.de>


jotoeri (approved):

susnux (approved):
:+1:

Chartman123:
Must have slipped through... Although it was even visible on my screenshot inside the PR 😆

PR #1478, opened 2023-02-02
Drop `v-clipboard` in favor of native browser API

susnux:
`v-clipboard` does not work for copying the embedding code in #1353 thats why I switch to `vue-clipboard2`.
But both packages are quite buggy or outdated.

So instead of using a package for this, we should use the browser clipboard API, similar to the server which is switching to that API too: https://github.com/nextcloud/server/pull/36450


jotoeri (approved):
Much better 👍

Chartman123 (approved):

PR #1479, opened 2023-02-02
Add lastUpdated property to Form

### LINKED ISSUE #1462: Navigation: Sort forms by recently edited / responses received
- Nextcloud-Version: 25.0.3
- Forms-Version: 3.0.3

Currently the forms in the navigation are sorted by creation date, which makes forms which have been created some time ago but are very active very difficult to navigate to.

It would make most sense to sort the forms by most recent activity, be it editing or having received responses. Basically similar to Talk or Notes. Newly created forms should still show up at the very top then (fitting the logic).

**Additional context**
In this example, "Nextcloud Include applications" is by far the most active form, but it stays on the bottom because I created it first:
image

Chartman123:
This fixes #1462 

### DIFF
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -103,11 +103,12 @@
 </template>
 
 <script>
-import { emit } from '@nextcloud/event-bus'
+import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { generateOcsUrl } from '@nextcloud/router'
 import { loadState } from '@nextcloud/initial-state'
 import { showError } from '@nextcloud/dialogs'
 import axios from '@nextcloud/axios'
+import moment from '@nextcloud/moment'
 
 import NcAppContent from '@nextcloud/vue/dist/Components/NcAppContent.js'
 import NcAppNavigation from '@nextcloud/vue/dist/Components/NcAppNavigation.js'
@@ -222,6 +223,14 @@ export default {
 
 	beforeMount() {
 		this.loadForms()
+	},
+
+	mounted() {
+		subscribe('forms:last-updated:set', (id) => this.onLastUpdatedByEventBus(id))
+	},
+
+	unmounted() {
+		unsubscribe('forms:last-updated:set', (id) => this.onLastUpdatedByEventBus(id))
 	},
 
 	methods: {
@@ -347,6 +356,23 @@ export default {
 			// Redirect if current form has been deleted
 			if (deletedHash === this.routeHash) {
 				this.$router.push({ name: 'root' })
+			}
+		},
+
+		/**
+		 * Update last updated timestamp in given form
+		 *
+		 * @param {number} id the form id
+		 */
+		onLastUpdatedByEventBus(id) {
+			const formIndex = this.forms.findIndex(form => form.id === id)
+			if (formIndex !== -1) {
+				this.forms[formIndex].lastUpdated = moment().unix()
+				this.forms.sort((b, a) => a.lastUpdated - b.lastUpdated)
+			} else {
+				const sharedFormIndex = this.sharedForms.findIndex(form => form.id === id)
+				this.sharedForms[sharedFormIndex].lastUpdated = moment().unix()
+				this.sharedForms.sort((a, b) => a.lastUpdated - b.lastUpdated)
 			}
 		},
 	},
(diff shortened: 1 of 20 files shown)


Chartman123:
@jotoeri could you please point me in the right direction, why my setLastUpdatedTimestamp function in the forms service doesn't work?

Chartman123:
> And you first need to store the form entity you get from findById. wink

🤦🏻 



> $this->formMapper->update($form);

Thanks, this was the missing part.

Chartman123:
@jotoeri The backend stuff should be ready now. However, I'm not sure how to proceed with the frontend, so that a changed form for example will also result in a reordered forms list in the AppNavigation. Shall we split this into a second PR?

jotoeri:
Hmm, to avoid confusion, i think we should shift it up, once updated/edited. That will of course not cover submissions, but will make clear the principle.
- Should suffice to just use the timestamp on the Frontend within a session, no need to have the exact timestamp of the backend.
- To not pass the event everywhere through our components you can also use the nextcloud/event-bus package.

And i wouldn't do it separately. Its closely connected.

Chartman123:
@jotoeri It's already working for updated form properties. However, I still get some errors when updating question properties and question options. Perhaps you could already have a look at it... I will only be able to get back to work tomorrow in the evening.

susnux:
> I still get some errors when updating question properties and question options.

Do you still get those errors? For me it works perfectly when changing question properties like *required* or *shuffle options*.

Chartman123:
@susnux I figured out the question related problems... now just the options/answers are missing.

I'm a little clueless what is triggered by the $emit('update:options',...) in dropdown and multiple questions. There is no @update:options in any of the vue components...

susnux:
> I'm a little clueless what is triggered by the $emit('update:options',...) in dropdown and multiple questions. There is no @update:options in any of the vue components...

I think it is used in `views/Create.vue`: `v-bind.sync="form.questions[index]"`. As this should bind `:options.sync`.

Chartman123:
@susnux @jotoeri I've now added the re-sorting for changed options. My last problem is that this also moves the form up just on clicking inside the "add new option" input box. At this time, the options haven't change in the database, so basically the sorting is "wrong".

Chartman123:
@susnux Your new ApiControllerTest is killing me... 😆 Could you please help me fix the failing checks here?

susnux:
You need to mock the `time()` function as the controller sets the `lastUpdated` value based on it. (see my commit)

Chartman123:
Thank you! Now I can polish the integration tests and then everything should be fine :)

src/Forms.vue
```diff
@@ -224,6 +225,14 @@ export default {
 		this.loadForms()
 	},
 
+	mounted() {
+		subscribe('change-last-updated', (id) => this.onLastUpdatedByEventBus(id))
```
susnux, line 229: I think for custom events we should stick to the event-bus naming scheme: `app-id:object:verb` e.g. `forms:last-updated:changed` or similar. (also to prevent naming clashes)

Chartman123: Done :)

jotoeri (approved):
So - now got into this one. Two small comments, maybe also squash a bit, but apart from that, it should be fine by me.

Chartman123:
> So - now got into this one. Two small comments, maybe also squash a bit, but apart from that, it should be fine by me.

Fixed them and squashed the commits. I think it's ready to go now :)

jancborchardt (commented):
Good stuff, thank you @Chartman123! :) (Can't review the code, so only leaving comment instead of approval.)

susnux (approved):
:+1:

susnux:
@Chartman123 
I tested this using mariaDB, so I did not noticed it, but this breaks updating a server that runs on SQLite.
The `comment` option on the new database column does not work with SQLite:
https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/schema-representation.html#common-options

```
In ExceptionConverter.php line 82:
                                                                                                     
  An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1 incomplete input  
                                                                                                     

In Exception.php line 30:
                                                      
  SQLSTATE[HY000]: General error: 1 incomplete input  
                                                      

In Connection.php line 72:
                                                      
  SQLSTATE[HY000]: General error: 1 incomplete input  
                                                      
```

I do not know why a clean install work (no forms installed before), it only fails if a partial migration is required.

Chartman123:
Strange, I copied the code from some older migration... 🙈

PR #1480, opened 2023-02-03
Add test for method `newForm` of `ApiController`

susnux:
I want to add some features on the API controller, but as not all methods of the controller have tests the PR would get crowded so I want to submit the added tests (for the current state) before submitting any other feature.

### DIFF
--- a/tests/Unit/Controller/ApiControllerTest.php
+++ b/tests/Unit/Controller/ApiControllerTest.php
@@ -79,6 +79,8 @@ class ApiControllerTest extends TestCase {
 	private $request;
 	/** @var IUserManager|MockObject */
 	private $userManager;
+	/** @var IL10N|MockObject */
+	private $l10n;
 
 	public function setUp(): void {
 		$this->activityManager = $this->createMock(ActivityManager::class);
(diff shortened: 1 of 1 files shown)


tests/Unit/Controller/ApiControllerTest.php
```diff
@@ -282,4 +289,90 @@ public function testExportSubmissions() {
...
+				'submitMultiple' => false,
+				'showExpiration' => false
+			]]
+		];
+	}
+	/**
+	 * @dataProvider dataTestCreateNewForm()
+	 */
+	public function testCreateNewForm($expectedForm) {
+		$validateForm = function ($expected) {
```
jotoeri, line 325: Function within function within function. Is that really necessary? You know, simple code wins on the long term? ;)

jotoeri (approved):
Anyways - Test is better than no test. :man_shrugging: :wink:

tests/Unit/Controller/ApiControllerTest.php
susnux: I reduced the nesting one level, but this helps on the further tests to reduce code duplication.

Chartman123 (approved):

PR #1488, opened 2023-02-05
Added test for `cloneForm` in `ApiController`

susnux:
Follow up on #1480, also submitted as single PR as the `cloneForm` method is quite huge and therefor requires more testing code.


susnux:
I read the migration guide from 7.x to 8.x but did not spot any point about throwing in anonymous functions, because it seems this is not allowed in php7? With php8 this works perfectly...

jotoeri (approved):

Chartman123 (approved):

PR #1489, opened 2023-02-05
Add Nextcloud 26 support

susnux:
Nextcloud 26 will require PHP 8.2 support (added with NC26 beta 2), but it looks like we could keep 3.x on both NC25 and NC26.

BTW all of my PRs are tested on NC26 as my local docker runs only the latest NC pre release, meaning the current main branch works fine on NC26.

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -18,7 +18,7 @@ jobs:
       # do not stop on another job's failure
       fail-fast: false
       matrix:
-        php-versions: ["8.0", "8.1"]
+        php-versions: ["8.0", "8.1", "8.2"]
         # Upload coverage only once (out of php-versions)
         coverage-on: ['8.1']
         databases: ['sqlite']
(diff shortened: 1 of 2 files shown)


jotoeri:
Regarding the max-version i'd wait until the first RC is out. Then we can check https://github.com/nextcloud/server/issues/34692 at once and be sure we dont miss anything.

ChristophWurst:
> Regarding the max-version i'd wait until the first RC is out. Then we can check [nextcloud/server#34692](https://github.com/nextcloud/server/issues/34692) at once and be sure we dont miss anything.

The feature freeze happened already. Ref https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule.

It would be cool to have this app ready for 26 by beta 5. We are aiming to have as many apps ready as possible by then to test them in production.

A prerelease on https://apps.nextcloud.com/apps/forms would be sufficient.

jotoeri (approved):
Rebased & checked the server issue - looks fine atm.

PR #1491, opened 2023-02-07
Add subtypes for short input, like email, phone or custom regex

### LINKED ISSUE #177: Validation of singleline/short response, url, regex...

susnux:
* Resolves #177 

## Summary
Added sub types for the short input, optional types (fallback to normal text input):
* numeric input
* telephone number
* email address
* custom regular expression

Added validation fallbacks on the front end as not every browser supports those input types, so if they are not supported we still get some validation.
Regex are validated on front and back end to be compatible with both PCRE (PHP) **and** ECMA (JS).

## Front end
https://user-images.githubusercontent.com/1855448/217513452-25ab0343-c58e-4541-b63d-e92ee398b780.mp4

## Known issues
* The "subtype" button overlays the input field when clicked, this will be fixed with the next `@nextcloud/vue` release, see nextcloud/nextcloud-vue#3724

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -101,6 +101,34 @@ class Constants {
 	];
 
 	/**
+	 * !! Keep in sync with src/models/ValidationTypes.js !!
+	 */
+
+	// Allowed short input types
+	public const SHORT_INPUT_TYPES = [
+		'phone',
+		'email',
+		'regex',
+		'number'
+	];
+
+	// This are allowed extra settings
+	public const EXTRA_SETTINGS_DROPDOWN = [
+		'allowOtherAnswer',
+		'shuffleOptions',
+	];
+
+	public const EXTRA_SETTINGS_MULTIPLE = [
+		'allowOtherAnswer',
+		'shuffleOptions',
+	];
+
+	public const EXTRA_SETTINGS_SHORT = [
+		'validationType',
+		'validationRegex'
+	];
+
+	/**
 	 * !! Keep in sync with src/mixins/ShareTypes.js !!
 	 */
 	public const SHARE_TYPES_USED = [
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -601,6 +601,10 @@ class ApiController extends OCSController {
 			throw new OCSForbiddenException('Please use reorderQuestions() to change order');
 		}
 
+		if (key_exists('extraSettings', $keyValuePairs) && !$this->formsService->areExtraSettingsValid($keyValuePairs['extraSettings'], $question->getType())) {
+			throw new OCSBadRequestException('Invalid extraSettings, will not update.');
+		}
+
 		// Create QuestionEntity with given Params & Id.
 		$question = Question::fromParams($keyValuePairs);
 		$question->setId($id);
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -516,4 +516,86 @@ class FormsService {
 		$form->setLastUpdated(time());
 		$this->formMapper->update($form);
 	}
+
+	/*
+	 * Validates the extraSettings
+	 *
+	 * @param array $extraSettings input extra settings
+	 * @param string $questionType the question type
+	 * @return bool if the settings are valid
+	 */
+	public function areExtraSettingsValid(array $extraSettings, string $questionType) {
+		if (count($extraSettings) === 0) {
+			return true;
+		}
+
+		// Ensure only allowed keys are set
+		switch ($questionType) {
+			case Constants::ANSWER_TYPE_DROPDOWN:
+				$allowed = Constants::EXTRA_SETTINGS_DROPDOWN;
+				break;
+			case Constants::ANSWER_TYPE_MULTIPLE:
+			case Constants::ANSWER_TYPE_MULTIPLEUNIQUE:
+				$allowed = Constants::EXTRA_SETTINGS_MULTIPLE;
+				break;
+			case Constants::ANSWER_TYPE_SHORT:
+				$allowed = Constants::EXTRA_SETTINGS_SHORT;
+				break;
+			default:
+				$allowed = [];
+		}
+		// Number of keys in extraSettings but not in allowed (but not the other way round)
+		$diff = array_diff(array_keys($extraSettings), $allowed);
+		if (count($diff) > 0) {
+			return false;
+		}
+
+		// Special handling of short input for validation
+		if ($questionType === Constants::ANSWER_TYPE_SHORT && isset($extraSettings['validationType'])) {
+			// Ensure input validation type is known
+			if (!in_array($extraSettings['validationType'], Constants::SHORT_INPUT_TYPES)) {
+				return false;
+			}
+
+			// For custom validation we need to sanitize the regex
+			if ($extraSettings['validationType'] === 'regex') {
+				// regex is required for "custom" input validation type
+				if (!isset($extraSettings['validationRegex'])) {
+					return false;
+				}
+
+				// regex option must be a string
+				if (!is_string($extraSettings['validationRegex'])) {
+					return false;
+				}
+
+				// empty regex matches every thing, this happens also when a new question is created
+				if (strlen($extraSettings['validationRegex']) === 0) {
+					return true;
+				}
+
+				// general pattern of a valid regex
+				$VALID_REGEX = '/^\/(.+)\/([smi]{0,3})$/';
+				// pattern to look for unescaped slashes
+				$REGEX_UNESCAPED_SLASHES = '/(?<=(^|[^\\\\]))(\\\\\\\\)*\\//';
+
+				$matches = [];
+				// only pattern with delimiters and supported modifiers (by PHP *and* JS)
+				if (@preg_match($VALID_REGEX, $extraSettings['validationRegex'], $matches) !== 1) {
+					return false;
+				}
+
+				// We use slashes as delimters, so unescaped slashes within the pattern are **not** allowed
+				if (@preg_match($REGEX_UNESCAPED_SLASHES, $matches[1]) === 1) {
+					return false;
+				}
+
+				// Try to compile the given pattern, `preg_match` will return false if the pattern is invalid
+				if (@preg_match($extraSettings['validationRegex'], 'some string') === false) {
+					return false;
+				}
+			}
+		}
+		return true;
+	}
 }
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -48,7 +48,7 @@ use OCP\IL10N;
 use OCP\IUser;
 use OCP\IUserManager;
 use OCP\IUserSession;
-
+use OCP\Mail\IMailer;
 use Psr\Log\LoggerInterface;
 
 class SubmissionService {
@@ -83,6 +83,9 @@ class SubmissionService {
 	/** @var IUser */
 	private $currentUser;
 
+	/** @var IMailer */
+	private $mailer;
+
 	public function __construct(FormMapper $formMapper,
 		QuestionMapper $questionMapper,
 		SubmissionMapper $submissionMapper,
@@ -92,7 +95,8 @@ class SubmissionService {
 		IL10N $l10n,
 		LoggerInterface $logger,
 		IUserManager $userManager,
-		IUserSession $userSession) {
+		IUserSession $userSession,
+		IMailer $mailer) {
 		$this->formMapper = $formMapper;
 		$this->questionMapper = $questionMapper;
 		$this->submissionMapper = $submissionMapper;
@@ -102,6 +106,7 @@ class SubmissionService {
 		$this->l10n = $l10n;
 		$this->logger = $logger;
 		$this->userManager = $userManager;
+		$this->mailer = $mailer;
 
 		$this->currentUser = $userSession->getUser();
 	}
@@ -342,6 +347,11 @@ class SubmissionService {
 						}
 					}
 				}
+
+				// Handle custom validation of short answers
+				if ($question['type'] === Constants::ANSWER_TYPE_SHORT && !$this->validateShortQuestion($question, $answers[$questionId][0])) {
+					return false;
+				}
 			}
 		}
 
@@ -366,5 +376,40 @@ class SubmissionService {
 	private function validateDateTime(string $dateStr, string $format) {
 		$d = DateTime::createFromFormat($format, $dateStr);
 		return $d && $d->format($format) === $dateStr;
+	}
+
+	/**
+	 * Validate short question answers if special validation types are set
+	 */
+	private function validateShortQuestion(array $question, string $data): bool {
+		if (!isset($question['extraSettings']) || !isset($question['extraSettings']['validationType'])) {
+			// No type defined, so fallback to 'text' => no special handling
+			return true;
+		}
+
+		switch ($question['extraSettings']['validationType']) {
+			case 'email':
+				return $this->mailer->validateMailAddress($data);
+			case 'number':
+				return is_numeric($data);
+			case 'phone':
+				// some special characters are used (depending on the locale)
+				$sanitized = str_replace([' ', '(', ')', '.', '/', '-', 'x'], '', $data);
+				// allow leading + for international numbers
+				if (str_starts_with($sanitized, '+')) {
+					$sanitized = substr($sanitized, 1);
+				}
+				return preg_match('/^[0-9]{3,}$/', $sanitized) === 1;
+			case 'regex':
+				// empty regex matches everything
+				if (!isset($question['extraSettings']['validationRegex'])) {
+					return true;
+				}
+				return preg_match($question['extraSettings']['validationRegex'], $data) === 1;
+			default:
+				$this->logger->error('Invalid input type for question detected, please report this issue!', ['validationType' => $question['extraSettings']['validationType']]);
+				// The result is just a non-validated text on the results, but not a fully declined submission. So no need to throw but simply return false here.
+				return false;
+		}
 	}
 }
--- a/src/components/Questions/QuestionShort.vue
+++ b/src/components/Questions/QuestionShort.vue
@@ -37,42 +37,189 @@
 				dir="auto"
 				:maxlength="maxStringLengths.answerText"
 				minlength="1"
-				type="text"
+				:type="validationObject.inputType"
+				:step="validationObject.inputType === 'number' ? 'any' : undefined"
 				@input="onInput"
 				@keydown.enter.exact.prevent="onKeydownEnter">
+			<NcActions v-if="edit"
+				:id="validationTypeMenuId"
+				:aria-label="t('forms', 'Input types')"
+				:container="`#${validationTypeMenuId}`"
+				:open.sync="isValidationTypeMenuOpen"
+				class="validation-type-menu__toggle"
+				type="tertiary-no-background">
+				<template #icon>
+					<component :is="validationObject.icon" :size="20" />
+				</template>
+				<NcActionRadio v-for="(validationTypeObject, validationTypeName) in validationTypes"
+					:key="validationTypeName"
+					:checked="validationType === validationTypeName"
+					:name="validationTypeName"
+					@update:checked="onChangeValidationType(validationTypeName)">
+					{{ validationTypeObject.label }}
+				</NcActionRadio>
+				<NcActionInput v-if="validationType === 'regex'"
+					ref="regexInput"
+					:label="t('forms', 'Regular expression for input validation')"
+					:value="validationRegex"
+					@input="onInputRegex"
+					@submit="onSubmitRegex">
+					<template #icon>
+						<IconRegex :size="20" />
+					</template>
+					/^[a-z]{3}$/i
+					<!-- ^ Some example RegExp for the placeholder text -->
+				</NcActionInput>
+			</NcActions>
 		</div>
 	</Question>
 </template>
 
 <script>
+import { splitRegex, validateExpression } from '../../utils/RegularExpression.js'
+
+import validationTypes from '../../models/ValidationTypes.js'
 import QuestionMixin from '../../mixins/QuestionMixin.js'
+
+import IconRegex from 'vue-material-design-icons/Regex.vue'
+import NcActions from '@nextcloud/vue/dist/Components/NcActions.js'
+import NcActionInput from '@nextcloud/vue/dist/Components/NcActionInput.js'
+import NcActionRadio from '@nextcloud/vue/dist/Components/NcActionRadio.js'
 
 export default {
 	name: 'QuestionShort',
 
+	components: {
+		IconRegex,
+		NcActions,
+		NcActionInput,
+		NcActionRadio,
+	},
+
 	mixins: [QuestionMixin],
 
+	data() {
+		return {
+			validationTypes,
+			typeMenuOpen: false,
+		}
+	},
 	computed: {
 		submissionInputPlaceholder() {
-			if (this.readOnly) {
-				return this.answerType.submitPlaceholder
+			if (this.edit) {
+				return this.validationObject.createPlaceholder || this.answerType.createPlaceholder
 			}
-			return this.answerType.createPlaceholder
+			return this.validationObject.submitPlaceholder || this.answerType.submitPlaceholder
+		},
+		/**
+		 * Current user input validation type
+		 */
+		validationObject() {
+			return validationTypes[this.validationType]
+		},
+		/**
+		 * Name of the current validation type, fallsback to 'text'
+		 */
+		validationType() {
+			return this.extraSettings?.validationType || 'text'
+		},
+		/**
+		 * Id of the validation type menu
+		 */
+		validationTypeMenuId() {
+			return 'q' + this.$attrs.index + '__validation_menu'
+		},
+		/**
+		 * The regular expression
+		 */
+		validationRegex() {
+			return this.extraSettings?.validationRegex || ''
 		},
 	},
 
 	methods: {
 		onInput() {
+			/** @type {HTMLObjectElement} */
 			const input = this.$refs.input
-			this.$emit('update:values', [input.value])
+			/** @type {string} */
+			const value = input.value
+
+			input.setCustomValidity('')
+
+			// Only check non empty values, this question might not be required, if not already invalid
+			if (value) {
+				// Then check native browser validation (might be better then our)
+				// If the browsers validation succeeds either the browser does not implement a validation
+				// or it is valid, so we double check by running our custom validation.
+				if (!input.checkValidity() || !this.validationObject.validate(value, splitRegex(this.validationRegex))) {
+					input.setCustomValidity(this.validationObject.errorMessage)
+				}
+			}
+
+			this.$emit('update:values', [value])
+		},
+
+		/**
+		 * Change input type
+		 *
+		 * @param {string} validationType new input type
+		 */
+		onChangeValidationType(validationType) {
+			if (validationType === 'regex') {
+				// Make sure to also submit a regex (even if empty)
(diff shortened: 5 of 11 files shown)


jancborchardt (approved):
Very cool, looks good design-wise! :)

Optional detail: "Telephone number" sounds a bit off and could be shortened to "Phone number".

susnux:
> Very cool, looks good design-wise! :)

Thank you :)

> Optional detail: "Telephone number" sounds a bit off and could be shortened to "Phone number".

I agree! Changed it.

src/utils/RegularExpression.js
```diff
@@ -0,0 +1,45 @@
...
+	if (regex.length === 0) return true
+
+	// Validate regex has delimters
+	if (!REGEX_WITH_DELIMITERS.test(regex)) return false
+
+	// Check pattern is escaped
+	const [pattern, modifiers] = splitRegex(regex)
+	if (REGEX_UNESCAPED_SLASH.test(pattern)) return false
+
+	// Validate regular expression
```
jotoeri, line 27: Reasonable information on the comment, please. The whole function does the validation, but this does specifically test the compilation to check if usable.

src/utils/RegularExpression.js
```diff
@@ -0,0 +1,45 @@
...
+	try {
+		(() => new RegExp(pattern, modifiers))()
```
jotoeri, line 29: Why as a function? 🤔 
To disable linting? Then `// eslint-disable-next-line no-new` with proper comment should do the trick.

src/utils/RegularExpression.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Split an enclosed regular expression into pattern and modifiers
+ *
+ * @param {string} regex regular expression with delimiters
+ * @return {[string, string]} pattern and modifiers
+ */
+export function splitRegex(regex) {
+	const [, pattern, modifiers] = regex.match(REGEX_WITH_DELIMITERS) || ['', '', '']
+	return [pattern, modifiers]
```
jotoeri, line 44: using object?

jotoeri (commented):
Huhh, that was a big one.
Sorry for that mass of comments, but it feels necessary. 🙈 
Especially on the two contants files, there is on each one a bigger structure-comment. Maybe check them first. Most of the others are smaller things.

src/utils/RegularExpression.js
susnux: I am using IIFE syntax to make clear this is not just an unused expression, but I expect some sideeffects (exception if invalid).

lib/Service/SubmissionService.php
```diff
@@ -363,4 +374,42 @@ private function validateDateTime(string $dateStr, string $format) {
...
+			case 'number':
+				return is_numeric($data);
+			case 'tel':
+				// some special characters are used (depending on the locale)
+				$sanitized = str_replace([' ', '(', ')', '.', '/', '-', 'x'], '', $data);
+				// allow leading + for international numbers
+				if (str_starts_with($sanitized, '+')) {
+					$sanitized = substr($sanitized, 1);
+				}
+				return preg_match('/^[0-9]{3,}$/', $sanitized) === 1;
```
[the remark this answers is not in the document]
susnux: Validating a phone number is quite hard, there are sooo many different patterns.

The basically removes the most common separators like `0123 456`, `012345-6`, `0123/456`, `(012) 3456` or the `x` is used in the US like `1-234-56 x 7`. And then validates the rest are only numbers.

src/components/Questions/QuestionShort.vue
```diff
@@ -44,44 +44,245 @@
...
+			/** @type {HTMLObjectElement} */
 			const input = this.$refs.input
-			this.$emit('update:values', [input.value])
+			/** @type {string} */
+			const value = input.value
+
+			input.setCustomValidity('')
+
+			// Only check non empty values, this question might not be required, if not already invalid
+			if (!!value && input.checkValidity()) {
```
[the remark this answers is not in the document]
susnux: >  Also i'd say we don't need a fallback: https://caniuse.com/?search=input.checkValidity

The fallback is required for e.g. `tel` as Firefox does not validate a phone number.

So the idea is: If the browser can validate it and sets `invalid` we do not need custom validation as it is already invalid.
If the browser says "it's valid" we can not be sure it really validated the input so for "valid" input we double check by using our own validation.

Of cause we could just use our own validation, but then we can not make use of (maybe better) browser validation.

src/components/Questions/QuestionShort.vue
```diff
@@ -44,44 +44,245 @@
...
+		 * Change input type
+		 *
+		 * @param {string} type new input type
+		 */
+		onChangeType(type) {
+			if (type !== 'custom') {
+				// keep menu open to enter regex if custom input type
+				this.typeMenuOpen = false
+				// remove regex if set
+				this.onExtraSettingsChange({ type, regex: undefined })
```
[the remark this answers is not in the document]
susnux: :+1:

src/components/Questions/QuestionShort.vue
```diff
@@ -44,44 +44,245 @@
...
+			}
 		},
 	},
 }
 </script>
 
 <style lang="scss" scoped>
-// Using type to have a higher order than the input styling of server
-.question__input[type=text] {
+input.question__input {
```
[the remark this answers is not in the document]
susnux: It is required, as we can no longer use the `[type=text]` to get the higher priority.

lib/Constants.php
```diff
@@ -99,6 +99,33 @@ class Constants {
...
+		'text',
+		'tel',
+		'email',
+		'custom',
+		'number'
+	];
+
+	// This are allowed extra settings
+	public const EXTRA_SETTINGS_DROPDOWN = [
+		'shuffleOptions' => true
```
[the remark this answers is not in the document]
susnux: The idea was to also validate the value type, but I did not finished that. So I think I will undo that and simply use an array of allowed keys.

lib/Controller/ApiController.php
```diff
@@ -655,6 +655,10 @@ public function updateQuestion(int $id, array $keyValuePairs): DataResponse {
 			throw new OCSForbiddenException('Please use reorderQuestions() to change order');
 		}
 
+		if (key_exists('extraSettings', $keyValuePairs) && ! $this->formsService->validateExtraSettings($keyValuePairs['extraSettings'], $question->getType())) {
```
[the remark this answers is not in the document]
susnux: maybe a CS fixer bug ;-)

lib/Service/FormsService.php
```diff
@@ -539,4 +539,80 @@ public function setLastUpdatedTimestamp(int $formId): void {
...
 	}
+
+	/*
+	 * Validates the extraSettings
+	 *
+	 * @param array $extraSettings input extra settings
+	 * @param string $questionType the question type
+	 * @return bool if the settings are valid
+	 */
+	public function validateExtraSettings(array $extraSettings, string $questionType) {
```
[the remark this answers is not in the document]
susnux: I agree :+1:

lib/Service/FormsService.php
```diff
@@ -539,4 +539,80 @@ public function setLastUpdatedTimestamp(int $formId): void {
...
+					// only pattern with delimiters and supported modifiers (by PHP *and* JS)
+					return false;
+				}
+
+				if (@preg_match($REGEX_UNESCAPED_SLASHES, $matches[1]) === 1) {
+					// unescaped slashes
+					return false;
+				}
+
+
```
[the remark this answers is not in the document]
susnux: `preg_match` return the number of match or `false` if the pattern is invalid, I use this to ensure the pattern is correct (while not being interested in the action matches).

Added a comment for this.

lib/Service/FormsService.php
```diff
@@ -539,4 +539,80 @@ public function setLastUpdatedTimestamp(int $formId): void {
...
+			case Constants::ANSWER_TYPE_SHORT:
+				$allowed = Constants::EXTRA_SETTINGS_SHORT;
+				break;
+			default:
+				$allowed = [];
+		}
+		$diff = array_diff_key($extraSettings, $allowed);
+		if (count($diff) > 0) {
+			return false;
+		}
```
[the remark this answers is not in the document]
susnux: No it does not, as it only returns the keys of the first array (`$extraSettings`) which are not provided by any other array (`$allowed`).

So it ensures every key of the extra settings array is allowed, but not the opposite (it does not enforce that all allowed keys are present).

src/models/constants.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const SHORT_INPUT_TYPES = [
+	'text',
+	'tel',
+	'email',
+	'custom',
+	'number',
+]
+
+// Keep in sync with `SHORT_INPUT_TYPES`
```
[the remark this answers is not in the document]
susnux: Good idea!

src/mixins/QuestionMixin.js
```diff
@@ -173,12 +173,14 @@ export default {
...
+		 * Either an object containing the changed settings can be given or
+		 * the first paremter defines the settings name and the second parameter the value.
+		 *
+		 * @param {[object]|[string, any]} args changed settings
 		 */
-		onExtraSettingsChange: debounce(function(parameter, value) {
-			const newSettings = Object.assign({}, this.extraSettings)
-			newSettings[parameter] = value
+		onExtraSettingsChange: debounce(function(...args) {
+			const newSetting = args.length === 1 ? args[0] : { [args[0]]: args[1] }
```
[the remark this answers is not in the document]
susnux: sound reasonable, I think allowing multiple settings to be changed at once is better for possible future usage.

src/components/Questions/QuestionShort.vue
```diff
@@ -44,44 +44,245 @@
...
+		 * Ensures the regex is enclosed with delimters, as required for PCRE,
+		 * and regex is only using modifiers supported by JS *and* PHP
+		 *
+		 * @param {InputEvent|SubmitEvent} event input event
+		 * @return {boolean} true if the regex is valid
+		 */
+		onInputRegex(event) {
+			if (event?.isComposing) return false
+
+			const input = this.$refs.regexInput.$el.querySelector('input')
```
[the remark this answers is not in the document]
susnux: No, because the `ref` is only pointing to the `NcActionInput` which is a `<li>` element, so we need to get the inner input element to trigger the custom validity.

susnux:
@jotoeri thank you for that long review!

I resolved all comments which I fixed and answered the other comments :)

src/components/Questions/QuestionShort.vue
```diff
@@ -44,44 +44,197 @@
...
+
+			this.$emit('update:values', [value])
+		},
+
+		/**
+		 * Change input type
+		 *
+		 * @param {string} type new input type
+		 */
+		onChangeType(type) {
```
jotoeri, line 170: validationType

jotoeri (commented):
🤷‍♂️

src/components/Questions/QuestionShort.vue
```diff
@@ -44,44 +44,197 @@
...
+				:id="typeMenuId"
+				:aria-label="t('forms', 'Input types')"
+				:container="`#${typeMenuId}`"
+				:open.sync="typeMenuOpen"
+				class="validation-toggle"
+				type="tertiary-no-background">
+				<template #icon>
+					<component :is="type.icon" :size="20" />
+				</template>
+				<NcActionRadio v-for="(optionValue, optionName) in validationTypes"
```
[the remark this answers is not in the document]
susnux: > Variable 'validationObject' is already declared in the upper scope

I now renamed it to the very verbose name `validationTypeObject`

src/utils/RegularExpression.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Validate a regex, ensures enclosed with delimiters and only supported modifiers by PHP *and* JS
+ */
+const REGEX_WITH_DELIMITERS = /^\/(.+)\/([smi]{0,3})$/
+/**
+ * Find unescaped slashes within a string (ignores a slash at the beginning as this is the delimiter)
```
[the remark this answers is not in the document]
susnux: The comment was misleading, sorry :see_no_evil:

src/mixins/QuestionMixin.js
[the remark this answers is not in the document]
susnux: :facepalm: sry

src/components/Questions/QuestionShort.vue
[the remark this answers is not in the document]
susnux: So you recommend to not use the browser built in validation but only use our own?

src/components/Questions/QuestionShort.vue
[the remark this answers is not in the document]
susnux: I checked this again and I think we can just remove the `input`, even the type was not required anymore.
Probably from some old NC version styling?

lib/Service/SubmissionService.php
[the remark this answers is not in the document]
susnux: I know, but there are only two possible ways:
1. Try to match as many valid pattern as possible, surely there might be invalid ones like your example.
2. Only allow a specific pattern, e.g. only ISO formatted.

1 might validate invalid pattern and 2 will probably annoy users which accustomed to use a locale pattern.

In the end I personally think we should allow many pattern, but form creators should, as always, validate the numbers. Because even if a pattern matches it does not mean it is a valid number (best example: it might not be assigned).

But I am open for your suggestions.

lib/Constants.php
```diff
@@ -99,6 +99,33 @@ class Constants {
 		self::ANSWER_TYPE_TIME => 'H:i'
 	];
 
+	/**
+	 * !! Keep in sync with src/constants.js !!
+	 */
+
+	// Allowed short input types
+	public const SHORT_INPUT_TYPES = [
+		'text',
```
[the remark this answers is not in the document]
susnux: You are right, `text` and no validation is the same, I think having an explicit `text` validation type makes it more clear and you can simply fallback to it if no validation is set.

> but show different UI

Where do they differ? As soon as no validation is set everything should fallback as if `text` was set as the validation type.

lib/Service/SubmissionService.php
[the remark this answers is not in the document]
Chartman123: @jotoeri @susnux what about using ~~ISO~~ [E.164](https://en.wikipedia.org/wiki/E.164) notation here for telephone numbers and tell the users to use the custom regex for special number formats? So then the creator of the form can choose to have the simple solution with E.164 checking and more complex solution with custom number formats in the form?

src/models/ValidationTypes.js
```diff
@@ -0,0 +1,121 @@
...
+	/**
+	 * Email address validation
+	 *
+	 * @type {ValidationType}
+	 */
+	email: {
+		icon: IconEMail,
+		inputType: 'email',
+		label: t('forms', 'Email address'),
+		validate: (input) => /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/.test(input),
```
Chartman123, line 88: With this validation pattern we're not matching all valid mail addresses... For example + is not allowed in your validation, but in RFC

```suggestion
		validate: (input) => /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,24}))$/.test(input),
```
I got this one from https://codepen.io/CSWApps/pen/MmpBjV

However, most sites suggest to use an API for address validation as the complexity needed for a complete regex validation hits performance. But I'm not sure if we want to leave NC for another service here.

Chartman123: Also some more input on this from Mozilla: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#validation

lib/Service/SubmissionService.php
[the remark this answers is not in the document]
Chartman123: Browser by the way don't provide a validation of telephone numbers because it's not reasonable: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel

irdiOL:
Are these fields already in forms because i can not see them in the latest version of the app?

Chartman123:
As you can see this PR is still open and thus not merged into the main branch.

irdiOL:
Thanks for taking the time to answer my question.

Sent from Proton Mail mobile

-------- Original Message --------

src/models/ValidationTypes.js
susnux: Ok I would vote for making the frontend code simpler. On the backend we already use the Nextcloud core service to validate the email.

So on frontend we should just make sure it somehow looks like an email. how about:
`[^@]+@[^@]+`
This one will only forbid one case: quoted `@` symobol (as something like `"foo@bar"@example.com` is valid.

lib/Constants.php
```diff
@@ -100,6 +100,33 @@ class Constants {
...
+	];
+
+	// This are allowed extra settings
+	public const EXTRA_SETTINGS_DROPDOWN = [
+		'shuffleOptions'
+	];
+
+	public const EXTRA_SETTINGS_MULTIPLE = [
+		'shuffleOptions'
+	];
```
Chartman123, line 123: For those two question types we have the second allowed extra setting "Allow other answer" now :)

lib/Service/SubmissionService.php
```diff
@@ -342,6 +347,11 @@ public function validateSubmission(array $questions, array $answers): bool {
 						}
 					}
 				}
+
+				// Handle custom validation of short answers
+				if ($question['type'] === Constants::ANSWER_TYPE_SHORT && !$this->validateShortQuestion($question, $answers[$questionId][0])) {
```
Chartman123, line 352: Why don't you check for the set extraSettings -> validationType already here?

lib/Service/SubmissionService.php
```diff
@@ -367,4 +377,42 @@ private function validateDateTime(string $dateStr, string $format) {
...
+	/**
+	 * Validate short question answers if special validation types are set
+	 */
+	private function validateShortQuestion(array $question, string $data): bool {
+		if (!isset($question['extraSettings']) || !isset($question['extraSettings']['validationType'])) {
+			// No type defined, so fallback to 'text' => no special handling
+			return true;
+		}
+
+		switch ($question['extraSettings']->validationType) {
```
Chartman123, line 390:
```suggestion
		switch ($question['extraSettings']['validationType']) {
```

lib/Service/SubmissionService.php
```diff
@@ -367,4 +377,42 @@ private function validateDateTime(string $dateStr, string $format) {
...
+				if (str_starts_with($sanitized, '+')) {
+					$sanitized = substr($sanitized, 1);
+				}
+				return preg_match('/^[0-9]{3,}$/', $sanitized) === 1;
+			case 'regex':
+				// empty regex matches everything
+				if (!isset($question['extraSettings']['validationRegex'])) {
+					return true;
+				}
+				return preg_match($question['extraSettings']->validationRegex, $data) === 1;
```
Chartman123, line 411:
```suggestion
				return preg_match($question['extraSettings']['validationRegex'], $data) === 1;
```

tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -632,6 +637,42 @@ public function dataValidateSubmission() {
 				// Expected Result
 				false
 			],
+			'invalid-short-phone' => [
+				// Questions
+				[
+					['id' => 1, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['type' => 'tel']]
```
Chartman123, line 643:
```suggestion
					['id' => 1, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'tel']]
```
Shouldn't the parameters be `validationType` instead of `type`

lib/Service/SubmissionService.php
susnux: :see_no_evil:

tests/Unit/Service/SubmissionServiceTest.php
susnux: Rebase issue :see_no_evil:

lib/Constants.php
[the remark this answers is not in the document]
susnux: Text removed as this is the same as empty :heavy_check_mark:

lib/Service/SubmissionService.php
susnux: Because that is part of the implementation of `validateShortQuestion`, no need for other functions to use that "internal" logic

susnux: (meaning `validationType` is only for short answer type, so `validateSubmission` does not need to know any logic required for a specific answer type except the validator function `validateShortQuestion`).

lib/Service/SubmissionService.php
```diff
@@ -367,4 +377,39 @@ private function validateDateTime(string $dateStr, string $format) {
...
+				}
+				return preg_match('/^[0-9]{3,}$/', $sanitized) === 1;
+			case 'regex':
+				// empty regex matches everything
+				if (!isset($question['extraSettings']['validationRegex'])) {
+					return true;
+				}
+				return preg_match($question['extraSettings']['validationRegex'], $data) === 1;
+			default:
+				$this->logger->error('Invalid input type for question detected, please report this issue!', ['validationType' => $question['extraSettings']->validationType]);
```
Chartman123, line 410:
```suggestion
				$this->logger->error('Invalid input type for question detected, please report this issue!', ['validationType' => $question['extraSettings']['validationType']]);
```
Another one that slipped through...

src/components/Questions/QuestionShort.vue
```diff
@@ -37,42 +37,188 @@
...
+		return {
+			validationTypes,
+			isValidationTypeMenuOpen: false,
+		}
+	},
 	computed: {
 		submissionInputPlaceholder() {
-			if (this.readOnly) {
-				return this.answerType.submitPlaceholder
+			if (this.edit) {
```
Chartman123, line 108:
```suggestion
			if (this.!readOnly) {
```

susnux: I could bet I grepped the file for "edit"....

lib/Service/SubmissionService.php
```diff
@@ -367,4 +377,39 @@ private function validateDateTime(string $dateStr, string $format) {
...
+				return is_numeric($data);
+			case 'phone':
+				// some special characters are used (depending on the locale)
+				$sanitized = str_replace([' ', '(', ')', '.', '/', '-', 'x'], '', $data);
+				// allow leading + for international numbers
+				if (str_starts_with($sanitized, '+')) {
+					$sanitized = substr($sanitized, 1);
+				}
+				return preg_match('/^[0-9]{3,}$/', $sanitized) === 1;
+			case 'regex':
```
Chartman123, line 403:
```suggestion
			case 'custom':
```

tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -664,9 +705,15 @@ public function dataValidateSubmission() {
...
-					['id' => 9, 'type' => 'multiple_unique', 'isRequired' => true, 'extraSettings' => ['allowOtherAnswer' => true], 'options' => [
+					['id' => 9, 'type' => 'multiple_unique', 'isRequired' => true, 'extraSettings' => [
+						'allowOtherAnswer' => true,
+					], 'options' => [
 						['id' => 3]
 					]],
+					['id' => 10, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'email']],
+					['id' => 11, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'number']],
+					['id' => 12, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'tel']],
+					['id' => 13, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'custom', 'validationRegex' => '/[a-z]{3}[0-9]{3}/']],
```
Chartman123, line 716:
```suggestion
					['id' => 13, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'regex', 'validationRegex' => '/[a-z]{3}[0-9]{3}/']],
```

tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -632,6 +637,42 @@ public function dataValidateSubmission() {
 				// Expected Result
 				false
 			],
+			'invalid-short-phone' => [
+				// Questions
+				[
+					['id' => 1, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'tel']]
```
Chartman123, line 643:
```suggestion
					['id' => 1, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'phone']]
```

tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -632,6 +637,42 @@ public function dataValidateSubmission() {
...
+				[
+					'1' => ['0800 NEXTCLOUD']
+				],
+				// Expected Result
+				false
+			],
+			'invalid-short-regex-not-matching' => [
+				// Questions
+				[
+					['id' => 1, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'custom', 'validationRegex' => '/[a-z]{4}/']]
```
Chartman123, line 655:
```suggestion
					['id' => 1, 'type' => 'short', 'isRequired' => false, 'extraSettings' => ['validationType' => 'regex', 'validationRegex' => '/[a-z]{4}/']]
```

tests/Unit/Service/SubmissionServiceTest.php
susnux: This is weird as in this case the test should fail but succeeded.

Chartman123: You expect `false` which is returned by the validation if a wrong type is used. Or am I wrong?

Chartman123 (approved):
I will try it in my dev instance this evening, but codewise looks good now :)

src/components/Questions/Question.vue
```diff
@@ -330,6 +330,10 @@ export default {
 	> * {
 		cursor: pointer;
 	}
+	input[disabled] {
+		// forward clicks to parent so we can toggle the edit mode
+		pointer-events: none;
+	}
```
Chartman123, line 337: Is this still needed with reworked create/edit mode?

src/components/Questions/QuestionShort.vue
```diff
@@ -37,42 +37,188 @@
...
+			/** @type {string} */
+			const value = input.value
+
+			input.setCustomValidity('')
+
+			// Only check non empty values, this question might not be required, if not already invalid
+			if (value) {
+				// Then check native browser validation (might be better then our)
+				// If the browsers validation succeeds either the browser does not implement a validation
+				// or it is valid, so we double check be running our custom validation.
```
Chartman123, line 152:
```suggestion
				// or it is valid, so we double check by running our custom validation.
```

src/components/Questions/QuestionShort.vue
```diff
@@ -37,42 +37,188 @@
 				dir="auto"
 				:maxlength="maxStringLengths.answerText"
 				minlength="1"
-				type="text"
+				:type="validationObject.inputType"
```
Chartman123, line 40:
```suggestion
				:type="validationObject.inputType"
                :step="validationObject.inputType === 'number' ? 'any' : null"
```

PR #1493, opened 2023-02-10
fix: Regression if user is not logged in `getPermissions`

susnux:
If the user is not logged in `getPermissions` will try to call `getUID` on `null`.
This makes public shares impossible (`getPublicForm` calls `getForm` which calls `getPermissions`).

### DIFF
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -271,9 +271,11 @@ class FormsService {
 		}
 
 		$permissions = [];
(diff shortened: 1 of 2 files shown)


Chartman123:
@susnux Is this addressing the same issue as #1492 ?

susnux:
Oh yes, looks like @jotoeri and I find that bug the same time :see_no_evil:

jotoeri:
Well, i was 5hrs earlier 😛 .
Also this version has the issue, that you add the permission submit by the permitAll-Setting, which is only thought for logged in users.

PR #1496, opened 2023-02-11
Notify users on new submissions for shared forms

susnux:
Allows users to receive notifications for new submissions of shared forms, meaning forms of which the user has *results* permissions.
Currently only the owner is notified but not other users.

Use cases: Not only you but the whole team should be informed of new submissions.

I have added a new notification setting / type for this, so users can disable e.g. mail notifications for shared forms but still get them for their own forms.


lib/Service/FormsService.php
```diff
@@ -493,6 +493,25 @@ public function notifyNewShares(Form $form, Share $share): void {
 		}
 	}
 
+	/**
+	 * Creates activities for new submissions on a form
+	 *
+	 * @param Form $form Related Form
+	 * @param string $submitter The ID of the user who submitted the form. Can also be our 'anon-user-'-ID
+	 */
+	public function notifyNewSubmission(Form $form, string $submitter): void {
```
jotoeri, line 502: Can we put this into the activityManager?
That is purely activity-related and the formsService is already quite crowded...

lib/Activity/ActivityManager.php
```diff
@@ -109,21 +110,58 @@ public function publishNewGroupShare(Form $form, string $groupId) {
...
+	 * @param Form $form The affected Form
+	 * @param string $submitterID ID of the User who submitted the form. Can also be our 'anon-user-'-ID
+	 */
+	public function publishNewSharedSubmission(Form $form, int $shareType, string $shareWith, string $submitterID) {
+		$users = [];
+		switch ($shareType) {
+			case IShare::TYPE_USER:
+				$users[] = $shareWith;
+				break;
+			case IShare::TYPE_GROUP:
```
jotoeri, line 143: Add circles here directly? 😉

tests/Unit/Activity/ActivityManagerTest.php
```diff
@@ -175,4 +176,100 @@ public function testPublishNewSubmission() {
...
+		])->willReturn($event);
+
+		$this->manager->expects($this->exactly(count($expected)))
+			->method('publish')
+			->with($event);
+
+		$this->activityManager->publishNewSharedSubmission($form, $shareType, $shareWith, $submitterId);
+	}
+
+	public function publishNewSharedSubmission(Form $form, int $shareType, string $shareWith, string $submitterID) {
```
jotoeri, line 245: rm

lib/Activity/ActivityManager.php
susnux: My plan was to either do it here or in the circles PR, depending on which one is merged first (and do in the later one).

lib/Service/FormsService.php
susnux: I used used the pattern already used, see function above (`notifyNewShares`).
But I could try to move both into the activity manager and see how it work out if you like.

tests/Unit/Activity/ActivityManagerTest.php
susnux: :facepalm:

lib/Service/FormsService.php
susnux: @jotoeri I can move it but it would introduce a dependency on the activity manager in the ShareAPIController.
Do we want that additional dependency and remove one layer of indirection?

Chartman123: I'd suggest to keep it the way the current notifications are implemented. We could move that code to ActivityManager in a later PR

Chartman123 (approved):
Looks good :) Just fix the failing workflows and the we can get this merged.

PR #1529, opened 2023-02-21
Add information about API to appinfo

susnux:
I think this information is missing on the app store page, as it is a nice feature we can advertise.


Chartman123 (approved):

jotoeri (approved):

jancborchardt (commented):
Nice addition! :)

PR #1531, opened 2023-02-21
Make timestamp in csv export ISO 8601 compliant

### LINKED ISSUE #1482: Timestamp in CSV export should follow ISO 8601
When exporting query answers as a CSV file, the timestamp column contains the date and time of the respective answers. These are long strings, like, e. g., “**Donnerstag, 26. Januar 2023 um 13:39:50 Mitteleuropäische Normalzeit**”.

However, date and time should rather be expressed following the ISO 8601 standard, which exists just for that purpose. The timestamp from my example above would then be “**2023-01-26T13:39:50+01:00**”. This format is still human-readable, but it can also be used for sorting, and it can be easier and more generally be parsed, because there’s no language dependency.

**Nextcloud**
- Nextcloud-Version: Can’t find out on the system I am using, but it should be 25.something.
- Forms-Version: Can’t find out on the system I am using, but it should be 3.0.3 or so.

(Before filing this issue, I wondered whether this is a bug or a feature request, but I believe it is rather a bug.)

Chartman123:
Fixes #1482 

- [x] Fix csv export
- [x] Adjust tests

### DIFF
--- a/tests/Unit/Service/SubmissionServiceTest.php
+++ b/tests/Unit/Service/SubmissionServiceTest.php
@@ -40,7 +40,6 @@ use OCP\Files\Folder;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 use OCP\IConfig;
-use OCP\IDateTimeFormatter;
 use OCP\IL10N;
 use OCP\IUser;
 use OCP\IUserManager;
@@ -74,9 +73,6 @@ class SubmissionServiceTest extends TestCase {
 	/** @var IConfig|MockObject */
 	private $config;
 
-	/** @var IDateTimeFormatter|MockObject */
-	private $dateTimeFormatter;
-
 	/** @var IL10N|MockObject */
 	private $l10n;
 
@@ -94,7 +90,6 @@ class SubmissionServiceTest extends TestCase {
 		$this->answerMapper = $this->createMock(AnswerMapper::class);
 		$this->storage = $this->createMock(IRootFolder::class);
 		$this->config = $this->createMock(IConfig::class);
-		$this->dateTimeFormatter = $this->createMock(IDateTimeFormatter::class);
 		$this->l10n = $this->createMock(IL10N::class);
 		$this->logger = $this->getMockBuilder('Psr\Log\LoggerInterface')->getMock();
 		$this->userManager = $this->createMock(IUserManager::class);
@@ -121,7 +116,6 @@ class SubmissionServiceTest extends TestCase {
 			$this->answerMapper,
 			$this->storage,
 			$this->config,
-			$this->dateTimeFormatter,
 			$this->l10n,
 			$this->logger,
 			$this->userManager,
@@ -285,6 +279,7 @@ class SubmissionServiceTest extends TestCase {
 					[
 						'id' => 1,
 						'userId' => 'user1',
+						'timestamp' => 123456789,
 						'answers' => [
 							['questionId' => 1, 'text' => 'Q1A1'],
 							['questionId' => 2, 'text' => 'Q2A1']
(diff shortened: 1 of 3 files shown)


tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -511,9 +505,9 @@ private function setUpCsvTest(array $questions, array $submissions, string $csvT
 			]));
 
 		// Just using any timestamp here
-		$this->dateTimeFormatter->expects($this->any())
-			->method('formatDateTime')
-			->willReturn('01.01.01, 01:01');
+		//$this->dateTimeFormatter->expects($this->any())
+		//	->method('formatDateTime')
+		//	->willReturn('1970-01-01T01:00:00+01:00');
```
Chartman123, line 510: @jotoeri As I'm not that much familiar with PHPUnit, I don't know exactly what to do here to adjust it to the new formatting. Could you please help me a little?

Chartman123:
@jotoeri should we put this one also into 3.1.1?

tests/Unit/Service/SubmissionServiceTest.php
jotoeri: In fact - nothing. 😅 
Actually, this code was mocking the procedure of the Nextcloud-DateTimeFormatter, that is not existent during the test. Since you now only use php-native functions, they are available and just do, what they should do. So you can just remove the commented code.

Interestingly the tests fail locally for me with `date_timestamp_set(): Passing null to parameter #2 ($timestamp) of type int is deprecated`. What we currently don't do, is to set a timestamp at all on the submissions. So to improve the tests, we could just set a timestamp like `123456789` on the submissions in dataGetSubmissionsCsv and accordingly adapt the expected timestamps to `1973-11-29T22:33:09+01:00`.

jotoeri:
> @jotoeri should we put this one also into 3.1.1?

Hmm. I think i wouldn't do this. It's actually not clear if this is a bug, or just a decision we took. I think we even talked about somewhen. And if people are somehow used to the current format, then this shouldn't change just by a patch release. So i think i'd keep it for 3.2.

tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -288,22 +282,24 @@ public function dataGetSubmissionsCsv() {
 						'answers' => [
 							['questionId' => 1, 'text' => 'Q1A1'],
 							['questionId' => 2, 'text' => 'Q2A1']
-						]
+						],
+						'timestamp' => 123456789
```
jotoeri, line 286: 👀 Logically, i would place it above the answers.
But thats a minor thing.

tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -446,7 +446,7 @@ private function setUpSimpleCsvTest(): string {
 			// Expected CSV-Result
 			'
 			"User ID","User display name","Timestamp","Question 1"
-			"user1","User 1","01.01.01, 01:01","Q1A1"
+			"user1","User 1","1970-01-01T01:00:00+01:00","Q1A1"
```
jotoeri, line 449: Also timestamp here?

Chartman123:
@jotoeri Yes, makes more sense to have it in the same order as in the csv 👍🏻 

I've worked in your comments now

jotoeri (approved):

susnux (approved):

jotoeri:
Just found, that we merged enhancements already. So it'll anyways become 3.2.0 🙈

susnux:
> we merged enhancements already

The `lastUpdated` property was an enhancement I guess?

Issue #1263, opened 2022-07-20
Lighthouse accessibility failures


PVince81:
Run Chrome's Lighthouse tool after creating a simple form with radio buttons and dropdowns:

<img width="423" alt="image" src="https://user-images.githubusercontent.com/277525/180016791-fc8a460a-9e64-4217-96f6-de55def778ff.png">

Note: the avatardiv is unrelated as it's in the header.

This was observed in forms version 2.5.1 on Nextcloud 23.0.7

susnux:
Context: https://dequeuniversity.com/rules/axe/4.4/aria-allowed-attr?utm_source=lighthouse&utm_medium=devtools

Issue #317, opened 2020-04-30
Reorder question via arrows


skjnldsv:
Instead of drag and drop, support arrox reordering when focusing the drag handle.
Might also be interesting to show two arrows buttons when hovering/focusing the handle that can also be clicked?

cc @jancborchardt

jancborchardt:
Yep, both suggestions are good enhancements. For the "support arrow reordering when focusing the drag handle" did you check if that is a somewhat standard in a way that it’s used by other apps too?

skjnldsv:
We talked abou this already xD
Not done specifically, but A11y have a lot of specifications where using arrows key to change items are used and recommended.

ei8fdb:
:100: agree. Reordering arrows would would much better here. I had the same issue. See timestamp 06:05 to 07:32 in [this testing video](https://www.flickr.com/photos/runningwithbulls/49862099823/in/datetaken/).

skjnldsv:
> See timestamp 06:05 to 07:32 in [this testing video](https://www.flickr.com/photos/runningwithbulls/49862099823/in/datetaken/).

I'm guessing you mean this video https://www.flickr.com/photos/runningwithbulls/49862099753/
Can you open an issue? Drag&drop should scroll the window indeed :+1:

PR #1532, opened 2023-02-22
Allow reordering questions using the keyboard

susnux:
### Part 1: Adding missing aria role
Resolves: #1263 
`div`s require a role when `aria-label` is set (see context in linked issue)

### Part 2: Add buttons
Resolves: #317 
Added buttons for changing the order. Buttons can be focused and triggered using the keyboard.
This also preserves the drag and drop handling, but also allows using buttons or the keyboard.

The buttons are only shown if hovered / focused, but they are always accessible for screen readers.

If a button is clicked / triggered the focus is kept on the button (the vanishing focus ring in the following video happens only because disabled buttons from `@nextcloud/vue` do not have a visual focus ring)

https://user-images.githubusercontent.com/1855448/220498206-3a1befaf-6def-45e1-9be7-980e8ee4b277.mp4

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -22,17 +22,41 @@
 
 <template>
 	<li v-click-outside="disableEdit"
-		:class="{ 'question--edit': edit }"
+		:class="{
+			'question': true,
+			'question--edit': edit,
+			'question--editable': !readOnly
+		}"
 		:aria-label="t('forms', 'Question number {index}', {index})"
-		class="question"
 		@click="enableEdit">
 		<!-- Drag handle -->
 		<!-- TODO: implement arrow key mapping to reorder question -->
 		<div v-if="!readOnly"
-			class="question__drag-handle"
-			:class="{'question__drag-handle--shiftup': shiftDragHandle}"
-			:aria-label="t('forms', 'Drag to reorder the questions')">
+			:class="{
+				'question__drag-handle': true,
+				'question__drag-handle--shiftup': shiftDragHandle
+			}">
+			<NcButton ref="buttonUp"
+				:aria-label="t('forms', 'Move question up')"
+				:disabled="!canMoveUp"
+				class="question__drag-handle-button"
+				type="tertiary-no-background"
+				@click.stop="onMoveUp">
+				<template #icon>
+					<IconArrowUp :size="20" />
+				</template>
+			</NcButton>
 			<IconDragHorizontalVariant :size="20" />
+			<NcButton ref="buttonDown"
+				:aria-label="t('forms', 'Move question down')"
+				:disabled="!canMoveDown"
+				class="question__drag-handle-button"
+				type="tertiary-no-background"
+				@click.stop="onMoveDown">
+				<template #icon>
+					<IconArrowDown :size="20" />
+				</template>
+			</NcButton>
 		</div>
 
 		<!-- Header -->
@@ -113,8 +137,11 @@ import NcActions from '@nextcloud/vue/dist/Components/NcActions.js'
 import NcActionButton from '@nextcloud/vue/dist/Components/NcActionButton.js'
 import NcActionCheckbox from '@nextcloud/vue/dist/Components/NcActionCheckbox.js'
 import NcActionInput from '@nextcloud/vue/dist/Components/NcActionInput.js'
+import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 
 import IconAlertCircleOutline from 'vue-material-design-icons/AlertCircleOutline.vue'
+import IconArrowDown from 'vue-material-design-icons/ArrowDown.vue'
+import IconArrowUp from 'vue-material-design-icons/ArrowUp.vue'
 import IconDelete from 'vue-material-design-icons/Delete.vue'
 import IconDragHorizontalVariant from 'vue-material-design-icons/DragHorizontalVariant.vue'
 import IconIdentifier from 'vue-material-design-icons/Identifier.vue'
@@ -128,6 +155,8 @@ export default {
 
 	components: {
 		IconAlertCircleOutline,
+		IconArrowDown,
+		IconArrowUp,
 		IconDelete,
 		IconDragHorizontalVariant,
 		IconIdentifier,
@@ -135,6 +164,7 @@ export default {
 		NcActionButton,
 		NcActionCheckbox,
 		NcActionInput,
+		NcButton,
 	},
 
 	inject: ['$markdownit'],
@@ -187,6 +217,14 @@ export default {
 		warningInvalid: {
 			type: String,
 			default: t('forms', 'This question needs a title!'),
+		},
+		canMoveDown: {
+			type: Boolean,
+			default: false,
+		},
+		canMoveUp: {
+			type: Boolean,
+			default: false,
 		},
 	},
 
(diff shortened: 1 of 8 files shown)


jancborchardt (changes requested):
Very nice enhancement! :) Makes it very obvious and accessible how to reorder, good stuff.

Only one thing: When reordering via these buttons, whether it be via keyboard or mouse click, it would be nice to have a quick animation of the movement using `var(--animation-quick)` or `var(--animation-slow)` depending on what looks better. :)

susnux:
> Very nice enhancement! :) Makes it very obvious and accessible how to reorder, good stuff.

Thank you :)

> Only one thing: When reordering via these buttons, whether it be via keyboard or mouse click, it would be nice to have a quick animation of the movement using `var(--animation-quick)` or `var(--animation-slow)` depending on what looks better. :)

Done. Added the animation, I choose `slow` because `quick` for long questions looks too "jumpy".


https://user-images.githubusercontent.com/1855448/220657783-6cda0693-8d11-4d95-baa1-de646dce0d8f.mp4

jancborchardt (approved):
Looks really amazing now @susnux, great work! :)

Chartman123:
@susnux I get lots of errors in the browser console like this:

```
[Vue warn]: Invalid prop: type check failed for prop "setReturnFocus". Expected Object, String, Function, Boolean, got HTMLButtonElement 

found in

---> <NcPopover>
       <NcActions>
         <RouterLink>
           <NcListItem>
             <AppNavigationForm> at src/components/AppNavigationForm.vue
               <NcAppNavigation>
                 <NcContent>
                   <Forms> at src/Forms.vue
                     <Root> 2 forms-main.js:134698:21
```

The lower part of the message varies...

And at least for me the focus ring always stays on the ⬇️ button after clicking on it with the mouse. The ⬆️ button doesn't show the ring.

susnux:
@Chartman123 the errors should be fixed with nextcloud Vue 7.7.0:
https://github.com/nextcloud/nextcloud-vue/pull/3803

Chartman123:
@susnux: With nc-vue 7.7 the errors are gone. But now I found some new ones: canMoveUp and canMoveDown are missing in the view mode and thus throwing errors.

```
[Vue warn]: Missing required prop: "canMoveUp"

found in

---> <Question> at src/components/Questions/Question.vue
       <QuestionDropdown> at src/components/Questions/QuestionDropdown.vue
         <NcAppContent>
           <Submit> at src/views/Submit.vue
             <NcContent>
               <Forms> at src/Forms.vue
                 <Root>
```

susnux:
> With nc-vue 7.7 the errors are gone. But now I found some new ones: canMoveUp and canMoveDown are missing in the view mode and thus throwing errors.

@Chartman123 those issues are resolved. And the focus issue is also resolved.

Chartman123:
> > With nc-vue 7.7 the errors are gone. But now I found some new ones: canMoveUp and canMoveDown are missing in the view mode and thus throwing errors.
> 
> @Chartman123 those issues are resolved. And the focus issue is also resolved.

Nice :) I will check it this evening

src/components/Questions/Question.vue
```diff
@@ -31,8 +31,29 @@
 		<div v-if="!readOnly"
 			class="question__drag-handle"
 			:class="{'question__drag-handle--shiftup': shiftDragHandle}"
-			:aria-label="t('forms', 'Drag to reorder the questions')">
+			:aria-label="t('forms', 'Drag to reorder the questions')"
+			role="button">
+			<NcButton ref="buttonU"
```
Chartman123, line 36:
```suggestion
			<NcButton ref="buttonUp"
```
shouldn't this be buttonUp?

src/components/Questions/Question.vue
```diff
@@ -218,6 +253,18 @@ export default {
...
+		/**
+		 * Reorder question but keep focus on the button
+		 */
+		onMoveDown() {
+			this.$emit('move-down')
+			this.$nextTick(() => this.$refs.buttonDown.$el.focus())
+		},
+		onMoveUp() {
+			this.$emit('move-up')
+			this.$nextTick(() => this.$refs.buttonU.$el.focus())
```
Chartman123, line 265:
```suggestion
			this.$nextTick(() => this.$refs.buttonUp.$el.focus())
```
Same here

src/components/Questions/QuestionDate.vue
```diff
@@ -30,10 +30,10 @@
 		:max-string-lengths="maxStringLengths"
 		:title-placeholder="answerType.titlePlaceholder"
 		:warning-invalid="answerType.warningInvalid"
+		v-on="$listeners"
 		@update:text="onTitleChange"
 		@update:description="onDescriptionChange"
-		@update:isRequired="onRequiredChange"
-		@delete="onDelete">
```
Chartman123, line 36: Can we remove the onDelete method from Question.vue if you remove it from every question type?

susnux: No. We still need to emit the `delete` event from the `Question`.

I just deleted the `onDelete` from the "parent" questions (like `QuestionDate`), as it is already included within the `v-on="$listeners"`.

src/views/Create.vue
```diff
@@ -85,15 +85,21 @@
...
-				<Questions :is="answerTypes[question.type].component"
-					v-for="(question, index) in form.questions"
-					ref="questions"
-					:key="question.id"
-					:answer-type="answerTypes[question.type]"
-					:index="index + 1"
-					:max-string-lengths="maxStringLengths"
-					v-bind.sync="form.questions[index]"
-					@delete="deleteQuestion(question)" />
+				<transition-group :move-class="isDragging ? 'nomove' : undefined">
```
jotoeri, line 88: Any reason for that nomove? A transition was desired here (see 200ms on Draggable). - Even that the old transition seems to get broken with this PR, too. 😕

susnux: If you drag it, a transition looks quite bad as when you drop the question you would expect it to be just where you dropped it and not on its previous position and moving to the new.

jotoeri: That's true - but while dragging the transition should be maintained as it is on master currently.

src/components/Questions/Question.vue
```diff
@@ -31,8 +31,29 @@
 		<div v-if="!readOnly"
 			class="question__drag-handle"
 			:class="{'question__drag-handle--shiftup': shiftDragHandle}"
-			:aria-label="t('forms', 'Drag to reorder the questions')">
+			:aria-label="t('forms', 'Drag to reorder the questions')"
+			role="button">
```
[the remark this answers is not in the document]
jancborchardt: cc accessibility experts @JuliaKirschenheuter @Pytal – what would you recommend?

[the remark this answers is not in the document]
Pytal: Since reordering can be done via keyboard with the move buttons already then there is already feature parity for keyboard-only users

We can still keep the draggable functionality for mouse users but the `role` and `aria-label` should be removed as they are never read by screenreaders anyways with the non-focusable `div`

src/views/Create.vue
susnux: Fixed

src/components/Questions/Question.vue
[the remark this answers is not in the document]
susnux: :+1: removed

susnux:
- Rebased
- Fixed a11y issues
- Fixed transition for both: Keyboard and mouse sorting


https://github.com/nextcloud/forms/assets/1855448/935cb36d-28c6-46c8-b64d-9b64e359fbe0

src/components/Questions/Question.vue
```diff
@@ -319,19 +373,32 @@ export default {
 		position: absolute;
 		display: flex;
 		inset-inline-start: 0;
+		flex-direction: column;
+		justify-content: space-around;
```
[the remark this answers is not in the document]
susnux: I agree, but then we need a `gap` otherwise they are too close. I will go for `12px` :)

Chartman123 (approved):

PR #1542, opened 2023-03-01
Fix paragraph rendering

### LINKED ISSUE #1537: Paragraphs in Descriptions Fields Are Not Possible

**Describe the bug**
It seems to be impossible to add paragraph breaks into description fields.

**To Reproduce**
1. Write some text in a description field with empty rows which should add paragraph breaks according to Markdown syntax.
2. Leave the edit field so it's rendered.
3. All text will be stuck together as a bit block of text without any paragraph breaks.

**Expected behavior**
I can insert paragraph breaks into descriptions to structure the description text a bit and improve readability.

**Screenshots**

_Markup:_
----
image
----

_Result:_
----
image
----

- Nextcloud-Version: 25.0.4
- Forms-Version: 3.1.0

 - OS: KDE neon (Ubuntu 22.04)
 - Browser Chromium
 - Version 110.0.5481.100 (Offizieller Build) Built on Ubuntu , running on neon 22.04 (64-Bit)

jotoeri:
`render` converts multi-linebreaks to paragraphs, `renderInline` just keeps them as multiple breaks.

Fixes #1537

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -83,7 +83,7 @@
 					class="question__header__description__input"
 					@update:value="onDescriptionChange" />
 				<!-- eslint-disable-next-line vue/no-v-html -->
-				<p v-else class="question__header__description__output" v-html="computedDescription" />
+				<div v-else class="question__header__description__output" v-html="computedDescription" />
 			</div>
 		</div>
 
@@ -184,7 +184,7 @@ export default {
 		},
 
 		computedDescription() {
-			return this.$markdownit.renderInline(this.description)
+			return this.$markdownit.render(this.description)
 		},
 
 		/**
@@ -247,6 +247,8 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+@import '../../scssmixins/markdownOutput';
+
 .question {
 	position: relative;
 	display: flex;
@@ -342,7 +344,7 @@ export default {
 
 			&__input {
 				margin: 0px;
-				min-height: 1.3em;
+				min-height: 1.5em;
 				border-width: 2px;
 				position: relative;
 				left: -12px;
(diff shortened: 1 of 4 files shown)


jotoeri:
Urgh, just saw that empty lines are now doubled on their first rendering. Need to have a look again tomorrow.

susnux:
It depends on what we want to archive if this is the correct solution:

1. Use common markdown: This is like we do on the form description, we do not need to change any code, just add a css rule: `.form-desc p:not(:first-child) { margin-top: 1.5em;}`. This will allow paragraph breaks as well as hard breaks.
2. Do not use paragraphs at all: This is like we do it on the question description. Just hard breaks within one paragraph.

*1* would be most common markdown rendering, allowing for more addition in the future like tables or other blocks. *2* would be inline rendering like you proposed, it is simple and sufficient for most inline content like formatting, links or images.

> Urgh, just saw that empty lines are now doubled on their first rendering. Need to have a look again tomorrow.

My guess is that this is related to the `breaks` option converting newlines to `<br>` elements, combined with this bug: https://github.com/nextcloud/nextcloud-vue/issues/3836
If so we need to either fix that bug or switch back to `<textarea>` input.

jotoeri:
Hmm, i'd say that most of the end users are in fact non-technical users, that are not used to markdown as you/we are. And i think, they'd consider it more of the annoying 'bug', if multiple newlines are just squashed into one 'newline'/paragraph. This is why i would have liked this way of keeping all newlines. - In fact, i'd say that a user not using markdown at all but just plain text, should ideally get the pixel-perfect result of the input. 🤷‍♂️ 

But indeed, if that kills us the tables on md, it's also not the best solution.

jotoeri:
So - now both as `render` and with a common css. Just kills more than one empty line now.

Chartman123 (approved):
I think, this is the best we can currently get. Would it be possible to remove the extra new lines on storing/switching to view mode or would this break some other rendering?

jotoeri:
> Would it be possible to remove the extra new lines on storing/switching to view mode or would this break some other rendering?

Don't think so, currently. This really is a bug from the vue lib, that the newlines are doubled. You can just enter empty newlines in https://nextcloud-vue-components.netlify.app/#/Components/NcRichContenteditable to see it. Just removing doubled newlines on our side wouldn't fully help, so let's better try to get rid of the source bug.

susnux (approved):
> I think, this is the best we can currently get.

I agree.

But I also think we should consider to switch back to `textarea`

PR #1548, opened 2023-03-04
[stable2] Drop stable22 & stable23 tests

jotoeri:
Dropped minversion 22 during release, since it's far out of support.


Chartman123 (approved):

jotoeri:
Just corrected it - It was 22 AND 23, that i dropped due to support-end. 🙈

Issue #1040, opened 2021-09-06
Add technical identifier / name for fields


j-dimension:
**Is your feature request related to a problem? Please describe.**
When retrieving a forms submission through the API, to further process the submission data an external application needs to "understand" the responses to map individual attributes to its own data model. E.g. I need to know which field is "street" and which is "city". 

**Describe the solution you'd like**
Nextcloud forms could distinguish between a display name of a field and an optional technical name / identifier, that is not just an integer but some meaningful value.
The technical name should be part of the data that is provided via the RESTful API.

**Describe alternatives you've considered**
Alternatives would be to just rely on the display name, but that would make an integration very fragile - display names may change and might not be suited to serve as a base for such mapping.

**Additional context**
No additional logic needs to be applied, an additional attribute at the field level would be sufficient.

PR #1553, opened 2023-03-13
Add technical identifiers for questions

susnux:
* Resolves: #1040

### Summary
Technical identifiers allow to map form questions to custom data.
On the frontend those identifiers are implemented as `name` props on the form elements.

### Screenshots
Name input:
name input on question actions menu

### Other notes
The name attribute for technical identifier should be mapped to all options so a `fieldset` is introduced around radio / checkbox buttons.
Added required aria label for `fieldset` (labeled by the question title).

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -140,6 +140,7 @@ Returns the full-depth object of the requested form (without submissions).
       "type": "dropdown",
       "isRequired": false,
       "text": "Question 1",
+      "name": "something",
       "options": [
         {
           "id": 1,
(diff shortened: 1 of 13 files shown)


Chartman123 (approved):
Could you please put the two commits in one before merging? I don't think that it's really needed to have two commits here :) Rest seems fine.

susnux:
@Chartman123 Shure, done. The second commit was just for review purpose to see why I changed it :)

Chartman123:
@susnux Please also add the new prop to the API docs :)

susnux:
> Please also add the new prop to the API docs :)

Done :heavy_check_mark:

PR #1554, opened 2023-03-13
Place actions popover so it does not overflow the page

susnux:
As the action button is placed on the very right, the popup menu should be placed <del>on the left of it.</del> so that it does not overflow the page (causes horizontal scrolling).

This is especially a problem with longer translations and will get a bigger problem with new features like #1553 or #1491 (because of the input elements within the actions popup).

### Before
https://user-images.githubusercontent.com/1855448/224829786-2c9f41ae-96b5-42ad-9d49-13601a9b4bb6.mp4

### After
after | after (mobile)
---|---
image | image

**First version:**
<details>

after | after mobile
---|---
popup is placed on the left of the menu so no hor. scrolling is needed | mobile view the popup is still accessible

</details>

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -59,6 +59,7 @@
 					:id="actionsId"
 					:container="'#' + actionsId"
 					:force-menu="true"
+					placement="bottom-end"
 					class="question__header__title__menu">
 					<NcActionCheckbox :checked="isRequired"
 						@update:checked="onRequiredChange">


jotoeri:
Mhm. I think it looks strange like this. Especially on 'normal'-sized screens. Especially, when the menu gains height by more features...
- Is there the possiblity for a placement like left-bottom?
- Otherwise maybe have the left-placement only with `isMobile`?

-> But i'll leave that decision up to @jancborchardt 😉

Chartman123:
> * Otherwise maybe have the left-placement only with `isMobile`?

That was also my first thought, to only have it on the left side if there is not enough space on the right.

susnux:
> Is there the possiblity for a placement like left-bottom?

@jotoeri like the following? I think this looks better:
actions placed on the bottom end

jotoeri (approved):
❤️ 👍

Chartman123 (approved):

jancborchardt (commented):
@susnux sorry, missed this one somehow! I'd agree with @jotoeri it would be best if the left-placement is specific to isMobile, if possible. :) Then it looks nice on both narrow and large screens.

PR #1559, opened 2023-03-15
Response view switcher: Make accessible for keyboard and add hover effect 

susnux:
I think with a hover effect it looks a bit better:

before | after
---|---
no hover effect on the switcher | with hover effect

Or maybe better to see on this screen recording:

https://user-images.githubusercontent.com/1855448/225323518-caa2b6a8-3494-4eb9-8a0a-85705c79ab99.mp4

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -373,6 +373,10 @@ export default {
 				font-weight: bold;
 				background-color: var(--color-background-dark);
 
+				&:hover {
+					background-color: var(--color-background-hover);
+				}
+
 				&:first-of-type {
 					border-top-right-radius: 0;
 					border-bottom-right-radius: 0;
@@ -387,7 +391,11 @@ export default {
 
 				&--active {
 					background-color: var(--color-primary);
-					color: var(--color-primary-text)
+					color: var(--color-primary-text);
+
+					&:hover {
+						background-color: var(--color-primary-hover);
+					}
 				}
 			}
 		}
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -42,28 +42,26 @@
 
 			<!-- View switcher between Summary and Responses -->
 			<div class="response-actions">
-				<div class="response-actions__radio">
-					<input id="show-summary--true"
-						v-model="showSummary"
-						type="radio"
-						:value="true"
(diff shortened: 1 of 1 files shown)


Chartman123 (approved):

jancborchardt (changes requested):
Nice! Just one thing, any hover effect should be present for `:focus` as well so it is accessible to people who use keyboard only. :)

susnux:
> Nice! Just one thing, any hover effect should be present for `:focus` as well so it is accessible to people who use keyboard only. :)

@ jancborchardt thank you! Because of your comment I tried to use keyboard navigation and noticed that it currently did not even work. Not only there was no hover effect but I could no focus the toggle to switch the view!

So I pushed another commit allowing to use keyboard navigation. Now it looks like this when using keyboard navigation:


https://user-images.githubusercontent.com/1855448/225338640-204cd3f0-cf14-472c-9384-93622970519a.mp4

jotoeri:
Shouldnt we generally change it, so it looks similar to the views-switch? Maybe just a bit smaller than the other one...

susnux:
> Shouldnt we generally change it, so it looks similar to the views-switch? Maybe just a bit smaller than the other one...

I think this UI was heavily inspired by google forms (see below), so as we have a custom view selector by now we can reuse that of cause.

google forms selector

susnux:
> Shouldnt we generally change it, so it looks similar to the views-switch? Maybe just a bit smaller than the other one...

@jotoeri what do you think about doing it like this?

https://user-images.githubusercontent.com/1855448/225457953-8c3b969d-fa66-44cd-9591-e7f2f174016b.mp4

jotoeri:
Hmm, yes, at least better than completely different from the other switch. Maybe a bit less in height? And a bit of margin around? 😉 @jancborchardt ?

susnux:
> Maybe a bit less in height?

If I am right the current height of 44px is already the minimum size in terms of accessibility? (At least from [WCAG](https://w3c.github.io/wcag/understanding/target-size-enhanced))

> And a bit of margin around?

Of cause, that screen recording was just a quick mockup.

jotoeri:
Superseded by #1576

PR #1560, opened 2023-03-15
WIP: Add collapsible submissions view

### LINKED ISSUE #1497: Make responses collapsible
- Nextcloud-Version: 25 / 26
- Forms-Version: 3

One some kind of forms, e.g. booking forms, when having multiple submissions the responses view gets crowded.
Meaning it is hard to find the submission you want  to look into or to keep track of it.

Make it possible to collapse all responses and expand a selected response.
E.g. it could be collapsed to only show the header e.g.:
```
Expand button (+)   |  User / Anonymous  |  Date  |  Actions 
```

susnux:
* Resolves: #1497 

This is still work in progress as there are a few questions for which I would like to have some feedback:

* Currently there is a "unread" notification, to have this work across browser sessions we would have to save it on the back end, so:
  * Do we want this feature? (I would like to have it as it allows better overview on already processed responses)
  * One global "viewed" flag per submission or one per user? (I would prefer it per user)
  * Saved within the submission? Or in the share? Or a new table? Somewhere else?
* I implemented this currently as a separate view, should we keep this as a third view or replace the "responses" view with this?

Here is a screen recording on how it looks like currently:

https://user-images.githubusercontent.com/1855448/225367124-2e790876-c3ee-4f91-9e16-e6b789a4f49b.mp4

### DIFF
--- a/src/components/Results/Answer.vue
+++ b/src/components/Results/Answer.vue
@@ -59,6 +59,7 @@ export default {
 
 	&__text {
 		white-space: pre-line;
+		margin-left: 1rem;
 	}
 }
 
--- a/src/components/Results/SubmissionItem.vue
+++ b/src/components/Results/SubmissionItem.vue
@@ -1,0 +1,163 @@
+<!--
+  - @copyright Copyright (c) 2023 Ferdinand Thiessen <rpm@fthiessen.de>
+  -
+  - @author Ferdinand Thiessen <rpm@fthiessen.de>
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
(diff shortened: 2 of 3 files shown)


Chartman123:
> * Currently there is a "unread" notification, to have this work across browser sessions we would have to save it on the back end, so:   
>   * Do we want this feature? (I would like to have it as it allows better overview on already processed responses)

Unread flag would be nice, this could also be represented in the list of forms on the left

>   * One global "viewed" flag per submission or one per user? (I would prefer it per user)

I would say that it's enough to have it per submission, but per user would be nice

>   * Saved within the submission? Or in the share? Or a new table? Somewhere else?

Depends on the point above... If per submission, then with the submission. If per user, probably somewhere else

> * I implemented this currently as a separate view, should we keep this as a third view or replace the "responses" view with this?

A separate view isn't necessary in my opinion... We're showing the same information as in the responses view. So in conjunction with the "unread" flag you could for example collapse all read submissions by default and have a collapse/expand all action on top.

Chartman123:
@susnux As this PR is quite old, should we keep it open?

Issue #1566, opened 2023-03-16
Markdown: Missing top margin on block elements


gohrner:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**
It is not possible to have an emtpy row before a Markdown code block.

**To Reproduce**
1. Enter description field in edit mode.
2. Enter text followed by a few empty lines, followed by three-backtick-codeblock.
3. Leave edit mode.
4. No space will be between the text and the code block.

**Expected behavior**
At least one empty line should stay between the text and the code block.

**Screenshots**
![Peek 2023-03-16 22-23](https://user-images.githubusercontent.com/26199042/225755868-239561b3-e78d-4ed2-a495-5b9eda9f3bbe.gif)

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 25.0.4
- Forms-Version: 3.2.0

**Desktop (please complete the following information):**
 - OS: KDE neon (Ubuntu 22.04)
 - Browser: Chromium
 - Version 111.0.5563.64 (Offizieller Build) Built on Ubuntu , running on neon 22.04 (64-Bit)

susnux:
> It is not possible to have an emtpy row before a Markdown code block.

That is not supported by Markdown, markdown treats multiple empty lines as one single line.
Meaning regardless of the number of empty lines, they will only result in a new paragraph.

So this is not a bug but by definition.

FYI: It is possible to have empty lines within one block:
```markdown
some text\
\
\
\
3 empty lines before this
```

Looks like this:
some text\
\
\
\
3 empty lines before this

susnux:
This is a styling issue, I will submit a fix for this asap.

PR #1568, opened 2023-03-17
fix(markdown): Add margin for all new paragraphs

susnux:
* Resolved: #1566

### Summary

Before this only `<p>` elements had some top margin applied, but blocks like code blocks (`pre`) would not have any margin and were rendered like if they belong to the previous paragraph.

### Screenshots

before | after
---|---
no margin between paragraphs | margin between all block elements

Markdown code:
````markdown
Text


Text with\
hard break


Text with\
\
\
empty lines


```
code
```
````

### DIFF
--- a/src/scssmixins/markdownOutput.scss
+++ b/src/scssmixins/markdownOutput.scss
@@ -22,7 +22,7 @@
 
 @mixin markdown-output {
 	::v-deep {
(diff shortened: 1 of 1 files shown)


jotoeri (approved):
👍

Chartman123 (approved):

Issue #1564, opened 2023-03-16
Markdown: Special HTML Characters Are Converted Into Their HTML Entity Codes Within Codeblocks


gohrner:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**
Special HTML Characters Are Converted Into Their HTML Entity Codes Within Codeblocks.
This renders code like e.g. Java - or worse, XML or similar - unreadable.

See attached screencast.

**To Reproduce**
1. Enter a code block into a description field, and add e.g. angle brackets into it.
2. Leave the edited field.
3. HTML entity codes will be rendered instead of the angle brackets.

**Expected behavior**
Angle brackets should be entered, just as entered into the text field.

**Screenshots**
![Peek 2023-03-16 22-05](https://user-images.githubusercontent.com/26199042/225752554-29b61947-3678-4952-98c1-66cc4d28c289.gif)

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 25.0.4
- Forms-Version: 3.2.0

**Desktop (please complete the following information):**
 - OS: KDE neon (Ubuntu 22.04)
 - Browser: Chromium
 - Version 111.0.5563.64 (Offizieller Build) Built on Ubuntu , running on neon 22.04 (64-Bit)

susnux:
This is an upstream bug: https://github.com/nextcloud/nextcloud-vue/issues/3264

As this and #1562 both are caused by the NcRichContenteditable component, I vote for replacing it with a simple textarea input.

Issue #1563, opened 2023-03-16
Markdown: First Line Of "Triple Backtick" Code Blocks is Sometimes "Absorbed" Into Triple-Backtick-Line


gohrner:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**
First Line Of "Triple Backtick" Code Blocks is Sometimes "Absorbed" Into Triple-Backtick-Line.

See screencast.

**To Reproduce**
1. Enter a (in this example one-line) Markdown code block into a description field
2. Leave edit mode.
3. The description field will be empty.
4. Re-entering edit mode will show why: the code block content was merged into the first "backtick line".

**Expected behavior**
The first line in the actual code block should NOT be merged into the backtick line.

**Screenshots**
![Peek 2023-03-16 22-12](https://user-images.githubusercontent.com/26199042/225753959-9ad9333a-9799-4e92-af07-2bdfd70d6846.gif)

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 25.0.4
- Forms-Version: 3.2.0

**Desktop (please complete the following information):**
 - OS: KDE neon (Ubuntu 22.04)
 - Browser: Chromium
 - Version 111.0.5563.64 (Offizieller Build) Built on Ubuntu , running on neon 22.04 (64-Bit)

susnux:
I can only reproduce this with Chrome. Using Firefox everything works as expected.

I think this is another issue with NcRichContenteditable.

susnux:
BTW this an instance of nextcloud/nextcloud-vue#3836

Issue #1562, opened 2023-03-16
Markdown: Pasting code "loses" pasted content


gohrner:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**
Pasting the following code block into a description field in Forms simply "loses" the Generic Type in angle brackets:

    ```java
    Test<Object> myMethod() {
        if (1 == 2) {
            // code
        }
    }
    ```

**To Reproduce**
1. Paste the code block above into an empty description field.
2. The method's return type will be the raw type `Test` instead of `Test<Object>`.

**Expected behavior**
The code block should be pasted as-is.

**Screenshots**
Screencast of the issue:

![Peek 2023-03-16 21-54](https://user-images.githubusercontent.com/26199042/225750499-b26a276a-47d1-4d33-bbf2-7fc6e279eb26.gif)

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 25.0.4
- Forms-Version: 3.2.0

**Desktop (please complete the following information):**
 - OS: KDE neon (Ubuntu 22.04)
 - Browser: Chromium
 - Version 111.0.5563.64 (Offizieller Build) Built on Ubuntu , running on neon 22.04 (64-Bit)

susnux:
Upstream issue: nextcloud/nextcloud-vue#3908

PR #1574, opened 2023-03-20
fix: Replace `NcRichContenteditable` with `textarea`

susnux:
* Resolves: #1562 
* Resolves: #1563
* Resolves: #1564 

Those issues are basically bugs of the `NcRichContenteditable` component.
So the simplest fix is just using a textarea as before, downside: No fancy emoji input (but I think we can cope with that).

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -231,8 +231,12 @@ export default {
 		resizeDescription() {
 			// next tick ensures that the textarea is attached to DOM
 			this.$nextTick(() => {
-				const rows = this.$refs.description.value.split('\n').length
-				this.$refs.description.setAttribute('rows', rows)
+				const textarea = this.$refs.description
+				if (textarea) {
+					textarea.style.cssText = 'height: 0'
+					// include 2px border
+					textarea.style.cssText = `height: ${textarea.scrollHeight + 4}px`
+				}
 			})
 		},
 
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -210,7 +210,10 @@ export default {
 			}
 		},
 	},
-
+	// Ensure description is sized correctly on initial render
+	mounted() {
+		this.$nextTick(() => this.resizeDescription())
+	},
 	methods: {
 		onTitleChange({ target }) {
 			this.$emit('update:text', target.value)
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -76,13 +76,13 @@
 				</NcActions>
 			</div>
 			<div v-if="hasDescription || edit || !questionValid" class="question__header__description">
-				<NcRichContenteditable v-if="edit || !questionValid"
-					:multiline="true"
+				<textarea v-if="edit || !questionValid"
+					ref="description"
 					:value="description"
 					:placeholder="t('forms', 'Description (formatting using Markdown is supported)')"
 					:maxlength="maxStringLengths.questionDescription"
 					class="question__header__description__input"
-					@update:value="onDescriptionChange" />
+					@input="onDescriptionChange" />
 				<!-- eslint-disable-next-line vue/no-v-html -->
 				<div v-else class="question__header__description__output" v-html="computedDescription" />
 			</div>
@@ -99,7 +99,6 @@ import { directive as ClickOutside } from 'v-click-outside'
 import NcActions from '@nextcloud/vue/dist/Components/NcActions.js'
 import NcActionButton from '@nextcloud/vue/dist/Components/NcActionButton.js'
 import NcActionCheckbox from '@nextcloud/vue/dist/Components/NcActionCheckbox.js'
-import NcRichContenteditable from '@nextcloud/vue/dist/Components/NcRichContenteditable.js'
 
 import IconAlertCircleOutline from 'vue-material-design-icons/AlertCircleOutline.vue'
 import IconDelete from 'vue-material-design-icons/Delete.vue'
@@ -119,7 +118,6 @@ export default {
 		NcActions,
 		NcActionButton,
 		NcActionCheckbox,
-		NcRichContenteditable,
 	},
 
 	inject: ['$markdownit'],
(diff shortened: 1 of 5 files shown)


jotoeri (commented):
Please check the css, so it looks similar before/after.

Chartman123:
Isn't it possible to fix the issues in the vue component? I mean we would lose the quick picker which would make adding links or other things quite easy for our users...

jotoeri:
Without text, the placeholder is too light and the box-sizes don't match. Testing means to check not only a single case, but all (at least nearly)!

susnux:
I will have to check to this, but if this is the case, then that is a server or Vue bug.
Because both should have the same styling.

susnux:
It turns out it is indeed a server bug: https://github.com/nextcloud/server/pull/37522

susnux:
@jotoeri 
> Without text, the placeholder is too light

This is a "fix" as the placeholder is indended to be `--color-text-maxcontrast`

> and the box-sizes don't match.

Which box sized do not match? I could not see any difference.

before | after
---|---
image |  image

jotoeri:
> This is a "fix" as the placeholder is indended to be `--color-text-maxcontrast`

Hmm, at least on my firefox it looks more like not so much contrast... 🤔 @jancborchardt is that how it should look like?
grafik


> Which box sized do not match? I could not see any difference.

Also check the questions! This one increased. Also, i just realised, that editing the question description fails at all.

susnux:
> Also check the questions! This one increased. Also, i just realised, that editing the question description fails at all.

I did, it looks like this:

before | after
---|---
image | image

As you can see the color of description placeholder is *now* the same as the question input one.
What indeed changed is the border color, but it is now consistent with the other borders (so an improvement I guess?).

But I now also added that fix for the server issue, as we need to fix it for our self, because otherwise we would have to depend on the upcoming minor releases of NC.

susnux:
Any feedback? Would be nice to have this fixed to create a bugfix release :)

Chartman123:
> Any feedback? Would be nice to have this fixed to create a bugfix release :)

Sorry, I'm very busy at the moment, but I will try to find some time in the next few days 👍🏻

susnux:
Server PRs have been merged:
* https://github.com/nextcloud/server/pull/37522
* https://github.com/nextcloud/server/pull/37829

(and I have rebased the PR onto current main branch)

susnux:
I also added a commit fixing description overflowing the page and looks different on edit vs view.

Chartman123:
@susnux Looks quite good now, just a little text movement left:
 
[Peek 2023-05-06 23-23.webm](https://user-images.githubusercontent.com/16020496/236647105-ed148780-482c-4ca5-a7f5-7442da057741.webm)

jotoeri:
NC 25.0.6
Firefox 113.0.1

|before|after|
|---|---|
|grafik|grafik|

susnux:
> just a little text movement left (question titles and button text)

@Chartman123 That's because of CSS math... When calculating the height of the input element I used `min-height: calc(25px + 1.5em);` which is correct but will be evaluated by the browser to `47.5px` while the actual input element is `48px` so this causes .5px jumping.

> NC 25.0.6
> Firefox 113.0.1
> :framed_picture:  :framed_picture: 

@jotoeri Caused by missing initial resizing of the textarea, added a `mounted` hook to resolve the issue.

src/views/Create.vue
```diff
@@ -411,20 +441,6 @@ export default {
...
-		async autoSizeTitle() {
-			this.$nextTick(() => {
-				const textarea = this.$refs.title
-				if (textarea) {
-					textarea.style.cssText = 'height:auto'
-					// include 2px border
-					textarea.style.cssText = `height: ${textarea.scrollHeight + 4}px`
-				}
-			})
-		},
```
jotoeri, line 427: Why?? Its in fact the same code, but more changed lines within the PR...!?

src/components/Questions/Question.vue
```diff
@@ -205,20 +203,39 @@ export default {
...
 
 		onRequiredChange(isRequired) {
 			this.$emit('update:isRequired', isRequired)
 		},
 
+		resizeDescription() {
+			// next tick ensures that the textarea is attached to DOM
+			this.$nextTick(() => {
+				const rows = this.$refs.description.value.split('\n').length
+				this.$refs.description.setAttribute('rows', rows)
```
jotoeri, line 235: Maybe just use the old code as in Create.vue?

jotoeri (approved):
Still the same issue for me.
But i give up now. 5 reviews is enough. 🐘

src/views/Create.vue
Chartman123: For me the behavior changed a little: before the text jumped when there was no or one line. Now it jumps for three lines in the input.

Didn't we have some old code before that computed the correct size for the field height? Or can't that be used anymore because of the new formatting capabilities?

src/components/Questions/Question.vue
susnux: Just reordered to be more logical when reading the file, but I can revert it if you prefer small visual diff.

src/views/Create.vue
susnux: > Why?? Its in fact the same code, but more changed lines within the PR...!?

Just reordered to be more logical when reading the file, but I can revert it if you prefer small visual diff.

susnux: > For me the behavior changed a little: before the text jumped when there was no or one line. Now it jumps for three lines in the input.

@Chartman123 
I finally fixed the jumping content :tada: (at least I could not find any :wink: ). The problem was that the line height was set to 1.5em which is 22.5px, now the rendered size was jumping around 1px.
Setting the line height to 22px fixed the problem for me.

src/components/Questions/Question.vue
susnux: > Maybe just use the old code as in Create.vue?

Jep. Changed that.

susnux:
> Still the same issue for me.

@jotoeri See comment above. Should be finally fixed

https://github.com/nextcloud/forms/assets/1855448/ff399754-028e-465e-9b36-9c4633e36800

src/views/Create.vue
```diff
@@ -478,13 +498,14 @@ export default {
 
 		.form-desc {
 			color: var(--color-text-maxcontrast);
-			line-height: 1.5em;
-			min-height: calc(25px + 1.5em); // one line
+			line-height: 22px;
+			min-height: 48px; // one line (25px padding + 22px text height)
```
Chartman123, line 502: @susnux 25 + 22 is 47 😉 but as long as it works correctly... Which I haven't checked by now. I will have a look at it this evening 👍🏻

Chartman123 (approved):
For me it's fine now :)

jotoeri:
@susnux still please squash your commits within PRs. Ideally such, that you change a line/routine only once per PR. The commit-stats are no contest!

PR #1658, opened 2023-06-27
Reduce white space between questions

### LINKED ISSUE #1234: Reduce Vertical Dead Space
Hello,

I have recently started using this app, and so far think that the flexibility it affords is great! However, I have noticed that the forms take up a lot of vertical dead space, reducing the amount of information and interactive elements on the page at one time.

image

Would it be possible to reduce the amount of vertical space between elements? I personally think that it would be beneficial to apply that for everyone, but a setting to allow it would be great too.

Thanks!

Chartman123:
Fixes #1234 by reducing the margin on the bottom of the questions from 80px to 64px

|80px | 70px | 60px | 50px |
|---|---|---|---|
|grafik|grafik |grafik |grafik |

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -302,7 +302,7 @@ export default {
 	align-items: stretch;
 	flex-direction: column;
 	justify-content: stretch;
-	margin-bottom: 80px;
+	margin-bottom: 64px;
 	padding-left: 44px;
 	user-select: none;
 	background-color: var(--color-main-background);


susnux (approved):
I think this is fine, but for me 50px is the minimum.

Chartman123:
> I think this is fine, but for me 50px is the minimum.

I also experimented with smaller values and evaluated 50 px to be the smallest that still looks ok.

nimishavijay (commented):
Great screenshots! :) I would vote for slightly more than 50px, maybe 64px, what do you think? This is 8px more than the height of each navigation item on the left. We don't want to get rid of too much whitespace either, especially on mobile where it can easily look too cramped. Also, we should make sure that this spacing accurately reflects for all the modes (view, edit and results)

Chartman123:
@nimishavijay thank you for the review. View/edit use the same spacing between the questions. For the results the elements are placed differently and I think that we can leave this the way it is currently. So I will change it to 64px and then we should be good to go here 😇

PR #1659, opened 2023-06-28
feat: Implement custom submission message

### LINKED ISSUE #385: Allow for custom message and/or redirect after submission
When creating a form, there's currently no flexibililty to configure the content/action that occurs after successful form submission.

Would it be beneficial to include a "Submission message" (text/richtext?) and/or "Redirect URL"  as part of Form configurations? These options would help with i18n for multilingual audiences, and offer improved UX for form creators.

As far as I can tell, there's no mechanism made available in the package to faciliate any customization [here](https://github.com/nextcloud/forms/blob/95cd4c0fb2fb46ce4b42f0020636249aa79c8300/src/views/Submit.vue#L67) or [here](https://github.com/nextcloud/forms/blob/95cd4c0fb2fb46ce4b42f0020636249aa79c8300/src/views/Submit.vue#L160-L175)?

susnux:
* Resolves #385 

## :information_source:  Summary
Users can configure a custom text which is shown if a users submits the form (custom "Thank you message").
It can be formatted using markdown, so it also allows the requested link feature (e.g. to link back to a website).

## :framed_picture: Screenshots

| Form submitted | Editing the message | Preview the message |
|---|---|---|
|image|image|image|

### DIFF
--- a/src/components/SidebarTabs/SettingsSidebarTab.vue
+++ b/src/components/SidebarTabs/SettingsSidebarTab.vue
@@ -59,6 +59,35 @@
 				{{ t('forms', 'Show expiration date on form') }}
 			</NcCheckboxRadioSwitch>
 		</div>
+		<NcCheckboxRadioSwitch :checked="hasCustomSubmissionMessage"
+			type="switch"
+			@update:checked="onUpdateHasCustomSubmissionMessage">
+			{{ t('forms', 'Custom submission message') }}
+		</NcCheckboxRadioSwitch>
+		<div v-show="hasCustomSubmissionMessage"
+			class="settings-div--indent submission-message"
+			:tabindex="editMessage ? undefined : '0'"
+			@focus="editMessage = true">
+			<textarea v-if="editMessage || !form.submissionMessage"
+				v-click-outside="() => { editMessage = false }"
+				aria-describedby="forms-submission-message-description"
+				:aria-label="t('forms', 'Custom submission message')"
+				:value="form.submissionMessage"
+				:maxlength="maxStringLengths.submissionMessage"
+				:placeholder="t('forms', 'Message to show after a user submitted the form (formatting using Markdown is supported)')"
+				class="submission-message__input"
+				@blur="editMessage = false"
+				@change="onSubmissionMessageChange" />
+			<!-- eslint-disable vue/no-v-html -->
+			<div v-else
+				:aria-label="t('forms', 'Custom submission message')"
+				class="submission-message__output"
+				v-html="submissionMessageHTML" />
+			<!-- eslint-enable vue/no-v-html -->
+			<div id="forms-submission-message-description" class="submission-message__description">
+				{{ t('forms', 'Message to show after a user submitted the form. Please note that the message will not be translated!') }}
+			</div>
+		</div>
 	</div>
 </template>
 
@@ -68,13 +97,22 @@ import NcCheckboxRadioSwitch from '@nextcloud/vue/dist/Components/NcCheckboxRadi
 import NcDatetimePicker from '@nextcloud/vue/dist/Components/NcDatetimePicker.js'
 import ShareTypes from '../../mixins/ShareTypes.js'
 
+import { directive as ClickOutside } from 'v-click-outside'
+import { loadState } from '@nextcloud/initial-state'
+
 export default {
 	components: {
 		NcCheckboxRadioSwitch,
 		NcDatetimePicker,
 	},
 
+	directives: {
+		ClickOutside,
+	},
+
 	mixins: [ShareTypes],
+
+	inject: ['$markdownit'],
 
 	props: {
 		form: {
@@ -89,10 +127,20 @@ export default {
 				stringify: this.stringifyDate,
 				parse: this.parseTimestampToDate,
 			},
+			maxStringLengths: loadState('forms', 'maxStringLengths'),
+			/** If custom submission message is shown as input or rendered markdown */
+			editMessage: false,
 		}
 	},
 
 	computed: {
+		/**
+		 * If the form has a custom submission message or the user wants to add one (settings switch)
+		 */
+		hasCustomSubmissionMessage() {
+			return this.form?.submissionMessage !== undefined && this.form?.submissionMessage !== null
+		},
+
 		/**
 		 * Submit Multiple is disabled, if it cannot be controlled.
 		 */
@@ -124,6 +172,12 @@ export default {
 		},
 		expirationDate() {
 			return moment(this.form.expires, 'X').toDate()
+		},
+		/**
+		 * The submission message rendered as HTML
+		 */
+		submissionMessageHTML() {
+			return this.$markdownit.render(this.form.submissionMessage || '')
 		},
 	},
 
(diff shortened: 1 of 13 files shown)


Chartman123:
@susnux Nice addition :) Should we hide the message input field behind a switch? [ ] Show custom message

susnux:
> Should we hide the message input field behind a switch? [ ] Show custom message

Unsure about this, what do you think looks better? (For me both variants have equal pros and cons)

Chartman123:
> Unsure about this, what do you think looks better? (For me both variants have equal pros and cons)

Using the switch we would surely need one more column in the DB, but then could also have an empty submission message. I think this is also a question for the designers. If we add the switch, we should also indent the text box under the text of the switches. Would you like to bring it up in the design team chat?

jancborchardt (commented):
Nice idea @susnux! And yep, I think @Chartman123’s suggestion about doing it as a switch is good, keeps it cleaner too since most people are likely to keep the original submission message. And then we can also add some grey hint text when it’s enabled saying something like:
"Message to show after form submission. Note that this can not be translated."

susnux:
@Chartman123 @jancborchardt 

Implemented the checkbox:
Screenshot_20230826_101816

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -59,6 +59,27 @@
...
+		<div v-show="hasCustomSubmissionMessage"
+			class="settings-div--indent submission-message"
+			:tabindex="editMessage ? undefined : '0'"
+			@focus="editMessage = true">
+			<textarea v-if="editMessage || !form.submissionMessage"
+				v-click-outside="() => { editMessage = false }"
+				:aria-label="t('forms', 'Custom submission message')"
+				:value="form.submissionMessage"
+				:maxlength="maxStringLengths.submissionMessage"
+				:placeholder="t('forms', 'Message to show after an user submitted the form (formatting using Markdown is supported)')"
```
Chartman123, line 76:
```suggestion
				:placeholder="t('forms', 'Message to show after a user submitted the form (formatting using Markdown is supported)')"
```

It's `a user` as it's pronounced `ˈjuːzɐ`

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -89,10 +119,21 @@ export default {
 				stringify: this.stringifyDate,
 				parse: this.parseTimestampToDate,
 			},
+			maxStringLengths: loadState('forms', 'maxStringLengths'),
+			/** If custom submission message is shown as input or rendered markdown */
+			editMessage: false,
+			showSubmissionMessageInput: false
```
Chartman123, line 125:
```suggestion
			showSubmissionMessageInput: false,
```
Lint warning

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -59,6 +59,27 @@
...
+				v-click-outside="() => { editMessage = false }"
+				:aria-label="t('forms', 'Custom submission message')"
+				:value="form.submissionMessage"
+				:maxlength="maxStringLengths.submissionMessage"
+				:placeholder="t('forms', 'Message to show after an user submitted the form (formatting using Markdown is supported)')"
+				class="submission-message__input"
+				@blur="editMessage = false"
+				@change="onSubmissionMessageChange" />
+			<!-- eslint-disable-next-line vue/no-v-html -->
+			<div v-else :aria-label="t('forms', 'Custom submission message')" class="submission-message__output" v-html="submissionMessageHTML" />
```
Chartman123, line 81: Please fix lint error

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -59,6 +59,27 @@
 				{{ t('forms', 'Show expiration date on form') }}
 			</NcCheckboxRadioSwitch>
 		</div>
+		<NcCheckboxRadioSwitch :checked="hasCustomSubmissionMessage"
```
Chartman123, line 62: I thought of having a separate prop here and not just querying the custom submission message itself.

Chartman123 (commented):
I had another look at the corresponding issue and saw that there was also a request for customizing also the title message of the submission page. Would you like to add this into this PR as well?

Other than that, please update the API docs :)

src/components/SidebarTabs/SettingsSidebarTab.vue
susnux: Do you mean another db property? I do not think that is needed, a null value will mean that there is no custom message. Or what do you mean?

Chartman123: Ok so that we can distinguish between `null` and `"" `? Works for me as well...

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -59,6 +59,35 @@
...
+				@change="onSubmissionMessageChange" />
+			<!-- eslint-disable vue/no-v-html -->
+			<div v-else
+				:aria-label="t('forms', 'Custom submission message')"
+				class="submission-message__output"
+				v-html="submissionMessageHTML" />
+			<!-- eslint-enable vue/no-v-html -->
+			<div id="forms-submission-message-description" class="submission-message__description">
+				{{ t('forms', 'Message to show after a user submitted the form. Please note that the message will not be translated!') }}
+			</div>
```
Chartman123, line 89: Do we really need this? It's just that we nowhere else have any description like this. Perhaps something for @jancborchardt

susnux: Yes for accessibility.  We should not misuse placeholders for this. We have the same problem e.g. with the comments for files where it will solved like this. 

I wanted to already implement it correctly for new features:)

Chartman123: Ok, then this is fine :)

src/views/Submit.vue
```diff
@@ -68,6 +68,10 @@
 			<template #icon>
 				<IconCheck :size="64" />
 			</template>
+			<template v-if="typeof form.submissionMessage === 'string'" #description>
+				<!-- eslint-disable-next-line vue/no-v-html -->
+				<p class="submission-message" v-html="submissionMessageHTML" />
```
Chartman123, line 73: This doesn't work correctly for me:
grafik

The message gets rendered as HTML as you can see in the Vue devtools, however, it's not shown after the submission. Perhaps the page is rendered before the HTML code gets rendered and so it just displays '' as the description?

Chartman123: The message in the screenshot could also come from the settings sidebar. I added some logger.debug calls to the method in Submit.vue but they don't show up in the browser console

susnux: Ah broke that. Sorry, fixed it again.

susnux:
@jancborchardt the setting do look like this now:
image

src/views/Submit.vue
```diff
@@ -365,6 +380,10 @@ export default {
 		}
 	}
 
+	.submission-message {
+		@include markdown-output;
```
Chartman123, line 384:
```suggestion
		@include markdown-output;
		text-align: center;
```

Without the `text-align: center;` the output looks odd when you have more than one line.

Chartman123 (dismissed):

Chartman123:
Looks good now. So only the following things are remaining:

- [ ] add the new prop to docs/API.md and docs/DataStructure.md, add info that prop is only available with higher API level
- [ ] raise API level?
- [ ] What to do about the title? Always keep the default or make it customizable, too?
- [ ] Squash commits into one, do a rebase

susnux:
@Chartman123 done with the todos

> raise API level?

I would say no as we did not change it, there are no new routes, we did not increase the API level for new option keys before (e.g. `extraSettings`).

> What to do about the title? Always keep the default or make it customizable, too?

can we do this in a follow up? I would like to discuss some options before and this PR at least provides the functionality requested.

Chartman123:
> we did not increase the API level for new option keys before (e.g. `extraSettings`).

I'm no expert in API level definitions... But fine for me to stay with 2.1

> can we do this in a follow up?

Fine for me 👍🏻

src/views/Submit.vue
Chartman123: @susnux What about this suggestion here?

susnux: Good point!

PR #1678, opened 2023-07-24
Add healtcheck options to PHPUnit for OCI

Chartman123:


susnux (approved):

Chartman123:
@susnux Hm... that didn't fix the "file exists" error in the workflows... ☹️ Any other ideas?

nickvergessen:
For now you need to retry it manually until it works, or you change the base image from latest to 22.04 so our custom github runners are not used (not recommended).

We are waiting on a new update of some lib to fix support for ZFS or something alike (forgot the details).

Issue #93, opened 2019-10-04
Allow custom/other answer for 'multiple' questions (radio & checkbox)


Sohalt:
**Which Forms version are you running?** (see apps page)
1.0.3

**Nextcloud or ownCloud?:**
Nextcloud

It would be cool to be able to enable an "other" option with a free form text input for radio buttons and checkboxes, to give the user the possibility to provide other answers than those that are explicitly given.

rami999999999:
I would like to add that it would also be great that what users type in "other" should then be available as an options to other users

ei8fdb:
:100: agree with this request. The _other_ option would allow for short number of words to be typed as answer.

Sohalt:
This should not be the default though, because it could leak sensitive information.

ei8fdb:
> > I would like to add that it would also be great that what users type in "other" should then be available as an options to other users
> 
> This should not be the default though, because it could leak sensitive information.

To me this ("what users type in "other" should then be available as an options to other users") seems strange behaviour for a form to support. My thinking:

the form is (usually) a request from the creator to others. The..."information flow" is from the person completing the form to the creator, not between towards others completing the form

I agree with @Sohalt that it shouldn't be default behaviour.

jdittrich:
This is also feature I need often and it currently is a bit of a blocker to not have it. 

Use cases:
* For questions like: "What is your favorite activity when using X" and allowing an answer that the survey creator did not think of or where including all possible answers would make the form really really long. 
* For "What is your gender?" although the normal "other" label is not nice here (think of "othering"). It would be cool if one could adjust the label and "other" would be just the default, so one could replace it e.g. with "prefer to self describe"

jdittrich:
If there is a more elaborate (BDD or so) specification and/or mockup needed, I can create it.

hanswundersam:
Is the "other" text field still something on the table for the forms app? It would be incredibly useful indeed for Nextcloud Forms to be a true replacement for external forms providers.

jotoeri:
It is, just not with highest priority currently. Also see the Milestone.

PR #1694, opened 2023-08-19
feat: Add 'Other' option for radio/checkbox questions.

AIlkiv:
I have implemented a feature to provide other answer for radio buttons and checkboxes.

Related to #93

Demo:

image

image

image

### DIFF
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -309,7 +309,11 @@ class SubmissionService {
 			$questionAnswered = array_key_exists($questionId, $answers);
 
 			// Check if all required questions have an answer
-			if ($question['isRequired'] && (!$questionAnswered || !array_filter($answers[$questionId], 'strlen'))) {
+			if ($question['isRequired'] &&
+				(!$questionAnswered ||
+				!array_filter($answers[$questionId], 'strlen') ||
+				(!empty($question['extraSettings']->allowOtherAnswer) && !array_filter($answers[$questionId], fn ($value) => $value !== Constants::QUESTION_EXTRASETTINGS_OTHER_PREFIX)))
+			) {
 				return false;
 			}
 
@@ -328,9 +332,9 @@ class SubmissionService {
 					!$this->validateDateTime($answers[$questionId][0], Constants::ANSWER_PHPDATETIME_FORMAT[$question['type']])) {
 					return false;
 				}
-	
+
 				// Check if all answers are within the possible options
-				if (in_array($question['type'], Constants::ANSWER_TYPES_PREDEFINED)) {
+				if (in_array($question['type'], Constants::ANSWER_TYPES_PREDEFINED) && empty($question['extraSettings']->allowOtherAnswer)) {
 					foreach ($answers[$questionId] as $answer) {
 						// Search corresponding option, return false if non-existent
 						if (array_search($answer, array_column($question['options'], 'id')) === false) {
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -43,6 +43,10 @@
 				@update:checked="onShuffleOptionsChange">
 				{{ t('forms', 'Shuffle options') }}
 			</NcActionCheckbox>
+			<NcActionCheckbox :checked="extraSettings?.allowOtherAnswer"
+				@update:checked="onAllowOtherAnswerChange">
+				{{ t('forms', 'Add "other"') }}
+			</NcActionCheckbox>
 		</template>
 		<template v-if="!edit">
 			<fieldset :name="name || undefined" :aria-labelledby="titleId">
@@ -57,6 +61,22 @@
 					@keydown.enter.exact.prevent="onKeydownEnter">
 					{{ answer.text }}
 				</NcCheckboxRadioSwitch>
+				<div v-if="allowOtherAnswer" class="question__other-answer">
+					<NcCheckboxRadioSwitch :checked.sync="questionValues"
+						:value="valueOtherAnswer"
+						:name="`${id}-answer`"
+						:type="isUnique ? 'radio' : 'checkbox'"
+						:required="checkRequired('other-answer')"
+						class="question__label"
+						@update:checked="onChange"
+						@keydown.enter.exact.prevent="onKeydownEnter">
+						{{ t('forms', 'Other:') }}
+					</NcCheckboxRadioSwitch>
+					<NcInputField :placeholder="placeholderOtherAnswer"
+						:required="hasRequiredOtherAnswerInput"
+						:value.sync="inputOtherAnswer"
+						class="question__input" />
+				</div>
 			</fieldset>
 		</template>
 
@@ -76,7 +96,15 @@
 						@delete="deleteOption"
 						@update:answer="updateAnswer" />
 				</template>
-
+				<li v-if="allowOtherAnswer" class="question__item">
+					<div :is="pseudoIcon" class="question__item__pseudoInput" />
+					<input :placeholder="t('forms', 'Other')"
+						class="question__input"
+						:maxlength="maxStringLengths.optionText"
+						minlength="1"
+						type="text"
+						:readonly="edit">
+				</li>
 				<li v-if="(edit && !isLastEmpty) || hasNoAnswer" class="question__item">
 					<div :is="pseudoIcon" class="question__item__pseudoInput" />
 					<input :aria-label="t('forms', 'Add a new answer')"
@@ -100,6 +128,7 @@ import { generateOcsUrl } from '@nextcloud/router'
 import axios from '@nextcloud/axios'
 import NcActionCheckbox from '@nextcloud/vue/dist/Components/NcActionCheckbox.js'
 import NcCheckboxRadioSwitch from '@nextcloud/vue/dist/Components/NcCheckboxRadioSwitch.js'
+import NcInputField from '@nextcloud/vue/dist/Components/NcInputField.js'
 import IconCheckboxBlankOutline from 'vue-material-design-icons/CheckboxBlankOutline.vue'
 import IconRadioboxBlank from 'vue-material-design-icons/RadioboxBlank.vue'
 
@@ -117,6 +146,7 @@ export default {
 		IconRadioboxBlank,
 		NcActionCheckbox,
 		NcCheckboxRadioSwitch,
+		NcInputField,
 	},
 
 	mixins: [QuestionMixin],
@@ -124,10 +154,19 @@ export default {
 	data() {
 		return {
 			questionValues: this.values,
+			inputOtherAnswer: this.valueToInputOtherAnswer(),
+			QUESTION_EXTRASETTINGS_OTHER_PREFIX: 'system-other-answer:',
 		}
 	},
 
 	computed: {
+		placeholderOtherAnswer() {
+			if (this.readOnly) {
+				return this.answerType.submitPlaceholder
+			}
+			return this.answerType.createPlaceholder
+		},
+
 		contentValid() {
 			return this.answerType.validate(this)
 		},
@@ -160,6 +199,19 @@ export default {
 		titleId() {
 			return `q${this.$attrs.index}_title`
 		},
+
+		allowOtherAnswer() {
+			return this.extraSettings?.allowOtherAnswer
+		},
+
+		valueOtherAnswer() {
+			return this.QUESTION_EXTRASETTINGS_OTHER_PREFIX + this.inputOtherAnswer
+		},
+
+		hasRequiredOtherAnswerInput() {
+			const checkedOtherAnswer = this.questionValues.filter(item => item.startsWith(this.QUESTION_EXTRASETTINGS_OTHER_PREFIX))
+			return checkedOtherAnswer[0] !== undefined
+		},
 	},
 
 	watch: {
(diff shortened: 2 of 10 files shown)


Chartman123:
Hi @AIlkiv

thanks for contributing :) I'll have a closer look at it in the next few days. In the meantime you could already go on and fix the errors from our check workflows.

@nextcloud/designers what do you think about the implementation? I'd propose to remove the `Other:` label in front of the input field and have it has the placeholder inside the field.

AIlkiv:
Fixed

src/components/Questions/QuestionMultiple.vue
```diff
@@ -57,6 +61,16 @@
...
 				</NcCheckboxRadioSwitch>
+				<NcCheckboxRadioSwitch v-if="isShowOtherAnswer"
+					:checked.sync="questionValues"
+					:value="valueOtherAnswer"
+					:name="`${id}-answer`"
+					:type="isUnique ? 'radio' : 'checkbox'"
+					:required="checkRequired('other-answer')"
+					@update:checked="onChange"
+					@keydown.enter.exact.prevent="onKeydownEnter">
+					{{ t('forms', 'Other:') }}&nbsp;<NcInputField placeholder="answer..." :value.sync="inputOtherAnswer" class="question__input" />
```
Chartman123, line 72:
```suggestion
					<NcInputField placeholder={{ t('forms', 'Other...') }} :value.sync="inputOtherAnswer" class="question__input" />
```

Please use the text directly as the placeholder. This is also the way Google handles it in Google Forms.

And please adjust the styling of the Input box to matcht the width of the other question types.

Last, as you also use two different placeholders for view/edit/submit mode, we should just move the placeholder text to AnswerTypes.js like for the other question types.

src/components/Questions/QuestionMultiple.vue
```diff
@@ -76,7 +90,16 @@
 						@delete="deleteOption"
 						@update:answer="updateAnswer" />
 				</template>
-
+				<li v-if="isShowOtherAnswer" class="question__item">
+					<div :is="pseudoIcon" class="question__item__pseudoInput" />
+					<input :aria-label="t('forms', 'Add a new answer')"
+						:placeholder="t('forms', 'Field for Other answer')"
```
Chartman123, line 96: See other comment about placeholders

src/components/Questions/QuestionMultiple.vue
```diff
@@ -160,6 +187,14 @@ export default {
 		titleId() {
 			return `q${this.$attrs.index}_title`
 		},
+
+		isShowOtherAnswer() {
```
Chartman123, line 191:
```suggestion
		hasOtherAnswer() {
```

Better wording for method

src/components/Results/ResultsSummary.vue
```diff
@@ -92,6 +92,13 @@ export default {
 				percentage: 0,
 			}))
 
+			// Also record 'Other'
+			questionOptionsStats.unshift({
+				text: t('forms', 'Other'),
+				count: 0,
+				percentage: 0,
+			})
```
Chartman123, line 100: This shows the "Other" option also for dropdown questions.

tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -561,6 +561,18 @@ public function dataValidateSubmission() {
 				// Expected Result
 				false
 			],
+			'required-emty-other-answer' => [
```
Chartman123, line 564:
```suggestion
			'required-empty-other-answer' => [
```

Typo

tests/Unit/Controller/ApiControllerTest.php
Chartman123, line 1: Please only add the test that belong to your PR and add the other tests in a separate PR so that we don't mix several things in one PR.

Chartman123 (changes requested):
I had a closer look at your changes and left some comments. :-)

marcoambrosini:
Thanks for the contribution @AIlkiv
@Chartman123 we cannot use placeholders to deliver important information because of accessibility problems.

I think that this text field that is shown when other is selected should appear below the radio button and span the whole width of the container.

Also I think it would be nice to see what these "other" results are in the answers page.

src/components/Questions/QuestionMultiple.vue
Chartman123: > Thanks for the contribution @AIlkiv @Chartman123 we cannot use placeholders to deliver important information because of accessibility problems.
> 
> I think that this text field that is shown when other is selected should appear below the radio button and span the whole width of the container.

@marcoambrosini Thanks for your input 🙂 

Here are two screenshots how Google handles it currently:

|Edit|View|
|---|---|
|image|image|

So it's a combination of what I suggested and @AIlkiv implemented. As they don't have a border around the input box it looks fine with the input box in line with the radio/checkbox button.

How would Allkiv's version look like with a RTL language?

Chartman123:
> Also I think it would be nice to see what these "other" results are in the answers page.

@marcoambrosini This could blow up the results even for the "same" answers if they are typed differently, but I also thought about this...

marcoambrosini:
> @marcoambrosini This could blow up the results even for the "same" answers if they are typed differently, but I also thought about this...

Maybe we could think about a way to "dive" into answers within the "others" bucket?

AIlkiv:
> Also I think it would be nice to see what these "other" results are in the answers page.

@marcoambrosini You can see the value of 'Other' in the Responses section.

src/components/Questions/QuestionMultiple.vue
AIlkiv: > Please use the text directly as the placeholder. This is also the way Google handles it in Google Forms.
> 
> And please adjust the styling of the Input box to match the width of the other question types.
> 
> Last, as you also use two different placeholders for view/edit/submit mode, we should just move the placeholder text to AnswerTypes.js like for the other question types.
@Chartman123 

image

image

image

> How would Allkiv's version look like with a RTL language?

image

When editing, the layout of some fields is displayed incorrectly, but this is the topic of a separate task:

image

AIlkiv:
@Chartman123  It looks like I fixed everything.

marcoambrosini:
@nextcloud/designers what do you think about this? I think it's still best to keep the input field below the radio. 
Looks better and works better on mobile screens.

src/components/Questions/QuestionMultiple.vue
Chartman123: > When editing, the layout of some fields is displayed incorrectly, but this is the topic of a separate task:

Yes, there's already another PR open to improve RTL support :)

nimishavijay:
Very cool! I don't have a strong opinion on the placement of the field, we can show it below like @marcoambrosini said so that it's nicely left aligned with the rest of the options :) 

Rest of design looks great, some wording suggestions
- [ ] "People can enter a other answer" --> "People can submit a different answer" 
- [ ] "Enter your other answer" --> "Enter your answer" (to avoid duplication of the word "Other")
- [ ] "Other..." --> "Other" (we usually use ellipses when there is some text that is cut off or hidden)

Chartman123:
Two more things: 
- First, I think that we should make sure, that if "other" is selected as answer, the input field should have a value. Currently you can select it but don't have to enter a value. Google allows this, too, but I don't think that this makes sense.
- I just saw that the hover effect doesn't look good with the text field, for both the in-line version and the new-line version. But I'm not sure what would be the best way to deal with it.

@marcoambrosini If we just add the line break like in the screenshots above, I don't like the look as it moves the text higher than the checkbox next to it. So did you think of a completely new line with the left alignment like for a normal input field? In this case we could also limit the hover effect to just the checkbox and the "Other: " label.

However, I don't think it's too problematic on mobile screens with the in-line field: (compared to the short answer field below)
grafik

src/components/Questions/QuestionMultiple.vue
```diff
@@ -76,7 +92,15 @@
 						@delete="deleteOption"
 						@update:answer="updateAnswer" />
 				</template>
-
+				<li v-if="hasOtherAnswer" class="question__item">
+					<div :is="pseudoIcon" class="question__item__pseudoInput" />
+					<input :placeholder="t('forms', 'Other...')"
+						class="question__input"
+						:maxlength="maxStringLengths.optionText"
```
Chartman123, line 99: Don't we have to add the string length of the prefix to the length of the input field for storing the value in the database? So we need to make sure the input is only maxStringLenghts.optionText (1024) - QUESTION_EXTRASETTINGS_OTHER_PREFIX.length() (20) = 1004

However, this is the wrong place to check this, as this is only used for the edit mode, where we can't add any text to the input field. So you have to move this maxlenght limitation to the input field in view mode.

marcoambrosini (commented):
Regardless of the positioning of the text field, it shouldn't be within the radio button's bundaries. If we do it inline we need to make sure it wraps on mobile screens

Screenshot 2023-08-24 at 07 35 46

AIlkiv:
> Незалежно від розташування текстового поля, воно не повинно бути в межах перемикача. Якщо ми робимо це вбудовано, нам потрібно переконатися, що воно переносить екрани мобільних пристроїв

@marcoambrosini In this version, in my opinion, there is no understanding that the radio button and the input field are one field.

In Google Forms on mobile devices, it looks normal when everything is in one row.

marcoambrosini:
Google Forms' layout breaks on a mobile screen though.
 
Screenshot 2023-08-24 at 15 13 11

AIlkiv:
> Regardless of the positioning of the text field, it shouldn't be within the radio button's boundaries. If we do it inline we need to make sure it wraps on mobile screens
> 
> Скріншот 2023-08-24 на 07 35 46

@marcoambrosini  Your example has an additional label, making it appear well-structured. I attempted to break it into two lines, but it seems less intuitive to me (I see four options and a separate option to enter an answer). What are your thoughts?

image

image

Chartman123:
@AIlkiv could you try to align the text inside the input field with the text of the other options? And to make it more clear that they both belong together you could change the label to `Other:` again

AIlkiv:
> @AIlkiv could you try to align the text inside the input field with the text of the other options? And to make it more clear that they both belong together you could change the label to `Other:` again

@Chartman123 

image

image

image

Chartman123:
@marcoambrosini ok, you're right. 👍🏻 

@AIlkiv you can use `v-if="isMobile"` here

marcoambrosini:
> you can use v-if="isMobile" here

@Chartman123  I don't think we can use `isMobile` here. `isMobile` just tells us that the total width of the page is below `1024px`, in which the inline input would still look fine.
Conversely, we might have a 4k widescreen but have this form placed in a tiny area of the page or in a dialog. 

So what we really care about is the width of the container, but that's a bit hard to "watch". I suggest we put the button and input in a flexbox and give a `min-width` of `300px` to the input with `flex-wrap` set to `wrap`

AIlkiv:
image

image

image

@Chartman123 @marcoambrosini Is it normal?

In order to prevent the "hover" effect from overlapping the input field, I had to add a margin after the label "Other:"

src/components/Questions/QuestionMultiple.vue
AIlkiv: We only store values in the database.

AIlkiv:
@Chartman123 Fixed everything.

src/components/Questions/QuestionMultiple.vue
Chartman123: My bad... sorry, you're of course correct :)

Chartman123:
@AIlkiv looks good from my side 👍🏻  just one last thing: you could shorten your commit message to `feat: ...` if you want to :)

Chartman123 (approved):

lib/Service/SubmissionService.php
```diff
@@ -309,7 +309,8 @@ public function validateSubmission(array $questions, array $answers): bool {
 			$questionAnswered = array_key_exists($questionId, $answers);
 
 			// Check if all required questions have an answer
-			if ($question['isRequired'] && (!$questionAnswered || !array_filter($answers[$questionId], 'strlen'))) {
+			if ($question['isRequired'] &&
+				(!$questionAnswered || !array_filter($answers[$questionId], 'strlen') || !array_filter($answers[$questionId], fn ($value) => $value !== Constants::QUESTION_EXTRASETTINGS_OTHER_PREFIX))) {
```
susnux, line 313: Not sure if we want to limit the new check to `in_array($question['type'], Constants::ANSWER_TYPES_PREDEFINED))`.
Otherwise also text input fails if it includes that string. (should not happen, but why not?)

susnux (approved):
Very nice! Thank you :tada: 

Looks good overall, I do not like the `otherAnswer` name, but should be ok.
So, this good from my point :)

lib/Service/SubmissionService.php
Chartman123: Yes, makes sense in my opinion 👍🏻

Chartman123:
> I do not like the otherAnswer name

Do you have something better in mind? We could still change it before merging...

susnux:
> Do you have something better in mind? We could still change it before merging...

Not really, I just think it is a bit misleading. It sounds like it is possible to have multiple different answers.
Maybe like `isRequired` we could use `allowCustomAnswer`?

Chartman123:
@AIlkiv Before merging, could you please add the new prop to the docs at docs/API.md and docs/DataStrutcture.md?

lib/Service/SubmissionService.php
AIlkiv: Add `!empty($question['extraSettings']->allowedOtherAnswer)`

AIlkiv:
> Maybe like `isRequired` we could use `allowCustomAnswer`?

I modified the name to 'allowOtherAnswer' in the 'extraSettings'.

susnux:
> > Maybe like `isRequired` we could use `allowCustomAnswer`?
> 
> I modified the name to 'allowOtherAnswer' in the 'extraSettings'.

Nice, please push the changes and add it to the api docs, then we can merge this :)

AIlkiv:
@Chartman123 @susnux Finished

Chartman123 (approved):

src/components/Questions/QuestionMultiple.vue
```diff
@@ -402,4 +472,24 @@ export default {
 		}
 	}
 }
+
+.question__other-answer::v-deep {
```
susnux, line 476: Is the deep sector required? If yes we should use `:deep()` as v-deep id deprecated

marcoambrosini:
> I had to add a margin after the label "Other:"

@AIlkiv could you add a bigger margin, say 8px?

src/components/Questions/QuestionMultiple.vue
AIlkiv: This is required for a single selector. Divided into parts with :deep() and without :deep().

AIlkiv:
> @AIlkiv could you add a bigger margin, say 8px?

@marcoambrosini  Changed code and push

image

AIlkiv:
> Sorry @AIlkiv, without the checked feedback the input field looks too far from the label "Other", so could we dial back the margin to 4px. Thank you :)

Ok, refreshed commit.

image

marcoambrosini:
Thanks a lot for your contributions @AIlkiv and @Chartman123 ❤️

AIlkiv:
@marcoambrosini @Chartman123 @susnux  Thank you for the collaboration!

PR #1696, opened 2023-08-22
Make components compatible to nextcloud-vue 8

Chartman123:
This adds compatibility to nextcloud-vue v8 following the pinned post there:

https://github.com/nextcloud-libraries/nextcloud-vue/issues/4223

### DIFF
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -57,7 +57,7 @@
 		<!-- No forms & loading emptycontents -->
 		<NcAppContent v-if="loading || !hasForms || !routeHash || !routeAllowed">
 			<NcEmptyContent v-if="loading"
-				class="emptycontent"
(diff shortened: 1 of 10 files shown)


Chartman123:
grafik

@nextcloud/designers Do we need to invert the icon in the navigation for the selected form?

@susnux Could you please have a look at my changes regarding `NcEmptyContent`? I think that in some cases it's not properly centered.

susnux:
> @susnux Could you please have a look at my changes regarding NcEmptyContent? I think that in some cases it's not properly centered.

Pushed a commit, basically `emptycontent` is already a class name used by the server, so it interferes (thats why I do not like scoped styles but really recommend css modules for (new) projects).

susnux (approved):

PR #1698, opened 2023-08-24
Make possible hide fields from header

vitormattos:
## Motivation

Make possible use a form embedded in a site as contact form by example. A contact form sometimes haven't a description.

## Before

Screenshot_20230825_130148

## After
Screenshot_20230825_130158

**How I implemented?**
I searched by `show?(expiration)` to look all places with this property and added the changes

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -14,9 +14,9 @@
 	]]></description>
 
 	<licence>agpl</licence>
-	<version>3.3.1</version>
+	<version>3.4.0</version>
 	<dependencies>
-		<nextcloud min-version="25" max-version="27" />
+		<nextcloud min-version="25" max-version="28" />
 	</dependencies>
 
 	<author>Affan Hussain</author>
@@ -35,6 +35,7 @@
 	<author>Stephan Link</author>
 	<author>Tim Sattizahn</author>
 	<author>Vinzenz Rosenkranz</author>
+	<author>Vitor Mattos</author>
 
 	<category>tools</category>
 	<category>social</category>
--- a/docs/API.md
+++ b/docs/API.md
@@ -116,7 +116,6 @@ Returns the full-depth object of the requested form (without submissions).
   "hash": "em4djk8B9BpXnkYG",
   "showHeader": true,
   "title": "Form 1",
-  "showTitle": true,
   "description": "Description Text",
   "ownerId": "jonas",
   "created": 1611240961,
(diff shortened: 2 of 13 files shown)


Chartman123:
Hi @vitormattos

thank you for your contribution. However, I don't get your point as the description will only be shown if it's set. Why don't you just leave it empty when you don't want to show it?

vitormattos:
Hi @Chartman123,
Maybe the admin will want to put a description to herself to identify the purpose of form or any other reason, and don't want to display the description to users.

vitormattos:
This PR can complement the use cases generated by #1353

vitormattos:
I did a more push to make possible completely hide the header.

vitormattos:
@jotoeri maybe we can maintain only the option to completely hide the header.

> I'd rather try to keep the UI simple instead of configuring every tiny bit.
> 
> > description to herself to identify the purpose of form or any other reason
> 
> That is more a mis-usage of the intended description field. So you'd basically with this only introduce an option to mis-use a functionality. With the drawback of a more complex UI. That's not the intention of the app.

jotoeri:
> @jotoeri maybe we can maintain only the option to completely hide the header.

Would be at least less disturbing, yes.
->  But still it's an option for a specific use-case, that you might have for embedding. 90% of the users just see a very strange option to create a head-less form. Why not just hiding the header via css on your specific embedding?

vitormattos:
Hiding only by CSS will make necessary touch the code creating a new app or implementing the CSS in another place. This  could have problems in a future change of id or class used to generate the CSS selector making very hardy to maintain the CSS customization.

Other point is that isn't all user that know how to use CSS to hide elements.

Exists an app to inject CSS without touching the code creating a new app to only hide the header of a form, but  this also is a overhead because when more app we have, more slow our Nextcloud instance will be because is more apps that need to be initialized.

Chartman123:
@jotoeri @vitormattos

As this is intended mainly towards embedded forms, what about moving the setting to the menu of an embedding enabled link?

And I'd also like to have only one setting for the whole header if we agree on having it added :)

vitormattos:
Hi @jotoeri & @Chartman123 ,

After our messages yesterday, I changed my PR reverting the implementation of toggle to title and description and kept only the toggle of header.

Could look if now is OK?

jancborchardt (changes requested):
Design-wise I would reject this as it goes against one of our design guidelines which is to not offer settings for every little thing – as was mentioned already. :)

Chartman123:
Ok, so I'll close this. And as we already pointed out theres a workaround using custom css to hide the information when a form is embedded into another site.

PR #1704, opened 2023-08-31
Added test for insertSubmission in ApiController

AIlkiv:


tests/Unit/Controller/ApiControllerTest.php
```diff
@@ -669,4 +669,67 @@ public function testInsertSubmission_answers() {
...
+		$this->apiController->insertSubmission(1, [], '');
+	}
+
+	public function dataForCheckForbiddenException() {
+		return [
+			[false, true, true],
+			[true, true, true],
+			[true, false, false],
+		];
+	}
```
Chartman123, line 689: Could you please add a description to the function how the test array is built? Just to make it clearer what the bools mean 🙂

Chartman123 (approved):
Just a small comment, rest looks good to me 👍🏻

tests/Unit/Controller/ApiControllerTest.php
AIlkiv: Added comment.

PR #1705, opened 2023-08-31
fix: Enhance extraSettings handling and fix XML output

### LINKED ISSUE #1469: API: Can not get form in XML format
**Describe the bug**
Currently it is not possible to get a form, using the API, in the `xml` format, only JSON is possible.
The response a HTTP 200, with empty content and following error in the server log:
```json
{
  "reqId": "lDSN6tYXk1tWzlUNgifS",
  "level": 3,
  "time": "2023-01-29T14:59:32+00:00",
  "remoteAddr": "172.18.0.1",
  "user": "admin",
  "app": "PHP",
  "method": "GET",
  "url": "/ocs/v2.php/apps/forms/api/v2/form/1",
  "message": "TypeError: XMLWriter::writeElement(): Argument #2 ($content) must be of type ?string, stdClass given at /var/www/html/lib/private/AppFramework/OCS/BaseResponse.php#148",
  "userAgent": "curl/7.87.0",
  "version": "26.0.0.2",
  "data": {
    "app": "PHP"
  }
}
```

**To Reproduce**
1. `curl -v -u admin:admin -X GET 'http://localhost:8080/ocs/v2.php/apps/forms/api/v2/form/1' -H "OCS-APIRequest: true"`
2. See empty response

**Expected behavior**
OCS response with form in XML format.

- Nextcloud-Version: 26 (current master)
- Forms-Version: current main

**Additional context**

Chartman123:
This fixes #1469

The getExtraSettings() function now returns an associative array so that the XML response of the API is working again.

Furthermore, we now store only set (`true`) extraSettings to the DB.

### DIFF
--- a/docs/DataStructure.md
+++ b/docs/DataStructure.md
@@ -60,7 +60,7 @@ This document describes the Object-Structure, that is used within the Forms App 
 | text           | String          | max. 2048 ch. | The question-text |
 | name           | String          |              | Technical identifier of the question, e.g. used as HTML name attribute |
 | options        | Array of [Options](#option) | | Array of options belonging to the question. Only relevant for question-type with predefined options. |
-| extraSettings  | StdClass        |              | Additional settings for the question. For dropdown/radio/checkbox, there can be a 'shuffleOptions': true - the list of options should be shuffled. For radio/checkbox, 'allowOtherAnswer': true - allows the user to specify their own option. |
+| extraSettings  | [Extra Settings](#extra-settings)        |              | Additional settings for the question.  |
 ```
 {
   "id": 1,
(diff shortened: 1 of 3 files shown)


Chartman123:
@jotoeri @susnux I dug into this now and found that XML output is broken because of our extraSettings JSON of the questions. However, I have by now not been able to fix it with the hints from @provokateurin

Could someone of you please have a look at my code and point me in the right direction? Thank you very much in advance :)

Chartman123:
Ok, I think I found the reason why the XML output breaks... It's the getExtraSettings() in Question.php that returns an object. I think we must refactor the code to work with an associative array here just like for getAccess() in Form.php

AlexM4H:
Can someone else possibly review the PR as well, or is there something else preventing the merge?

PR #1707, opened 2023-09-01
Optimization of FormMapper::findById calls

AIlkiv:
Motivation:

There is a method called ApiController::getSharedForms, which is invoked on the main page of the module.

This method retrieves all web forms without exception.

And then this method invokes various other methods to obtain the Form using its formId. In essence, it performs N (representing the total number of forms in the system) additional requests to the database. Even if these requests are somehow cached, they still amount to N unnecessary cache requests.

For other methods, the number of findById calls is reduced.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -155,7 +155,7 @@ class ApiController extends OCSController {
 
 		$result = [];
 		foreach ($forms as $form) {
-			$result[] = $this->formsService->getPartialFormArray($form->getId());
+			$result[] = $this->formsService->getPartialFormArray($form);
 		}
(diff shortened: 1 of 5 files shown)


AIlkiv:
@Chartman123 could you kindly perform a code review, please?

Chartman123:
Hi @AIlkiv 

sorry, I don't have much time at the moment. Will try to review your PR next week.

susnux (approved):
Very nice work,  code looks good and make sense

Chartman123 (approved):
Fine for me as well :) Could you please do another rebase on latest main and then I think we can merge this one.

PR #1727, opened 2023-09-27
Move to min-version 26 and drop PHP 7.4

Chartman123:


composer.json
```diff
@@ -5,7 +5,7 @@
 		"classmap-authoritative": true,
 		"sort-packages": true,
 		"platform": {
-			"php": "7.4"
+			"php": "8.2"
```
susnux, line 8:
```suggestion
			"php": "8.0"
```

Chartman123: Just for my understanding: why use the oldest supported version here?

susnux: It defines the minimum requirements that will be checked when installing the dependencies. 

Ref: https://getcomposer.org/doc/06-config.md#platform

susnux (approved):

PR #1730, opened 2023-09-29
fix: Incorrect type for empty extraSettings in frontend

Chartman123:
This fixes an error introduced with the XML fix in #1705 

For questions where extraSettings weren't set we got a type mismatch in the extraSettings prop (Object expected, array provided)

### DIFF
--- a/src/mixins/QuestionMixin.js
+++ b/src/mixins/QuestionMixin.js
@@ -116,10 +116,8 @@ export default {
 
 		/**
 		 * Extra settings
-		 * Currently only contains whether options should be shuffled
 		 */
 		extraSettings: {
-			type: Object,
 			default: () => {
 				return {}
 			},


susnux:
Not sure if this is a proper fix or just silencing this error. As the error has a reason: The given type a an array instead of the expected object.

The reason is if the extra settings are empty, php will return an empty assoc. array which will be converted to `[]` instead of `{}` for the json.

susnux (approved):
But I think there is no other good solution, because either we have to fix it when the form is loaded in the frontend.

Or drop XML support as it seems broken.

Chartman123:
> But I think there is no other good solution, because either we have to fix it when the form is loaded in the frontend.
> 
> Or drop XML support (no idea if used, but we should not drop it now).

Yes, with this fix the extraSettings prop is either an array[0] or an object depending of the result from the database... Do you see any chance to fix this in the backend and always return an object in JSON?

PR #1731, opened 2023-09-29
fix: Warning about missing label for other answer

Chartman123:
NcInputField for other answer throws a warning about missing label prop. As label also works as placeholder, this simply changes the prop for the placeholder text from placeholder to label.

### DIFF
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -72,7 +72,7 @@
 						@keydown.enter.exact.prevent="onKeydownEnter">
 						{{ t('forms', 'Other:') }}
 					</NcCheckboxRadioSwitch>
-					<NcInputField :placeholder="placeholderOtherAnswer"
+					<NcInputField :label="placeholderOtherAnswer"
 						:required="hasRequiredOtherAnswerInput"
 						:value.sync="inputOtherAnswer"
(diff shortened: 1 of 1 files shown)


src/components/Questions/QuestionMultiple.vue
```diff
@@ -72,7 +72,7 @@
 						@keydown.enter.exact.prevent="onKeydownEnter">
 						{{ t('forms', 'Other:') }}
 					</NcCheckboxRadioSwitch>
-					<NcInputField :placeholder="placeholderOtherAnswer"
+					<NcInputField :label="placeholderOtherAnswer"
```
susnux, line 75: No, this makes no sense for accessibility, as the label of the input would be "Enter your answer".
It would make more sense to have a separate label like "Custom answer" or "Other answer" and keep "Enter your answer" as the placeholder.

Chartman123: Or we could change the placeholder text for submitPlaceholder to "Enter your own answer"? This way we wouldn't need another string...

susnux: Lets do this for now. As soon as we use nextcloud/vue 8 we could just use the inner labels and rework this :)

susnux (approved):

PR #1467, opened 2023-01-28
Feature: Allow to share forms with Circles

susnux:
* Resolves #1012 

## Summary
This allows to share forms with circles.
The circle manager is not directly injected as it is possible that users disabled the circles app, so only the container is injected and then used to query the manager (if available).

The front end changes are minimal, just changed the placeholder and added the `display-name` property to the `NcAvatar` to fix circle avatars (if not added the avatar would be generated from the ID which would have wrong initials).

PR #1746, opened 2023-10-16
feat(ActivityManager): Send notification about new submissions to circle memebers

susnux:
Follow up on #1496 and #1467

### DIFF
--- a/tests/Unit/Activity/ActivityManagerTest.php
+++ b/tests/Unit/Activity/ActivityManagerTest.php
@@ -241,7 +241,13 @@ class ActivityManagerTest extends TestCase {
 				'sharedGroup',
 				[['user1'], ['user2']],
 				['user1', 'user2'],
-			]
+			],
+			'circle-share' => [
+				IShare::TYPE_CIRCLE,
+				'sharedCircle',
+				[['user1'], ['user2']],
+				['user1', 'user2'],
+			],
 		];
 	}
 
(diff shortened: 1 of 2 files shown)


tests/Unit/Activity/ActivityManagerTest.php
```diff
@@ -259,9 +265,12 @@ public function testPublishNewSharedSubmission(int $shareType, string $shareWith
 		$form->setOwnerId('formOwner');
 		$submitterId = 'submittingUser';
 
-		if ($groupUsers === null) {
-			$this->groupManager->expects($this->never())->method('get');
-		} else {
+		if ($shareType === IShare::TYPE_CIRCLE) {
+			$this->circlesService->expects($this->once())
+				->method('getCircleUsers')
+				->with('sharedCircles')
```
Chartman123, line 271:
```suggestion
				->with('sharedCircle')
```

For groups we also only use the singular in the with-method (see line 241)

tests/Unit/Activity/ActivityManagerTest.php
```diff
@@ -259,9 +265,12 @@ public function testPublishNewSharedSubmission(int $shareType, string $shareWith
...
+				->willReturn($groupUsers);
```
Chartman123, line 272: I wouldn't expect `$groupUsers` here... Is this intended?

tests/Unit/Activity/ActivityManagerTest.php
Chartman123: This is also the reason for the failing PHPUnit checks

tests/Unit/Activity/ActivityManagerTest.php
susnux: renamed the argument so it does not confuse anymore ;-)

tests/Unit/Activity/ActivityManagerTest.php
susnux: Thank you, typo :see_no_evil:

Chartman123 (approved):

Issue #1256, opened 2022-07-15
Focus order issues


PVince81:
When adding checkboxes to a form there’s an issue with focus order. When we opened the options for the question focus moved through the options as expected, but after that, it jumped to higher up on the page.

When finished moving through an opened area such as this focus should move to the next interactive object after the one that opened the area. It should not jump to a completely different part of the page.

![image](https://user-images.githubusercontent.com/277525/179245629-a0eea75c-e69b-49ac-8580-ed037ac65cfd.png)

The headings in the form are interactive but aren’t possible to use with the keyboard. When you make them interactive for keyboard users, take care that they’re placed logically in the focus order.

![image](https://user-images.githubusercontent.com/277525/179245679-864822f3-d14c-4a7e-bf54-2f1ac9ead2d6.png)

When opening the fly-out in the form focus isn’t moved into the new area, but back to the main content.

![image](https://user-images.githubusercontent.com/277525/179245704-f598dd00-df41-40ff-8043-3ea5a46cf73f.png)

PR #1750, opened 2023-10-19
fix: Make form editable with keyboard

### LINKED ISSUE #416: Impossible to edit questions/answers with keyboard only after initial writing

Chartman123:
This fixes #416 by always showing the edit mode for all fields in Edit/Create view

grafik

To-Do:
- [x] Fix tabbing from multilple choice questions to next question. Probably due to switching between normal Input and AnswerInput.vue (also tracked in #1256)
- [x] Focus after adding question not in title
- [x] emit onInput event after creating a new option only works after first input

### DIFF
--- a/src/components/Questions/AnswerInput.vue
+++ b/src/components/Questions/AnswerInput.vue
@@ -1,5 +1,5 @@
 <template>
-	<li class="question__item">
+	<li class="question__item" @focusout="handleTabbing">
 		<div :is="pseudoIcon"
 			v-if="!isDropdown"
 			class="question__item__pseudoInput" />
@@ -15,7 +15,7 @@
 			dir="auto"
 			@input="onInput"
 			@keydown.delete="deleteEntry"
-			@keydown.enter.prevent="addNewEntry">
+			@keydown.enter.prevent="focusNextInput">
 
 		<!-- Delete answer -->
 		<NcActions>
@@ -98,6 +98,10 @@ export default {
 	},
 
 	methods: {
+		handleTabbing() {
+			this.$emit('tabbed-out')
+		},
+
 		/**
 		 * Focus the input
 		 */
@@ -133,8 +137,8 @@ export default {
 		/**
 		 * Request a new answer
 		 */
-		addNewEntry() {
-			this.$emit('add')
+		focusNextInput() {
+			this.$emit('focus-next', this.index)
 		},
 
 		/**
@@ -221,13 +225,14 @@ export default {
 	.question__input {
 		width: 100%;
 		position: relative;
-		margin-inline-end: 2px !important;
+		inset-inline-start: -12px;
+		margin-inline-end: -12px !important;
 
 		&--shifted {
-			inset-inline-start: -30px;
+			inset-inline-start: -34px;
 			inset-block-start: 1px;
-			margin-inline-end: -30px !important;
-			padding-inline-start: 32px !important;
+			margin-inline-end: -34px !important;
+			padding-inline-start: 36px !important;
 		}
 	}
 }
--- a/src/components/Questions/QuestionDropdown.vue
+++ b/src/components/Questions/QuestionDropdown.vue
@@ -33,7 +33,7 @@
 				{{ t('forms', 'Shuffle options') }}
 			</NcActionCheckbox>
 		</template>
-		<NcSelect v-if="!edit"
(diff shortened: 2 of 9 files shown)


Chartman123:
@susnux @jotoeri While reviewing, please especially have a closer look at my second and third commit. They make the code work the way I wanted, but I'm not sure if there weren't better solutions :)

src/components/Questions/AnswerInput.vue
```diff
@@ -15,7 +15,8 @@
 			dir="auto"
 			@input="onInput"
 			@keydown.delete="deleteEntry"
-			@keydown.enter.prevent="addNewEntry">
+			@keydown.enter.prevent="addNewEntry"
+			@focusout="leaveField">
```
susnux, line 19: I think we do not need this, besides proxying a native HTML event with its HTML name seems to be not a good idea, this is not needed because `focusout` bubbles and therefor we can catch it on the parent without proxying.

susnux (commented):
Code looks good, one comment.
But it seems that adding new options on pressing enter is broken now

src/components/Questions/AnswerInput.vue
Chartman123: @susnux I think I only added this as it didn't work with just the `@focusout` in the QuestionDropdown\Multiple.vue

Chartman123:
> But it seems that adding new options on pressing enter is broken now

@susnux Yes, thats true. I will have another look at this...

Do you have an idea how I can fix the problem that new options will only be added to the database when you have at least more than one input events on a new option field? I couldn't find any working solution myself by now...

src/components/Questions/AnswerInput.vue
Chartman123: @susnux I just checked it again, without the @focusout here and then the emit in leaveField the checkValidOptions in QuestionDropdown isn't triggered at all. But Enter/Delete start working again...

susnux (approved):

src/components/Questions/QuestionDropdown.vue
```diff
@@ -56,17 +56,18 @@
 				:max-option-length="maxStringLengths.optionText"
 				@add="addNewEntry"
 				@delete="deleteOption"
-				@update:answer="updateAnswer" />
+				@update:answer="updateAnswer"
+				@focusout="checkValidOption" />
```
susnux, line 60: Should it not be `@tabbed-out`?

Chartman123: Sure, did you add it only to Multiple in your commit? I didn't have a closer look :-/

src/components/Questions/QuestionMultiple.vue
```diff
@@ -252,16 +252,18 @@ export default {
 			return this.areNoneChecked
 		},
 
-		/**
-		 * Remove any empty options when leaving an option
-		 */
```
susnux: why removing the doc block?

src/components/Questions/QuestionMultiple.vue
```diff
@@ -252,16 +252,18 @@ export default {
...
+					return false
+				}
+				return true
+			})
+
 			// update parent
-			if (options.length !== this.options.length) {
-				this.updateOptions(options)
-			}
+			this.updateOptions(options)
```
susnux, line 266: Ah I see the delete was doing more than updating :see_no_evil: 

But the `this.updateOptions(options)` is not needed as it is part of `deleteOption` function, so the same will be emitted twice. So I think we can simplify to:
```js
// When leaving edit mode, filter and delete empty options
this.options.forEach(option => {
	if (!option.text) {
		this.deleteOption(option.id)
	}
})
```

susnux (approved):
Works but one comment

src/components/Questions/QuestionMultiple.vue
Chartman123: This was not intended...

PR #1690, opened 2023-08-18
feat: allow editing of submission by the user

tpokorra:
related to #1110 

- [x] show checked radio buttons and check boxes with previous answer
- [x] update submission and answers, don't delete and create submission or answers
- [x] allow deletion of submission
- [x] rebase to main branch 
- [x] resolve Lint and Unit Test Job Failures
- [x] add Unit tests for new functionality

PR #1758, opened 2023-10-22
Added possibility to link spreadsheet for automatic submission export in multiple formats

Koc:
Hello everyone!

I've added possibility to link form with spreadsheet in csv/ods/xlsx format and automatically export it once new submission added. Inspired from #1403 but with more formats. Also we edit already existent file, which give us possibility to add additional columns with notes/statuses/etc.

Here you can find few video/screenshots:

https://github.com/nextcloud/forms/assets/191082/4e1426dc-17dd-4328-81e4-7e84bd43e414

<details>
<summary>Exported file</summary>

image
</details>

What is out of scope right now but can be added in upcoming PRs:
- handling of responses deletion
- handling of adding/removal of the questions

Todo:
- [x] fix old tests and add new

### DIFF
--- a/.github/workflows/lint-php-cs.yml
+++ b/.github/workflows/lint-php-cs.yml
@@ -30,6 +30,7 @@ jobs:
           php-version: 8.2
           coverage: none
           ini-file: development
+          extensions: gd
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -4,6 +4,7 @@
  *
  * @author affan98 <affan98@gmail.com>
  * @author Christian Hartmann <chris-hartmann@gmx.de>
+ * @author Ferdinand Thiessen <opensource@fthiessen.de>
  * @author John Molakvoæ (skjnldsv) <skjnldsv@protonmail.com>
  * @author Jonas Rittershofer <jotoeri@users.noreply.github.com>
  *
@@ -73,7 +74,7 @@ return [
 			'verb' => 'OPTIONS',
 			'requirements' => [
 				'path' => '.+',
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 
@@ -83,7 +84,7 @@ return [
 			'url' => '/api/{apiVersion}/forms',
 			'verb' => 'GET',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -91,7 +92,7 @@ return [
 			'url' => '/api/{apiVersion}/form',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -107,7 +108,7 @@ return [
 			'url' => '/api/{apiVersion}/form/clone/{id}',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		// TODO: Remove POST in next API release
@@ -124,7 +125,7 @@ return [
 			'url' => '/api/{apiVersion}/form/update',
 			'verb' => 'PATCH',
 			'requirements' => [
-				'apiVersion' => 'v2\.[2-3]'
+				'apiVersion' => 'v2\.[2-4]'
 			]
 		],
 		[
@@ -132,7 +133,7 @@ return [
 			'url' => '/api/{apiVersion}/form/transfer',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2\.[2-3]'
+				'apiVersion' => 'v2\.[2-4]'
 			]
 		],
 		[
@@ -140,7 +141,7 @@ return [
 			'url' => '/api/{apiVersion}/form/{id}',
 			'verb' => 'DELETE',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -148,7 +149,7 @@ return [
 			'url' => '/api/{apiVersion}/partial_form/{hash}',
 			'verb' => 'GET',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -156,7 +157,7 @@ return [
 			'url' => '/api/{apiVersion}/shared_forms',
 			'verb' => 'GET',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 
@@ -166,7 +167,7 @@ return [
 			'url' => '/api/{apiVersion}/question',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		// TODO: Remove POST in next API release
@@ -183,7 +184,7 @@ return [
 			'url' => '/api/{apiVersion}/question/update',
 			'verb' => 'PATCH',
 			'requirements' => [
-				'apiVersion' => 'v2\.[2-3]'
+				'apiVersion' => 'v2\.[2-4]'
 			]
 		],
 		// TODO: Remove POST in next API release
@@ -200,7 +201,7 @@ return [
 			'url' => '/api/{apiVersion}/question/reorder',
 			'verb' => 'PUT',
 			'requirements' => [
-				'apiVersion' => 'v2\.[2-3]'
+				'apiVersion' => 'v2\.[2-4]'
 			]
 		],
 		[
@@ -208,7 +209,7 @@ return [
 			'url' => '/api/{apiVersion}/question/{id}',
 			'verb' => 'DELETE',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -216,7 +217,7 @@ return [
 			'url' => '/api/{apiVersion}/question/clone/{id}',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2.3'
+				'apiVersion' => 'v2\.[3-4]'
 			]
 		],
 
@@ -226,7 +227,7 @@ return [
 			'url' => '/api/{apiVersion}/option',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		// TODO: Remove POST in next API release
@@ -243,7 +244,7 @@ return [
 			'url' => '/api/{apiVersion}/option/update',
 			'verb' => 'PATCH',
 			'requirements' => [
-				'apiVersion' => 'v2\.[2-3]'
+				'apiVersion' => 'v2\.[2-4]'
 			]
 		],
 		[
@@ -251,7 +252,7 @@ return [
 			'url' => '/api/{apiVersion}/option/{id}',
 			'verb' => 'DELETE',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 
@@ -261,7 +262,7 @@ return [
 			'url' => '/api/{apiVersion}/share',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -269,7 +270,7 @@ return [
 			'url' => '/api/{apiVersion}/share/{id}',
 			'verb' => 'DELETE',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		// TODO: Remove POST in next API release
@@ -286,7 +287,7 @@ return [
 			'url' => '/api/{apiVersion}/share/update',
 			'verb' => 'PATCH',
 			'requirements' => [
-				'apiVersion' => 'v2\.[2-3]'
+				'apiVersion' => 'v2\.[2-4]'
 			]
 		],
 
@@ -296,7 +297,7 @@ return [
 			'url' => '/api/{apiVersion}/submissions/{hash}',
 			'verb' => 'GET',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -320,7 +321,7 @@ return [
 			'url' => '/api/{apiVersion}/submissions/{formId}',
 			'verb' => 'DELETE',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -328,7 +329,7 @@ return [
 			'url' => '/api/{apiVersion}/submission/insert',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -336,7 +337,7 @@ return [
 			'url' => '/api/{apiVersion}/submission/{id}',
 			'verb' => 'DELETE',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		// Submissions linking with file in cloud
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -99,7 +99,7 @@ return [
 			'url' => '/api/{apiVersion}/form/{id}',
 			'verb' => 'GET',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -304,7 +304,7 @@ return [
 			'url' => '/api/{apiVersion}/submissions/export/{hash}',
 			'verb' => 'GET',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -312,7 +312,7 @@ return [
 			'url' => '/api/{apiVersion}/submissions/export',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.[1-3])?'
+				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
 		[
@@ -339,5 +339,23 @@ return [
 				'apiVersion' => 'v2(\.[1-3])?'
 			]
 		],
+		// Submissions linking with file in cloud
+		[
+			'name' => 'api#linkFile',
+			'url' => '/api/{apiVersion}/form/link/{fileFormat}',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2.4',
+				'fileFormat' => 'csv|ods|xlsx'
+			]
+		],
+		[
+			'name' => 'api#unlinkFile',
+			'url' => '/api/{apiVersion}/form/unlink',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2.4',
+			]
+		]
 	]
 ];
--- a/docs/API.md
+++ b/docs/API.md
@@ -28,9 +28,16 @@ This file contains the API-Documentation. For more information on the returned D
 - Completely new way of handling access & shares.
 
 ### Other API changes
-- In API version 2.1 the endpoint `/api/v2.1/share/update` was added to update a Share
-- In API version 2.2 the endpoint `/api/v2.2/form/transfer` was added to transfer ownership of a form
+- In API version 2.4 the following endpoints were introduced:
+  - `POST /api/2.4/form/link/{fileFormat}` to link form to a file
+  - `POST /api/2.4/form/unlink` to unlink form from a file
+- In API version 2.4 the following endpoints were changed:
+  - `GET /api/v2.4/submissions/export/{hash}` was extended with optional parameter `fileFormat` to export submissions in different formats
+  - `GET /api/v2.4/submissions/export` was extended with optional parameter `fileFormat` to export submissions to cloud in different formats
+  - `GET /api/v2.4/form/{id}` was extended with optional parameters `fileFormat`, `fileId`, `filePath` to link form to a file
 - In API version 2.3 the endpoint `/api/v2.3/question/clone` was added to clone a question
+- In API version 2.2 the endpoint `/api/v2.2/form/transfer` was added to transfer ownership of a form
+- In API version 2.1 the endpoint `/api/v2.1/share/update` was added to update a Share
 
 ## Form Endpoints
 ### List owned Forms
@@ -111,7 +118,7 @@ See next section, 'Request full data of a form'
 
 ### Request full data of a form
 Returns the full-depth object of the requested form (without submissions).
-- Endpoint: `/api/v2.2/form/{id}`
+- Endpoint: `/api/v2.4/form/{id}`
 - Url-Parameter:
   | Parameter | Type    | Description |
   |-----------|---------|-------------|
@@ -132,6 +139,9 @@ Returns the full-depth object of the requested form (without submissions).
     "showToAllUsers": false
   },
   "expires": 0,
+  "fileFormat": "csv",
+  "fileId": 157,
+  "filePath": "foo/bar",
   "isAnonymous": false,
   "submitMultiple": true,
   "showExpiration": false,
@@ -250,6 +260,37 @@ Transfer the ownership of a form to another user
 "data": 3
 ```
 
+### Link a form to a file
+- Endpoint: `/api/v2.4/form/link/{fileFormat}`
+- Url-Parameter:
+  | Parameter    | Type    | Description  |
+  |--------------|---------|--------------|
+  | _fileFormat_ | String  | csv|ods|xlsx |
+- Method: `POST`
+- Parameters:
+  | Parameter | Type    | Description                                |
+  |-----------|---------|--------------------------------------------|
+  | _hash_    | String  | Hash of the form to update                 |
+  | _path_    | String  | Path within User-Dir, to store the file to |
+- Response: The new question object.
+```
+"data": {
+  "fileFormat": "csv",
+  "fileId": 157,
+  "filePath": "foo/bar",
+  "fileName": "Form 1 (responses).csv"
+}
+```
+
+### Unlink file from form
+- Endpoint: `/api/v2.4/form/unlink`
+- Method: `POST`
+- Parameters:
+  | Parameter | Type    | Description                |
+  |-----------|---------|----------------------------|
+  | _hash_    | String  | Hash of the form to update |
+- Response: **Status-Code OK**
+
 ## Question Endpoints
 Contains only manipulative question-endpoints. To retrieve questions, request the full form data.
 
@@ -262,7 +303,7 @@ Contains only manipulative question-endpoints. To retrieve questions, request th
   | _formId_  | Integer |          | ID of the form, the new question will belong to |
   | _type_    | [QuestionType](DataStructure.md#question-types) |  | The question-type of the new question |
   | _text_    | String  | yes      | *Optional* The text of the new question. |
-- Response: The new question object. 
+- Response: The new question object.
 ```
 "data": {
   "id": 3,
@@ -434,7 +475,7 @@ Update a single or all properties of an option-object
   |------------------|----------|-------------|
   | _id_             | Integer  | ID of the share to update |
   | *keyValuePairs*¹ | Array    | Array of key-value pairs to update |
-  
+
   ¹Currently only the _permissions_ can be updated.
 - Method: `PATCH`
 - *Method: `POST` deprecated*
@@ -542,13 +583,14 @@ Get all Submissions to a Form
 
 ### Get Submissions as csv (Download)
 Returns all submissions to the form in form of a csv-file.
-- Endpoint: `/api/v2.2/submissions/export/{hash}`
+- Endpoint: `/api/v2.4/submissions/export/{hash}`
 - Url-Parameter:
-  | Parameter | Type    | Description |
-  |-----------|---------|-------------|
-  | _hash_    | String  | Hash of the form to get the submissions for |
+  | Parameter    | Type    | Description |
+  |--------------|---------|-------------|
+  | _hash_       | String  | Hash of the form to get the submissions for |
+  | _fileFormat_ | String  | csv|ods|xlsx |
 - Method: `GET`
-- Response: A Data Download Response containg the headers `Content-Disposition: attachment; filename="Form 1 (responses).csv"` and `Content-Type: text/csv;charset=UTF-8`. The actual data contains all submissions to the referred form, formatted as comma separated and escaped csv.
+- Response: A Data Download Response containing the headers `Content-Disposition: attachment; filename="Form 1 (responses).csv"` and `Content-Type: text/csv;charset=UTF-8`. The actual data contains all submissions to the referred form, formatted as comma separated and escaped csv.
 ```
 "User display name","Timestamp","Question 1","Question 2"
 "jonas","Friday, January 22, 2021 at 12:47:29 AM GMT+0:00","Option 2","Answer"
@@ -557,13 +599,14 @@ Returns all submissions to the form in form of a csv-file.
 
 ### Export Submissions to Cloud (Files-App)
 Creates a csv file and stores it to the cloud, resp. Files-App.
-- Endpoint: `/api/v2.2/submissions/export`
+- Endpoint: `/api/v2.4/submissions/export`
 - Method: `POST`
 - Parameters:
-  | Parameter | Type    | Description |
-  |-----------|---------|-------------|
-  | _hash_    | String  | Hash of the form to get the submissions for |
-  | _path_    | String  | Path within User-Dir, to store the file to |
+  | Parameter    | Type    | Description |
+  |--------------|---------|-------------|
+  | _hash_       | String  | Hash of the form to get the submissions for |
+  | _path_       | String  | Path within User-Dir, to store the file to |
+  | _fileFormat_ | String  | csv|ods|xlsx |
 - Response: Stores the file to the given path and returns the fileName.
 ```
 "data": "Form 2 (responses).csv"
@@ -592,7 +635,7 @@ Store Submission to Database
   | _formId_  | Integer | ID of the form to submit into |
   | _answers_ | Array   | Array of Answers |
   | _shareHash_ | String | optional, only neccessary for submissions to a public share link |
-  
+
   The Array of Answers has the following structure:
   - QuestionID as key
   - An **array** of values as value --> Even for short Text Answers, wrapped into Array.
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -48,12 +48,10 @@ use OCP\AppFramework\Db\IMapperException;
 use OCP\AppFramework\Http\DataDownloadResponse;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCS\OCSBadRequestException;
-use OCP\AppFramework\OCS\OCSException;
 use OCP\AppFramework\OCS\OCSForbiddenException;
 use OCP\AppFramework\OCS\OCSNotFoundException;
 use OCP\AppFramework\OCSController;
 use OCP\Files\NotFoundException;
-use OCP\Files\NotPermittedException;
 use OCP\IL10N;
 use OCP\IRequest;
 use OCP\IUser;
@@ -378,7 +376,7 @@ class ApiController extends OCSController {
 		}
 
 		$user = $this->userManager->get($uid);
-		if($user == null) {
+		if ($user == null) {
 			$this->logger->debug('Could not find new form owner');
 			throw new OCSBadRequestException('Could not find new form owner');
 		}
@@ -393,7 +391,7 @@ class ApiController extends OCSController {
 
 		// Update changed Columns in Db.
 		$this->formMapper->update($form);
-		
+
 		return new DataResponse($form->getOwnerId());
 	}
 
@@ -1106,6 +1104,18 @@ class ApiController extends OCSController {
 		//Create Activity
 		$this->formsService->notifyNewSubmission($form, $submission->getUserId());
 
+		if ($form->getFileId() !== null) {
+			try {
+				$filePath = $this->formsService->getFilePath($form);
+				$fileFormat = $form->getFileFormat();
+				$ownerId = $form->getOwnerId();
+
+				$this->submissionService->writeFileToCloud($form, $filePath, $fileFormat, $ownerId);
+			} catch (NotFoundException $e) {
+				$this->logger->notice('Form {formId} linked to a file that doesn\'t exist anymore', ['formId' => $formId]);
+			}
+		}
+
 		return new DataResponse();
 	}
 
@@ -1190,11 +1200,12 @@ class ApiController extends OCSController {
 	 * Export submissions of a specified form
 	 *
 	 * @param string $hash the form hash
+	 * @param string $fileFormat File format used for export
 	 * @return DataDownloadResponse
 	 * @throws OCSBadRequestException
 	 * @throws OCSForbiddenException
 	 */
-	public function exportSubmissions(string $hash): DataDownloadResponse {
+	public function exportSubmissions(string $hash, string $fileFormat = Constants::DEFAULT_FILE_FORMAT): DataDownloadResponse {
 		$this->logger->debug('Export submissions for form: {hash}', [
 			'hash' => $hash,
 		]);
@@ -1203,7 +1214,7 @@ class ApiController extends OCSController {
 			$form = $this->formMapper->findByHash($hash);
 		} catch (IMapperException $e) {
 			$this->logger->debug('Could not find form');
-			throw new OCSBadRequestException();
+			throw new OCSNotFoundException();
 		}
 
 		if (!$this->formsService->canSeeResults($form)) {
@@ -1211,8 +1222,10 @@ class ApiController extends OCSController {
 			throw new OCSForbiddenException();
 		}
 
-		$csv = $this->submissionService->getSubmissionsCsv($hash);
-		return new DataDownloadResponse($csv['data'], $csv['fileName'], 'text/csv');
+		$submissionsData = $this->submissionService->getSubmissionsData($form, $fileFormat);
+		$fileName = $this->formsService->getFileName($form, $fileFormat);
+
+		return new DataDownloadResponse($submissionsData, $fileName, Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat]);
 	}
 
 	/**
@@ -1223,21 +1236,17 @@ class ApiController extends OCSController {
 	 *
 	 * @param string $hash of the form
 	 * @param string $path The Cloud-Path to export to
+	 * @param string $fileFormat File format used for export
 	 * @return DataResponse
 	 * @throws OCSBadRequestException
 	 * @throws OCSForbiddenException
 	 */
-	public function exportSubmissionsToCloud(string $hash, string $path) {
-		$this->logger->debug('Export submissions for form: {hash} to Cloud at: /{path}', [
-			'hash' => $hash,
-			'path' => $path,
-		]);
-
+	public function exportSubmissionsToCloud(string $hash, string $path, string $fileFormat = Constants::DEFAULT_FILE_FORMAT) {
 		try {
 			$form = $this->formMapper->findByHash($hash);
 		} catch (IMapperException $e) {
 			$this->logger->debug('Could not find form');
-			throw new OCSBadRequestException();
+			throw new OCSNotFoundException();
 		}
 
 		if (!$this->formsService->canSeeResults($form)) {
@@ -1245,14 +1254,86 @@ class ApiController extends OCSController {
 			throw new OCSForbiddenException();
 		}
 
-		// Write file to cloud
+		$file = $this->submissionService->writeFileToCloud($form, $path, $fileFormat);
+
+		return new DataResponse($file->getName());
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @param string $hash of the form
+	 */
+	public function unlinkFile(string $hash): DataResponse {
 		try {
-			$fileName = $this->submissionService->writeCsvToCloud($hash, $path);
-		} catch (NotPermittedException $e) {
-			$this->logger->debug('Failed to export Submissions: Not allowed to write to file');
-			throw new OCSException('Not allowed to write to file.');
+			$form = $this->formMapper->findByHash($hash);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
+			throw new OCSNotFoundException();
 		}
 
-		return new DataResponse($fileName);
+		if ($form->getOwnerId() !== $this->currentUser->getUID()) {
+			$this->logger->debug('This form is not owned by the current user');
+			throw new OCSForbiddenException();
+		}
+
+		if (!$form->getFileId()) {
+			$this->logger->debug('Form not linked to file');
+			throw new OCSBadRequestException();
+		}
+
+		$form->setFileId(null);
+		$form->setFileFormat(null);
+
+		$this->formMapper->update($form);
+
+		return new DataResponse($hash);
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * Export Submissions to the Cloud and Link the FileId to the form
+	 *
+	 * @param string $hash of the form
+	 * @param string $path The Cloud-Path to export to
+	 * @param string $fileFormat File format used for export
+	 * @return DataResponse
+	 * @throws OCSBadRequestException
+	 * @throws OCSForbiddenException
+	 */
+	public function linkFile(string $hash, string $path, string $fileFormat): DataResponse {
+		$this->logger->debug('Linking form {hash} to file at /{path} in format {fileFormat}', [
+			'hash' => $hash,
+			'path' => $path,
+			'fileFormat' => $fileFormat,
+		]);
+
+		try {
+			$form = $this->formMapper->findByHash($hash);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
+			throw new OCSNotFoundException();
+		}
+		if ($form->getOwnerId() !== $this->currentUser->getUID()) {
+			$this->logger->debug('This form is not owned by the current user');
+			throw new OCSForbiddenException();
+		}
+
+		$file = $this->submissionService->writeFileToCloud($form, $path, $fileFormat);
+
+		$form->setFileId($file->getId());
+		$form->setFileFormat($fileFormat);
+
+		$this->formMapper->update($form);
+
+		$filePath = $this->formsService->getFilePath($form);
+
+		return new DataResponse([
+			'fileId' => $file->getId(),
+			'fileFormat' => $fileFormat,
+			'fileName' => $file->getName(),
+			'filePath' => $filePath,
+		]);
 	}
 }
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -27,33 +27,36 @@ namespace OCA\Forms\Service;
 use DateTime;
 use DateTimeZone;
 
-use League\Csv\EncloseField;
-use League\Csv\EscapeFormula;
-use League\Csv\Reader;
-use League\Csv\Writer;
 use OCA\Forms\Constants;
 use OCA\Forms\Db\Answer;
 
 use OCA\Forms\Db\AnswerMapper;
+use OCA\Forms\Db\Form;
 use OCA\Forms\Db\FormMapper;
 use OCA\Forms\Db\QuestionMapper;
 use OCA\Forms\Db\Submission;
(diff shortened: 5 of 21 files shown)


Chartman123:
@Koc thank you very much for your contribution! We'll have a look at it 🙂 

@nextcloud/designers Could you please do a quick review of the screenshots? I'm not really convinced of the UI in the menu. Looks a bit crowded...

@hamza221 fine for you to close your PR and keep working on this one?

susnux (commented):
Thank you very much! Nice work!

Some first thought about your code, besides there are quite a lot of linter issues to fix.
I think we do not need to implement this as a breaking change API wise but can make this compatible with the current API by just extending.

Moreover I do not know if it makes much sense to store the file path in the database. We probably should just work with the file id.

lib/Controller/ApiController.php
```diff
@@ -1125,12 +1145,91 @@ public function exportSubmissionsToCloud(string $hash, string $path) {
...
+			'hash' => $hash,
+			'path' => $path,
+			'fileFormat' => $fileFormat,
+		]);
+
+		try {
+			$form = $this->formMapper->findByHash($hash);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find form');
+			throw new OCSBadRequestException();
```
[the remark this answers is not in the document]
Koc: it's copy-pasted from other places, like `getPartialForm/getForm/cloneForm/updateForm/` etc. But I can use `OCSNotFoundException` at least in methods that I've touched

Koc:
Mates, any idea how to fix that?

> ```
> Error: lib/Service/SubmissionService.php:175:12: MissingDependency: OCP\Files\IRootFolder depends on class or interface oc\hooks\emitter that does not exist (see https://psalm.dev/157)
> ```

I haven't touched `\OCP\Files\IRootFolder $storage` dependency at all

lib/Controller/ApiController.php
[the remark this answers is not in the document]
Koc: fixed

lib/Migration/Version030200Date20231001200100.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Ferdinand Thiessen <rpm@fthiessen.de>
+ *
+ * @author Ferdinand Thiessen <rpm@fthiessen.de>
```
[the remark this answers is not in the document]
Koc: fixed :smile:

src/views/Results.vue
```diff
@@ -67,22 +67,113 @@
...
+
+					<template #icon>
+						<IconLink :size="20" />
+					</template>
+
+					<NcActionButton :close-after-click="true" @click="onLinkFile('csv')">
+						<template #icon>
+							<IconFileDelimited :size="20" />
+						</template>
+						{{ t('forms', 'CSV') }}
```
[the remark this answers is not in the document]
Koc: translations removed

src/views/Results.vue
```diff
@@ -67,22 +67,113 @@
...
 					</NcActionButton>
+
+					<NcActionCaption :title="t('forms', 'Save to Files')" />
+
+					<NcActionButtonGroup>
+						<NcActionButton :close-after-click="true" @click="onStoreToFiles('csv')">
+							<template #icon>
+								<IconFileDelimited :size="20" />
+							</template>
+							{{ t('forms', 'CSV') }}
```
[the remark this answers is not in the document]
Koc: translations removed

appinfo/routes.php
```diff
@@ -227,30 +227,57 @@
...
+			'url' => '/api/{apiVersion}/submissions/{hash}',
 			'verb' => 'GET',
 			'requirements' => [
 				'apiVersion' => 'v2(\.1)?'
 			]
 		],
 		[
 			'name' => 'api#exportSubmissionsToCloud',
-			'url' => '/api/{apiVersion}/submissions/export',
+			'url' => '/api/{apiVersion}/submissions/export/{fileFormat}',
```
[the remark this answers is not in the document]
Koc: ok, I've try to preserve BC for all already existent public methods in this controller. Please check

appinfo/routes.php
```diff
@@ -227,30 +227,57 @@
...
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.1)?'
+				'apiVersion' => 'v2(\.1)?',
+				'fileFormat' => 'csv|ods|xlsx'
+			]
+		],
+		[
+			'name' => 'api#exportSubmissions',
+			'url' => '/api/{apiVersion}/submissions/export/{hash}/{fileFormat}',
```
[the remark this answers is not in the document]
Koc: done

Koc:
@susnux 
> Moreover I do not know if it makes much sense to store the file path in the database. We probably should just work with the file id.

Yes, it's doable but little bit tricky because we can choose file or folder. If it is a blocker for merge - just say and I will rework. Because now it sounds like you not 100% sure :)

nimishavijay:
Super nice! :)
I have some questions:

- [ ] Is the spreadsheet automatically updated when new responses are added?
- [ ] What does "re-export spreadsheet" do? There are several actions on the screen that are closely related (create spreadsheet, save, download, re-export) so it could get a bit confusing to someone new

I do agree that the action menu looks a bit crowded, we can easily fix that :)
- [ ] Instead of disabling, we could hide the "Re-export" and "Unlink" actions if a spreadsheet has not been created yet
- [ ] For "Save" and "Download" we could use a second level action menu, similar to the send later action in mail

https://github.com/nextcloud/forms/assets/52440189/d4e967fa-080b-44cd-b650-282b93337ac3

What do you think?

susnux:
> > Error: lib/Service/SubmissionService.php:175:12: MissingDependency: OCP\Files\IRootFolder depends on class or interface oc\hooks\emitter that does not exist (see https://psalm.dev/157)

Can be silenced like this:
https://github.com/nextcloud/spreed/blob/55e5fa00ffcdafe3eb3a94bc03b25e6f2ee9982d/psalm.xml#L91

Koc:
@nimishavijay yes, we are automatically update spreadsheet once new response added. But sometimes we can get into trouble with a race conditions. Imagine that Employee edits file with submissions in a browser and overwrites changes that exported after new response added. In this case we should have mechanism to force re-export responses to file after Employee finishes his work.

Koc:
@nimishavijay I've added nested menu, you can find a video record for it in the PR description. Also disabling of non-active buttons has been replaced with hidding.

@susnux I've reworked PR, now we are storing `fileId, fileFormat` only, without `filePath`.

Also I've fixed all pipelines (I hope :crossed_fingers: ). Can we enable full pipeline for an every push commit for this PR?

Chartman123:
@Koc The workflows will run automatically once you have contributed to this repository. Until then we have to manually start it :)

Koc:
aah, sorry for that. 

How can I run integration tests locally? `composer test:integration` causes an errors like:

```
25) OCA\Forms\Tests\Integration\Api\ApiV2Test::testDeleteSingleSubmission with data set "deleteSingleSubmission" (array(array(array('user2', 'user2', 12345, array(array('This is another short answer.'), array('Option 2'))), array('user3', 'user3', 1234, array(array('')))), array(array('short', 'First Question?', true, '', 1, array(), 'Please answer this.', array()), array('multiple_unique', 'Second Question?', false, 'city', 2, array(array('Option 1'), array('Option 2'), array('')), '', array(true)))))
GuzzleHttp\Exception\ConnectException: cURL error 7: Failed to connect to localhost port 8080 after 0 ms: Connection refused (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://localhost:8080/ocs/v2.php/apps/forms/api/v2.1/submission/89
```

Chartman123:
@Koc you can have a look at https://github.com/nextcloud/forms/pull/1690#issuecomment-1775508838 and the two following comments. They might already help you. I currently don't run the tests locally, so I can't help you much. Perhaps @susnux can help you...

nimishavijay:
Ah okay! I understood. Wrapping my head around the different operations was a bit confusing. So I have some more suggestions:

- [ ] Rename "Save to Files" to "Save copy to Files"
- [ ] We could cut down on one step of selecting the filetype by having it in the file picker itself. Would that be possible? So when you click on "Create spreadsheet", the file picker opens, with 3 buttons (from right to left): Primary button `Create [file name].ods`, secondary button `Create CSV`, secondary button `Create XLSX`. (Reasoning: .ods is the default filetype for spreadsheets when you create a new one on Nextcloud. If there a better reasoning for showing another filetype as the primary button then I am totally ok with it :) The point is that there should be 1 primary button so the users doesn't have to make a big decision with what file type it is). So it would look something like this:
  
image

  - [ ] Change title of file picker modal to "Choose spreadsheet location"

Koc:
@nimishavijay it's little bit sad to hear that we need to rework file select dialog one more time :disappointed: . Last week you suggested to use nested menus to choose file format, I did all the changes and now we need start from the beginning.

BTW For some reason 3 buttons not works for me as expected (there are no text labels inside + we have warning in console). I'm using everything like in [docs](https://www.npmjs.com/package/@nextcloud/dialogs/v/4.2.1#use-the-filepickerbuilder)

```js
const picker = getFilePickerBuilder(t('forms', 'Choose spreadsheet location'))
	.setMultiSelect(false)
	.setModal(true)
	.setType(FilePickerType.Custom)
	.allowDirectories()
	.setMimeTypeFilter([
		'text/csv',
		'application/vnd.oasis.opendocument.spreadsheet',
		'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
	])
	.addButton({
		id: 'select-xlsx',
		label: t('forms', 'Create XLSX'), // docs suggest use `label` but warning in console suggests use `text` or `ariaLabel`
		text: t('forms', 'Create XLSX'),
		ariaLabel: t('forms', 'Create XLSX'),
		callback: (nodes) => {
			console.log(nodes)
			// this.applyFile(nodes[0]?.path)
		},
	})
	.addButton({
		id: 'select-csv',
		label: t('forms', 'Create csv'),
		text: t('forms', 'Create csv'),
		ariaLabel: t('forms', 'Create csv'),
		callback: (nodes) => {
			console.log(nodes)
			// this.applyFile(nodes[0]?.path)
		},
	})
	.addButton({
		id: 'select-ods',
		label: t('forms', 'Create ODS'),
		text: t('forms', 'Create ODS'),
		ariaLabel: t('forms', 'Create ODS'),
		callback: (nodes) => {
			console.log(nodes)
			// this.applyFile(nodes[0]?.path)
		},
		type: 'primary',
	})
	.build()
```

image

nimishavijay:
> I did all the changes and now we need start from the beginning.

Ah I didn't mean to change the nested dialog, only the create spreadsheet dialog :) The nested dialog does look great and works for our use case. `Save as > select filetype > select location` is a common pattern. But we could simplify the flow while creating a spreadsheet as the filetype doesn't matter many times, people just want to see a spreadsheet :) Does that make sense?

Koc:
@nimishavijay I did renaming "Save to Files" to "Save copy to Files".

Yeah, we can optimize create spreadsheet dialog, but as I've already mentioned - for some reason this not works for me and buttons has no labels. Maybe there is a bug in `nextcloud/dialogs` library or documentation is not up to date. I don't know. If you can provide me a code snippet that works - it would be helpful.

Koc:
test failure looks [unrelated](https://github.com/nextcloud/forms/actions/runs/6810633603/job/18519475679?pr=1758) :thinking:

Chartman123:
> test failure looks [unrelated](https://github.com/nextcloud/forms/actions/runs/6810633603/job/18519475679?pr=1758) 🤔

Yes, the unit tests sometimes fail. You normally can see in the logs if it's related to your code or some other technical problem while setting up the container or uploading the codecov results :)

Koc:
wow, we have completely green pipeline for the very first time! :tada:

Chartman123 (commented):
I had a look at the code, but not everything in-depth. Added some comments :) I will run the code in the next few days and see how it works.

lib/Service/SubmissionService.php
```diff
@@ -27,31 +27,40 @@
...
+use PhpOffice\PhpSpreadsheet\Writer\Csv;
+
 use Psr\Log\LoggerInterface;
 
 class SubmissionService {
+	public const SUPPORTED_EXPORT_FORMATS = [
+		'csv' => 'text/csv',
+		'ods' => 'application/vnd.oasis.opendocument.spreadsheet',
+		'xlsx' => 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+	];
```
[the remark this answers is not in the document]
Koc: done

appinfo/routes.php
```diff
@@ -227,19 +227,36 @@
...
-		// Submissions
+		// Submissions linking with file in cloud
 		[
-			'name' => 'api#getSubmissions',
-			'url' => '/api/{apiVersion}/submissions/{hash}',
+			'name' => 'api#linkFile',
+			'url' => '/api/{apiVersion}/submissions/link/{fileFormat}',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2(\.1)?',
```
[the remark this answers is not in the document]
Koc: done, please check

appinfo/routes.php
```diff
@@ -227,19 +227,36 @@
...
+				'fileFormat' => 'csv|ods|xlsx'
+			]
+		],
+		[
+			'name' => 'api#getLinkedFile',
+			'url' => '/api/{apiVersion}/submissions/linkedFile/{hash}',
```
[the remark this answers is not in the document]
Koc: done, please check

appinfo/routes.php
```diff
@@ -227,19 +227,36 @@
...
 			]
 		],
 		[
-			'name' => 'api#exportSubmissions',
-			'url' => '/api/{apiVersion}/submissions/export/{hash}',
+			'name' => 'api#unlinkFile',
+			'url' => '/api/{apiVersion}/submissions/unlink',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2(\.1)?'
```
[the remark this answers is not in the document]
Koc: done, please check

lib/Controller/ApiController.php
```diff
@@ -61,6 +60,8 @@
 use Psr\Log\LoggerInterface;
 
 class ApiController extends OCSController {
+	private const DEFAULT_FILE_FORMAT = 'csv';
+
```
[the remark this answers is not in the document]
Koc: done, please check

lib/Service/SubmissionService.php
```diff
@@ -150,60 +163,99 @@ public function getSubmissions(int $formId): array {
...
-		// Get Data
-		$csvData = $this->getSubmissionsCsv($hash);
+		$this->logger->debug('Export submissions for form: {hash} to Cloud at: /{path} in format {fileFormat}', [
+			'hash' => $hash,
+			'path' => $path,
+			'fileFormat' => $fileFormat,
+		]);
 
-		// If chosen path is a file, get folder, if file is csv, use filename.
+		$form = $this->formMapper->findByHash($hash);
```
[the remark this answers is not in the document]
Koc: done, please check

lib/Controller/ApiController.php
```diff
@@ -1101,36 +1114,102 @@ public function exportSubmissions(string $hash): DataDownloadResponse {
...
+	/**
+	 * @NoAdminRequired
+	 *
+	 * Return information about linked file
+	 *
+	 * @param string $hash of the form
+	 * @return DataResponse
+	 */
+	public function getLinkedFile(string $hash): DataResponse {
```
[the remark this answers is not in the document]
Koc: > Form might not have a linked file

And that's fine. We will respond with http 200 and `{fileId: null, path: null, fileFormat: null}`. Otherwise we will have a lot of noise of unresolved promises caused by http 400

image

> User might not be allowed to access file

Please suggest how can I check this (I'm not familiar with Nextclud API a lot yet) :pray:

Koc:
@Chartman123 I did all almost the changes, the only few are left:

- [comment 1](https://github.com/nextcloud/forms/pull/1758#issuecomment-1789368363)
  - as I've already mentioned - it seems like we have a bug with current version of the dialogs apo. Maybe problem is fixed in 5.x but it is major version, not sure that we need to bump it right now
  - even more suggested dialog will add more confusion: it is possible to select `a.csv` file and press `export as xlsx` button.
- [comment 2](https://github.com/nextcloud/forms/pull/1758#discussion_r1389236857)
  - how can I check is file readable or not? And what to do if it not readable?

susnux:
> [comment 1](https://github.com/nextcloud/forms/pull/1758#issuecomment-1789368363)
> 
>     * as I've already mentioned - it seems like we have a bug with current version of the dialogs apo. Maybe problem is fixed in 5.x but it is major version, not sure that we need to bump it right now
> 
>     * even more suggested dialog will add more confusion: it is possible to select `a.csv` file and press `export as xlsx` button.

What do you mean with bug?

But I would recommend to go with version 5 of the dialogs package, that allows to use a button factory with which you can show buttons based on the current directory or selected files. Meaning it would be possible to select a CSV file and hide the "select XLSX" button.

---

For the other comment:

> $hash might not exist

In this case return 404

> Form might not have a linked file

This should either be a 204 (no content) but I think this makes not much sense as this is not a successful request.
Could also a 404 (there is no such file).

I would go with 404, @Chartman123 ?

> User might not be allowed to access file

This one is harder, generally you could just return the file and the user will fail to query it as it has no permission (I prefer this).
But you could also return 403, which would require to check for permissions on that file. But I do not known when this can happen.

Currently I see 3 situations:
1: User selects a file that they have not access to -> selecting the file should fail
2: User selects a file in a share or groupfolder, that gets unshared afterwards -> what to do?
3: User selects a file but (as we do in future) a collaborating user does not have permission to that file.

In all this cases we should return that there is a connected file (case 3 is currently not important).

src/views/Results.vue
```diff
@@ -216,10 +401,22 @@ export default {
...
 
 	methods: {
+		async unlinkFile() {
+			await axios.post(generateOcsUrl('apps/forms/api/v2/submissions/unlink'), {
+				hash: this.form.hash,
+			})
+			await this.loadLinkedFile()
+		},
+		async loadLinkedFile() {
+			const response = await axios.get(generateOcsUrl(`apps/forms/api/v2/submissions/linkedFile/${this.form.hash}`))
```
susnux, line 416: You can check the errors for loading the linked file here, e.g. error 404:

```suggestion
			try {
				const response = await axios.get(generateOcsUrl(`apps/forms/api/v2/submissions/linkedFile/${this.form.hash}`))
			} catch (e) {
				// handle the error, e.g.:
				// if (e?.response?.status === 404) { // not found...
			}
```

Chartman123:
> I would go with 404, @Chartman123 ?

404 is also what's on my mind...

Koc:
@susnux I've described bug few [messages above](https://github.com/nextcloud/forms/pull/1758#issuecomment-1793700200), see attached screenshot with console errors about missing props.

> BTW For some reason 3 buttons not works for me as expected (there are no text labels inside + we have warning in console). I'm using everything like in [docs](https://www.npmjs.com/package/@nextcloud/dialogs/v/4.2.1#use-the-filepickerbuilder)

I've try to update dialogs library but it blocked by `@nextcloud/vue` version lock. Dialogs v5 expects Nextcloud v28 but current main branch coupled with v27.

```
devcontainer@28c61d60834b:/var/www/html/apps/forms$ npm install
npm notice 
npm notice New major version of npm available! 8.19.4 -> 10.2.4
npm notice Changelog: https://github.com/npm/cli/releases/tag/v10.2.4
npm notice Run npm install -g npm@10.2.4 to update!
npm notice 
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: forms@3.3.1
npm ERR! Found: @nextcloud/vue@7.12.7
npm ERR! node_modules/@nextcloud/vue
npm ERR!   @nextcloud/vue@"^7.12.7" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer @nextcloud/vue@"^8.2.0" from @nextcloud/dialogs@5.0.2
npm ERR! node_modules/@nextcloud/dialogs
npm ERR!   @nextcloud/dialogs@"^5.0.2" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
npm ERR! See /home/devcontainer/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/devcontainer/.npm/_logs/2023-11-21T23_43_53_711Z-debug-0.log
```

Chartman123:
@Koc could you please rebase your code onto the latest main? There are some conflicts after our latest code changes :)

jancborchardt:
@Koc very nice feature! :) Since one of your latest commits has "wip" in the text, I'm wondering if this is ready for review – or you'll say when? :)

lib/Controller/ApiController.php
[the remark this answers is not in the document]
Koc: done, please check one more time

src/views/Results.vue
Koc: done

Koc:
:spiral_notepad: Changes:
- rebased, fixed conflicts
- modal dialog reworked, applied suggestion with 3 buttons (it is possible with dialogs library V5)
- fixed other code review comments
- attached fresh video with demo functionality

ready to next round of code review @jancborchardt @Chartman123 @susnux @nimishavijay

.github/workflows/lint-php-cs.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install dependencies
-        run: composer i
+        run: composer install --ignore-platform-reqs=ext-gd
```
Koc, line 37: GD extension is needed for `phpoffice/phpspreadsheet`. But at the same time we are not using anything related to images, so we can ignore this dependency IMHO

jancborchardt (commented):
Looks really very nice @Koc! :)
Very small detail, a separator in the action menu above "Download" would be good to separate the "Linked spreadsheet" options from Download and Delete would be nice. Otherwise top notch!

src/views/Results.vue
```diff
@@ -89,29 +102,96 @@
...
+
+							<NcActionButton :close-after-click="true" @click="onDownloadFile('xlsx')">
+								<template #icon>
+									<IconFileExcel :size="20" />
+								</template>
+								XSLX
+							</NcActionButton>
+						</template>
+					</NcActions>
+
```
Chartman123, line 194:
```suggestion
```

src/views/Results.vue
```diff
@@ -89,29 +102,96 @@
...
+								{{ t('forms', 'Delete all responses') }}
+							</NcActionButton>
+						</template>
+
+						<template v-if="isDownloadActionOpened">
+							<NcActionButton :close-after-click="false" @click="hideSubActions">
+								<template #icon>
+									<IconChevronLeft :title="t('mail', 'Download')"
+													 :size="20" />
+									{{ t('forms', 'Download') }}
```
Chartman123, line 168:
```suggestion
									<IconChevronLeft :title="t('mail', 'Download')"
										:size="20" />
```

src/views/Results.vue
```diff
@@ -137,9 +217,9 @@
 </template>
 
 <script>
-import { generateOcsUrl } from '@nextcloud/router'
+import { generateOcsUrl, generateUrl } from '@nextcloud/router'
 import { getRequestToken } from '@nextcloud/auth'
-import { getFilePickerBuilder, showError, showSuccess } from '@nextcloud/dialogs'
+import {getFilePickerBuilder, showError, showSuccess, FilePickerType} from '@nextcloud/dialogs'
 import { emit } from '@nextcloud/event-bus'
```
Chartman123, line 223:
```suggestion
import { getFilePickerBuilder, showError, showSuccess } from '@nextcloud/dialogs'
```

src/views/Results.vue
```diff
@@ -155,6 +235,16 @@ import IconDownload from 'vue-material-design-icons/Download.vue'
...
+import IconFileDelimited from 'vue-material-design-icons/FileDelimited.vue'
+import IconFileExcel from 'vue-material-design-icons/FileExcel.vue'
+import IconLink from 'vue-material-design-icons/Link.vue'
+import IconLinkVariantOff from 'vue-material-design-icons/LinkVariantOff.vue'
+import IconChevronLeft from 'vue-material-design-icons/ChevronLeft.vue'
+import IconRefresh from 'vue-material-design-icons/Refresh.vue'
+import IconTable from 'vue-material-design-icons/Table.vue'
+import IconFileExcelSvg from '@mdi/svg/svg/file-excel.svg?raw'
+import IconFileDelimitedSvg from '@mdi/svg/svg/file-delimited.svg?raw'
+import IconTableSvg from '@mdi/svg/svg/table.svg?raw'
```
Chartman123, line 247: Please correct the sorting of the imports

src/views/Results.vue
```diff
@@ -166,10 +256,56 @@ import SetWindowTitle from '../utils/SetWindowTitle.js'
 import OcsResponse2Data from '../utils/OcsResponse2Data.js'
 import PermissionTypes from '../mixins/PermissionTypes.js'
 
-const picker = getFilePickerBuilder(t('forms', 'Save CSV to Files'))
+const SUPPORTED_FILE_FORMATS = {ods : IconTableSvg, csv: IconFileDelimitedSvg, xlsx: IconFileExcelSvg}
+
```
Chartman123, line 260:
```suggestion
const SUPPORTED_FILE_FORMATS = { ods: IconTableSvg, csv: IconFileDelimitedSvg, xlsx: IconFileExcelSvg }
```

src/views/Results.vue
```diff
@@ -166,10 +256,56 @@ import SetWindowTitle from '../utils/SetWindowTitle.js'
...
+					icon: SUPPORTED_FILE_FORMATS[extension],
+					callback() {
+						fileFormat = extension
+					},
+					type: 'primary',
+				}]
+			}
+
+			return [];
+		}
```
Chartman123, line 280:
```suggestion
	.setButtonFactory((selectedNodes, currentPath, currentView) => {
		if (selectedNodes.length === 1) {
			const extension = selectedNodes[0].extension.slice(1).toLowerCase()
			if (SUPPORTED_FILE_FORMATS[extension]) {
				return [{
					label: t('forms', 'Select {file}', { file: selectedNodes[0].basename }),
					icon: SUPPORTED_FILE_FORMATS[extension],
					callback() {
						fileFormat = extension
					},
					type: 'primary',
				}]
			}

			return []
```

src/views/Results.vue
```diff
@@ -181,6 +317,13 @@ export default {
 		IconFolder,
 		IconPoll,
 		IconShareVariant,
+		IconChevronLeft,
+		IconFileDelimited,
+		IconFileExcel,
+		IconLink,
+		IconRefresh,
+		IconTable,
+		IconLinkVariantOff,
```
Chartman123, line 326: Please adjust sorting

src/views/Results.vue
```diff
@@ -220,10 +381,31 @@ export default {
...
 	},
 
 	methods: {
+		async unlinkFile() {
+			await axios.post(generateOcsUrl('apps/forms/api/v2/submissions/unlink'), {
+				hash: this.form.hash,
+			})
+
+			this.linkedFile = {fileId: null, path: null, fileFormat: null}
+		},
```
Chartman123, line 395:
```suggestion
			this.linkedFile = { fileId: null, path: null, fileFormat: null }
```

src/views/Results.vue
```diff
@@ -261,6 +466,7 @@ export default {
 						const response = await axios.post(generateOcsUrl('apps/forms/api/v2.2/submissions/export'), {
 							hash: this.form.hash,
 							path,
+							fileFormat
 						})
```
Chartman123, line 470:
```suggestion
							fileFormat,
```

Chartman123 (commented):
Hi @Koc nice to see you back at work :) Hope you had some nice holidays!

I added a few comments and suggestions :)

@susnux: Could you please also have a look at my comments/suggestions regarding `submissions` vs `form` in the API routes and the methods in submissionService vs formsService

Koc:
@jancborchardt separator added
image

other changes are in progress

src/views/Results.vue
```diff
@@ -47,12 +47,25 @@
...
+						</template>
+						{{ t('forms', 'Share form') }}
+					</NcButton>
+
+					<NcActionButton v-if="!linkedFile.fileId"
+									type="tertiary-no-background"
+									:menu-name="t('forms', 'Create spreadsheet')"
+									:aria-label="t('forms', 'Create spreadsheet')"
+									@click="onLinkFile"
+					>
```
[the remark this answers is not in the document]
Koc: Any idea why this wasn't fixed automatically by `npm run lint:fix`? I run it multiple times before commit :confused:

src/views/Results.vue
Koc: why we don't have auto-imports like in [Nuxt](https://nuxt.com/docs/guide/concepts/auto-imports)?

Koc:
@Chartman123 changes done. I'm going to re-test everything after all this changes later today. Please check, did I miss anything?

appinfo/routes.php
```diff
@@ -339,5 +339,31 @@
 				'apiVersion' => 'v2(\.[1-3])?'
 			]
 		],
+		// Submissions linking with file in cloud
+		[
+			'name' => 'api#linkFile',
+			'url' => '/api/{apiVersion}/link/link/{fileFormat}',
```
Chartman123, line 345:
```suggestion
			'url' => '/api/{apiVersion}/form/link/{fileFormat}',
```

A little typo slipped in ;)

jancborchardt (commented):
Nice, design-wise this is good to go unless @nimishavijay has additional feedback :)

Chartman123:
@Koc I fixed the errors in FormsServiceTest.php, however there are still some test failures in SubmissionServiceTest.php where the expectation and the actual result don't match.

Koc:
@Chartman123 should be fixed now
image

Koc:
@Chartman123 @jancborchardt @susnux can you help me with this test error in integration tests?

> `1) OCA\Forms\Tests\Integration\Api\ApiV2Test::testGetSharedForms with data set "getTestforms" (array(array('abcdefghij', 'Title of a second Form', 0, 123456789, array('submit'), true)))
OCP\Files\NotPermittedException: Could not create folder "/test/files/foo"`

How can I set root storage path to something like `/tmp` instead of ``test/files/`?

Chartman123:
@Koc what I don't understand: why does the test try to create a folder while only getting the list of forms?

Koc:
@Chartman123 this is needed for `testLinkedFile` - we are asserting endpoint response. So, we should create dummy file because endpoint resolves file path by `fileId`

Chartman123:
@Koc test/files/ should be the folder inside the users storage directory... But I don't know very much about that. I just had a quick look how the Mail app handles attachment testing: 

https://github.com/nextcloud/mail/blob/main/tests/Unit/Service/Attachment/AttachmentStorageTest.php

Perhaps this will help you :)

Koc:
@Chartman123 no, you pointed out me to the unit test, where everything is mocked. But I stuck with functional test where we are using real dependencies

Chartman123:
@Koc I'm no expert in PHP unit / integration tests... What I found in our old versions of the workflow definitions was this: `./occ user:add --password-from-env --display-name="Test Displayname" test` in the Setup Nextcloud step. This is no longer in the current workflows. Perhaps thats why the permissions for the folders aren't set correctly.

@susnux @jotoeri Do you know more about that?

Chartman123:
@Koc i think the conflicts are mainly about the file picker. You can take most parts from @susnux. He also fixed a problem when the dialog is closed without selecting a file/folder

Chartman123:
@Koc I resolved the merge conflicts and added the fixes from @susnux

Please have a closer look if everything still works the way you wanted. A quick test here didn't show any errors in the frontend

Chartman123:
@Koc I found two bugs:

- When you load the results view without a linked file you get an error in the console. I think we shouldn't show this error in the console, as it is an expected behaviour when there's no linked file.
- When you link a file and then delete the linked file e.g. in the files app, you also get an error the next time you access the results view

Regarding the second error, I'm not really sure how to handle this: keep the file id in the database as it might be a temporary problem or remove the link?

Koc:
@Chartman123 thank you so much for conflicts resolving :pray: 

I have only this error in console which can't be hidden. We catching exception inside `loadLinkedFile`. Do you have other error? 
image

Chartman123:
@Koc You're welcome 😇 

> Do you have other error?

No, just the XHR error that you posted... What about not having a special linked file endpoint but just using the Form props?

Rergarding the missing/deleted files: There are lots of problems that you could think of, but I think that we will be able to handle all of them. In the frontend we could perhaps just add another dialog asking the user "linked file is not available, would you like to link a new file?" or something like that.

@susnux @nimishavijay @jancborchardt What do you think UX wise? Is there someone better to ask UX questions?

appinfo/routes.php
```diff
@@ -339,5 +339,31 @@
 				'apiVersion' => 'v2(\.[1-3])?'
 			]
 		],
+		// Submissions linking with file in cloud
+		[
+			'name' => 'api#linkFile',
+			'url' => '/api/{apiVersion}/form/link/{fileFormat}',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2(\.3)?',
```
Chartman123, line 348:
```suggestion
				'apiVersion' => 'v2.3',
```

The endpoints are only available in v2.3, not in v2. Same for the other two... 🙂

jancborchardt:
> In the frontend we could perhaps just add another dialog asking the user "linked file is not available, would you like to link a new file?" or something like that.

@Chartman123 yup, that sounds like a good solution!

appinfo/routes.php
Koc: it not works for me with your suggested change

image

Chartman123: You have to adjust the call in results.vue too ;)

docs/API.md
```diff
@@ -28,6 +28,14 @@ This file contains the API-Documentation. For more information on the returned D
 - Completely new way of handling access & shares.
 
 ### Other API changes
+#fixme: actualize after removal getLinkedFile endpoint
+- In API version 2.3 the following endpoints were introduced:
+  - `POST /api/2.3/form/link/{fileFormat}` to link form to a file
+  - `GET /api/2.3/form/linkedFile/{hash}` to retrieve linked file
+  - `POST /api/2.3/form/unlink` to unlink form from a file
```
Chartman123, line 35:
```suggestion
  - `POST /api/2.3/form/link/{fileFormat}` to link form to a file
  - `POST /api/2.3/form/unlink` to unlink form from a file
```

You can remove it here, too :) And of course further down as well

Koc: yeah, I left fixme for that. Actualized now

docs/API.md
```diff
@@ -544,11 +585,12 @@ Get all Submissions to a Form
...
-  | Parameter | Type    | Description |
-  |-----------|---------|-------------|
-  | _hash_    | String  | Hash of the form to get the submissions for |
+  | Parameter    | Type    | Description |
+  |--------------|---------|-------------|
+  | _hash_       | String  | Hash of the form to get the submissions for |
+  | _fileFormat_ | String  | csv|ods|xlsx |
 - Method: `GET`
-- Response: A Data Download Response containg the headers `Content-Disposition: attachment; filename="Form 1 (responses).csv"` and `Content-Type: text/csv;charset=UTF-8`. The actual data contains all submissions to the referred form, formatted as comma separated and escaped csv.
+- Response: A Data Download Response contains the headers `Content-Disposition: attachment; filename="Form 1 (responses).csv"` and `Content-Type: text/csv;charset=UTF-8`. The actual data contains all submissions to the referred form, formatted as comma separated and escaped csv.
```
Chartman123, line 593:
```suggestion
- Response: A Data Download Response containing the headers `Content-Disposition: attachment; filename="Form 1 (responses).csv"` and `Content-Type: text/csv;charset=UTF-8`. The actual data contains all submissions to the referred form, formatted as comma separated and escaped csv.
```

Chartman123:
@Koc I saw that the tests now all pass 🎉 So is it ready for review now?

Koc:
@Chartman123 yes!

Chartman123:
Great :) Could you please squash the commits into one? I try to have a look at it this week

Koc:
@Chartman123 commits squashed, coverage increased, also I've added fresh demo video which shows handling of the missing files

Chartman123 (approved):
Some typos to fix, but otherwise looks good from my side :) Let's wait for @susnux to also add a review :)

matthijskooijman:
Hey all. I've been following this PR for a while, since this is exactly what I've been missing for a particular workflow. Thanks for all the invested time and effort so far!

To provide some feedback, I've installed the version from this PR to a test nextcloud instance and played around with it a bit.

**Usecase**
The usecase I want to implement is a logbook for work on the electrical installation in a building. We currently have a spreadsheet in Nextcloud with a line for each change made, with some details about what, where, who, when, etc. To lower the bar for people to enter something in the logbook (and importantly, to allow unauthenticated users to *add* but not *edit* entries), I want to use a public form that adds entries to this logbook (and then put up QR-codes in the electrical cabinets so people can find the form).

I've collected some feedback below. I can imagine that some or most of these things might be out of scope for this PR and should be considered for followup improvements, so feel free to label them as such.

So, I've been testing the app as it is now, and overall it seems to support my usecase nearly completely. Only the first issue is a blocker for my usecase, the others would be nice to have, but not essential.

**Overwriting data**
I believe that now every time a new answer is added, the field names and *all* previous answers are also written to the spreadsheet (again). This means that I cannot:
 - Manually add a row in the spreadsheet - it will be overwritten by the next form response
 - Edit any form data in the spreadsheet - it will be reverted on the next form response

Both of these are problematic for my usecase - the reason I want to export to a spreadsheet is so I have the logbook in an easy to manage and edit form, and also so people that are more familiar with the tech can just add their entries directly in the spreadsheet. I suspect a fix for this would be (conceptually) simple: Just add only the new entry on the first empty row in the spreadsheet. Technically, this might be more complicated (might be prone to race conditions as well)?

**Inserting columns**
When I manually add extra data to the right of the form data, that is properly preserved as expected. However, if I insert a column in between the form data, then that data will be overwritten. This is expected now I understand how form results are written, but as a user I was a bit surprised. I guess my mental model was that the column headers would be used to decide where to put new values, so I could free reorder and insert columns). I guess this could be implemented, but it might not be easy, and breaks easily when the form is modified (this could maybe be solved by storing a history of titles for each question, but it is likely not worth effort, especially not at this stage).

**Adding data above the results**
I considered adding some instructions above the form results, to tell users how to work with the logbook (and what they should avoid to prevent things being overwritten). However, those instructions are now also overwritten with the field header line. It would be nice if this would not happen, though I'm not entirely sure how. You could write the field headers only to an empty file and afterwards only write new responses, but then the field headers get out of sync when the form is modified. This also complicates implementing reorderable columns as I suggested above. One way to implement this would be to add a "skip N rows" configuration that configures at what row the field headers should be written to?

**Multiple worksheets**
Instead of adding instructions above the form, I considered adding a second "README" worksheet to the spreadsheet with some instructions.

However, when the spreadsheet has multiple worksheets, results are written to the active worksheet (i.e. the one that was selected when the spreadsheet was saved), even when it was previously saved to another worksheet.

This was slightly unexpected, and makes adding a worksheet with instructions infeasible (unless users always make sure to switch back to the logbook worksheet, but I'm pretty sure that won't happen every time).

Maybe some worksheet ID could be saved when creating the spreadsheet and then used to save to the same worksheet every time (the worksheet title would not work so well, since users can change that in the spreadsheet).

**Fonts**
I've noticed that the font used for form data (Calibri) is different from the default font my libreoffice uses when adding data to empty cells (Liberation Sans). I'm not at all sure how this works and if this is a matter of system-dependent defaults or not (I have not tested with collabora, since I do not have it on my test Nextcloud instance), but it is a bit unexpected:

image

**Unclear arrow for technical name**
When clicking the dots menu for a question, there is a "Technical name" input field, with an "right arrow" button. However, it is unclear to me what that button is supposed to do. I thought maybe a "save" or "apply" button, but clicking it does not give any feedback, and even when not clicking it changes seems saved.


image


Writing this, I realized that this is probably not related to this PR at all, though, so maybe I should put this in a separate issue?

Chartman123:
Hi @matthijskooijman 

thanks for you elaborate thoughts about this PR. I think your "blocker" is indeed the easiest to accomplish by only exporting the latest submission to a linked file and only writing all submissions on clicking the "re-export spreadsheet". This would then of course need some more adjustments in the API and the SubmissionService.

From an UX point of view, this is probably what most users would expect it to be like. @Koc could you please have a look if this were a big deal to change? Please don't just do it, only give an estimation of the needed effort. I'd prefer to have this PR merged for releasing a 4.1 of Forms and we could then add some additions as a follow-up for the next version.

You're other proposals will definitely not make it into this PR anymore. So please open separate feature request issues for them. We can then discuss them there.

About the technical name: the little error should be indeed for saving the values and this function is available since a few versions already. So as you said not related. Please also add a new issue for that.

Chartman123:
My bad... I forgot to escape the `'` in one suggestion in ApiController.php

lib/Controller/ApiController.php
```diff
@@ -1106,6 +1104,18 @@ public function insertSubmission(int $formId, array $answers, string $shareHash
...
 
+		if ($form->getFileId() !== null) {
+			try {
+				$filePath = $this->formsService->getFilePath($form);
+				$fileFormat = $form->getFileFormat();
+				$ownerId = $form->getOwnerId();
+
+				$this->submissionService->writeFileToCloud($form, $filePath, $fileFormat, $ownerId);
+			} catch (NotFoundException $e) {
+				$this->logger->notice('Form {formId} linked to a file that doesn't exist anymore', ['formId' => $formId]);
```
Chartman123, line 1115:
```suggestion
				$this->logger->notice('Form {formId} linked to a file that doesn\'t exist anymore', ['formId' => $formId]);
```

lib/Controller/ApiController.php
```diff
@@ -1223,36 +1236,104 @@ public function exportSubmissions(string $hash): DataDownloadResponse {
...
+	 *
+	 * @param string $hash of the form
+	 * @param string $path The Cloud-Path to export to
+	 * @param string $fileFormat File format used for export
+	 * @return DataResponse
+	 * @throws OCSBadRequestException
+	 * @throws OCSForbiddenException
+	 */
+	public function linkFile(string $hash, string $path, string $fileFormat): DataResponse {
+		$this->logger->debug('Linking form {hash} to file at /{path} in format {fileFormat}', [
```
susnux, line 1306: see above the path should be absolute so you can get ride of the manual / before

lib/Migration/Version030200Date20231001200100.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Ferdinand Thiessen <rpm@fthiessen.de>
```
susnux, line 6: I am pretty sure I did not create that migration :wink:

lib/Migration/Version030200Date20231001200100.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace OCA\Forms\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version030200Date20231001200100 extends SimpleMigrationStep {
```
susnux, line 35: That version does not match! Needs to be adjusted (we are at version 4 atm).
Just generate a new one and replace the version here with the generated one `occ migrations:generate files 040010` or if you like just take something like this:

`Version040010Date20240122133700` :wink: 

(at least the version should be >= current version so it will execute)

lib/Service/SubmissionService.php
```diff
@@ -163,60 +172,97 @@ public function isUniqueSubmission(Submission $newSubmission): bool {
...
-		/** @var \OCP\Files\Folder|\OCP\Files\File $node */
-		$node = $this->storage->getUserFolder($this->currentUser->getUID())->get($path);
+	public function writeFileToCloud(Form $form, string $path, string $fileFormat, ?string $ownerId = null): File {
+		if (empty(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
+			throw new \InvalidArgumentException('Invalid file format');
+		}
 
-		// Get Data
-		$csvData = $this->getSubmissionsCsv($hash);
+		$this->logger->debug('Export submissions for form: {hash} to Cloud at: /{path} in format {fileFormat}', [
```
susnux, line 188: same as above probably should be absolute path so the extra / can be removed

lib/Service/SubmissionService.php
```diff
@@ -87,6 +90,9 @@ class SubmissionService {
 	/** @var IMailer */
 	private $mailer;
 
+	/** @var ITempManager */
+	private $tempManager;
+
```
susnux, line 95:
```suggestion
```

see next one below

lib/Service/SubmissionService.php
```diff
@@ -97,7 +103,9 @@ public function __construct(FormMapper $formMapper,
 		LoggerInterface $logger,
 		IUserManager $userManager,
 		IUserSession $userSession,
-		IMailer $mailer) {
+		IMailer $mailer,
+		ITempManager $tempManager,
+		private FormsService $formsService) {
```
susnux, line 108: Why use promotion on forms serve but not tempmanager?
Also add blank line to keep future diffs small.

```suggestion
		private ITempManager $tempManager,
		private FormsService $formsService,
	) {
```

lib/Service/SubmissionService.php
```diff
@@ -163,60 +172,97 @@ public function isUniqueSubmission(Submission $newSubmission): bool {
...
+			'path' => $path,
+			'fileFormat' => $fileFormat,
+		]);
 
-		// If chosen path is a file, get folder, if file is csv, use filename.
+		$fileName = $this->formsService->getFileName($form, $fileFormat);
+
+		/** @var \OCP\Files\Folder|File $node */
+		if ($ownerId) {
+			$node = $this->storage->getUserFolder($ownerId)->get($path);
```
susnux, line 198: As said above, I do not think relative paths make much sense, I would either use an absolute one and remove the leading / here.
Or as initally said just use the file id, you can simply get the file then with `$this->storage->getById(FILEID)`  (this returns an array if I am right, just take the first node then).

susnux (changes requested):
This works and I think we are close to merge!

I just have some nitpicks and some issues like wrong copyright, wrong version in migration and one more specific issue:
I still would like to see the API using fileids if possible. Or do you have some reason for using paths there?

appinfo/routes.php
```diff
@@ -339,5 +339,23 @@
 				'apiVersion' => 'v2(\.[1-3])?'
 			]
 		],
+		// Submissions linking with file in cloud
+		[
+			'name' => 'api#linkFile',
+			'url' => '/api/{apiVersion}/form/link/{fileFormat}',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2.3',
```
susnux, line 348: Oh we are already at `2.3` so this needs to be `2.4` (and also the one below)

docs/API.md
```diff
@@ -28,6 +28,13 @@ This file contains the API-Documentation. For more information on the returned D
 - Completely new way of handling access & shares.
 
 ### Other API changes
+- In API version 2.3 the following endpoints were introduced:
```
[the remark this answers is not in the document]
susnux: See other comment we are already at `2.3` so this needs to be `2.4`

lib/Service/SubmissionService.php
Koc: no, we can't use `fileId` here because `writeFileToCloud` method used in "save copy to files" to arbitrary path without linking

image

Koc: Also about relative vs absolute path: current codebase also uses [relative path](https://github.com/nextcloud/forms/blob/main/lib/Service/SubmissionService.php#L173). I'm using literally same method.

lib/Service/SubmissionService.php
Koc: it's written the [same like](https://github.com/nextcloud/forms/blob/main/lib/Controller/ApiController.php#L1231) it now in current main branch.

docs/API.md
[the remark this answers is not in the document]
Koc: version bumped, entries reordered, please check

lib/Migration/Version030200Date20231001200100.php
Koc: renamed to `Version040010Date20240122133700`

lib/Migration/Version030200Date20231001200100.php
susnux: No copyright belongs to you as the author :)

appinfo/routes.php
```diff
@@ -339,5 +339,23 @@
 				'apiVersion' => 'v2(\.[1-3])?'
```
susnux, line 339:
```suggestion
				'apiVersion' => 'v2(\.[1-4])?'
```

susnux: same with the rest

susnux (approved):
Nice work! Just fix the rest of the routes (API version) then this is good to go :tada:

lib/Service/SubmissionService.php
susnux: Right this PR is big enough lets do optional refactoring in the future and keep the current way :)

susnux:
@Koc Thank you very much for this fantastic work! :rocket: 
I really like the feature and I am very happy you kept on working on this, I hope we can release it with 4.1.0 soon!

Koc:
very happy to hear that, thanx! It was hard :smiley:

Chartman123:
Yes, thank you very much @Koc :) We'd be glad if you continue contributing here 🙂

jancborchardt:
Awesome work @Koc! Do you already have a guest account on our instance? We could add you there so you can join the Forms community Talk channel. :)
(I just need an email address to add you with)

Koc:
@jancborchardt no, I haven't. You can find my email address in [copyrights](https://github.com/nextcloud/forms/blob/main/lib/Migration/Version040010Date20240122133700.php#L8) :smiley:

Gamechiefx:
Having this error when attempting to create a spreadsheet:

"no app in context","method":"POST","url":"/ocs/v2.php/apps/forms/api/v2.4/form/link/ods","message":"Exception thrown: OCP\\Files\\NotFoundException","userAgent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36","version":"28.0.2.5","exception":{"Exception":"OCP\\Files\\NotFoundException","Message":"/admin/files/Documents/folder/TCC \nData Usage Assessment Form (responses).ods","Code":0,"Trace":[{"file":"/var/www/html/lib/private/Files/Node/Folder.php","line":135,"function":"get","class":"OC\\Files\\Node\\Root","type":"->","args":["/TCC/files/Documents/folder/TCC \nData Usage Assessment Form (responses).ods"]},{"file":"/var/www/html/custom_apps/forms/lib/Service/SubmissionService.php","line":216,"function":"get","class":"OC\\Files\\Node\\Folder","type":"->","args":["TCC \nData Usage Assessment Form (responses).ods"]},{"file":"/var/www/html/custom_apps/forms/lib/Controller/ApiController.php","line":1324,"function":"writeFileToCloud","class":"OCA\\Forms\\Service\\SubmissionService","type":"->","args":[["OCA\\Forms\\Db\\Form",1],"/Documents/folder","ods"]},{"file":"/var/www/html/lib/private/AppFramework/Http/Dispatcher.php","line":230,"function":"linkFile","class":"OCA\\Forms\\Controller\\ApiController","type":"->","args":["LZyTtsgDfqtFWbNo","/Documents/folder","ods"]},{"file":"/var/www/html/lib/private/AppFramework/Http/Dispatcher.php","line":137,"function":"executeController","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->","args":[["OCA\\Forms\\Controller\\ApiController"],"linkFile"]},{"file":"/var/www/html/lib/private/AppFramework/App.php","line":184,"function":"dispatch","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->","args":[["OCA\\Forms\\Controller\\ApiController"],"linkFile"]},{"file":"/var/www/html/lib/private/Route/Router.php","line":315,"function":"main","class":"OC\\AppFramework\\App","type":"::","args":["OCA\\Forms\\Controller\\ApiController","linkFile",["OC\\AppFramework\\DependencyInjection\\DIContainer"],["v2.4","ods","ocs.forms.api.linkFile"]]},{"file":"/var/www/html/ocs/v1.php","line":65,"function":"match","class":"OC\\Route\\Router","type":"->","args":["/ocsapp/apps/forms/api/v2.4/form/link/ods"]},{"file":"/var/www/html/ocs/v2.php","line":23,"args":["/var/www/html/ocs/v1.php"],"function":"require_once"}],"File":"/var/www/html/lib/private/Files/Node/Root.php","Line":207,"CustomMessage":"Exception thrown: OCP\\Files\\NotFoundException"},"id":"65c302dad2cd8"}

Chat GPT Findings:

The error log you've provided outlines a NotFoundException occurring within a Nextcloud instance, specifically when trying to access or link to an .ods file located in a user's documents folder. Here's a detailed breakdown of what happened, the probable cause, and how to address it:

Error Summary

Exception Type: OCP\Files\NotFoundException
Affected Component: The Nextcloud Forms app (custom_apps/forms), particularly its SubmissionService and ApiController.
User Action: Attempting to link an .ods file ("TCC Data Usage Assessment Form (responses).ods") to a form submission.
System Response: The system was unable to find the specified .ods file in the path provided.

Probable Cause

The error was triggered during a POST request to link an .ods file to a form submission. The path to the file includes a line break (\n) which might be unintentional or misinterpreted by the system as part of the file path. File paths typically do not include line breaks, and their presence can cause the system to search for a file in a non-existent location.

Chartman123:
@Gamechiefx please create a new issue for that and use the template for bug reports. Thank you!

matthijskooijman:
> You're other proposals will definitely not make it into this PR anymore. So please open separate feature request issues for them. We can then discuss them there.

> About the technical name: the little error should be indeed for saving the values and this function is available since a few versions already. So as you said not related. Please also add a new issue for that.

Done, I've opened #1970, #1971, #1972 and #1973 for these issues.

JocelynDelalande:
Hello, I think this feature is what I look for, but I don't understand how to use it. Is it documented somewhere ? Thanks by advance.

Chartman123:
@JocelynDelalande Just have a look at the video in the OP. :)

JocelynDelalande:
@Chartman123 Thanks for your answer, and nevermind: I was just using a version of NC that was too old for forms 4.1. Can't wait to test the new Forms :)

sbernard31:
Hi,

I understand that file should be automatically added when new form entry are submitted.

> Is the spreadsheet automatically updated when new responses are added?
> > yes, we are automatically update spreadsheet once new response added. 

I tested recently with nextcloud (29.0.7) and Forms (4.3.13)
And this is not done automatically, I need to click manually on "Re-Export" button.
Is it normal  ? or should it work ? maybe it is available on most recent version ?  :pray: 


Another question.   
Style and formula seems to be lost after a re-export when using ods.
With xlsx there are not lost. same questions. :pray:

Chartman123:
@sbernard31 the export is done by a background job, so it will take some time until the linked spreadsheet will be updated automatically.

Style and formula should also get lost with other formulas as we don't only add new responses but do a full re-export of all responses. So all changes to the linked files will be lost after the next export.

PR #1760, opened 2023-10-25
Fix submission of other answers

Chartman123:

### DIFF
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -52,7 +52,7 @@
 			<fieldset :name="name || undefined" :aria-labelledby="titleId">
 				<NcCheckboxRadioSwitch v-for="(answer) in sortedOptions"
 					:key="answer.id"
-					:checked.sync="questionValues"
+					:checked.sync="values"
 					:value="answer.id.toString()"
 					:name="`${id}-answer`"
 					:type="isUnique ? 'radio' : 'checkbox'"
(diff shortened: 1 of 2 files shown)


Chartman123:
@AIlkiv Perhaps you could also have another look at this. Current main is broken if you allow custom answers and also the validation on the server side. Statistics are also not counted correctly.

@susnux This is my first attempt to fix the problem that I told you about.

I'm currently not really sure where this is coming from, might be my change that fixed the output for XML in #1705

src/components/Questions/QuestionMultiple.vue
```diff
@@ -62,7 +62,7 @@
 					{{ answer.text }}
 				</NcCheckboxRadioSwitch>
 				<div v-if="allowOtherAnswer" class="question__other-answer">
-					<NcCheckboxRadioSwitch :checked.sync="questionValues"
+					<NcCheckboxRadioSwitch :checked.sync="values"
```
susnux, line 65: This is not valid, you shall not modify properties (`values` is a property so you should instead only set it like `:checked="values"` but handle the update with a method like `@update:values="onValuesChange"`

susnux:
> I'm currently not really sure where this is coming from, might be my change that fixed the output for XML in #1705

Yes thats at least the reason for the backend issues, for the front end the code was not that error resilient.

AIlkiv:
@Chartman123 Is the problem still there? Can you provide examples of how to reproduce?

Chartman123:
@AIlkiv we could already fix it :) it was introduced with my change of the backend to fix XML output

PR #1763, opened 2023-10-25
fix: Handle questions props as props and not as attrs

susnux:
Fixed errors reported in dev console:
```
could not invoke event listener TypeError: this.sharedForms[sharedFormIndex] is undefined
```

We should handle properties as properties and no as attributes.
This also refactors the code a bit to remove code duplication across questions.


src/components/Questions/QuestionMultiple.vue
```diff
@@ -147,6 +141,11 @@ export default {
 
 	mixins: [QuestionMixin],
 
+	/**
+	 * Inherit all attributes, meaning every attr that is not a prop is forwarded to the root of the template (the Question)
+	 */
+	inheritAttrs: true,
```
susnux, line 147: I should remove this as this is already the default, thats why you do not need something like `v-bind=$attrs"`

Chartman123: Is this meant as a task that you still want to do before merging?

Chartman123 (approved):
Looks good, couldn't find any problems when testing the frontend.

What I noticed in edit view: As soon as you enter edit mode for the title or a question, all fields switch to edit mode. Is this intended as a preparation of implementing the a11y improvements?

src/components/Questions/QuestionMultiple.vue
susnux: yes makes no sense to have that in code^^

susnux:
Squashed and fixed the commend, good to merge now :+1:

PR #1764, opened 2023-10-26
fix: Make sure "other" answers are correctly handled

susnux:
* Alternative to https://github.com/nextcloud/forms/pull/1760

In the backend it handling the "other" answer setting was not changed after switching from object to array for extra settings.

In the frontend the value handling of the checkboxes or radio switches was not correct as the initial values is always an empty array. This lead to issues with radio switch answers.


Chartman123 (approved):
Code wise this looks good :) I'll test it in the evening on my local instance

Chartman123:
@susnux Could you please have another look at ResultsSummary.vue? For radio questions there is a separate line now for each custom value. They should be aggregated under "other" like for checkboxes.

PR #1769, opened 2023-10-28
Move parameter typing for extraSettings

Chartman123:


lib/Db/Question.php
```diff
@@ -79,8 +79,6 @@ public function getExtraSettings(): array {
 	 * @param object|array $extraSettings
 	 */
 	public function setExtraSettings($extraSettings) {
```
susnux, line 81:
```suggestion
	public function setExtraSettings(array $extraSettings) {
```

I think the todo meant this

Chartman123: Ah ok, I thought it was about the param description above...

susnux: No php 8 allowed typings like `public function foo(array|object $bar)` for older PHP versions you needed that comment

Chartman123: So should we keep the `array|object` typing? And can we remove the comment, too?

lib/Db/Question.php
```diff
@@ -75,12 +75,7 @@ public function getExtraSettings(): array {
 		return json_decode($this->getExtraSettingsJson() ?: '{}', true); // assoc=true, => Convert to associative Array
 	}
 
-	/**
-	 * @param object|array $extraSettings
-	 */
-	public function setExtraSettings($extraSettings) {
-		// TODO: When the php requirement is >= 8.0 change parameter typing to `object|array` to allow assoc. arrays from `Question::fromParams`
-		
+	public function setExtraSettings(object|array $extraSettings) {
```
susnux, line 78:
```suggestion
	public function setExtraSettings(array $extraSettings) {
```

Is it anytime an object? I think this is a leftover from the object -> associative array migration

Chartman123:
```
{
  "id":87,
  "keyValuePairs":{
    "extraSettings":{
      "shuffleOptions":true,
      "allowOtherAnswer":false
    }
  }
}
```
This is at least what we send from the frontend on updating the extraSettings of a question...

susnux: Yes but I normally JS objects should be encoded as associative arrays on PHP.

Chartman123: @susnux yes, you're right: `Array ( [extraSettings] => Array ( [shuffleOptions] => 1 [allowOtherAnswer] => 1) ) `

susnux (approved):

PR #1785, opened 2023-11-20
fix: selected state after loading from localStorage

### LINKED ISSUE #1780: Checkboxes/Radio buttons don't update in UI if values are loaded from LocalStorage

**Describe the bug**
Checkboxes/Radio buttons don't update in Submit view if values are loaded from LocalStorage. If on empty form, everything works fine.

**To Reproduce**
1. Fill some values in an empty form including checkbox/radio questions in Submit view
2. Leave form
3. Come back to form in Submit view
4. Click on checkbox/radio options
5. UI doesn't adjust, focus hangs on checkbox/radio, Local Storage gets updated
6. Reload page, see changes in UI

**Expected behavior**
Checkboxes/Radios get updated on click/change

- Nextcloud-Version: 27.1.3
- Forms-Version: latest main

**Browser log**
```
Nothing in browser log
```

Chartman123:
This fixes #1780

### DIFF
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -41,7 +41,7 @@
 			<fieldset :name="name || undefined" :aria-labelledby="titleId">
 				<NcCheckboxRadioSwitch v-for="(answer) in sortedOptions"
 					:key="answer.id"
-					:checked="questionValues"
+					:checked.sync="questionValues"
 					:value="answer.id.toString()"
 					:name="`${id}-answer`"
 					:type="isUnique ? 'radio' : 'checkbox'"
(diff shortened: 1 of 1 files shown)


susnux:
What is the difference except from making is not reactive anymore?
Currently onChange is called which already should update the  values in the submit view which then should propagade back through the props to the question.
This create a copy of the value and drops the reactivity as the values are never loaded back from the props if changed after mount.

susnux:
Well I see why this is working: The question was never updated from Submit as the answers object in Submit is not reactive and thus the changed answers were not updated back on the question -> UI seems to be stuck.

This fixes it by creating a local copy of the values, but therefor loose any reactivity.

PR #1786, opened 2023-11-21
fix(Submit): Make `answers` reactive and fix invalid mutation of computed property

### LINKED ISSUE #1780: Checkboxes/Radio buttons don't update in UI if values are loaded from LocalStorage

susnux:
* Alternative to resolve #1780 

This fixes the root cause for the question component not updating: Non reactive answers in `Submit`, meaning after assigning a new value to `answers[key]` the watch effect was never triggered and thus the question component was never updated.
So instead use a `onUpdate` method to update the `answers` object in a way that will trigger any watch effect.

Also make `getFormValuesFromLocalStorage` a method instead a computed value. Previously that computed value were assigned a new value which is invalid as computed values should be constant (except for computed setter but this was none).

### DIFF
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -32,7 +32,7 @@
 				@update:checked="onShuffleOptionsChange">
 				{{ t('forms', 'Shuffle options') }}
 			</NcActionCheckbox>
-			<NcActionCheckbox :checked="extraSettings?.allowOtherAnswer"
+			<NcActionCheckbox :checked="allowOtherAnswer"
 				@update:checked="onAllowOtherAnswerChange">
 				{{ t('forms', 'Add "other"') }}
 			</NcActionCheckbox>
@@ -52,7 +52,7 @@
 				</NcCheckboxRadioSwitch>
 				<div v-if="allowOtherAnswer" class="question__other-answer">
 					<NcCheckboxRadioSwitch :checked="questionValues"
-						:value="valueOtherAnswer"
+						:value="otherAnswer ?? QUESTION_EXTRASETTINGS_OTHER_PREFIX"
 						:name="`${id}-answer`"
 						:type="isUnique ? 'radio' : 'checkbox'"
 						:required="checkRequired('other-answer')"
@@ -61,10 +61,10 @@
 						@keydown.enter.exact.prevent="onKeydownEnter">
 						{{ t('forms', 'Other:') }}
 					</NcCheckboxRadioSwitch>
-					<NcInputField :label="placeholderOtherAnswer"
-						:required="hasRequiredOtherAnswerInput"
-						:value.sync="inputOtherAnswer"
-						class="question__input" />
+					<NcInputField class="question__input"
+						:label="placeholderOtherAnswer"
+						:required="otherAnswer !== undefined"
+						:value.sync="otherAnswerText" />
 				</div>
 			</fieldset>
 		</template>
@@ -96,10 +96,9 @@
 				<li v-if="!isLastEmpty || hasNoAnswer" class="question__item">
 					<div :is="pseudoIcon" class="question__item__pseudoInput" />
 					<input ref="pseudoInput"
-						v-model="inputValue"
+						class="question__input"
 						:aria-label="t('forms', 'Add a new answer')"
 						:placeholder="t('forms', 'Add a new answer')"
-						class="question__input"
 						:maxlength="maxStringLengths.optionText"
 						minlength="1"
 						type="text"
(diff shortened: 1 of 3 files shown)


susnux:
Works also with "other" answer:

https://github.com/nextcloud/forms/assets/1855448/1943f40b-4cd7-4d66-b452-dda2e37ac2aa

src/components/Questions/QuestionMultiple.vue
```diff
@@ -311,12 +317,10 @@ export default {
 			options.push({
 				id: GenRandomId(),
 				questionId: this.id,
-				text: this.inputValue,
+				text: '',
 				local: true,
 			})
 
-			this.inputValue = ''
-
```
Chartman123, line 319: This is needed for the new create view behaviour... I need to get the input from the pseudo input for new options and pass them to the new AnswerInput for the new option

src/components/Questions/QuestionMultiple.vue
```diff
@@ -197,29 +197,21 @@ export default {
...
-				values.push(this.valueOtherAnswer)
+		/**
+		 * Text of the "other" answer field
+		 */
+		otherAnswerText() {
+			if (this.otherAnswer) {
+				return this.otherAnswer.slice(QUESTION_EXTRASETTINGS_OTHER_PREFIX.length)
 			}
+			return ''
+		},
```
Chartman123, line 208: This changes the current behaviour: when you unselect the checkbox/radio for the custom answer, the text stays in the input box so that you don't have to re-type the text when you select the option again.

With this change the text gets removed as soon as you unselect the option

susnux: Do we want to keep the old behavior? Would require to cache the input (additional data and watcher)

src/components/Questions/QuestionMultiple.vue
susnux: Fixed:

https://github.com/nextcloud/forms/assets/1855448/01fdaeb4-3aeb-410b-aa5f-be185a671546

src/components/Questions/QuestionMultiple.vue
Chartman123: If it's not too much effort, I would prefer to keep the old behaviour... Might be an edge use case that someone checks/unchecks the other option but then might be really annoying to always have to re-type the answer.

susnux: Fixed

susnux:
Fixed the "other" answer cache, see:


https://github.com/nextcloud/forms/assets/1855448/09974ccd-1240-4139-b226-a513ebe94fd8

src/components/Questions/QuestionMultiple.vue
```diff
@@ -61,10 +61,11 @@
 						@keydown.enter.exact.prevent="onKeydownEnter">
 						{{ t('forms', 'Other:') }}
 					</NcCheckboxRadioSwitch>
-					<NcInputField :label="placeholderOtherAnswer"
-						:required="hasRequiredOtherAnswerInput"
-						:value.sync="inputOtherAnswer"
-						class="question__input" />
+					<NcInputField class="question__input"
+						:label="placeholderOtherAnswer"
+						:disabled="otherAnswer === undefined"
```
Chartman123, line 66: Is this necessary? With this change you have to first check the checkbox/radio, before you could start to write into the field and that triggered the checked state

susnux: Not really, but it simplifies the code and makes the flow easier as from the control structure the element is disabled when not selected. Meaning the accessibility properties should list it as disabled.

Chartman123: Ok, but from UX perspective I'd prefer the way it is e.g. also handled by Google: type into the input and auto check the correponding radio/checkbox

susnux: Fixed:

https://github.com/nextcloud/forms/assets/1855448/c43528ae-fd28-457a-9f01-d6badde03f51

Chartman123 (approved):

PR #1795, opened 2023-11-27
chore(phpunit): Switch to phpunit workflow from organization

susnux:


susnux:
Only integration is failing, but I do not understand:
The form that is exported is **not** anonymous, so all users should be returned by their ID / displayname.
Why did the test ever expect to return anonymous?

Chartman123:
> Why did the test ever expect to return anonymous?

In the current PHPUnit-WF we created one user `user1`, user2 and user3 did not exist in the test environment. Might this be the cause for this?

susnux:
Ah yes I see so it tests for user2 and user3 are deleted users...

susnux:
Not related but maybe we then need to anonymize reponses on user-deleted events, because otherwise they would be connected to new users with the same ID

Chartman123:
> Ah yes I see so it tests for user2 and user3 are deleted users... 

Yes, that's how I understand it

Chartman123:
> Not related but maybe we then need to anonymize reponses on user-deleted events, because otherwise they would be connected to new users with the same ID

Good idea :) could you please create a new issue for that?

Chartman123 (approved):

Issue #1781, opened 2023-11-17
Search for users by email in the sharing section


iFortunateSon:
Hello,

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 26.0.4
- Forms-Version: 3.3.1

**Describe the solution you'd like**
When entering an email into the user search box, the found user is displayed, as it is done when sharing access to files in nextcloud

**Additional context**
Providing access for files in NC:
![image](https://github.com/nextcloud/forms/assets/54754344/2502607c-ceb2-4e48-b569-b8f03236977a) 
The user is found by email

Providing access for Forms:
![image](https://github.com/nextcloud/forms/assets/54754344/ac62ca95-7fe3-4175-aa7b-e4ca836d967e)
User is not found by email

susnux:
Sounds good! I will have a look how this can be done :)

PR #1802, opened 2023-11-28
enh(sharing): Move away from deprecated icon classes and allow to search user by email

susnux:
1. Move away from deprecated css classes for icons and use SVGs
2. Add `subname` attribute to sharees, this allows to search users by their email address like on server core

Resolves https://github.com/nextcloud/forms/issues/1781


Chartman123 (approved):
Looks good :)

PR #1803, opened 2023-11-28
enh: Show confirmation dialog before submitting an empty form

### LINKED ISSUE #482: How to deal with empty responses
Sometimes during the surveys we had responses which were completely empty, that is none of the questions had any answer.

We should probably catch this, even when none of the fields are mandatory. It’s just likely an error with pressing the submit button and doesn’t really make sense to list.

I thought about cutting it out of the data at display, but what would make more sense is to already catch it on submit. What do you think @jotoeri?

susnux:
* Resolves #482

When submitting and empty form a confirmation dialog is shown, to make sure users submitted willingly an empty form.
Also please note the missing padding of the text in the dialog will be fixed with the next nextcloud-vue update (is a component bug not related to this PR).

https://github.com/nextcloud/forms/assets/1855448/47212d9c-31a8-422b-9fd4-cf83053a1685

Note: This can not be backported due to missing NcDialog in nextcloud-vue 7.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,5 +8,6 @@ module.exports = {
 	rules: {
 		// We are using the @nextcloud/logger
 		'no-console': ['error', { allow: undefined }],
+		'import/no-unresolved': ['error', { ignore: ['\\?raw'] }],
 	},
 }
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -70,7 +70,7 @@
 			:name="t('forms', 'Thank you for completing the form!')"
 			:description="form.submissionMessage">
 			<template #icon>
-				<IconCheck :size="64" />
+				<NcIconSvgWrapper :svg="IconCheckSvg" :size="64" />
 			</template>
 			<template v-if="submissionMessageHTML" #description>
 				<!-- eslint-disable-next-line vue/no-v-html -->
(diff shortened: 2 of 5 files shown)


src/views/Submit.vue
```diff
@@ -70,7 +70,7 @@
 			:name="t('forms', 'Thank you for completing the form!')"
 			:description="form.submissionMessage">
 			<template #icon>
-				<IconCheck :size="64" />
+				<NcIconSvgWrapper :svg="IconCheckSvg" :size="64" />
```
susnux, line 73: This is theoretically unrelated, but we can simply also use the SVG here to drop one dependency of this component (the NcIconSvgWrapper will be anyways included by NcDialog / NcDialogButton). Meaning we can get rid of one unneeded component (IconCheck).

.eslintrc.js
```diff
@@ -8,5 +8,6 @@ module.exports = {
 	rules: {
 		// We are using the @nextcloud/logger
 		'no-console': ['error', { allow: undefined }],
+		'import/no-unresolved': ['error', { ignore: ['\\?raw'] }],
```
Chartman123, line 11: Shouldn't this be in a separate PR? It's not related to the feature itself.

Chartman123 (approved):

.eslintrc.js
susnux: It is otherwise eslint fails because of the raw queries

szaimen:
> * There is very little outside padding, especially visible on the left of the content text and on the right and bottom of the buttons. Is this not handled by the component cc @marcoambrosini @szaimen ?

Indeed the confirm modal looks strange. @susnux any idea why?

susnux:
> * We generally don’t use the wording "Abort" but instead "Cancel"
> 
> * The "Cancel" button does not need an icon as that calls for too much attention again despite being secondary :)

Ok will change this , thank you for the feedback :)

PR #1805, opened 2023-11-28
enh: Allow to set `results_delete` permission on the frontend

susnux:
Allow to set the permission for deleting submissions. Also fix the permission check when deleting submission to allow any user with permission to delete a submission instead of restricting to form owner.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -1016,8 +1016,9 @@ class ApiController extends OCSController {
 			throw new OCSBadRequestException();
 		}
 
-		if ($form->getOwnerId() !== $this->currentUser->getUID()) {
-			$this->logger->debug('This form is not owned by the current user');
+		// The current user has permissions to remove submissions
+		if (!$this->formsService->canDeleteResults($form)) {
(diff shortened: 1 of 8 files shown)


lib/Controller/ApiController.php
```diff
@@ -1016,9 +1016,13 @@ public function deleteSubmission(int $id): DataResponse {
...
 
+		// Either the current user needs to be the owner
 		if ($form->getOwnerId() !== $this->currentUser->getUID()) {
-			$this->logger->debug('This form is not owned by the current user');
-			throw new OCSForbiddenException();
+			// Or has permissions to remove submissions
+			if (!$this->formsService->canDeleteResults($form)) {
+				$this->logger->debug('This form is not owned by the current user and user has no results_delete permission');
+				throw new OCSForbiddenException();
+			}
```
Chartman123, line 1025: Is the double if needed here? The owner of the form always has the permission to delete submissions, doesn't he? So it would suffice to only query for !canDeleteResults... Or am I wrong?

susnux: Yes this can be simplified :)

Chartman123 (approved):

PR #1809, opened 2023-11-29
enh: use PUT for updating and move to API v2.2

### LINKED ISSUE #1464: Move POST endpoints that update data to PUT/PATCH
Currently we're only using GET, POST and DELETE as verbs in our API. 

According to https://www.restapitutorial.com/lessons/httpmethods.html PUT/PATCH should be used for updating data.

This will break the current API and needs to be planned thoroughly.

Chartman123:
This fixes #1464 by adding new PUT/PATCH routes for current POST routes that update data.

Additionally, this changes the API level to v2.2 and marks the POST endpoints as deprecated. The will be removed in v3 of the API as a breaking change.


appinfo/routes.php
```diff
@@ -82,63 +89,72 @@
...
 			]
 		],
+		// TODO: Remove POST in next API release
 		[
 			'name' => 'api#updateForm',
 			'url' => '/api/{apiVersion}/form/update',
 			'verb' => 'POST',
 			'requirements' => [
-				'apiVersion' => 'v2(\.1)?'
+				'apiVersion' => 'v2(\.[1-2])?'
```
susnux, line 125: Not sure here if we can drop it from newer API but keep it as 2.1?

appinfo/routes.php
```diff
@@ -31,11 +32,17 @@
...
 		[
 			'name' => 'config#updateAppConfig',
 			'url' => '/config/update',
 			'verb' => 'POST'
 		],
+		[
+			'name' => 'config#updateAppConfig',
+			'url' => '/config/update',
+			'verb' => 'PATCH'
+		],
```
susnux, line 45:
```suggestion
		[
			'name' => 'config#updateAppConfig',
			'url' => '/config/update',
			'verb' => 'PATCH'
		],
```

I think we can safely remove this as this is not part of our public API

appinfo/routes.php
susnux: (same would apply to other places below, but I am not really sure).

susnux (approved):
Looks good to me, two comments but overall :broom:

appinfo/routes.php
Chartman123: If we would drop it here I would prefer to raise the API level to version 3 for the new methods. I wanted to make transition to the new methods as easy as possible. That's why I kept everything also for 2.2

Chartman123:
You were a little bit too quick with merging... I wanted to rebase this again and have the two commits in one ;)

PR #1812, opened 2023-11-30
fix mutating prop directly by removing loading spinner

Chartman123:


susnux:
I would say for feedback we should keep the loading spinners.

PR #1832, opened 2023-12-14
Fix typo in PR feedback action

zoul:


Chartman123 (approved):
Thanks for the fix :)

PR #1841, opened 2023-12-21
fix: Prevent race condition on unique-submission forms

susnux:
Currently a series of requests can be crafted so that multiple submissions of the same user are inserted.
As the unique property of a submission is only checked once so that if multiple requests arrive at the server right in the same time they will all be inserted.

To prevent this the time window is minimized by placing the check right before the insert and check afterwards if the submission really is unique.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -1055,11 +1055,6 @@ class ApiController extends OCSController {
 			throw new OCSForbiddenException('This form is no longer taking answers');
 		}
 
-		// Does the user have permissions to submit
-		if (!$this->formsService->canSubmit($form)) {
-			throw new OCSForbiddenException('Already submitted');
-		}
-
 		// Is the submission valid
 		if (!$this->submissionService->validateSubmission($questions, $answers)) {
 			throw new OCSBadRequestException('At least one submitted answer is not valid');
@@ -1078,9 +1073,22 @@ class ApiController extends OCSController {
 			$submission->setUserId($this->currentUser->getUID());
 		}
 
+		// Does the user have permissions to submit
+		// This is done right before insert so we minimize race conditions for submitting on unique-submission forms
+		if (!$this->formsService->canSubmit($form)) {
+			throw new OCSForbiddenException('Already submitted');
+		}
+
 		// Insert new submission
 		$this->submissionMapper->insert($submission);
 		$submissionId = $submission->getId();
+
+		// Ensure the form is unique if needed.
+		// If we can not submit anymore then the submission must be unique
+		if (!$this->formsService->canSubmit($form) && !$this->submissionService->isUniqueSubmission($submission)) {
+			$this->submissionMapper->delete($submission);
+			throw new OCSForbiddenException('Already submitted');
+		}
 
 		// Process Answers
 		foreach ($answers as $questionId => $answerArray) {
--- a/lib/Db/SubmissionMapper.php
+++ b/lib/Db/SubmissionMapper.php
@@ -110,16 +110,22 @@ class SubmissionMapper extends QBMapper {
 	}
 
 	/**
+	 * Count submissions by form and optionally also by userId
+	 * @param int $formId ID of the form to count submissions for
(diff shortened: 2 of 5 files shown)


Chartman123:
@susnux Looks good :) Just a little question: You check all submissions against the new submission for uniqueness. If we have a form with lots of submissions, isn't this very resource consuming?

susnux:
@Chartman123 I pushed a performance fix, what do you think?

Chartman123:
Yes, something like this was on my mind 👍🏻

lib/Controller/ApiController.php
```diff
@@ -987,10 +982,23 @@ public function insertSubmission(int $formId, array $answers, string $shareHash
...
+			throw new OCSForbiddenException('Already submitted');
+		}
+
 		// Insert new submission
 		$this->submissionMapper->insert($submission);
 		$submissionId = $submission->getId();
 
+		// Ensure the form is unique if needed.
+		// If we can not submit anymore then the submission must be unique
+		if (!$this->formsService->canSubmit($form) && !$this->submissionService->isUniqueSubmission($submission)) {
```
Fenn-CS, line 997:
```suggestion
		if (!$this->formsService->canSubmit($form) && !$this->submissionService->isUniqueSubmission($submission)) {
```

Since the first part of this condition already exists up exactly and throws an exception, if this exception exists this part of the code would never be reached?

Or does `canSubmit` result change after the first insertion?

Fenn-CS (commented):
Left and inline remark/question to help me review.

lib/Controller/ApiController.php
Chartman123: This is a calculated form prop https://github.com/nextcloud/forms/blob/main/lib%2FService%2FFormsService.php#L271

susnux: <del>The idea is here if the user still can submit after insert then multiple submissions are allowed</del>

While writing this: Is this not a (unrelated to this PR) bug? If unique submissions are enabled why should the owner be allowed to submit multiple?

Chartman123: > why should the owner be allowed to submit multiple?

Good question, the only reason I could think of: The owner fills the form for someone else who can't access the online form for some reason but needs to?

This part of the code was added by @jotoeri 2 years ago. So let's ask him what was his intention :)

jotoeri: Hmm, i'm not !00% sure, but i think we shortly discussed that somewhere that once. Otherwise that change wouldn't have gone through just like that.^^
Anyways: I would interpret the Owner as some kind of an administrator for the form, being allowed to do everything with it. The owner could also just remove the submitOnce, submit and then re-set it, which is just annoying handling. So why should we block the owner of doing whatever he/she wants to do with the form?
In fact it's mostly a UX-Question wrt. the owner, no?

Chartman123:
> by placing the check right before the insert

@susnux One thing that just came to my mind: by moving the check right before the insert, we would check for valid submission data even if a user were no longer allowed to submit. This creates some unnecessary checks in that case.

susnux:
> @susnux One thing that just came to my mind: by moving the check right before the insert, we would check for valid submission data even if a user were no longer allowed to submit. This creates some unnecessary checks in that case.

Well this only happens when using API and not our frontend as the frontend already checks that. Meaning only API usage from external application will cause this.
But yes both cases are not triggered by our frontend but only by (malicious) external usage, so we can discuss both ways.

If we keep the order of the checks the time frame for the attack is a bit wider but as we sanitize afterwards it would be reverted anyways.

Chartman123:
For me it's ok to keep it the way you implemented it currently 👍🏻

Chartman123 (approved):
Just squash the fixup commit, then we can merge it from my point of view :)

lib/Controller/ApiController.php
Chartman123: @Fenn-CS do the comments help you to finish your review?

susnux:
> Just squash the fixup commit, then we can merge it from my point of view :)

Done

PR #1851, opened 2023-12-24
fix(Question): close question menu after copy question

Chartman123:
This fixes some focus jumping after copying a question by closing the action menu after the click.

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -99,7 +99,8 @@
 						</template>
 						{{ t('forms', 'Technical name') }}
 					</NcActionInput>
-					<NcActionButton @click="onClone">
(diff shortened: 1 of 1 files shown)


susnux (approved):

src/components/Questions/Question.vue
```diff
@@ -99,7 +99,8 @@
 						</template>
 						{{ t('forms', 'Technical name') }}
 					</NcActionInput>
-					<NcActionButton @click="onClone">
+					<NcActionButton :close-after-click="true"
```
susnux, line 102: For boolean properties this is not needed and depends on your preference / styleguide as this is the same like:
```suggestion
					<NcActionButton close-after-click
```

PR #1853, opened 2023-12-26
fix: Use separate chunk for `vue` and `vue-router` to prevent multiple imports

### LINKED ISSUE #1842: Export Resultat as CSV to Files not functional
**Describe the bug**
The survey results cannot be exported as CSV to the files app, at least in my instance.

**To Reproduce**
1. Go to 'results'
2. Click on 'three dot menu'
3. Choose "Export CSV to Files"
4. No file created 

**Expected behavior**
File should be created 

**Screenshots**

- Nextcloud-Version: [ 28.0.0.1]
- Forms-Version: latest as of time of reporting 

 - OS: [Debian 12]
 - Browser Brave
 - Version latest as of time of reporting

susnux:
Fixes #1842

For some reason vue and vue-router are imported multiple times by different chunks, this causes errors as vue2 is not allowed to be used more than one time in one application.
So it is fixed by create a manual chunk just for vue, vue-router and its dependencies.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -10,6 +10,13 @@ export default createAppConfig({
 	config: {
 		build: {
 			cssCodeSplit: false,
+			rollupOptions: {
+				output: {
+					manualChunks: {
+						vendor: ['vue', 'vue-router'],
+					},
+				},
+			},
 		},
 	},
 })


Chartman123:
@susnux If you close the picker without selecting a file/folder, there's another error message in the console:

`Uncaught (in promise) Error: FilePicker: No nodes selected`

Do you think you can add fix for that to this PR as well or should we create a separate PR for that as this also appears on instances where the original issue isn't showing up?

Chartman123 (approved):
I just checked it on my productive instance where I had the issue. The file picker opens like expected.

PR #1880, opened 2024-01-17
enh: Use NcDialog to confirm leaving

Chartman123:
This uses NcDialog for warning the user before leaving an unsubmitted form instead of the default confirm dialog of the browsers. Due to limitations of the browsers this only works for route changes within the Forms app. When switching to other apps, another website or closing the tab/window, the default dialog will still be used.

### DIFF
--- a/src/components/TopBar.vue
+++ b/src/components/TopBar.vue
@@ -85,6 +85,7 @@
 </template>
 
 <script>
+import logger from '../utils/Logger.js'
 import { useIsMobile } from '@nextcloud/vue'
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import IconEye from 'vue-material-design-icons/Eye.vue'
@@ -162,7 +163,7 @@ export default {
 					params: {
 						hash: this.$route.params.hash,
 					},
-				})
+				}).catch((error) => logger.debug('Navigation cancelled', { error }))
 			}
 		},
 
@@ -173,7 +174,7 @@ export default {
 					params: {
 						hash: this.$route.params.hash,
 					},
-				})
+				}).catch((error) => logger.debug('Navigation cancelled', { error }))
 			}
 		},
 
@@ -184,7 +185,7 @@ export default {
 					params: {
 						hash: this.$route.params.hash,
 					},
-				})
+				}).catch((error) => logger.debug('Navigation cancelled', { error }))
 			}
 		},
 
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -1,6 +1,7 @@
 <!--
  - @copyright Copyright (c) 2020 John Molakvoæ <skjnldsv@protonmail.com>
  -
+ - @author Christian Hartmann <chris-hartmann@gmx.de>
  - @author John Molakvoæ <skjnldsv@protonmail.com>
  - @author Michael Schmidmaier
  -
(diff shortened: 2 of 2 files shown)


src/views/Submit.vue
```diff
@@ -174,7 +179,7 @@ export default {
 	beforeRouteUpdate(to, from, next) {
 		// This navigation guard is called when the route parameters changed (e.g. form hash)
 		// continue with the navigation if there are no changes or the user confirms to leave the form
-		if (this.confirmLeaveForm()) {
+		if (this.confirmLeaveForm(next)) {
```
susnux, line 182: Maybe make the function async and await that method here, meaning `if (await this.confirm...)`

src/views/Submit.vue
```diff
@@ -449,11 +471,11 @@ export default {
 		Check if the form contains unsaved changes, returns true if the the form can be leaved safely, false if the navigation should be canceled.
 		 */
 		confirmLeaveForm() {
-			return (
-				this.submitForm
-				|| Object.keys(this.answers).length === 0
-				|| confirm(t('forms', 'You have unsaved changes! Do you still want to leave?'))
-			)
+			if (!this.submitForm && Object.keys(this.answers).length !== 0) {
```
susnux, line 474: and here then use something like 
```js
return new Promise((resolve) => {
    this.confirmButtonCallback = resolve 
}
```

src/views/Submit.vue
```diff
@@ -301,6 +307,22 @@ export default {
...
+		confirmLeaveFormButtons() {
+			return [{
+				label: t('forms', 'Abort'),
+				icon: IconCancelSvg,
+				callback: () => false,
+			}, {
+				label: t('forms', 'Leave'),
+				icon: IconCheckSvg,
+				type: 'primary',
+				callback: () => true,
```
susnux, line 323: And here something like:
```js
callback: () => this.confirmButtonCallback(true)
```

susnux (commented):
Maybe like this?

susnux:
I get the dialog even if no answers were entered?


https://github.com/nextcloud/forms/assets/1855448/c81e6289-1101-4318-9059-a1c5eef7b8df

Chartman123:
> I get the dialog even if no answers were entered?

Should be fixed now :)

src/views/Submit.vue
```diff
@@ -302,6 +311,22 @@ export default {
...
 			}]
 		},
+
+		/**
+		 * Buttons for the "confirm leave unsubmitted form" dialog
+		 */
+		confirmLeaveFormButtons() {
+			return [{
+				label: t('forms', 'Abort'),
+				icon: IconCancelSvg,
```
susnux, line 321: I think we do not use button for those dialogs on Nextcloud, so maybe remove both.
And maybe instead make the `leave` button `type="error"`?

Chartman123: You mean 'icon' instead of 'button', don't you?

susnux: Yes Icons 😅

Chartman123: Ok, `error` doesn't look right next to the default button. So for now I would stick with primary and also the icons. You're using also the icons in the confirmation dialog for submitting an empty form. :)

susnux: Well then lets stick for the moment with that :)

src/components/TopBar.vue
```diff
@@ -162,7 +163,7 @@ export default {
 					params: {
 						hash: this.$route.params.hash,
 					},
-				})
+				}).catch((err) => logger.debug('Navigation cancelled', { err }))
```
susnux, line 166: There is special handling for 'error' in the context so I would rename 'err' to 'error'

susnux (approved):

PR #1892, opened 2024-01-24
stop filtering result while sharing

avinash-0007:
- Share only works by display name right now

- Search by username to share form is broken completely as there is a filterBy option in [vue-select](https://vue-select.org/api/props.html#filterby) whose return value defaults to return (label || '').toLocaleLowerCase().indexOf(search.toLocaleLowerCase()) > -1

   - label in our code is set to displayName so only search by displayName works
- Since we already get our results by searching(OCS), I don't see why we need the filterBy option in the first place
   - So I just make the return value true
   - In case we do need the option, we should make it check for both displayName and username attributes


avinash-0007:
this PR is based on this PR for NcSelect https://github.com/nextcloud/calendar/pull/5645

src/components/SidebarTabs/SharingSearchDiv.vue
```diff
@@ -30,6 +30,7 @@
 			:options="options"
 			:placeholder="t('forms', 'Search for user, group or circle …')"
 			:user-select="true"
+			:filter-by="filterResults"
```
susnux, line 33: I am not sure if a simple `() => true` would not be better here.
As we do not add any logic but stub the function.

Would also help when reading the component to understand what is happening, I would say.

src/mixins/UserSearchMixin.js
```diff
@@ -66,6 +66,17 @@ export default {
 		},
 	},
 	methods: {
+	/**
+	 * Function to filter results in NcSelect
```
susnux, line 70: Otherwise (see above) if we keep this way, then please add a comment **why** you are just returning `true`.

susnux (commented):
Thank you!
See comments and please fix the lint issues :)

src/mixins/UserSearchMixin.js
avinash-0007: @susnux yeah.your suggestion looks fine to me. Implemented.

Chartman123 (approved):

susnux (approved):

Chartman123:
And squash them into one single commit :)

avinash-0007:
@Chartman123 @susnux - created new PR here:https://github.com/nextcloud/forms/pull/1895

PR #1904, opened 2024-01-25
fix empty content public page view

### LINKED ISSUE #1903: "Form not found" page looks broken in 28
open e.g. `/apps/forms/s`

grafik

Chartman123:
This fixes #1903 by adding NcAppContent to the FormsEmptyContent template

### DIFF
--- a/src/emptyContent.js
+++ b/src/emptyContent.js
@@ -24,6 +24,9 @@ import Vue from 'vue'
 
 import FormsEmptyContent from './FormsEmptyContent.vue'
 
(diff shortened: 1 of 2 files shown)


src/emptyContent.js
Chartman123, line 1: Unsure if this is needed, but it is in the other .js files in our src root

susnux (approved):

src/emptyContent.js
susnux: Needed for older browsers

PR #1906, opened 2024-01-26
WIP: Make show forms to all users in sidebar an admin setting

akhil1508:
- Permit to all can be controlled by an admin setting
- Makes sense to me to control also "show to all" using an admin setting


Chartman123:
Hi @akhil1508 

are you still working on this? What's still missing? Do you need any help? :)

PR #1911, opened 2024-01-26
use isMenu for Download action

Chartman123:


src/views/Results.vue
```diff
@@ -536,13 +530,6 @@ export default {
 			return submissions
 		},
 
-		hideSubActions() {
-			// close event is fired too early, which lead to showing main menu for few ms, so we need to wait a bit
```
Koc, line 540: Have you seen this comment before removal? Also menu works buggy after this changes, see attached recording

https://github.com/nextcloud/forms/assets/191082/28554021-6d4b-4e1b-87de-671d23f5dfb5

Chartman123: The bug is caused by something else... I tried to fix it with that. But it's probably in the Vue lib.

Chartman123: @Koc btw this shouldn't be needed... Mail also uses the submenu handling like this and they don't have any timer for switching between menu and submenu.

susnux (approved):

PR #1920, opened 2024-01-27
Hide linked file actions from users who have access to results view

Koc:
Use case:
1. Share Form with User, give him access to view results

Current behavior
- User can see `Link file/Unlink file/Open file` buttons (but he not allowed to perform this actions due to checks on BE side)

Behavior after fix
- Only Form owner can see mentioned buttons

<details><summary>Screenshots</summary>

image
image
</details>

### DIFF
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -259,6 +259,16 @@ class FormsService {
 	}
 
 	/**
+	 * Can the current user edit a form
+	 *
+	 * @param Form $form
+	 * @return boolean
+	 */
+	public function canEditForm(Form $form): bool {
+		return in_array(Constants::PERMISSION_EDIT, $this->getPermissions($form));
+	}
+
+	/**
 	 * Can the current user see results of a form
 	 *
 	 * @param Form $form
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -62,7 +62,7 @@
 						{{ t('forms', 'Share form') }}
 					</NcButton>
 
-					<NcButton v-if="!form.fileId" type="tertiary-no-background" @click="onLinkFile">
+					<NcButton v-if="canEditForm && !form.fileId" type="tertiary-no-background" @click="onLinkFile">
 						<template #icon>
 							<IconLink :size="20" />
 						</template>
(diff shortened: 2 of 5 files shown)


Koc:
Test failures looks unrelated

susnux:
Not sure if we need a new permission or if it should be within the `PERMISSION_EDIT`

Koc:
@susnux I've updated PR with reusing already existent `PERMISSION_EDIT` permission instead of introduction of the new one

susnux (approved):

lib/Service/FormsService.php
```diff
@@ -258,6 +258,16 @@ public function getPermissions(Form $form): array {
...
 	}
 
+	/**
+	 * Can the current user edit a form
+	 *
+	 * @param Form $form
+	 * @return boolean
+	 */
+	public function canEditForm(Form $form): bool {
+		return in_array(Constants::PERMISSION_EDIT, $this->getPermissions($form), true);
```
Chartman123, line 268: The other permission queries in the FormsService don't use the `strict` comparison. Is this really needed here? I'd prefer to have them all the same...

src/views/Results.vue
```diff
@@ -112,28 +112,30 @@
...
+						</NcActions>
+					</template>
 
 					<!-- Action menu for cloud export and deletion -->
 					<NcActions :aria-label="t('forms', 'Options')"
 						:force-menu="true"
 						@close="isDownloadActionOpened = false">
 						<template v-if="!isDownloadActionOpened">
-							<template v-if="form.fileId">
+							<template v-if="canEditForm && form.fileId">
```
Chartman123, line 138: Why shouldn't someone having only the results permission but not the edit permission not be able to do a re-export?

@susnux What do you think about this?

src/views/Results.vue
```diff
@@ -112,28 +112,30 @@
...
-					<NcButton v-if="form.fileId" :href="fileUrl" type="tertiary-no-background">
-						<template #icon>
-							<IconTable :size="20" />
-						</template>
-						{{ t('forms', 'Open spreadsheet') }}
-					</NcButton>
-
-					<NcActions v-else type="tertiary-no-background" :force-name="true">
-						<NcActionButton @click="onLinkFile">
+					<template v-if="canEditForm">
```
Chartman123, line 115: Connected to my other comment: Someone with results permission but not with edit permission can't open the linked spreadsheet with the results. If this person has access to the file, I'd say they should also be able to click on this button...

@susnux What do you think?

src/views/Results.vue
Koc: For now `edit` permission granted only to the Form owner. It is not possible to share form with `edit` permission

susnux: > Why shouldn't someone having only the results permission but not the edit permission not be able to do a re-export?

The file can only linked by the owner, so if you have submissions permission you are not guranteed to have access to the results. You would be able to overwrite files not owned by you nor shared to you.

Chartman123: Yes, but this would also be true for people with `edit` permissions... So if we follow that logic it would only make sense to keep this buttons visible to only the owner.

susnux: No someone with `edit` permissions would at least be able to replace the linked file with an owned one.
Meaning someone with `results` permissions currently is read-only but would be promoted to write actions here

Chartman123: True, but still some problems possible if they change the linked file to some other path... 🙈

How do other form apps behave? Did someone check that?

Chartman123: Fine for me :)

lib/Service/FormsService.php
Koc: fixed

src/views/Results.vue
Koc: so, as far I understand - nothing needed from my side, right?

Chartman123 (approved):

Chartman123:
Ok, then just do a rebase and squash the commits into one :)

PR #1932, opened 2024-02-01
Update API version to v2.4

Chartman123:
This pushes all API calls in the app to the new 2.4 level


susnux:
That are the deprecated routes, I think we do not need or should not update?

Chartman123:
Yes, I was unsure... In the docs we have stated, that they will be removed with v3, that's why updated them as well

susnux:
Well personally I think we should support it in every version until it is removed?

susnux (approved):

PR #2011, opened 2024-03-16
chore: update deps

Chartman123:
Update npm and composer dependencies

### DIFF
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -655,7 +655,7 @@ class FormsService {
 	}
 
 	public function getFileName(Form $form, string $fileFormat): string {
-		if (empty(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
+		if (!isset(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
 			throw new \InvalidArgumentException('Invalid file format');
 		}
 
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -178,7 +178,7 @@ class SubmissionService {
 	 * @throws NotPermittedException
 	 */
 	public function writeFileToCloud(Form $form, string $path, string $fileFormat, ?string $ownerId = null): File {
-		if (empty(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
+		if (!isset(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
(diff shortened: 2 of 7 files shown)


Chartman123:
@susnux After updating npm and composer deps we get some Psalm static analysis errors regarding the check if the file type array returns an empty value for undefined file types. Is there a better way to fix this error than to just mute it like I did in my second commit?

lib/Service/FormsService.php
```diff
@@ -655,6 +655,7 @@ public function getFilePath(Form $form): ?string {
 	}
 
 	public function getFileName(Form $form, string $fileFormat): string {
+		/** @psalm-suppress TypeDoesNotContainType */
 		if (empty(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
```
susnux, line 659: Instead of silence the message:

```suggestion
		if (isset(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
```

lib/Service/SubmissionService.php
```diff
@@ -178,6 +178,7 @@ public function isUniqueSubmission(Submission $newSubmission): bool {
 	 * @throws NotPermittedException
 	 */
 	public function writeFileToCloud(Form $form, string $path, string $fileFormat, ?string $ownerId = null): File {
+		/** @psalm-suppress TypeDoesNotContainType */
 		if (empty(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
```
susnux, line 182: instead 
```suggestion
		if (isse(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
```

lib/Service/SubmissionService.php
```diff
@@ -239,6 +240,7 @@ public function writeFileToCloud(Form $form, string $path, string $fileFormat, ?
 	 * @return string File content
 	 */
 	public function getSubmissionsData(Form $form, string $fileFormat, ?File $file = null): string {
+		/** @psalm-suppress TypeDoesNotContainType */
 		if (empty(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
```
susnux, line 244:
```suggestion
		if (isset(Constants::SUPPORTED_EXPORT_FORMATS[$fileFormat])) {
```

lib/Service/FormsService.php
Chartman123: Thank you :)

susnux (approved):

PR #2013, opened 2024-03-18
Show expired shared forms to users with results permission

### LINKED ISSUE #1991: Shared expired forms disappear even when "access results" permission is granted

**Describe the bug**
When a form expires, *all* shares on the form stop working. The form disappears from the navigation bar of the share receivers. This includes shares with "access results" permission. Hence, it is impossible for collaborators to access the results of the form once the form expires.

It makes sense to hide the form from users that have received the share just to *fill* the form, but less so for users that are expected to work with the results.

A workaround is for the form creator to export the results to CSV and then share that CSV with all of their collaborators.

**To Reproduce**
1. Create any form
2. Share the form with any user and tick the checkbox for the "access results" permission
3. Confirm that the user you shared the form with can indeed access the form and its results page
4. Set the expiration date to a date in the past
5. The user you shared the form with can no longer see it, so they can't access the results page anymore

**Expected behavior**
Users with the "access results" share permission should still be able to access the results when the form expires.

- Nextcloud-Version: 27.1.6 Enterprise
- Forms-Version: 3.4.6

Chartman123:
This fixes #1991 by checking whether a user has the permission to see results.

### DIFF
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -389,8 +389,8 @@ class FormsService {
 			return false;
 		}
 
-		// Dont show expired forms.
-		if ($this->hasFormExpired($form)) {
+		// Dont show expired forms if user isn't allowed to see results.
+		if ($this->hasFormExpired($form) && !$this->canSeeResults($form)) {
 			return false;
 		}
 


Chartman123 (commented):
/backport to stable3

susnux (approved):
thank you!

Issue #1699, opened 2023-08-24
Navigation icons disappear on mobile after showing settings sidebar


Chartman123:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**

**To Reproduce**
1. Open form on mobile browser
2. Open settings sidebar
3. Close settings sidebar
4. See disappeared icons

**Expected behavior**
Icons still shown

**Screenshots**
https://github.com/nextcloud/forms/assets/16020496/993bca7a-4d44-4a42-baeb-950a8939a66e

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 27.0.2
- Forms-Version: 3.3.1

**Smartphone (please complete the following information):**
 - Device: Oneplus Nord 3
 - OS: Android 13
 - Browser Firefox

**Browser log**
```
Open your console, reload your page and/or do the action leading to this issue and copy/paste the log in this thread.
```

<details>
<summary>How to access your browser console (Click to expand)</summary>

# Chrome
- Press either CTRL + SHIFT + J to open the “console” tab of the Developer Tools.
- Alternative method:
    1. Press either CTRL + SHIFT + I or F12 to open the Developer Tools.
    2. Click the “console” tab.

# Safari
- Press CMD + ALT + I to open the Web Inspector.
- See Chrome’s step 2. (Chrome and Safari have pretty much identical dev tools.)

# IE9
1. Press F12 to open the developer tools.
2. Click the “console” tab.

# Firefox
- Press CTRL + SHIFT + K to open the Web console (COMMAND + SHIFT + K on Macs).
- or, if Firebug is installed (recommended):
    1. Press F12 to open Firebug.
    2. Click on the “console” tab.

# Opera
1. Press CTRL + SHIFT + I to open Dragonfly.
2. Click on the “console” tab.
</details>

**Additional context**

Chartman123:
I found a nextcloud-vue issue that sounds a lot like this, but should be fixed already for some time.

https://github.com/nextcloud-libraries/nextcloud-vue/issues/3231

PR #2015, opened 2024-03-19
fix(TopBar): provide optional button text via new slot syntax

ShGKme:
* Fix: https://github.com/nextcloud/forms/issues/1699

This is an issue on `NcButton` side, but can also be fixed here. Passing conditional content always works better with explicit no-passing slot completely.

It is an old known issue in Vue 2 with deprecated slot syntax and checking `this.$slots`. Because there is a content passed via old slot syntax (not using `v-slot` directive), it handles `$slots` differently. It has icon rendering function in `$scopedSlots`, but the rendering result does not exist in `$slots`.

Unfortunately, I don't remember the Vue 2 issue link...

### DIFF
--- a/src/components/TopBar.vue
+++ b/src/components/TopBar.vue
@@ -33,7 +33,7 @@
 				<template #icon>
 					<IconEye :size="20" />
 				</template>
-				<template v-if="!isMobile">
+				<template v-if="!isMobile" #default>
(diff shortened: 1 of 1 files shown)


Chartman123 (approved):
Can confirm that this fixes it :) Thank you very much @ShGKme!

PR #2029, opened 2024-03-23
Get only forms shared with user from database

Chartman123:


susnux:
This does not work for group shares, team shares etc :(

Chartman123:
Ok... yes, I didn't think about that... and for direct shares I'd need to add share_type = 0...

But wouldn't it be possible to resolve groups/teams with the current user and then also add this to DB query?

Chartman123:
@susnux I think we could use `getUserGroups()` for the groups and `probeCircles()` for teams. and then check if the entries in the shares table for groups/teams are within the returned groups/teams. What do you think?

lib/Db/FormMapper.php
```diff
@@ -95,16 +96,56 @@ public function findByHash(string $hash): Form {
...
+			);
+		}
+		// public forms
+		if ($publicForms) {
+			$memberships->add($qb->expr()->like('forms.access_json', $qb->createNamedParameter('%"showToAllUsers":true%')));
+		}
+
+		$qb->select('forms.*')
+			->from($this->getTableName(), 'forms')
+			->leftJoin('forms', 'forms_v2_shares', 'shares', $qb->expr()->eq('forms.id', 'shares.form_id'))
```
susnux, line 141: I am not a DB expert, so the question if an inner join would be faster?

Chartman123: We need the left join because we need all rows from the forms table and not only those where we also have an entry in the shares table.

susnux: Makes sense! I was thinking like "get the inner join on those conditions + forms where access is..." but probably thats either very complex or not possible in one sql query.

lib/Db/FormMapper.php
```diff
@@ -95,16 +96,56 @@ public function findByHash(string $hash): Form {
...
+			$memberships->add(
+				$qb->expr()->andX(
+					$qb->expr()->eq('shares.share_type', $qb->createNamedParameter(IShare::TYPE_CIRCLE)),
+					$qb->expr()->in('shares.share_with', $teams),
+				),
+			);
+		}
+		// public forms
+		if ($publicForms) {
+			$memberships->add($qb->expr()->like('forms.access_json', $qb->createNamedParameter('%"showToAllUsers":true%')));
```
susnux, line 136: I do not know why this was made a JSON field, but I think performance wise it is worse than adding two columns to the table, no?
Meaning everything that we filter on DB level makes more sense as real column, no?

lib/Db/FormMapper.php
Chartman123: But do we really need to put it into one statement? We could also do several queries and collect all the results...

1) Shared with all users
2) Shared directly with one user
3) Shared with a group
4) If enabled, shared with a team

lib/Db/FormMapper.php
Chartman123: At least until every DB can handle JSON fields directly...

lib/Db/FormMapper.php
susnux: I think this is the purpose of a relational database, to handle complex queries on the database and just return the results. (hopefully) they are even optimized for this.

lib/Db/FormMapper.php
susnux: yes, this is more like a comment. I think this is fine for now.

lib/Service/FormsService.php
```diff
@@ -383,37 +383,40 @@ public function hasUserAccess(Form $form): bool {
...
+		);
+
+		$publicForms = [];
+		if ($this->configService->getAllowPermitAll()) {
+			$publicForms = $this->formMapper->findPublicForms($this->currentUser->getUID());
+		}
+
+		$forms = array_merge($forms, $publicForms);
+		$forms = array_filter($forms, fn (Form $form): bool => $this->shouldShowFormAsShared($form));
+		return $forms;
```
Chartman123, line 404: Should we do another ordering of the forms here before giving them back to the APIController? Otherwise they will be sorted this way: latest to oldest directly shared forms, latest to oldest publicly shared forms.

lib/Service/FormsService.php
```diff
@@ -383,37 +383,40 @@ public function hasUserAccess(Form $form): bool {
...
 	 * @param Form $form
 	 * @return bool
 	 */
-	public function isSharedFormShown(Form $form): bool {
-		// Dont show here to owner, as its in the owned list anyways.
-		if ($form->getOwnerId() === $this->currentUser->getUID()) {
-			return false;
-		}
-
+	private function shouldShowFormAsShared(Form $form): bool {
```
Chartman123, line 413: I'd keep the old naming here...

lib/Service/FormsService.php
Chartman123: On the other hand the sorting would probably be faster on the DB. So should we still merge the DB queries into one?

Chartman123: Ok, I found a way to put into one (and a half) queries. Doing it with selectDistinct didn't work for OCI because one of the result columns obviously is a CLOB that doesn't work with distinct...

susnux (approved):

Chartman123:
Thanks for the review. I'm gonna squash the commits and merge it then :)

PR #2040, opened 2024-04-01
Add support for file question

Koc:
This PR introduces possibility add new File Question type:
<details><summary>Preview</summary>

image

image

image

image

image

image

image

image

image

image

image

image

image

</details>

This is how it works under the hood:
1. when user select file during form fill - we're uploading it to form owner's folder `forms/unsubmitted/{formId}. {formName}/{questionId}. {questionName}` and store inside `oc_forms_v2_uploaded_files` table.
2. once form submitted we are moving file to `forms/{formId}. {formName}/{submissionId}/{questionId}. {questionName}`, store original file name as `oc_forms_v2_answers.text` and reference to uploaded file as `oc_forms_v2_answers.file_id`. And remove row from `oc_forms_v2_uploaded_files`

There are validations for:
- file mime types/extensions
- max file size
- max files count

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -55,6 +55,10 @@
 		<nextcloud min-version="28" max-version="29" />
 	</dependencies>
 
+	<background-jobs>
+		<job>OCA\Forms\BackgroundJob\CleanupUploadedFilesJob</job>
+	</background-jobs>
+
 	<settings>
 		<admin>OCA\Forms\Settings\Settings</admin>
 		<admin-section>OCA\Forms\Settings\SettingsSection</admin-section>
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -332,6 +332,14 @@ return [
 			]
 		],
 		[
+			'name' => 'api#uploadFiles',
+			'url' => '/api/{apiVersion}/uploadFiles/{formId}/{questionId}',
+			'verb' => 'POST',
+			'requirements' => [
+				'apiVersion' => 'v2.5'
+			]
+		],
+		[
 			'name' => 'api#insertSubmission',
 			'url' => '/api/{apiVersion}/submission/insert',
 			'verb' => 'POST',
--- a/lib/BackgroundJob/CleanupUploadedFilesJob.php
+++ b/lib/BackgroundJob/CleanupUploadedFilesJob.php
@@ -1,0 +1,112 @@
+<?php
+/**
+ * @copyright Copyright (c) 2024 Kostiantyn Miakshyn <molodchick@gmail.com>
+ *
+ * @author Kostiantyn Miakshyn <molodchick@gmail.com>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Forms\BackgroundJob;
+
+use OCA\Forms\Constants;
+use OCA\Forms\Db\FormMapper;
+use OCA\Forms\Db\UploadedFileMapper;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\TimedJob;
+use OCP\Files\Folder;
+use OCP\Files\IRootFolder;
+use Psr\Log\LoggerInterface;
+
+class CleanupUploadedFilesJob extends TimedJob {
+	private const FILE_LIFETIME = '-1 hour';
+
+	public function __construct(
+		private IRootFolder $storage,
+		private FormMapper $formMapper,
+		private UploadedFileMapper $uploadedFileMapper,
+		private LoggerInterface $logger,
+		ITimeFactory $time) {
+		parent::__construct($time);
+
+		$this->setInterval(60 * 60);
+	}
+
+	/**
+	 * @param array $argument
+	 */
+	public function run($argument): void {
+		$dateTime = new \DateTimeImmutable(self::FILE_LIFETIME);
+
+		$this->logger->info('Deleting files that were uploaded before {before} and still not submitted.', [
+			'before' => $dateTime->format(\DateTimeImmutable::ATOM),
+		]);
+
+		$uploadedFiles = $this->uploadedFileMapper->findUploadedEarlierThan($dateTime);
+
+		$deleted = 0;
+		$usersToCleanup = [];
+		foreach ($uploadedFiles as $uploadedFile) {
+			$this->logger->info('Deleting uploaded file "{originalFileName}" for form {formId}.', [
+				'originalFileName' => $uploadedFile->getOriginalFileName(),
+				'formId' => $uploadedFile->getFormId(),
+			]);
+
+			$form = $this->formMapper->findById($uploadedFile->getFormId());
+			$usersToCleanup[$form->getOwnerId()] = true;
+			$userFolder = $this->storage->getUserFolder($form->getOwnerId());
+
+			$nodes = $userFolder->getById($uploadedFile->getFileId());
+
+			if (!empty($nodes)) {
+				$node = $nodes[0];
+				$node->delete();
+			} else {
+				$this->logger->warning('Could not find uploaded file "{fileId}" for deletion.', [
+					'fileId' => $uploadedFile->getFileId(),
+				]);
+			}
+
+			$this->uploadedFileMapper->delete($uploadedFile);
+
+			$deleted++;
+		}
+
+		$this->logger->info('Deleted {deleted} uploaded files.', ['deleted' => $deleted]);
+
+		// now delete empty folders in user folders
+		$deleted = 0;
+		foreach (array_keys($usersToCleanup) as $userId) {
+			$this->logger->info('Cleaning up empty folders for user {userId}.', ['userId' => $userId]);
+			$userFolder = $this->storage->getUserFolder($userId);
+
+			$unsubmittedFilesFolder = $userFolder->get(Constants::UNSUBMITTED_FILES_FOLDER);
+			if (!$unsubmittedFilesFolder instanceof Folder) {
+				continue;
+			}
+
+			foreach ($unsubmittedFilesFolder->getDirectoryListing() as $node) {
+				if ($node->getName() < $dateTime->getTimestamp()) {
+					$node->delete();
+					$deleted++;
+				}
+			}
+		}
+
+		$this->logger->info('Deleted {deleted} folders.', ['deleted' => $deleted]);
+	}
+}
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -91,6 +91,7 @@ class Constants {
 	public const ANSWER_TYPE_DATE = 'date';
 	public const ANSWER_TYPE_DATETIME = 'datetime';
 	public const ANSWER_TYPE_TIME = 'time';
+	public const ANSWER_TYPE_FILE = 'file';
 
 	// All AnswerTypes
 	public const ANSWER_TYPES = [
@@ -101,7 +102,8 @@ class Constants {
 		self::ANSWER_TYPE_LONG,
 		self::ANSWER_TYPE_DATE,
 		self::ANSWER_TYPE_DATETIME,
-		self::ANSWER_TYPE_TIME
+		self::ANSWER_TYPE_TIME,
+		self::ANSWER_TYPE_FILE,
 	];
 
 	// AnswerTypes, that need/have predefined Options
@@ -155,6 +157,21 @@ class Constants {
 		'validationRegex' => ['string'],
 	];
 
+	public const EXTRA_SETTINGS_FILE = [
+		'allowedFileTypes' => ['array'],
+		'allowedFileExtensions' => ['array'],
+		'maxAllowedFilesCount' => ['integer'],
+		'maxFileSize' => ['integer'],
+	];
+
+	// should be in sync with FileTypes.js
+	public const EXTRA_SETTINGS_ALLOWED_FILE_TYPES = [
+		'image',
+		'x-office/document',
+		'x-office/presentation',
+		'x-office/spreadsheet',
+	];
+
 	/**
 	 * !! Keep in sync with src/mixins/ShareTypes.js !!
 	 */
@@ -204,4 +221,8 @@ class Constants {
 	];
 
 	public const DEFAULT_FILE_FORMAT = 'csv';
+
+	public const UNSUBMITTED_FILES_FOLDER = self::FILES_FOLDER . '/unsubmitted';
+
+	public const FILES_FOLDER = 'Forms';
 }
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -40,6 +40,8 @@ use OCA\Forms\Db\QuestionMapper;
 use OCA\Forms\Db\ShareMapper;
 use OCA\Forms\Db\Submission;
 use OCA\Forms\Db\SubmissionMapper;
+use OCA\Forms\Db\UploadedFile;
+use OCA\Forms\Db\UploadedFileMapper;
 use OCA\Forms\Service\ConfigService;
 use OCA\Forms\Service\FormsService;
 use OCA\Forms\Service\SubmissionService;
@@ -48,10 +50,13 @@ use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\IMapperException;
 use OCP\AppFramework\Http\DataDownloadResponse;
 use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\Http\Response;
 use OCP\AppFramework\OCS\OCSBadRequestException;
 use OCP\AppFramework\OCS\OCSForbiddenException;
 use OCP\AppFramework\OCS\OCSNotFoundException;
 use OCP\AppFramework\OCSController;
+use OCP\Files\IMimeTypeDetector;
+use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 use OCP\IL10N;
 use OCP\IRequest;
@@ -81,6 +86,9 @@ class ApiController extends OCSController {
 		private IL10N $l10n,
 		private LoggerInterface $logger,
 		private IUserManager $userManager,
+		private IRootFolder $storage,
+		private UploadedFileMapper $uploadedFileMapper,
+		private IMimeTypeDetector $mimeTypeDetector,
 	) {
 		parent::__construct($appName, $request);
 		$this->currentUser = $userSession->getUser();
@@ -438,6 +446,7 @@ class ApiController extends OCSController {
 
 		$response = $question->read();
 		$response['options'] = [];
+		$response['accept'] = [];
 
 		$this->formsService->setLastUpdatedTimestamp($formId);
 
@@ -696,6 +705,7 @@ class ApiController extends OCSController {
 
 		$response = $newQuestion->read();
 		$response['options'] = [];
+		$response['accept'] = [];
 
 		foreach ($sourceOptions as $sourceOption) {
 			$optionData = $sourceOption->read();
@@ -918,14 +928,19 @@ class ApiController extends OCSController {
 	/**
 	 * Insert answers for a question
 	 *
+	 * @param Form $form
 	 * @param int $submissionId
 	 * @param array $question
-	 * @param array $answerArray [arrayOfString]
+	 * @param string[]|array<array{uploadedFileId: string, uploadedFileName: string}> $answerArray
 	 */
-	private function storeAnswersForQuestion($submissionId, array $question, array $answerArray) {
+	private function storeAnswersForQuestion(Form $form, $submissionId, array $question, array $answerArray) {
 		foreach ($answerArray as $answer) {
-			$answerText = '';
+			$answerEntity = new Answer();
+			$answerEntity->setSubmissionId($submissionId);
+			$answerEntity->setQuestionId($question['id']);
 
+			$answerText = '';
+			$uploadedFile = null;
 			// Are we using answer ids as values
 			if (in_array($question['type'], Constants::ANSWER_TYPES_PREDEFINED)) {
 				// Search corresponding option, skip processing if not found
@@ -935,6 +950,25 @@ class ApiController extends OCSController {
 				} elseif (!empty($question['extraSettings']['allowOtherAnswer']) && strpos($answer, Constants::QUESTION_EXTRASETTINGS_OTHER_PREFIX) === 0) {
 					$answerText = str_replace(Constants::QUESTION_EXTRASETTINGS_OTHER_PREFIX, "", $answer);
 				}
+			} elseif ($question['type'] === Constants::ANSWER_TYPE_FILE) {
+				$uploadedFile = $this->uploadedFileMapper->getByUploadedFileId($answer['uploadedFileId']);
+				$answerEntity->setFileId($uploadedFile->getFileId());
+
+				$userFolder = $this->storage->getUserFolder($form->getOwnerId());
+				$path = $this->formsService->getUploadedFilePath($form, $submissionId, $question['id'], $question['name'], $question['text']);
+
+				if ($userFolder->nodeExists($path)) {
+					$folder = $userFolder->get($path);
+				} else {
+					$folder = $userFolder->newFolder($path);
+				}
+				/** @var \OCP\Files\Folder $folder */
+
+				$file = $userFolder->getById($uploadedFile->getFileId())[0];
+				$name = $folder->getNonExistingName($file->getName());
+				$file->move($folder->getPath() . '/' . $name);
+
+				$answerText = $name;
 			} else {
 				$answerText = $answer; // Not a multiple-question, answerText is given answer
 			}
@@ -943,12 +977,130 @@ class ApiController extends OCSController {
 				continue;
 			}
 
-			$answerEntity = new Answer();
-			$answerEntity->setSubmissionId($submissionId);
-			$answerEntity->setQuestionId($question['id']);
 			$answerEntity->setText($answerText);
 			$this->answerMapper->insert($answerEntity);
+			if ($uploadedFile) {
+				$this->uploadedFileMapper->delete($uploadedFile);
+			}
 		}
+	}
+
+	/**
+	 * @CORS
+	 * @NoAdminRequired
+	 * @PublicPage
+	 *
+	 * Uploads a temporary files to the server during form filling
+	 *
+	 * @return Response
+	 */
+	public function uploadFiles(int $formId, int $questionId, string $shareHash = ''): Response {
+		$this->logger->debug('Uploading files for formId: {formId}, questionId: {questionId}',
+			['formId' => $formId, 'questionId' => $questionId]);
+
+		$uploadedFiles = [];
+		foreach ($this->request->getUploadedFile('files') as $key => $files) {
+			foreach ($files as $i => $value) {
+				$uploadedFiles[$i][$key] = $value;
+			}
+		}
+
+		if (!count($uploadedFiles)) {
+			throw new OCSBadRequestException('No files provided');
+		}
+
+		$form = $this->loadFormForSubmission($formId, $shareHash);
+
+		if (!$this->formsService->canSubmit($form)) {
+			throw new OCSForbiddenException('Already submitted');
+		}
+
+		try {
+			$question = $this->questionMapper->findById($questionId);
+		} catch (IMapperException $e) {
+			$this->logger->debug('Could not find question with id {questionId}', ['questionId' => $questionId]);
+			throw new OCSBadRequestException(previous: $e instanceof \Exception ? $e : null);
+		}
+
+		$path = $this->formsService->getTemporaryUploadedFilePath($form, $question);
+
+		$response = [];
+		foreach ($uploadedFiles as $uploadedFile) {
+			$error = $uploadedFile['error'] ?? 0;
+			if ($error !== UPLOAD_ERR_OK) {
+				$this->logger->error('Failed to get the uploaded file. PHP file upload error code: ' . $error,
+					['file_name' => $uploadedFile['name']]);
+
+				throw new OCSBadRequestException(sprintf('Failed to upload the file "%s".', $uploadedFile['name']));
+			}
+
+			if (!is_uploaded_file($uploadedFile['tmp_name'])) {
+				throw new OCSBadRequestException('Invalid file provided');
+			}
+
+			$userFolder = $this->storage->getUserFolder($form->getOwnerId());
+			$userFolder->getStorage()->verifyPath($path, $uploadedFile['name']);
+
+			$extraSettings = $question->getExtraSettings();
+			if (($extraSettings['maxFileSize'] ?? 0) > 0 && $uploadedFile['size'] > $extraSettings['maxFileSize']) {
+				throw new OCSBadRequestException(sprintf('File size exceeds the maximum allowed size of %s bytes.', $extraSettings['maxFileSize']));
+			}
+
+			if (!empty($extraSettings['allowedFileTypes']) || !empty($extraSettings['allowedFileExtensions'])) {
+				$mimeType = $this->mimeTypeDetector->detectContent($uploadedFile['tmp_name']);
+				$aliases = $this->mimeTypeDetector->getAllAliases();
+
+				$valid = false;
+				foreach ($extraSettings['allowedFileTypes'] ?? [] as $allowedFileType) {
+					if (str_starts_with($aliases[$mimeType] ?? '', $allowedFileType)) {
+						$valid = true;
+						break;
+					}
+				}
+
+				if (!$valid && !empty($extraSettings['allowedFileExtensions'])) {
+					$mimeTypesPerExtension = method_exists($this->mimeTypeDetector, 'getAllMappings')
+						? $this->mimeTypeDetector->getAllMappings() : [];
+					foreach ($extraSettings['allowedFileExtensions'] as $allowedFileExtension) {
+						if (isset($mimeTypesPerExtension[$allowedFileExtension])
+							&& in_array($mimeType, $mimeTypesPerExtension[$allowedFileExtension])) {
+							$valid = true;
+							break;
+						}
+					}
+				}
+
+				if (!$valid) {
+					throw new OCSBadRequestException(sprintf('File type is not allowed. Allowed file types: %s',
+						implode(', ', array_merge($extraSettings['allowedFileTypes'] ?? [], $extraSettings['allowedFileExtensions'] ?? []))
+					));
+				}
+			}
+
+			if ($userFolder->nodeExists($path)) {
+				$folder = $userFolder->get($path);
+			} else {
+				$folder = $userFolder->newFolder($path);
+			}
+			/** @var \OCP\Files\Folder $folder */
+
+			$fileName = $folder->getNonExistingName($uploadedFile['name']);
+			$file = $folder->newFile($fileName, file_get_contents($uploadedFile['tmp_name']));
+
+			$uploadedFileEntity = new UploadedFile();
+			$uploadedFileEntity->setFormId($formId);
+			$uploadedFileEntity->setOriginalFileName($fileName);
+			$uploadedFileEntity->setFileId($file->getId());
+			$uploadedFileEntity->setCreated(time());
+			$this->uploadedFileMapper->insert($uploadedFileEntity);
+
+			$response[] = [
+				'uploadedFileId' => $uploadedFileEntity->getId(),
+				'fileName' => $fileName,
+			];
+		}
+
+		return new DataResponse($response);
 	}
 
 	/**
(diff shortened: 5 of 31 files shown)


lib/Controller/ApiController.php
```diff
@@ -28,6 +28,7 @@
 
 namespace OCA\Forms\Controller;
 
+use OC\Files\Filesystem;
```
Chartman123, line 31: @susnux Am I correct that we shouldn't use things from OC namespace?

susnux: Yes, good point. `OC` is private and should not be used in apps. Only use `OCP` API.

susnux:
> While using Nextcloud's webdav would be pretty simple for this, I am not that sure how to do authentication there - so without any research a custom endpoint is currently probably easier...

What I could imagine is creating shares for the forms folders with upload only permission (or upload+view when having form-results permissions). Then one could directly upload to that folder using webdav.

Koc:
Mates, I've see one problem with multiple files: there is no possibility to set multiple hyperlinks inside single cell. see https://answers.microsoft.com/en-us/msoffice/forum/all/multiple-hyperlinks-in-one-cell-excel/7af82648-b7a0-4c9c-b4d7-c485b256dc06 for the reference. Even if it possible with LibreOffice Calc, PhpSpreadsheet does not supports this. 

Any ideas how to workaround that?

Koc:
> I am not sure why we need a new table? Would it be not sufficent to just store the filename in the answers? As everything else is already stored in Nextcloud's file database.

@susnux the problem is that we need upload file and store it id somewhere before form submission. Of course, I can use only `\OCP\Files\Node::getId` but:
- it's not safe from security perspective. User can provide arbitrary fileId during form submission as fileId is simple integer. Uuid fixes this problem
- we want have original user's file name
- would be nice to cleanup uploaded files for unsubmitted forms after some delay

Chartman123:
> Any ideas how to workaround that?

What about creating a subfolder per submission and then just linking the folder instead of the single files?

Koc:
Can anybody help me with styles? :pray: 
image
image

lib/Constants.php
```diff
@@ -135,6 +137,73 @@ class Constants {
...
+		'maxFileSizeValue',
+		'maxFileSizeUnit',
+	];
+
+	// should be in sync with
+	public const EXTRA_SETTINGS_ALLOWED_FILE_TYPES = [
+		'document',
+		'presentation',
+		'spreadsheet',
+		'drawing',
```
Koc, line 153: do we really need drawing as we've already have image?

lib/Constants.php
```diff
@@ -135,6 +137,73 @@ class Constants {
...
+		'pdf',
+		'image',
+		'video',
+		'audio',
+	];
+
+	public const EXTRA_SETTINGS_ALLOWED_FILE_TYPES_MIME = [
+		'document' => [
```
Koc, line 161: can anybody review list of allowed mime types per file types and suggest changes?

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,163 @@
...
+		</div>
+	</Question>
+</template>
+
+<script>
+import QuestionMixin from '../../mixins/QuestionMixin.js'
+import axios from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+import OcsResponse2Data from '../../utils/OcsResponse2Data.js'
+import NcActionCaption from "@nextcloud/vue/dist/Components/NcActionCaption.js";
```
Koc, line 89: why we haven't auto-import of the components, like in [NuxtJS](https://nuxt.com/docs/guide/concepts/auto-imports)? :thinking:

lib/Constants.php
Chartman123: I don't know exactly what the difference between image and drawing is for Google. This was just an example. So for me it would be totally fine to only have image :)

Chartman123:
> Can anybody help me with styles? 🙏 

pinging @jancborchardt Do you have a preferred way to do this design wise? 

I think we should at least have it in submenus like for the file linking. Another possibility would be a NcModal...

jancborchardt:
Yes, I would say go with a modal here since it's quite a bunch of different settings. The small dropdown is too cramped of a space for that.

And agree on that we should go with the actual file names, and also "Forms/Form name", without UUID. Since this will be visible in the file system there should be no IDs visible – if the name exists, we can append " (2)" and count up like we do in Files and elsewhere too.

jancborchardt (commented):
Yes, I would say go with a modal here since it's quite a bunch of different settings. The small dropdown is too cramped of a space for that.

And agree on that we should go with the actual file names, and also "Forms/Form name", without UUID. Since this will be visible in the file system there should be no IDs visible – if the name exists, we can append " (2)" and count up like we do in Files and elsewhere too.

@Koc and also, awesome feature! That will be very useful. :)

susnux:
> * it's not safe from security perspective. User can provide arbitrary fileId during form submission as fileId is simple integer. Uuid fixes this problem

Why should the UUID be safe? It is just obfuscation.
If you see a security issue here please inform me, but for me it should be pretty ok to have it by fileid, we just need to check it is located under the owner's `forms/{form}`.

> * we want have original user's file name

This one would be preserved, there is no need to rename the file at all, see the suggestion of subdirs for submissions.

> * would be nice to cleanup uploaded files for unsubmitted forms after some delay

Would also be possible.


---

Given `{user}/{formsfolder}/{formid}` we can create a folder `{submissionid}`, this folder holds all submission files.

The process could be:
1. Upload the files (`{user}/{formsfolder}/{formid}/{some temporary name}`)
2. Submit the form
3. The temporary files are renamed into the submission id
(4.) If not submitted a background job removes all temporary files

Koc:
@susnux 
> Why should the UUID be safe? It is just obfuscation. If you see a security issue here please inform me,

As `fileId` is a just integer - I can brutforce them and send `fileId` of File which was uploaded by another person to `insertSubmission`

Chartman123:
@Koc I don't know much about that, but I found this in the dev docs: https://docs.nextcloud.com/server/stable/developer_manual/basics/storage/appdata.html

susnux:
> I'd prefer app data

BTW this could be exploited by create own forms and upload files there to create files even if the quota is exceeded. Not sure if this would be a problem.

---

An other problem with app data would be that it does not allow any integration out of the box, e.g. someone submits a file -> you can only access it within the forms app with custom handling.
But you could not integrate it in workflows or use it in other apps.
To enable this integration we would need a mount of the appdata, like collectives is doing.

susnux:
But I also think app data is probably the best location as that data is not directly connected to an user (because in the future a form might be collaborative).
And we might want to also use it in the future for e.g. form attachments (embedded images or similar).

Chartman123:
> To enable this integration we would need a mount of the appdata, like collectives is doing.

Yes, that's what I meant, use app data folder and mountpoint Forms in root folder

lib/Controller/ApiController.php
```diff
@@ -1017,14 +1050,134 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+		$path = $this->formsService->getTemporaryUploadedFilePath($form, $question);
+
+		$response = [];
+		foreach ($uploadedFiles as $uploadedFile) {
+			$error = $uploadedFile['error'] ?? 0;
+			if ($error !== UPLOAD_ERR_OK) {
+				$this->logger->error('Failed to get the uploaded file. PHP file upload error code: ' . $error,
+					['file_name' => $uploadedFile['name']]);
+
+				return new JSONResponse(
```
Koc, line 1102: I would like to throw an `OCSBadRequestException` here, but seems like exceptions not supports translations and file parameters. So, should I pass already translated strings into `OCSBadRequestException`?

Chartman123: AFAIK the exceptions are always sent in English...

susnux: On server core we mostly use English -> Only for dev. And use the response code to show a translated error message to the user.

Some rare cases use translated error messages and directly show them to the user.
But I would say keep the error in English here.

Koc: That's not user friendly and leads to inconsistency :man_shrugging:

I would like pass translated strings now and improve DX later by introduction of something like:

```php
OCSBadRequestException::fromTranslation('File size exceeds the maximum allowed size of {maxFileSizeValue} {maxFileSizeUnit}.',
    [
        'maxFileSizeValue' => $extraSettings['maxFileSizeValue'], 
        'maxFileSizeUnit' => $extraSettings['maxFileSizeUnit',
    ]]);
```

Koc:
Mates, I've updated PR description with new screenshoots. Can we keep UI design as is for the first iteration?

I will try to use app folder, but it requires also adding mount. I have no experience in that :worried:

lib/Controller/ApiController.php
Chartman123: The parameters could be a problem, but the translation could be done in the frontend by reacting to the English error sent by the server like @susnux suggested.

Chartman123:
> Can we keep UI design as is for the first iteration?

What about putting the menu entries into two submenus? Otherwise this menu gets very long...

lib/Controller/ApiController.php
Koc: as you've already mentioned it will not work with parameters in translations. Plus I guess it's makes harder to extract translations from the code (because we will call `this.t(variableName)` instead of `this.t('some_plain_string')`

susnux: > That's not user friendly and leads to inconsistency 🤷‍♂️

Why would it?
Simply thrown an meaningful error code and show the error message on the frontend, the frontend should decide how to handle it.
This is purely API and the API consistently using English for status messages (as this are only meant for debugging).

See all other API functions?

lib/Controller/ApiController.php
```diff
@@ -1017,12 +1050,132 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+			if ($uploadedFile) {
+				$this->uploadedFileMapper->delete($uploadedFile);
+			}
+		}
+	}
+
+
+	/**
+	 * @CORS
+	 * @PublicCORSFix
```
susnux, line 1064: This does not exist anymore

lib/Controller/ApiController.php
```diff
@@ -1017,12 +1050,132 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+	 * @PublicPage
+	 *
+	 * Uploads a temporary files to the server during form filling
+	 *
+	 * @return Response
+	 */
+	public function uploadFiles(int $formId, int $questionId): Response {
+		$this->logger->debug('Uploading files for formId: {formId}, questionId: {questionId}',
+			['formId' => $formId, 'questionId' => $questionId]);
+
```
susnux, line 1075: *First* check that:
1. the form exists
2. the user has permissions to answer the form
3. the question exists

lib/Controller/ApiController.php
```diff
@@ -81,6 +89,9 @@ public function __construct(
 		private IL10N $l10n,
 		private LoggerInterface $logger,
 		private IUserManager $userManager,
+		private IRootFolder $storage,
+		private UploadedFileMapper $uploadedFileMapper,
+		private MimeTypes $mimeTypes,
```
susnux, line 94: No please do not use Symphony here directly this is very prone to break dependencies with Server core.

Instead use OCP\Files\IMimeTypeDetector

lib/Controller/ApiController.php
susnux: So even if you want to provide user feedback *what* is invalid, I think this should be done on the frontend **before** sending it

lib/Controller/ApiController.php
```diff
@@ -1017,12 +1050,132 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+
+			$uploadedFileId = Uuid::uuid4()->toString();
+			$uploadedFileEntity = new UploadedFile();
+			$uploadedFileEntity->setId($uploadedFileId);
+			$uploadedFileEntity->setFormId($formId);
+			$uploadedFileEntity->setOriginalFileName($fileName);
+			$uploadedFileEntity->setFileId($file->getId());
+			$uploadedFileEntity->setCreated(time());
+			$this->uploadedFileMapper->insert($uploadedFileEntity);
+
```
susnux, line 1171: I still do not think using the owner directory is a good decision, because it makes accessing the files much harder when we have shared results permissions and collaborative editing.

Instead I think we should use appdata like this:
```
(appdata forms)
  |--- forms
        |--- {form_id}
              |-- submissions
              |   |
              |   |--- {submission_id}
              |        |
              |        |--- {original file name}
              |
              |--- (this could be later used for form attachments)
```

susnux: I asked some other app developers for feedback, the common response:
Use user folder.

So I revert my comment partially.

Chartman123: > I asked some other app developers for feedback, the common response: Use user folder.

Did they also tell you why? Just curious

susnux: 1. Feedback from users on similar features that otherwise that does not count into user quota -> can be exploited
2. Security risks, there were issues in the past with e.g. Talk attachments being able to accessed by wrong users. Having it in the user directory prevents this.

susnux: I mean if we do it right we still can do it in appdata, this was just some feedback from dev that already did similar in the past.

lib/Controller/ApiController.php
Koc: done

lib/Controller/ApiController.php
Koc: done, I've extracted private method to reuse some checks

lib/Controller/ApiController.php
```diff
@@ -1017,12 +1050,132 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+				return new JSONResponse(
+					['data' => ['message' => $this->l10n->t('Invalid file provided')]],
+					Http::STATUS_BAD_REQUEST
+				);
+			}
+
+			$userFolder->getStorage()->verifyPath($path, $uploadedFile['name']);
+
+			$extraSettings = $question->getExtraSettings();
+			if ($extraSettings['maxFileSizeValue'] > 0 && $extraSettings['maxFileSizeUnit']) {
```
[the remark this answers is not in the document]
Koc: it makes sense. I've reworked storing of `maxFileSize` to use bytes under the hood but keep UI as it now

lib/Controller/ApiController.php
Koc: so, what's the final decision? AppData + mount or keep UserData as a temporary solution?

lib/Constants.php
```diff
@@ -153,6 +155,88 @@ class Constants {
...
+		'document',
+		'presentation',
+		'spreadsheet',
+		'pdf',
+		'image',
+		'video',
+		'audio',
+	];
+
+	public const EXTRA_SETTINGS_ALLOWED_FILE_EXTENSIONS = [
```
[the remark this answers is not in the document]
Koc: I will try to rework this part. We can open extra PR to add `pdf`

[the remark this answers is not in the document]
Koc: how do you see input of custom mime-types to upload from UI perspective? Textares or multuple inputs?

Koc:
I've updated UI for file types management. Please check

image
image
image

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,305 @@
...
+  -->
+
+<template>
+	<Question v-bind="questionProps"
+		:title-placeholder="answerType.titlePlaceholder"
+		:warning-invalid="answerType.warningInvalid"
+		v-on="commonListeners">
+		<template #actions>
+			<template v-if="!allowedFileTypesDialogOpened">
+				<NcActionLink @click.prevent="allowedFileTypesDialogOpened = true">
```
Chartman123, line 30: Why don't you use NcActionButtons here? Then you could use `:is-menu="true"` to make it even clearer that there's a submenu behind it. I think NcActionLink is more intended for real links... Same of course for the button to go back again further down in your code.

Chartman123:
> I've updated UI for file types management

Yes, that's better :)

src/components/Questions/QuestionFile.vue
Koc: done :ok_hand: 
image

Koc:
I've added a cron job that will cleanup files that were uploaded more than 2 days ago but still not submitted. And also there is a validation that file is exists during submission, here an example of the error message:

image

Koc:
I've updated docs and fixed Psalm issues. Tests on my way. Ready to another round of review, removed draft/wip labels.

@Chartman123 @susnux mates, I would like to ask you keep user data for the 1st iteration and improve this part in upcoming PRs :pray::
- in our organization we have very cheap and small server with tiny storage. That's why we prefer limit storage size per user and prevent growing of app data without control
- we are urgently need this feature asap (yesterday :smile: ). Another circle of development can significantly delay merge.

in ideal word we should add possibility to configure how to store attachments (in app data or in user data) via config.php or via form settings. But this can be done via separate PR

Koc:
Awesome, we have completely green pipeline tonight :partying_face:

lib/Migration/Version040010Date20240401123456.php
```diff
@@ -0,0 +1,80 @@
+<?php
```
Chartman123, line 1: You should always check if the tables/columns don't exist before creating them :)

Chartman123 (commented):
I didn't check all the code by now... So just a few things that came into my mind when I had a quick look.

appinfo/info.xml
```diff
@@ -13,7 +13,7 @@
 - **🙋 Get involved!** We have lots of stuff planned like more question types, collaboration on forms, [and much more](https://github.com/nextcloud/forms/milestones)!
 	]]></description>
 
-	<version>4.2.3</version>
+	<version>4.2.4</version>
```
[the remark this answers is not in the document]
Koc: done, reverted

lib/Migration/Version040010Date20240401123456.php
Koc: added checks, please review one more time

appinfo/routes.php
```diff
@@ -331,6 +331,14 @@
 				'apiVersion' => 'v2(\.[1-4])?'
 			]
 		],
+		[
+			'name' => 'api#uploadFiles',
+			'url' => '/api/{apiVersion}/uploadFiles/{formId}/{questionId}',
```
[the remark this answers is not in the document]
Koc: should I change urls now or keep as is?

lib/Controller/ApiController.php
```diff
@@ -943,12 +973,135 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
 			$answerEntity->setText($answerText);
 			$this->answerMapper->insert($answerEntity);
+			if ($uploadedFile) {
+				$this->uploadedFileMapper->delete($uploadedFile);
+			}
+		}
+	}
+
+
+	/**
```
Chartman123, line 985:
```suggestion
	}

	/**
```

lib/Controller/ApiController.php
```diff
@@ -943,12 +973,135 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+		$path = $this->formsService->getTemporaryUploadedFilePath($form, $question);
+
+		$response = [];
+		foreach ($uploadedFiles as $uploadedFile) {
+			$error = $uploadedFile['error'] ?? 0;
+			if ($error !== UPLOAD_ERR_OK) {
+				$this->logger->error('Failed to get the uploaded file. PHP file upload error code: ' . $error,
+					['file_name' => $uploadedFile['name']]);
+
+				throw new OCSBadRequestException($this->l10n->t('Failed to upload the file "%s".', [$uploadedFile['name']]));
```
Chartman123, line 1031: I'm still unsure if we really should send translated error messages to the client...

lib/Controller/ApiController.php
```diff
@@ -943,12 +973,135 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+			}
+
+			if (!is_uploaded_file($uploadedFile['tmp_name'])) {
+				throw new OCSBadRequestException($this->l10n->t('Invalid file provided'));
```
Chartman123, line 1035: Same here...

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,304 @@
+<!--
+- @copyright Copyright (c) 2024 Kostiantyn Miakshyn <molodchick@gmail.com>
+-
+- @author Kostiantyn Miakshyn <molodchick@gmail.com>
```
Chartman123, line 4:
```suggestion
  - @copyright Copyright (c) 2024 Kostiantyn Miakshyn <molodchick@gmail.com>
  -
  - @author Kostiantyn Miakshyn <molodchick@gmail.com>
```

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,304 @@
...
+  -->
+
+<template>
+	<Question v-bind="questionProps"
+		:title-placeholder="answerType.titlePlaceholder"
+		:warning-invalid="answerType.warningInvalid"
+		v-on="commonListeners">
+		<template #actions>
+			<template v-if="!allowedFileTypesDialogOpened">
+				<NcActionButton :is-menu="true" @click="allowedFileTypesDialogOpened = true">
```
Chartman123, line 30:
```suggestion
				<NcActionButton is-menu @click="allowedFileTypesDialogOpened = true">
```

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,304 @@
...
+					@option:created="onAllowedFileExtensionsAdded"
+					@option:deselected="onAllowedFileExtensionsDeleted" />
+
+				<NcActionSeparator />
+			</template>
+
+			<template v-if="!allowedFileTypesDialogOpened">
+				<NcActionInput type="number"
+					:value="maxAllowedFilesCount"
+					:label-outside="true"
```
Chartman123, line 71:
```suggestion
					label-outside
```

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,304 @@
...
+					:label="t('forms', 'Maximum number of files')"
+					:show-trailing-button="false"
+					@input="onMaxAllowedFilesCountInput($event.target.value)" />
+
+				<NcActionInput type="number"
+					:value="maxFileSizeValue"
+					:label-outside="true"
```
Chartman123, line 78:
```suggestion
					label-outside
```

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,304 @@
...
+			}
+
+			if (allowedFileTypes.length) {
+				return t('forms', 'Allowed file types: {fileTypes}.', { fileTypes: allowedFileTypes.join(', ') })
+			}
+
+			return t('forms', 'All file types are allowed.')
+		},
+	},
+	mounted() {
```
Chartman123, line 198:
```suggestion
	},

	mounted() {
```

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,304 @@
...
+			try {
+				this.fileLoading = true
+				response = await axios.post(url, formData, { headers: { 'Content-Type': 'multipart/form-data' } })
+			} catch (error) {
+				logger.error('Error while submitting the form', { error })
+				showError(t('forms', 'There was an error during submitting the file: {message}.',
+					{ message: error.response.data.ocs.meta.message }))
+
+				return
+			} finally {
```
Chartman123, line 236:
```suggestion
					{ message: error.response.data.ocs.meta.message }))
			} finally {
```

Is this `return` really necessary here?

src/components/Questions/QuestionFile.vue
Chartman123, line 1: I think it might be a little easier to read, if you added a blank line between the methods...

Chartman123 (commented):
Some more comments on the code :)

lib/Controller/ApiController.php
Koc: Not all of our users know English good enough. It's quite common practice to translate validation errors.

src/components/Questions/QuestionFile.vue
Koc: yes, because I don't want to execute next line:

```
this.$emit('update:values', [...this.values, ...OcsResponse2Data(response)])
```

if server responded with an error

Chartman123: I didn't know that you can trigger the return after running finally coming from the catch block when you insert a return there.

src/components/Questions/QuestionFile.vue
Koc: added line-breaks. I guess this stylish review can be automated by linters/cs tools

appinfo/routes.php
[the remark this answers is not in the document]
susnux: I do not think this should be on the *forms* or *question* collection, this is basically a submission.
So if the submission is created before the upload I would go for:
`/api/{apiVersion}/submission/{id}/files`

but if you need to upload first before the ID is known I would go for
`/api/{apiVersion}/submission/files`

lib/Controller/ApiController.php
susnux: > Not all of our users know English good enough. It's quite common practice to translate validation errors.

It is not about user errors, for users the UI should show a translated error. But for API responses we just use English messages.

composer.json
```diff
@@ -30,7 +30,8 @@
 		"phpunit/phpunit": "^9"
 	},
 	"require": {
-		"phpoffice/phpspreadsheet": "^2.0"
+		"phpoffice/phpspreadsheet": "^2.0",
+		"symfony/polyfill-uuid": "^1.29"
```
susnux, line 34: This can be really bad, because this can create race conditions with server core, PHP only allows one version of a dependency and then if depends if the server version or this is loaded first.
Meaning either server or forms can get random errors if the version used in server does not match the version from forms.

src/views/Submit.vue
```diff
@@ -551,7 +551,8 @@ export default {
 				emit('forms:last-updated:set', this.form.id)
 			} catch (error) {
 				logger.error('Error while submitting the form', { error })
-				showError(t('forms', 'There was an error submitting the form'))
+				showError(t('forms', 'There was an error submitting the form: {message}',
```
susnux, line 554: This is the translated error message topic, this would require us to translate all API messages.
Currently our approach is to validate the form before submitting. So this error message should only be shown if there is a network error or similar.

lib/Controller/ApiController.php
```diff
@@ -943,14 +973,136 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+
+		$uploadedFiles = [];
+		foreach ($this->request->getUploadedFile('files') as $key => $files) {
+			foreach ($files as $i => $value) {
+				$uploadedFiles[$i][$key] = $value;
+			}
+		}
+
+		if (!count($uploadedFiles)) {
+			throw new OCSBadRequestException($this->l10n->t('No file provided'));
```
susnux, line 1005: For example this is really easy to check before submitting, so no need to use the server message and instead validate before uploading.

lib/Db/UploadedFile.php
```diff
@@ -0,0 +1,69 @@
...
+	protected $formId;
+	protected $originalFileName;
+	protected $fileId;
+	protected $created;
+
+	/**
+	 * Answer constructor.
+	 */
+	public function __construct() {
+		$this->addType('id', 'string');
```
susnux, line 53: This conflicts with the `Entitiy` signature, where `setId(int id)` and `getId(): int` is set.

lib/Controller/ApiController.php
```diff
@@ -943,14 +973,136 @@ private function storeAnswersForQuestion($submissionId, array $question, array $
...
+				$folder = $userFolder->newFolder($path);
+			}
+			/** @var \OCP\Files\Folder $folder */
+
+			$fileName = $folder->getNonExistingName($uploadedFile['name']);
+			$file = $folder->newFile($fileName, file_get_contents($uploadedFile['tmp_name']));
+
+			$uploadedFileId = uuid_create(\UUID_TYPE_RANDOM);
+			$uploadedFileEntity = new UploadedFile();
+			$uploadedFileEntity->setId($uploadedFileId);
```
susnux, line 1090: I still do not see any reason to use a UUID here, you can also just use the auto incremented numerical ID of the entity.
Is there any benefit of using a UUID here? Some use case I might not see?

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,314 @@
...
+						<NcActionText>{{ uploadedFile.fileName }}</NcActionText>
+
+						<NcActionButton class="delete-button-wrapper"
+							@click="onDeleteUploadedFile(uploadedFile.uploadedFileId)">
+							<template #icon>
+								<IconDelete :size="20" />
+							</template>
+							{{ t('forms', 'Delete') }}
+						</NcActionButton>
+					</ul>
```
susnux, line 105: NcAction* are not meant for this usage, this might break in the future.
Also I do not think the inner list makes semantically sense, so instead I would use a NcListItem instead of the `ul.uploaded-file`.

src/components/Questions/QuestionFile.vue
```diff
@@ -0,0 +1,314 @@
...
+					this.maxFileSizeUnit = unit
+				}
+			})
+
+			this.maxFileSizeValue = this.extraSettings.maxFileSize / FILE_SIZE_UNITS[this.maxFileSizeUnit]
+		}
+	},
+
+	methods: {
+		async onFileInput() {
```
susnux, line 214: This function should probably validate the file input, so that we can should any issues before submitting

susnux:
@Koc I think the overall implementation is good, but there are a few points that I think are problemantic:
1. Validation: We validate all inputs before submitting, so also files should be validated (as good as possible) before submitting. This also makes it not necessary to translate API messages.
2. Why using UUIDs? We do not use them for any entity, if there is no good reason for introducing them I would simply go for numeric IDs.

But other than that is looks quite good :)

lib/BackgroundJob/CleanupUploadedFilesJob.php
```diff
@@ -0,0 +1,112 @@
...
+
+	public function __construct(
+		private IRootFolder $storage,
+		private FormMapper $formMapper,
+		private UploadedFileMapper $uploadedFileMapper,
+		private LoggerInterface $logger,
+		ITimeFactory $time) {
+		parent::__construct($time);
+
+		$this->setInterval(60 * 60 * 24);
```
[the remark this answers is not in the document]
Chartman123: Isn't this the interval of when the background job runs? The file lifetime is set some lines above with `-2 days`.

lib/Controller/ApiController.php
Koc: Here is use case what comes into my mind:

1. User U1 uploads file test.txt with id=100
2. User U2 submits form and Brut-forces id of the uploaded file
3. So, User U2 submits form with file that has been uploaded by User U1

Probability of the uuid Brut-force is much lower than integer.

lib/Db/UploadedFile.php
Koc: yeah, but psalm is ok with that

src/components/Questions/QuestionFile.vue
Koc: Done

image

lib/BackgroundJob/CleanupUploadedFilesJob.php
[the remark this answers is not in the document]
Koc: I've set both values to 1h

lib/Controller/ApiController.php
Koc: ok, mates, as you wish. I've removed all translations on BE side

src/views/Submit.vue
Koc: As we've removed API error messages translations on BE side, then we need to translate only single translation string `There was an error submitting the form: {message}`

src/components/Questions/QuestionFile.vue
Koc: - I've added file size validation.
- Also we've already have validation for files count (file input is disabled if uploaded enough files).
- But it is hard to add mime type validation because we've using not clear mime types but some kind of remapping provided by Nextcloud (`mimetypealiases.dist.json`).

lib/Controller/ApiController.php
Koc: this check removed. If file not selected on FE side - we don't call api

composer.json
Koc: I've added explanation about uuid [here](https://github.com/nextcloud/forms/pull/2040#discussion_r1593153518). As this is polyfill for a native PHP extension, I doubt that we can expect any kind of breaking changes between versions.

lib/Controller/ApiController.php
susnux: It is not about removing the check, we **always** should also check on backend.
But we only need to have localized errors on the frontend, the backend is only for API, but still needs the checks to prevent any malicious content from someone not using the frontend but the API directly.

lib/Db/UploadedFile.php
susnux: Yes because psalm can not check this here, but it will lead to runtime errors due to string passed to int attribute

src/components/Questions/QuestionFile.vue
susnux: > But it is hard to add mime type validation because we've using not clear mime types but some kind of remapping provided by Nextcloud (mimetypealiases.dist.json).

I think it would be enough to set the `accept` attribute of the file input on the frontend, we also do the validation on the backend so we should be good.

composer.json
Koc: dependency removed

lib/Controller/ApiController.php
Koc: I've rolled back validation removal.

lib/Controller/ApiController.php
Koc: newermind, I've changed type to integer

src/components/Questions/QuestionFile.vue
Koc: I've just added accept attribute, looks good for me

lib/Db/UploadedFile.php
Koc: changed field type to int

Koc:
@susnux @Chartman123 I did all changes

Koc:
I've added folder sharing with user when you're sharing form

image
image

appinfo/routes.php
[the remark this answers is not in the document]
Koc: Let's keep as is and change in upcoming major with all other routes?

Koc:
conflicts fixed

lib/Service/FormsService.php
```diff
@@ -696,8 +721,35 @@ public function getFileName(Form $form, string $fileFormat): string {
...
+			self::normalizeFileName($form->getId().'. '.$form->getTitle()),
+		]);
+	}
+
+	public function getUploadedFilePath(Form $form, int $submissionId, int $questionId, ?string $questionName, string $questionText): string {
+
+		return implode('/', [
+			$this->getFormUploadedFilesFolderPath($form),
+			$submissionId,
+			self::normalizeFileName($questionId.'. '.($questionName ?: $questionText))
```
susnux, line 739: This is a bit weird to me as `number. something` looks like a enumeration, but you will have something like:

```
14. some title
124. other title
8723. yet another title
```

I mean I am fine with that (just nit picking), but maybe instead `ID - TITLE`?

appinfo/routes.php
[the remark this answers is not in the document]
susnux: makes sense yes

susnux (approved):
Looks pretty good to me right now.
Yet I would like to give this a longer test (also security) before releasing in the stable channel :)

lib/Constants.php
```diff
@@ -204,4 +221,8 @@ class Constants {
 	];
 
 	public const DEFAULT_FILE_FORMAT = 'csv';
+
+	public const UNSUBMITTED_FILES_FOLDER = self::FILES_FOLDER . '/unsubmitted';
+
+	public const FILES_FOLDER = 'forms';
```
Chartman123, line 227: Now that I see it in action, I think we should use `Forms` with capital F :)

```suggestion
	public const FILES_FOLDER = 'Forms';
```

Chartman123 (approved):
I already tested some things like what happens, if the folder is already created manually by the user or shared by someone else. So far no problems (other than of course 'file not found' problems if you delete the folder (or subfolders).

So, you could just resolve the comment from my review and then also ready to go from my side :)

@susnux what do you mean with "longer testing"? Already merge it but don't release a new 4.3 version or still wait with merging?

Chartman123:
I get the following warning in my Nextcloud log:

```
{
  "reqId": "jWDW9tqYEAOi5IViSsUD",
  "level": 2,
  "time": "2024-06-02T20:27:32+00:00",
  "remoteAddr": "192.168.1.132",
  "user": "admin",
  "app": "forms",
  "method": "POST",
  "url": "/nextcloud/ocs/v2.php/apps/forms/api/v2.4/submission/insert",
  "message": "Controller OCA\\Forms\\Controller\\ApiController::insertSubmission executed 251 queries.",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64; rv:126.0) Gecko/20100101 Firefox/126.0",
  "version": "29.0.2.1",
  "data": {
    "app": "forms"
  },
  "id": "665ce985a209b"
}
```

Chartman123:
@Koc I don't know... It's the first time I saw such a warning. Perhaps @susnux can say more about it and where this comes from

lib/Constants.php
Koc: done

lib/BackgroundJob/CleanupUploadedFilesJob.php
Chartman123, line 1: @Koc In some cases empty subfolder in the unsubmitted folder don't get deleted by the background job. Please have another look at the deletion logic. Problably when a file within such a folder was deleted by hand... I'm unsure if we really should go through all folders with the name `./Forms/unsubmitted/*`, because this could take quite some time for larger installations. What do you think?

Koc: >  I'm unsure if we really should go through all folders with the name ./Forms/unsubmitted/*, because this could take quite some time for larger installations. 

As we're running this job hourly - it looks fine for me. Anyway we can improve it in next iteration. But if you have any ideas how can we do it in different way - please share.

Koc:
@susnux @Chartman123  so, there are 2 approvals, green pipeline. What else left to be merged?

susnux:
As @Koc  said we should look for issues and fix them in next iteration.

Great work @Koc nice to have feature! :rocket:

szaimen:
🎉🎉🎉🎉🎉

cloudkuba:
Hi,
Does this change only allow files to be uploaded once the survey is open, or can I just add the file to the survey so that users can view it?

I need this functionality - the user must download the file to be able to send the survey.

Chartman123:
@cloudkuba This PR is just about submitting files... Attaching files to the description of the form or a question can already be done with a markdown link: `[https://www.domain.tld](Text)`

cloudkuba:
> @cloudkuba This PR is just about submitting files... Attaching files to the description of the form or a question can already be done with a markdown link: `[Text](https://www.domain.tld/path/to/file)`

Hi, What I mean is more about verification that the file MUST be downloaded in order to proceed.

Chartman123:
@cloudkuba no, this is not in our scope

PR #2062, opened 2024-04-14
fix: don't filter expired forms in navigation

Chartman123:
Expired forms are no longer shown in the navigation. This removes the filter and shows the epxired forms again.

### DIFF
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -205,22 +205,16 @@ export default {
 		 * All own active forms
 		 */
 		ownedForms() {
-			const now = Date.now() / 1000
-
 			return this.forms
 				.filter((form) => form.state !== FormState.FormArchived)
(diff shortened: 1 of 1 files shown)


src/Forms.vue
```diff
@@ -205,22 +205,16 @@ export default {
...
 			return this.forms
 				.filter((form) => form.state !== FormState.FormArchived)
-				.filter((form) => form.expires === 0 || form.expires >= now)
 		},
 
 		/**
 		 * All active shared forms
 		 */
 		sharedForms() {
-			const now = Date.now() / 1000
```
susnux, line 219: Currently we do not show expired shared forms - except you have results permission. So I would rather keep it this way (so the has permissions check is missing here)

Chartman123: Yes but they should already be filtered server side...

susnux (approved):

PR #2065, opened 2024-04-15
fix: remove setup() and move code to data() in Results.vue

Chartman123:

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -305,8 +305,18 @@ export default {
 
 	mixins: [PermissionTypes, ViewsMixin],
 
-	setup() {
+	data() {
 		return {
+			activeResponseView: responseViews[0],
+
+			isDownloadActionOpened: false,
+			loadingResults: true,
+
+			picker: null,
+			showSummary: true,
+			showConfirmDeleteDialog: false,
+			showLinkedFileNotAvailableDialog: false,
+
 			isMobile: useIsSmallMobile(),
 
 			// non reactive props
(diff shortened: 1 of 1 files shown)


src/views/Results.vue
```diff
@@ -305,8 +305,18 @@ export default {
...
+
+			isDownloadActionOpened: false,
+			loadingResults: true,
+
+			picker: null,
+			showSummary: true,
+			showConfirmDeleteDialog: false,
+			showLinkedFileNotAvailableDialog: false,
+
 			isMobile: useIsSmallMobile(),
```
susnux, line 320: This **must** be in `setup`. (even if this specific case might work, composables should always and only be used in `setup` context).

Chartman123:
Ok, could you please fix my code then?

susnux:
> @[susnux](https://github.com/susnux) susnux merged commit [56f410a](https://github.com/nextcloud/forms/commit/56f410a34be953c7ea09e371bda34d460509cb97) into main [Apr 15, 2024](https://github.com/nextcloud/forms/pull/2065#event-12471846466)

I am pretty sure I merged the other PR not this one.

Chartman123:
Ok, should we revert it partially? I just read that setup() is only supported with vue3, that's why I moved everything into data().

susnux:
> Ok, should we revert it partially? I just read that setup() is only supported with vue3

No, `setup` is supported since Vue 2.7
Everything in `data` will be made reactive, things in `setup` not. In setup you use the Composition API to make things reactive (e.g. `const foo = ref(4)`).

But some things like composables need to be defined in `setup` to allow them to register all life-cycle hooks like `mounted`.

PR #2066, opened 2024-04-15
Set lastUpdated on link/unlink file

Chartman123:
Currently lastUpdated is not set when a file is (un)linked


susnux (approved):

susnux:
Maybe we should just have updated the `FormMapper::update` method to set always the lastupdated value on update and then call the parent `update` :thinking:

Chartman123:
Could be done as a refactoring for 4.3

PR #2069, opened 2024-04-15
fix(DB): Correctly fetch shared forms

susnux:
Adjusted FormMapper code (mostly reverting to a version we used in the PR before).
So this fixes following:
1. Also return directly shared forms even if the form is not public shared
2. If no filter was set we need to return **all** public forms not only those with *PERMIT* permission, but all that are not not-public ;)
3. Let the database software handle fetching the forms
4. Added database tests for shared forms.


Chartman123 (approved):
yep, no it looks good :) just squash the commits and then lgtm

PR #2073, opened 2024-04-16
Fix toggle between Summary and Responses Submission overview

Koc:
1. Go to Submission overview
2. Switch from Summary to Responses

Expected: view mode changed

Actual: nothing happened


Chartman123:
Good catch, even before someone reported the issue ;)

Chartman123 (approved):

PR #2077, opened 2024-04-16
feat: Switch to PlayWright for E2E and component tests

susnux:
* Chained on #2208 (<-- this must be merged first)

This should be more stable than Cypress + real open source and not open core...
Also this is faster.

I will add more tests later.


Chartman123:
@susnux could you please remove the last commit? Then we can merge this one as well :)

Chartman123 (approved):

Issue #616, opened 2020-09-28
Allow limiting/requiring a certain number of selected checkbox answers


EricKerby:
The checkbox question type should provide the following instruction/validation options:

- Require an exact number of checkbox answers be selected
- Require at least a certain number of checkbox answers be selected
- Limit selected checkbox answers to no more than a certain number

This is useful, for instance, if you need to have someone choose their "top 3" choices from a list of options.

ei8fdb:
> This is useful, for instance, if you need to have someone choose their "top 3" choices from a list of options.

Hmmm, what's your intended outcome? For the person to a minimum/maximum of X options? Or to express a preferred subset preference?

I'm struggling to see how the "top 3" example would work here. 

I only ask because I've done similar "top X" or choose what order" type questions by using 2 columns - left has all options, and right is the users preferred options (a subset of all options). 

The user then drags from left to right and the order denotes the preference.

EricKerby:
> Hmmm, what's your intended outcome? For the person to a minimum/maximum of X options? Or to express a preferred subset preference?

For my use case, it is to select (or vote for) a minimum/maximum of the options. No need to specify an order. Example: vote for 3 out of the following 5 candidates.

PR #2080, opened 2024-04-20
feat: Allow to limit multiple question type

susnux:
* Resolves: https://github.com/nextcloud/forms/issues/616

This allows to set a minimum and maximum of options that can or must be selected.
E.g. you can enforce max. 3 options to be selected.

Other changes needed:
1. Add custom validation
2. Add validation of extra settings parameter types

### Screenshots

settings | error
---|---
Screenshot_20240420_021429|Screenshot_20240420_021446

### DIFF
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -613,8 +613,15 @@ class FormsService {
 			}
 		}
 
-		// Special handling of short input for validation
-		if ($questionType === Constants::ANSWER_TYPE_SHORT && isset($extraSettings['validationType'])) {
+		if ($questionType === Constants::ANSWER_TYPE_MULTIPLE) {
+			// Ensure limits are sane
+			if (isset($extraSettings['optionsLimitMax']) && isset($extraSettings['optionsLimitMin'])
+				&& $extraSettings['optionsLimitMax'] < $extraSettings['optionsLimitMin']) {
+				return false;
+			}
+
+			// Special handling of short input for validation
+		} elseif ($questionType === Constants::ANSWER_TYPE_SHORT && isset($extraSettings['validationType'])) {
 			// Ensure input validation type is known
 			if (!in_array($extraSettings['validationType'], Constants::SHORT_INPUT_TYPES)) {
 				return false;
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -600,9 +600,17 @@ class FormsService {
 				$allowed = [];
 		}
 		// Number of keys in extraSettings but not in allowed (but not the other way round)
-		$diff = array_diff(array_keys($extraSettings), $allowed);
+		$diff = array_diff(array_keys($extraSettings), array_keys($allowed));
 		if (count($diff) > 0) {
 			return false;
+		}
+
+		// Check type of extra settings
+		foreach ($extraSettings as $key => $value) {
+			if (!in_array(gettype($value), $allowed[$key])) {
+				// Not allowed type
+				return false;
+			}
 		}
 
 		// Special handling of short input for validation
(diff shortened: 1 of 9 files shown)


susnux:
@jancborchardt I am not sure about using the note card for validation / showing the error. What do you think?

src/views/Submit.vue
```diff
@@ -509,13 +509,26 @@ export default {
...
+
+				// in case no answer is set or all are empty show the confirmation dialog
+				if (Object.keys(this.answers).length === 0 || Object.values(this.answers).every((answers) => answers.length === 0)) {
+					this.showConfirmEmptyModal = true
+				} else {
+					// otherwise do the real submit
+					this.onConfirmedSubmit()
+				}
+			} catch (error) {
+				logger.debug(error)
```
Chartman123, line 530: I get this in my console when I a submit with an invalid number of chosen checkbox options:

```
[DEBUG] forms: Unexpected Error "One question did not validate sucessfully"

Stack trace:
onSubmit@http://localhost/nextcloud/apps/forms/js/Submit-Ds8F_QcK.mjs:1567:4872
async*submit@http://localhost/nextcloud/apps/forms/js/Submit-Ds8F_QcK.mjs:1567:8207
```

Chartman123 (commented):
Some small suggestions and a Stack trace in a Debug error message... Unsure if this is intended.

Chartman123:
> @jancborchardt I am not sure about using the note card for validation / showing the error. What do you think?

I think it looks nice and we could also use this for the other questions :)

src/views/Submit.vue
susnux: Yes that is what is expected, its the debug message logged

Chartman123: Ok, I just thought because of the "Unexpected Error" message...

susnux: But then lets change it, it should not confuse :)

Chartman123:
Could you please add some documentation here? https://github.com/nextcloud/forms/blob/main/docs/DataStructure.md#extra-settings

Chartman123 (commented):
One more little typo in a comment.

And as I stated in the other comment we should add the new extraSettings to the docs. And we should make sure that the min number can't be bigger than the max number. :)

susnux:
> And as I stated in the other comment we should add the new extraSettings to the docs. And we should make sure that the min number can't be bigger than the max number. :)

Both fixed and also added a test for the options :)

lib/Service/FormsService.php
```diff
@@ -600,13 +600,28 @@ public function areExtraSettingsValid(array $extraSettings, string $questionType
...
+			if (!in_array(gettype($value), $allowed[$key])) {
+				// Not allowed type
+				return false;
+			}
+		}
+
+		if ($questionType === Constants::ANSWER_TYPE_MULTIPLE) {
+			// Ensure limits are sane
+			if (isset($extraSettings['optionsLimitMax']) && isset($extraSettings['optionsLimitMin'])
+				&& $extraSettings['optionsLimitMax'] < $extraSettings['optionsLimitMin']) {
```
Chartman123, line 619: I'd say they should be less or equal, so that you can use it for e.g. exactly 3 answers

Chartman123 (commented):
Yes looks good. But min and max could be the same so we should do the check with <=

lib/Service/FormsService.php
susnux: Yes exactly that is checked here "if max is smaller than min then fail" meaning if max is **greater** or **equal** it will succeed.

Chartman123: Oh yes, sure, sorry...

Chartman123 (approved):
Ok, I think we're good to merge this now :)

Could you do another rebase so that we also have a Codecov update here?

susnux:
> Could you do another rebase so that we also have a Codecov update here?

image

:wink:

Chartman123:
Yes but last time I did that something regarding the signing went wrong 🙈

PR #2090, opened 2024-04-20
playground: Investigate forms v3 api

susnux:


Chartman123:
@susnux as this new way of defining the routes directly in the controller is only available starting with NC29, should we do this as a second step and implement the new routes in the old way? Otherwise 4.3 should also be NC29+ only...

PR #2092, opened 2024-04-22
feat: Allow to reorder options of "checkbox" "radio" and "dropdown" question types in frontend

susnux:
* Resolves https://github.com/nextcloud/forms/issues/354

This adds support to reorder options instead of having to remove them.
It works but I had to modify a lot of places to keep smooth behavior of the list ordering.

Implementation things:
* I am not sure about the API endpoint, it does not match our current naming style
* Should we expose the "order" property or just keep it internally and only return options sorted?

@nextcloud/designers question:
* There is not much space for a "drag" handle + a11y button so I just used buttons, I do not think it looks super awesome, but it works. Do you have an idea how to improve? Especially because there is already the drag handle for the question itself.

## Screenshots

Shuffle options enabled = no sorting | No shuffle = you can sort
---|---
Screenshot 2024-04-22 at 12-31-10 fff9 - Forms - Nextcloud|Screenshot 2024-04-22 at 12-31-51 fff9 - Forms - Nextcloud


appinfo/routes.php
```diff
@@ -200,31 +200,42 @@
...
+		],
+
+		// Batch processing for reordering of options
+		[
+			'name' => 'api#reorderOptions',
+			'url' => '/api/{apiVersion}/question/{id}/options',
+			'verb' => 'PATCH',
+			'requirements' => [
+				'id' => '\d+',
+				'apiVersion' => 'v2(\.5)?'
```
Chartman123, line 238: I'd switch those two lines. Makes more sense alphabetically and also based on the appearance in the URL
```suggestion
				'apiVersion' => 'v2(\.5)?',
				'id' => '\d+'
```

docs/API.md
```diff
@@ -28,6 +28,7 @@ This file contains the API-Documentation. For more information on the returned D
 - Completely new way of handling access & shares.
 
 ### Other API changes
+- In API version 2.5 the endpoint `/api/v2.5/options` was added for `PATCH` requests to reorder options
```
Chartman123, line 31: Doesn't fit to the defined route.

lib/Controller/ApiController.php
```diff
@@ -930,11 +933,59 @@ public function deleteOption(int $id): DataResponse {
...
+
 		$this->formsService->setLastUpdatedTimestamp($form->getId());
 
 		return new DataResponse($id);
 	}
 
+	/**
+	 * Reorder options for a given question
+	 * @param int $id The question ID
+	 * @param int[] Order to use
```
Chartman123, line 951:
```suggestion
	 * @param int[] $order Order to use
```
Parameter name was missing

src/components/Questions/AnswerInput.vue
```diff
@@ -4,41 +4,63 @@
...
+					</NcButton>
 				</template>
-				{{ t('forms', 'Delete answer') }}
-			</NcActionButton>
-		</NcActions>
+				<NcButton type="tertiary"
+					:aria-label="t('forms', 'Delete answer')"
+					@click="deleteEntry">
+					<template #icon>
+						<IconClose :size="20" />
```
szaimen, line 46: I think this should rather use a delete icon instead of a close icon?

szaimen:
> * There is not much space for a "drag" handle + a11y button so I just used buttons, I do not think it looks super awesome, but it works. Do you have an idea how to improve? Especially because there is already the drag handle for the question itself.

I think it looks fine but had some suggestion :)

src/components/Questions/AnswerInput.vue
```diff
@@ -82,19 +113,41 @@ export default {
...
-			}, 500),
 		}
 	},
 
 	computed: {
+		ariaLabel() {
+			if (this.local) {
+				return t('forms', 'Add a new answer')
+			}
+			return t('forms', 'An answer for the {index} option', { index: this.index + 1 })
```
Chartman123, line 126:
```suggestion
			return t('forms', 'An answer for the option {index}', { index: this.index + 1 })
```
Wouldn't this result in "An answer for 1 option" which should be "1st option"?

Chartman123 (commented):
Some things I found by looking at the code, did not test it by now.

src/components/Questions/AnswerInput.vue
susnux: Basically copied / moved from here:
https://github.com/nextcloud/forms/pull/2092/files#diff-644e72f1327ea20409e3f75541d13240c8d29e59ddfdfd14e3c261d3df3ca7d1L7

But yes looks weird

docs/API.md
susnux: This was my first idea, see commit in initial post. I am not sure if the current fits best?

jancborchardt (commented):
> There is not much space for a "drag" handle + a11y button so I just used buttons, I do not think it looks super awesome, but it works. Do you have an idea how to improve? Especially because there is already the drag handle for the question itself.

@susnux what about using a drag handle that on click/enter shows a menu saying "Move up" and "Move down"? Then it would be 1 icon less, and both cases would be covered.

Questions:
- Should the drag handle be on left or right – I would tend towards left
- Which icon to use. I’d say the simple "drag" – or we use that for the whole question, and the "drag-horizontal-variant" for the individual replies?

Chartman123:
> drag handle that on click/enter shows a menu saying "Move up" and "Move down"

@jancborchardt This would on the other hand introduce one more click for every move operation...

> * Should the drag handle be on left or right – I would tend towards left

having the handle on the left would introduce one more difference between edit and view mode, but we already have some differences, so I'd be fine with that. Whatever looks better as a whole.

susnux:
The problem with a drag handle on the left is the conflict with the drag handle for the question itself.
We would then have two drag handles next to each other. Not sure about that.

Chartman123:
@susnux @jancborchardt What about having the arrow icons as an overlay at the end of the input fields that only shows up when the field has the focus / hover?

@susnux should I add your backend code to the API v3 PR or would you like to keep it in here and adjust it after v3 merge?

susnux:
> adjust it after v3 merge?

will adjust afterwards :)

jancborchardt (commented):
> What about having the arrow icons as an overlay at the end of the input fields that only shows up when the field has the focus / hover?

@Chartman123 sounds good! Then it’s not that much visual noise, and the fields are nicely aligned in width. :)

Chartman123:
This is now rebased to our current main. Design-wise there's still some room for improvement as I'm really no expert in CSS stuff. But I already made the arrows overlay the input fields.

@jancborchardt what do you think?

grafik

marcoambrosini:
@Chartman123 @jancborchardt is out on vacation and I'm replacing him. 
The arrows look good! Please make sure that their respective buttons are of `--clickable-area` size and that the icons are centered in them. Since these are tertiary buttons we don't need any extra margins. 
One question: shouldn't the arrows show only when the input field is focused? 

Other than that, for the left drag handle I would suggest this icon:
https://fonts.google.com/icons?selected=Material+Symbols+Outlined:drag_indicator:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=drag&icon.size=24&icon.color=%23e8eaed
Screenshot 2025-02-17 at 14 25 34

Chartman123:
> One question: shouldn't the arrows show only when the input field is focused?

@marcoambrosini yes, but this might perhaps also be an a11y issue? and we can also deliver it in one of the next minor/patch releases :)

> The arrows look good! Please make sure that their respective buttons are of `--clickable-area` size and that the icons are centered in them. Since these are tertiary buttons we don't need any extra margins.

Is the size not managed directly by using the `NCButton`?

Chartman123:
@marcoambrosini so instead of the arrows on the right a drag handle (also on the right) and the up/down actions behind the action menu?

marcoambrosini:
Yes exactly

src/mixins/QuestionMultipleMixin.ts
```diff
@@ -0,0 +1,293 @@
...
+				showError(t('forms', 'Error while saving options order'))
+			}
+		},
+
+		/**
+		 * Reorder option by moving it upwards the list
+		 * @param {number} index Option that should move up
+		 */
+		onOptionMoveUp(index: number) {
+			if (index > 0) {
```
AIlkiv, line 271: There is no need for this if, because the button that calls this logic will be disable when index = 0

src/mixins/QuestionMultipleMixin.ts
```diff
@@ -0,0 +1,293 @@
...
+				this.onOptionMoveDown(index - 1)
+			}
+		},
+
+		/**
+		 * Reorder option by moving it downwards the list
+		 * @param {number} index Option that should move down
+		 */
+		onOptionMoveDown(index: number) {
+			if (index === this.sortedOptions.length - 1) {
```
AIlkiv, line 281: There is no need for this if, because the button that calls this logic will be disable when index = this.sortedOptions.length - 1

src/components/Questions/AnswerInput.vue
```diff
@@ -233,6 +303,45 @@ export default {
...
+			this.$emit('move-up')
+			if (this.index > 1) {
+				this.$nextTick(() => this.$refs.buttonUp.$el.focus())
+			} else {
+				this.$nextTick(() => this.$refs.buttonDown.$el.focus())
+			}
+		},
+
+		/**
+		 * Handle compostion start event for IME inputs
```
AIlkiv, line 330: compostion > composition

src/components/Questions/AnswerInput.vue
```diff
@@ -233,6 +303,45 @@ export default {
...
+		 */
+		onCompositionStart() {
+			this.isIMEComposing = true
+		},
+
+		/**
+		 * Handle compostion end event for IME inputs
```
AIlkiv, line 337: compostion > composition

AIlkiv (changes requested):

AIlkiv:
I don't think you should overlap the input borders on hover. I tested on version 30.

image

Chartman123:
@max-nextcloud any idea why the e2e tests are failing without my changes? I mean we're still using aria-labels, even if they are now coming from a computed prop: https://github.com/nextcloud/forms/pull/2092/commits/acfe0a3f2dd4af19e8b5c4fd7fbba4b7f2b31675

susnux:
> Yes this has to be fixed, but not a blocker as the buttons will also be replaced in #2579 by a drag handle

But we still need buttons in addition so it can be moved with keyboard only

src/mixins/QuestionMultipleMixin.ts
susnux: still makes sense for this method to be defensive and not rely on template conditions

src/mixins/QuestionMultipleMixin.ts
susnux: still makes sense for this method to be defensive and not rely on template conditions

Chartman123:
> > Yes this has to be fixed, but not a blocker as the buttons will also be replaced in #2579 by a drag handle
> 
> But we still need buttons in addition so it can be moved with keyboard only

Sure, they are preserved and with better styling in the linked follow up PR

susnux:
> I don't think you should overlap the input borders on hover. I tested on version 30.

Resolved

Chartman123 (approved):

Chartman123:
Failing workflow unrelated (see fix in #2592)

PR #2093, opened 2024-04-22
fix: Parse momentFormat and storageFormat

### LINKED ISSUE #2091: Date fields don't show their previously entered values upon reloading
Steps to reproduce: 
1. Enter a date into a date field
2. Leave view and go back to view / reload window
3. See missing date

The date gets stored correctly in the local storage but isn't loaded back. Might be related to #1969

Chartman123:
This fixes #2091 by parsing the momentFormat `L` or the storageFormat `YYYY-MM-DD`.

### DIFF
--- a/src/components/Questions/QuestionDate.vue
+++ b/src/components/Questions/QuestionDate.vue
@@ -109,7 +109,7 @@ export default {
 		 * @return {Date}
 		 */
 		parse(dateString) {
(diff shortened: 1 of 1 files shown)


susnux (approved):

susnux:
We should have regression tests for things like this ;)

Chartman123:
> We should have regression tests for things like this ;)

Can this be achieved with E2E tests?

PR #2096, opened 2024-04-23
fix(refactor): Remove duplicated code from ApiController

susnux:
Reduce duplicated code from the controller


Chartman123:
Nice one :)

Chartman123 (approved):

PR #2099, opened 2024-04-25
fix: add required icon to question header in edit mode

Chartman123:


jancborchardt (commented):
The icon technically is an asterisk, but it does look a bit funky and unexpected.

What do you think instead about a little red circle overlaid on the 3-dot menu (much like when you have unread notifications), which calls attention to opening the menu which then also explains that it is required?

Then we also save 1 icon spot additionally.

Chartman123:
> What do you think instead about a little red circle overlaid on the 3-dot menu (much like when you have unread notifications), which calls attention to opening the menu which then also explains that it is required?

@jancborchardt thanks for the review. I think if we want to point out directly that one of settings of the submenu is enabled, it should be directly visible. Otherwise you wouldn't know what the red ring means and you'd have to look into the menu.

But I'm with you that it is probably a little bit too big...

Chartman123:
For text questions ok, but how would you do it for e.g. the new file question?

jotoeri:
Was just dropping by so here some greets and my 5 cent:
I like the mark on the three dot menu - but i'd think about using an asterisk there. This should be clearly connected to the required option. I could imagine one could do that similarly with specific signs of up to three options from the menu (or maybe even four on the four edges)? :wink: 

:wave:

susnux:
> I like the mark on the three dot menu - but i'd think about using an asterisk there.

I like the idea :+1:

Chartman123:
So we overlay the Icon in small over the three dot menu?

susnux:
Would look like this:
image

Chartman123:
I could live with that, but move it to the top left and don't use the red color. I think this is too much.

susnux:
image

or 

image

Chartman123:
Thanks for updating the screenshots and the code :) I only have little time right now...

I'd be fine with that second version without the accent color :) @jancborchardt what do you think?

jancborchardt (approved):
Yep, agree with @Chartman123 , second version without the accent is very nice. :) Thanks @susnux as well!

src/components/Questions/Question.vue
```diff
@@ -87,6 +87,14 @@
 					:force-menu="true"
 					placement="bottom-end"
 					class="question__header__title__menu">
+					<template #icon>
+						<IconOverlay>
+							<template v-if="!readOnly && isRequired" #overlay>
```
Chartman123, line 92: @susnux I think the `!readOnly` isn't required here as the whole `NcActions` already has the if...

PR #2151, opened 2024-05-08
fix(a11y): Add missing page headings

susnux:
We need a h1 on the page, otherwise every `h2` or higher is invalid.
The h1 is visually hidden and only added for accessibility.

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -24,7 +24,7 @@
   -->
 
 <template>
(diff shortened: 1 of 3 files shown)


Chartman123:
Should we backport these as well?

susnux:
> Should we backport these as well?

I would say stable4 is ok

Chartman123 (approved):

PR #2164, opened 2024-05-12
Use text editor for editing question description

Koc:
Currently it is hard to write rich question descriptions for non-tech people. Would be nice add UI editor for that. Here some draft of this functionality. If text app isn't installed we fallback to simple textarea like it was before.

https://github.com/nextcloud/forms/assets/191082/3df8edcf-3e7d-483d-aab4-d72966f48724

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -30,11 +30,13 @@ namespace OCA\Forms\AppInfo;
 
 use OCA\Forms\Capabilities;
 use OCA\Forms\FormsMigrator;
+use OCA\Forms\Listener\BeforeTemplateRenderedListener;
 use OCA\Forms\Listener\UserDeletedListener;
 use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
+use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
 use OCP\User\Events\UserDeletedEvent;
 
 class Application extends App implements IBootstrap {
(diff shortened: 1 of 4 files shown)


Koc:
does anybody know why it is not possible upload attachments into markdown?

here what we hare in current PR
image
this button does nothing. What do I wrong?

here same editor from Collectives
image

Chartman123:
> does anybody know why it is not possible upload attachments into markdown?

@Koc sorry, I can't help you much with this...

Regarding your questions:
- I don't think that we should make it a hard dependency... For most forms plain text descriptions will be enough, so I'm not sure if we really need the editor at all. On the other hand for most not technically gifted people it's much easier to add formatting like that.
- Yes, the white space below the editor is way too much. I didn't have time to look into your code, but you should try to get rid of it.

Koc:
@mejo- any idea? As far I can see you have experience with Collective + Text integration

mejo-:
> @mejo- any idea? As far I can see you have experience with Collective + Text integration

So regarding the attachment support: you'll have to provide your own implementation here, because you use the editor which is not file-based and instead gets the markdown content as a property via the API. The attachments support as you know it from Text and Collectives works because there the editor is file-based and attachments can be uploaded to a related attachments folder.

As far as I know the content-based editor (different than the file-based editor) so far doesn't support adding attachments and images. The deck app has support to upload attachments via the button you referenced, but these attachments don't show up in the text document either, they just get listed as attachments. See [here](https://github.com/nextcloud/deck/blob/f14ec12b3c7a0af8c1fa6d8ae7a0959214c3dc89/src/components/card/Description.vue#L196-L198) for the implementation.

Regarding the bottom padding of the editor, that's the standard CSS of the editor and it needs to be overwritten. Again see the deck app [how it's done there](https://github.com/nextcloud/deck/blob/f14ec12b3c7a0af8c1fa6d8ae7a0959214c3dc89/src/components/card/Description.vue#L368-L370).

Koc:
I've achieved some progress with styling. You can see it on attached video in PR description. Next goal - make image upload work.

Koc:
Implementation of the images uploading functionality requires much more effort than I expected :crying_cat_face: . At the same time for me it has no sense to provide UI editor without this functionality (as it was the root cause why I've started work on this task).

We can return to this later

PR #2201, opened 2024-06-09
Add support of multi-columns descriptions

Koc:
Now we can use multi-columns descriptions thankful to [markdown-it-container](https://github.com/markdown-it/markdown-it-container) plugin

```markdown
::::: columns
:::: column-double
first column has double relative width (ie 2/3 of total width) 
::::
:::: column-single
second column has single relative width (ie 1/3 of total width)
::::
:::::
```

<details>
  <summary>:mag:  Preview</summary>

  image
  image
</details>

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -16,3 +16,4 @@ l10n/
 # PHP
 lib/
 **/*.php
+composer.lock
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "crypto-js": "^4.2.0",
         "debounce": "^2.1.0",
         "markdown-it": "^14.1.0",
+        "markdown-it-container": "^4.0.0",
         "p-debounce": "^4.0.0",
         "p-queue": "^8.0.1",
         "v-click-outside": "^3.2.0",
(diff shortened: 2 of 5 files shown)


susnux:
I am not sure about this, I do not see a good use case and will make thing more complex.
I would probably stick with CommonMark to keep it simple.

Opinions @jancborchardt @Chartman123 ?

Chartman123:
Yeah same for me... another dependency for only a really small benefit.

Koc:
for our use case It has sense for students quizzes, e.g. we have some image/schema on the left column and a question on the right column

Koc:
Updates:
- properly handle columns on mobile
- added more real-life example in PR description
- fixed pipeline

Chartman123:
@Koc thanks for the updated description with the preview... How would this work together with your other PR to implement Text as a Markdown editor? As far as I could see Text doesn't support multi-columns.

Koc:
@Chartman123 I've plan to close #2164 because it requires much more work than I expected for adding images support. So there are no any conflicts with this 2 PRs anymore :smiley:

jancborchardt (commented):
If Text doesn’t support multi-columns, I am also hesitant to add this. As mentioned, we shouldn’t stray too much from a standard, and at least be internally consistent.

Chartman123:
@Koc you can follow the blocklist link in the last post to add your username to the list so that your PRs don't get this feedback post :)

PR #2221, opened 2024-06-22
Update npm and composer deps

Chartman123:

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -650,16 +650,16 @@
 		},
 		{
 			"name": "myclabs/deep-copy",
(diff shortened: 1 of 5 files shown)


susnux (approved):
dependabot failed to do it?

Chartman123:
> dependabot failed to do it?

the nextcloud-files update was part of another dependabot PR so he refused to update it in the package.json... I also did a general `npm update` which reduced the number of dependencies a lot. However, there are still some dependencies marked as 'deprecated', but it looks like we can't influence them directly...

Composer update also added some newer versions of our deps.

PR #2223, opened 2024-06-26
Drop NC27 support for dependabot

Chartman123:


susnux (approved):
Do we have to release a last v3?

Chartman123:
There are currently no fixes waiting for a release from our side, so this would only include newer deps...

PR #2225, opened 2024-06-27
Optimization method FormsService::canSubmit

AIlkiv:
Currently, to check if a user has filled out the form, all entities are retrieved, and then two loops are performed at the PHP. My version does the same thing but with a single simple query.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -1161,7 +1161,7 @@ class ApiController extends OCSController {
 
 		// Ensure the form is unique if needed.
 		// If we can not submit anymore then the submission must be unique
-		if (!$this->formsService->canSubmit($form) && !$this->submissionService->isUniqueSubmission($submission)) {
+		if (!$this->formsService->canSubmit($form) && $this->submissionMapper->hasMultipleFormSubmissionsByUser($form, $submission->getUserId())) {
 			$this->submissionMapper->delete($submission);
 			throw new OCSForbiddenException('Already submitted');
 		}
--- a/lib/Db/SubmissionMapper.php
+++ b/lib/Db/SubmissionMapper.php
@@ -86,36 +86,40 @@ class SubmissionMapper extends QBMapper {
 	}
 
 	/**
-	 * @param int $formId
-	 * @throws DoesNotExistException if not found
-	 * @return array
+	 * Сhecks if there are multiple form submissions by user
+	 * @param Form $form of the form to count submissions
+	 * @param string $userId ID of the user to count submissions
 	 */
-	public function findParticipantsByForm(int $formId): array {
-		$qb = $this->db->getQueryBuilder();
-
-		$qb->select('user_id')
-			->from($this->getTableName())
-			->where(
-				$qb->expr()->eq('form_id', $qb->createNamedParameter($formId, IQueryBuilder::PARAM_INT))
-			);
-
-		$submissionEntities = $this->findEntities($qb);
-
-		// From array of submissionEntities produce array of userIds.
-		$userIds = array_map(function ($submissionEntity) {
-			return $submissionEntity->getUserId();
-		}, $submissionEntities);
-
-		return $userIds;
+	public function hasMultipleFormSubmissionsByUser(Form $form, string $userId): bool {
+		return $this->countSubmissionsWithFilters($form->getId(), $userId, 2) >= 2;
 	}
 
 	/**
-	 * Count submissions by form and optionally also by userId
-	 * @param int $formId ID of the form to count submissions for
-	 * @param string|null $userId optionally limit submissions to the one of that user
+	 * Сhecks if there are form submissions by user
+	 * @param Form $form of the form to count submissions
+	 * @param string $userId ID of the user to count submissions
+	 */
+	public function hasFormSubmissionsByUser(Form $form, string $userId): bool {
+		return (bool)$this->countSubmissionsWithFilters($form->getId(), $userId, 1);
+	}
+
+	/**
+	 * Count submissions by form
+	 * @param int $formId ID of the form to count submissions
 	 * @throws \Exception
 	 */
-	public function countSubmissions(int $formId, ?string $userId = null): int {
+	public function countSubmissions(int $formId): int {
+		return $this->countSubmissionsWithFilters($formId, null, -1);
+	}
+
+	/**
+	 * Count submissions by form with optional filters
+	 * @param int $formId ID of the form to count submissions
+	 * @param string|null $userId optionally limit submissions to the one of that user
+	 * @param int $limit allows to limit the query selection. If -1, the restriction is ignored
+	 * @throws \Exception
+	 */
+	protected function countSubmissionsWithFilters(int $formId, ?string $userId = null, int $limit = -1): int {
 		$qb = $this->db->getQueryBuilder();
 
 		$query = $qb->select($qb->func()->count('*', 'num_submissions'))
@@ -123,6 +127,9 @@ class SubmissionMapper extends QBMapper {
 			->where($qb->expr()->eq('form_id', $qb->createNamedParameter($formId, IQueryBuilder::PARAM_INT)));
 		if (!is_null($userId)) {
 			$query->andWhere($qb->expr()->eq('user_id', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR)));
+		}
+		if ($limit !== -1) {
+			$query->setMaxResults($limit);
 		}
 
 		$result = $query->executeQuery();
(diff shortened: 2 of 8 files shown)


lib/Db/SubmissionMapper.php
```diff
@@ -87,26 +87,27 @@ public function findById(int $id): Submission {
 
 	/**
 	 * @param int $formId
-	 * @throws DoesNotExistException if not found
-	 * @return array
+	 * @param string $userId
+	 * @return bool
 	 */
-	public function findParticipantsByForm(int $formId): array {
+	public function hasSubmissions(int $formId, string $userId): bool {
```
Chartman123, line 93:
```suggestion
	public function hasFormSubmissionsByUser(int $formId, string $userId): bool {
```

What about including form and user in the function name? You could also please add some description of what it does to the PHPDoc comment.

Chartman123 (commented):
Nice one :) Just a little comment about the new name for the function :)

@susnux what do you think?

And it would be nice if you could add some more tests to satisfy the Codecov checks 😉

lib/Db/SubmissionMapper.php
AIlkiv: @Chartman123 As for the tests, I'm not sure if it makes sense to write tests for DB Mapper, since other classes of database work are not covered.

lib/Db/SubmissionMapper.php
```diff
@@ -86,44 +86,48 @@ public function findById(int $id): Submission {
...
+	 * @param int $formId ID of the form
+	 * @param string $userId UID of the user
+	 * @param bool $checkMultipleFormSubmissions if true, we check if there is more than one submissions, if false, at least one submissions
+	 * @return bool
 	 */
-	public function findParticipantsByForm(int $formId): array {
+	public function hasFormSubmissionsByUser(
+		int $formId,
+		string $userId,
+		bool $checkMultipleFormSubmissions
```
susnux, line 98: Hm maybe make this numeric and directly use for the `requireCountFormSubmissions` ?
If not I would make this false by default.

lib/Db/SubmissionMapper.php
```diff
@@ -86,44 +86,48 @@ public function findById(int $id): Submission {
...
+	): bool {
+		$requireCountFormSubmissions = $checkMultipleFormSubmissions ? 2 : 1;
+
 		$qb = $this->db->getQueryBuilder();
 
-		$qb->select('user_id')
+		$query = $qb->select('id')
```
susnux, line 104: Maybe just:
```suggestion
		$query = $qb->select($qb->createFunction('COUNT(`id`)'))
```

lib/Db/SubmissionMapper.php
```diff
@@ -86,44 +86,48 @@ public function findById(int $id): Submission {
...
 	/**
-	 * Count submissions by form and optionally also by userId
-	 * @param int $formId ID of the form to count submissions for
-	 * @param string|null $userId optionally limit submissions to the one of that user
+	 * Count submissions by form
+	 * @param int $formId ID of the form to count submissions
 	 * @throws \Exception
 	 */
-	public function countSubmissions(int $formId, ?string $userId = null): int {
+	public function countSubmissions(int $formId): int {
```
susnux, line 125: This is not related

susnux (commented):
Generally nice! Some notes :)

(and we should add a composite index on "formId userId"

lib/Db/SubmissionMapper.php
AIlkiv: countSubmissions with userId only used on isUniqueSubmission. isUniqueSubmission is removed in this PR.

lib/Db/SubmissionMapper.php
```diff
@@ -86,36 +86,39 @@ public function findById(int $id): Submission {
...
 	/**
-	 * Count submissions by form and optionally also by userId
-	 * @param int $formId ID of the form to count submissions for
+	 * Count submissions by form with optional filters
+	 * @param int $formId ID of the form to count submissions
 	 * @param string|null $userId optionally limit submissions to the one of that user
 	 * @throws \Exception
 	 */
-	public function countSubmissions(int $formId, ?string $userId = null): int {
+	protected function countSubmissionsWithFilters(int $formId, ?string $userId = null, int $limit = -1): int {
```
Chartman123, line 121: `@param int $limit` is missing in PHPDoc comment

AIlkiv: fix

lib/Db/SubmissionMapper.php
AIlkiv: fixed

lib/Db/SubmissionMapper.php
AIlkiv: I created two separate methods. I think this is more correct.

Chartman123:
@AIlkiv could you please rebase on current main?

lib/Controller/ApiController.php
```diff
@@ -1161,7 +1161,8 @@ public function insertSubmission(int $formId, array $answers, string $shareHash
 
 		// Ensure the form is unique if needed.
 		// If we can not submit anymore then the submission must be unique
-		if (!$this->formsService->canSubmit($form) && !$this->submissionService->isUniqueSubmission($submission)) {
+		if (!$this->formsService->canSubmit($form) && $this->submissionMapper->hasMultipleFormSubmissionsByUser($form, $submission->getUserId())
+		) {
```
Chartman123, line 1165: I'd like to keep the braces in the first line like everywhere else :)

```suggestion
		if (!$this->formsService->canSubmit($form) && $this->submissionMapper->hasMultipleFormSubmissionsByUser($form, $submission->getUserId())) {
```

lib/Db/SubmissionMapper.php
```diff
@@ -124,14 +128,16 @@ public function countSubmissions(int $formId, ?string $userId = null): int {
...
+			$query->setMaxResults($limit);
+		}
 
 		$result = $query->executeQuery();
 		$row = $result->fetch();
 		$result->closeCursor();
 
 		return (int) ($row['num_submissions'] ?? 0);
 	}
-
```
Chartman123, line 134: Please add this empty line again :)

lib/Migration/Version040300Date20240708163401.php
```diff
@@ -0,0 +1,52 @@
...
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('forms_v2_submissions')) {
+			$schema->getTable('forms_v2_submissions')
+				->addIndex(['form_id', 'user_id'], 'forms_submissions_form_user');
```
Chartman123, line 47: Please check if the index isn't already there before adding it

lib/Service/FormsService.php
```diff
@@ -333,15 +333,12 @@ public function canSubmit(Form $form): bool {
 		if ($this->currentUser->getUID() === $form->getOwnerId()) {
 			return true;
 		}
-
```
Chartman123, line 336: Please add empty line again

lib/Service/FormsService.php
```diff
@@ -333,15 +333,12 @@ public function canSubmit(Form $form): bool {
...
-		if (!$form->getSubmitMultiple()) {
-			$participants = $this->submissionMapper->findParticipantsByForm($form->getId());
-			foreach ($participants as $participant) {
-				if ($participant === $this->currentUser->getUID()) {
-					return false;
-				}
-			}
+		if (
+			!$form->getSubmitMultiple()
+			&& $this->submissionMapper->hasFormSubmissionsByUser($form, $this->currentUser->getUID())
```
Chartman123, line 339: I'd put the `&&` in the first line

```suggestion
			!$form->getSubmitMultiple() &&
            $this->submissionMapper->hasFormSubmissionsByUser($form, $this->currentUser->getUID())
```

Chartman123 (commented):
Some small comments, but rest seems ok to me :)

lib/Migration/Version040300Date20240708163401.php
AIlkiv: I checked on migrations and in my local database.

Chartman123: Yes but I mean that you should add an `if` for it and only create it when it's not already there. :)

AIlkiv: fixed

Chartman123 (approved):

Issue #2253, opened 2024-07-24
Layout issue with Upload type question.


TUCAOEVER:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**
Layout issue with upload type question.

**To Reproduce**
Add upload type question and check the layout

**Expected behavior**
There should be no blank space between the upload button and question

**Screenshots**
![image](https://github.com/nextcloud/forms/assets/43756134/d60cab54-cb02-463b-ad67-915054c2830a)

In editor view, there is spacing between the option and answer.

![image](https://github.com/nextcloud/forms/assets/43756134/02ac0a90-8954-48cd-a6f9-757777fde579)

In preview page, there is spacing under the upload file button, which may related to the layout issue.

**Nextcloud (please complete the following information):**

-   Nextcloud-Version: 29.0.3
-   Forms-Version: [[e.g. 2.0.0-beta4]](https://github.com/nextcloud/forms/issues/2220#issuecomment-2210256685)

**Desktop (please complete the following information):**

-   OS: Ubuntu
-   Browser: Edge
-   Version: 127.0.2651.61

Koc:
@TUCAOEVER I've opened PR with fix for it, please check/suggest changes

PR #2259, opened 2024-07-28
fix: Improve styles of layout for QuestionFile #2253

Koc:
See #2253 for more info. I'm not css-guy, so, if you have any suggestions how to fix it in a proper way - welcome :pray: 

Fixed version has same height and paddings line other input types for both edit/view versions:
image
image

### DIFF
--- a/src/components/Questions/QuestionFile.vue
+++ b/src/components/Questions/QuestionFile.vue
@@ -118,26 +118,40 @@
 						</NcActionButton>
 					</template>
 				</NcListItem>
+				<li v-if="fileLoading" class="question__loading">
+					<NcLoadingIcon v-show="fileLoading" />
+					{{ t('forms', 'Uploading …') }}
+				</li>
+				<li v-else-if="values.length < maxAllowedFilesCount">
+					<div class="question__input-wrapper">
+						<label>
+							{{ t('forms', 'Add new file as answer') }}
+							<input
+								ref="fileInput"
+								class="hidden-visually"
+								type="file"
+								:disabled="!readOnly"
+								:multiple="maxAllowedFilesCount > 1"
+								:name="name || undefined"
+								:accept="accept.length ? accept.join(',') : null"
+								@input="onFileInput" />
+						</label>
+						<NcButton
+							:disabled="
+								!readOnly || values.length >= maxAllowedFilesCount
+							"
+							type="tertiary-no-background"
+							@click="toggleFileInput">
+							<template #icon>
+								<IconUploadMultiple
+									v-if="maxAllowedFilesCount > 1"
+									:size="20" />
+								<IconUpload v-else :size="20" />
+							</template>
+						</NcButton>
+					</div>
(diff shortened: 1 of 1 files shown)


Chartman123:
@Koc thanks for looking into it. Perhaps @susnux can also have a look at it, as I'm not an CSS expert either...

Koc:
I guess for a long term it would be nice to add `type=file` support to [`NcActionInput`](https://nextcloud-vue-components.netlify.app/#/Components/NcActions?id=ncactioninput)

src/components/Questions/QuestionFile.vue
```diff
@@ -382,4 +385,20 @@ export default {
...
 	margin-left: 30px;
 }
+
+.question__input--wrapper {
+	&:has(input:disabled) {
+		margin-left: -13px;
+	}
+
+	input {
+		height: 44px;
```
susnux, line 395:
```suggestion
		height: var(--default-clickable-area);
```

susnux:
Just use the variables, but otherwise fine :)

Koc:
@susnux for some reason it not works for me with variables :confused: 

Input should have `height: 44px` for edit mode, but variable is only `34px`
image
image

susnux:
> Input should have `height: 44px` for edit mode, but variable is only `34px`

No starting with Nextcloud 30 all inputs should be 34px, that why we should only use the variables to be independent from the server version.

Koc:
ok, but how can I properly calculate paddings?

susnux:
> ok, but how can I properly calculate paddings?

Paddings should always be multiple of `--default-grid-baseline`, if it is about centering then probably something like this:
`calc((var(--default-clickable-area) - 1lh) / 2)` (this would be the padding needed to center something).

susnux:
I pushed another commit with unifies the style with other inputs form Nextcloud. Note that it might look not 100% equal currently (border width), but this is due to a missing fix for e.g. NcDatetimePicker which will be release this week.

Feel free to adjust or even drop my commit if you want to adjust :)

Screenshot 2024-08-14 at 18-58-14 oo - Formulare - Nextcloud

susnux (approved):

Koc:
@susnux mate, it looks really amazing, thank you. This css is too hard for backend developers like me :see_no_evil: . I've fixed formatting to pass pipeline

Chartman123 (approved):

Koc:
https://github.com/nextcloud-libraries/nextcloud-vue/issues/4339 

> I guess for a long term it would be nice to add `type=file` support to [`NcActionInput`](https://nextcloud-vue-components.netlify.app/#/Components/NcActions?id=ncactioninput)

we have ticket for that™ https://github.com/nextcloud-libraries/nextcloud-vue/issues/4339

Chartman123:
@Koc yeah nice, so we could switch in the future, when it's integrated into the lib

Until then: let's get this merged :)

PR #2261, opened 2024-07-28
fix: ignore composer.lock in prettier

Koc:
Actually `composer.lock` is autogenerated file by Composer. There is no any sense to re-format it. This is not a code that we control.

At the same time we can add [composer-normalize](https://github.com/ergebnis/composer-normalize) plugin to unify `compser.json` if it needed

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -16,3 +16,4 @@ l10n/
 # PHP
 lib/
 **/*.php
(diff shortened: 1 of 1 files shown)


Chartman123 (approved):
Yeah, was something I also wanted to change... Thanks!

Issue #1850, opened 2023-12-24
Insert form submission directly into Nextcloud Deck Card


elephantastyczny:
**Nextcloud (please complete the following information):**
- Nextcloud-Version: [28.0.0 +]
- Forms-Version: [4.0.0 +]

**Describe the solution you'd like**

Id like to see Forms to be also used as a "ticketing system".
We should be able to predefine, i.e. in dropdown type question, who is responsible for dealing with the issue and once from is submitted:

- new card is created in predefined deck and list
- correct person assigned and notified (predefined in question)
- correct tag is selected (predefined in question)
- card description copied from predefined title question (short text)
- other submitted data copied into card notes

Chartman123:
As this is really an edge case to use Forms, I don't see it coming any time soon. There are lots of ticketing systems available already and so I also don't think that we should abuse two apps designed for something else.

@susnux what do you think?

elephantastyczny:
I was rather thinking about internal solution for groups.

Imagine 100 employees pool using just Nextcloud and they would like to have a single point of reporting issues:



*	Ordering supplies IT/non IT
*	general maintenance (lights, faulty locks)
*	HR related items onboarding / offboarding / change job requests

 

I believe this could replace any 3rd party solution 😊

No pressure, just saying 😊

susnux:
Yes we can keep this open but with very low priority, we should probably first think about:
- #805 
- #1748

CelianGdfrd:
Hi, is there any advancement on this, please ? It would be incredibly useful for associations I work with

Chartman123:
I don't know why I didn't notice it earlier, but this is duplicate of  #1772

Issue #1748, opened 2023-10-17
Insert form submission directly into Nextcloud Tables App


jvantslot:
A form should be able to be configured to insert records into an existing Nextcloud Tables App table.  This would make it really easy to share form data and use the Nextcloud Analytics app.  This would negate the need to export data from the form app and import into the tables app.

Chartman123:
@jvantslot I think that this is an interesting feature request :) However, there are of course some further dependencies and requirements that need to be checked.

Just a quick brainstorming what this integration could be about:
- [ ] Create a new table
- [ ] Specify an already existing table 
  - [ ] Export all questions: number/type of columns must match the questions
  - [ ] Export only some questions: map questions to existing columns?
- [ ] For further exports: update the already linked tables
  - [ ] Automatically update the linked tables on every submission?

elephantastyczny:
Love the idea!

Please don't forget to add a field for image upload and display (not sure if already possible).

Rello:
https://github.com/nextcloud/forms/pull/2195

I am working to add Forms as a direct data source to Analytics.
stay tuned...

CATER-COM:
Very nice idea. I  think at the beginning it could work just like the "export to file" work, export the whole data to a tables. The difference would be that you don't need to re-export manually, it would create a table and after add row to table for each new answer. It will gi the powerfull of Tables app to Forms app :-)

Mannshoch:
I think it is strange that it is possible to link the form to xlsx, ods and csv but not with tables.

I would really like it if I could link it with tables. 

My cause would be to show this table also in collectives, so people have a central site where they could get all current news of a club activity. (I do not wish give them a table to edit)

The Integration of a spreadsheet is nice and also used for members from orga team but a spreadsheet is not for all members accessible enougth.

alimmroth:
This would be a very important feature to be working with form-submissions systematically. Very appreciated!

SirCartaux:
I'd love to see this in future versions of nextcloud!!!

Issue #1223, opened 2022-06-04
Trigger webhook after response 


1kastner:
**Nextcloud (please complete the following information):**
- Nextcloud-Version: newest
- Forms-Version: newest

**Describe the solution you'd like**

Google Forms offers an integration a form into a pipeline via webhooks. There, I can provide a URL that the user input is forwarded to once they finalize the form. Great for automation!

**Describe alternatives you've considered**
I could repeatedly pull the current responses via the API and check for any new responses.

codeHorse87:
I see this as the missing piece of the puzzle to utilize Nextcloud forms for various workflow automation tasks. Another alternative to a webhook function provided out of the app can be the option to write submissions directly into a file. 
As a result, the workflow system https://nextcloud.com/workflow/ with the webhook app https://apps.nextcloud.com/apps/webhooks can take over.

Chartman123:
The alternative mentioned by @codeHorse87 would be achieved by #1403

drfinkus:
Any plans to address this? It would be really good to be able to e.g. post to Slack when a form is filled in.

SamuelLHuber:
https://docs.nextcloud.com/server/stable/admin_manual/webhook_listeners/index.html contains information on forms x webhooks that may work. 

Though I haven't seen documentation on how to configure webhooks per form nicely in a User Interface like https://github.com/kffl/nextcloud-webhooks which hasn't been tested against the latest major versions according to the nextcloud store page

Issue #805, opened 2021-02-25
Flow and Forms


Spartachetto:
reading the issues  #790 and #789, I thought that it would be interesting to make a further step and integrate Forms with Flow.

There are a lot of possible conditions and actions that could be of interest, like for example:

- when someone answered a form (then send a Talk message to user x)
- when someone sends this message to the group y then answer with the csv file of the answers  
- when someone answered a form then send to that user a message (email, whatever, ...) thanking for the feedback

and so on...

An hypothesis of schema

**conditions**

- When someone create a form
- When someone answers a form
- When someone answers the form x

**actions**

- then share the form x with the user
- then send the files with the answers to the user y (via email, sharing the file, ...)
- then update an analytics report

I put only the actions specific to the Forms app, it is clear that all the other actions would be usable.
- When someone shares a form

susnux:
As I have a use case for this, I started looking into it, this will require quite a lot of work:

- [ ] Nextcloud events for  form creation / modification / removal, submissions
- [ ] Workflow entities mapping those Nextcloud events
- [ ] Custom workflow checks to filter those entities
- [ ] (optional) some actions like sending an email

gotjoshua:
@susnux 
> I started looking into it,

me too! and i [started a repo ](https://github.com/onezoomin/forms_to_deck_flow)where i will try to create a flow from forms to deck.

However, I am a bit unsure how to proceed:
1. never hacked on a nextcloud plugin before  = D
2. none of the flow specific events are being emitted
3. hoping that i can listen for [this activity thingy](https://github.com/nextcloud/forms/blob/da50512aa70687752026e9433741677019db3f07/lib/Activity/ActivityManager.php#L114)

is this a reasonable idea?
any help?

BarBaar44:
Yes! Very much support this!

I have an intake form for new clients. I would like to create a new user based on the information in the intake form. It would be a super simple flow:

1. Share unique link to form with a potential client
2. The client fills out the form and submits
3. NextCloud creates a user account (etc) based on the input of the form (username == e-mail address)

Rello:
https://github.com/nextcloud/forms/pull/2195

I am working to add Forms as a direct data source to Analytics.
stay tuned...

QuantenToast:
What is the current status on this? I would like to contribute to this issue and would like to know what I can do, for example: should I work off the branch @susnux created or is that one a "dead end"? If so, what progress has already been made?

PR #2265, opened 2024-08-06
feat: Allow listening to form submissions via events and webhooks

marcelklehr:
requires Nextcloud 30

This will allow users of the upcoming windmill workflow automation app to trigger automations based on form submissions.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -1092,7 +1092,7 @@ class ApiController extends OCSController {
 		$this->formsService->setLastUpdatedTimestamp($formId);
 
 		//Create Activity
-		$this->formsService->notifyNewSubmission($form, $submission->getUserId());
+		$this->formsService->notifyNewSubmission($form, $submission);
 
 		if ($form->getFileId() !== null) {
 			try {
@@ -2362,7 +2362,7 @@ class ApiController extends OCSController {
 		$this->formsService->setLastUpdatedTimestamp($formId);
 
 		//Create Activity
-		$this->formsService->notifyNewSubmission($form, $submission->getUserId());
+		$this->formsService->notifyNewSubmission($form, $submission);
 
 		if ($form->getFileId() !== null) {
 			try {
--- a/lib/Events/AbstractFormEvent.php
+++ b/lib/Events/AbstractFormEvent.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace OCA\Forms\Events;
+
+use OCA\Forms\Db\Form;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IWebhookCompatibleEvent;
+
+if (interface_exists(\OCP\EventDispatcher\IWebhookCompatibleEvent::class)) {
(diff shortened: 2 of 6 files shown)


susnux:
> requires Nextcloud 30

So this can only be part of the next planned major version 5.0 as we still support Nextcloud 29 an 28, while there are already plans for 5.0 we still need to work on some features / technical dept.

lib/Events/AbstractFormEvent.php
```diff
@@ -0,0 +1,26 @@
...
+use OCA\Forms\Db\Form;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IWebhookCompatibleEvent;
+
+abstract class AbstractFormEvent extends Event implements IWebhookCompatibleEvent {
+	protected Form $form;
+
+	public function __construct() {
+		parent::__construct();
+	}
```
come-nc, line 14:
```suggestion
	public function __construct(
		protected Form $form,
	) {
		parent::__construct();
	}
```
What is the reason for not setting this property in the constructor here rather than in the child class?

lib/Controller/ApiController.php
```diff
@@ -1180,7 +1180,7 @@ public function insertSubmission(int $formId, array $answers, string $shareHash
 		$this->formsService->setLastUpdatedTimestamp($formId);
 
 		//Create Activity
-		$this->formsService->notifyNewSubmission($form, $submission->getUserId());
+		$this->formsService->notifyNewSubmission($form, $submission->getUserId(), $submission);
```
susnux, line 1183: Duplicated information. If the whole submission is passed then we do not need the userid.
```suggestion
		$this->formsService->notifyNewSubmission($form, $submission);
```

So the `notifyNewSubmission` needs of course to be adjusted

susnux (commented):
Nice side effect: We can use this for **flow** support besides **windmill** :)

marcelklehr:
I removed the requirement of Nextcloud 30, so we can hopefully merge this sooner :pray:

Chartman123:
@marcelklehr nice :) please fix the errors, then we can merge it :)

Chartman123 (approved):

Chartman123:
@come-nc @marcelklehr Can this be used to solve issues #1850, #1772, #1748, #1223 and/or #805?

susnux:
Yes and no, this is about windmill webhooks based workflows.
That allows to do even more than with Nextcloud flow but requires you to also host Windmill, so this is nothing that can be easily done on "plain" hosting.


Meaning the first 3 issues would I still consider feature requests, similar to the analytics integration - if someone wants to create an integration (and add proper tests) we could integrate that.
The last one ... well the new windmill integration is called "flow" but that ticket is more about workflowengine flow which is the Nextcloud built-in alternative that does not require external API, maybe still something to consider but would also need work on our side.

only #1223 could be done with Nextcloud 30 due to the `webhook_listeners` app integrate with Nextcloud now.
You should be able to register webhooks based on the forms submitted event. I did not test this yet, but probably something we could document and then close the request :)

PR #2269, opened 2024-08-07
fix: Add brute force protection to form endpoints

susnux:
Endpoints that query for forms are now protected against brute force attacks to find valid forms, invalid hashes or IDs.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -1604,7 +1604,7 @@ class ApiController extends OCSController {
 		} finally {
 			// Now forbid, if no public share and no direct share.
 			if (!$isPublicShare && !$this->formsService->hasUserAccess($form)) {
-				throw new NoSuchFormException('Not allowed to access this form');
+				throw new NoSuchFormException('Not allowed to access this form', Http::STATUS_FORBIDDEN);
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


Chartman123:
Could you perhaps base this on my api PR?

Chartman123:
@susnux thanks to @provokateurin we could finally merge the OpenAPI PR, so I think that you can now go on with this PR and base it on the current main

susnux:
@Chartman123 rebased and adjusted to the new controller

Chartman123:
Should we add the brute force protection to the page controller for public share hashes too?

lib/Controller/ApiController.php
```diff
@@ -1634,7 +1604,7 @@ private function loadFormForSubmission(int $formId, string $shareHash): Form {
 		} finally {
 			// Now forbid, if no public share and no direct share.
 			if (!$isPublicShare && !$this->formsService->hasUserAccess($form)) {
-				throw new OCSForbiddenException('Not allowed to access this form');
+				throw new NoSuchFormException('Not allowed to access this form');
```
Chartman123, line 1607:
```suggestion
				throw new NoSuchFormException('Not allowed to access this form', Http::STATUS_FORBIDDEN);
```

provokateurin (commented):
Changes look good, but I have little knowledge about the Forms app, so I have to leave it to someone else.

lib/Controller/ApiController.php
susnux: makes sense!

Chartman123 (approved):
Should be fine now :) Just squash the commits and then we're good to go. And we can add the protection in the PageController in a follow-up :)

PR #2270, opened 2024-08-08
chore: unify __construct functions

Chartman123:


susnux (commented):
Super nice refactoring!

Chartman123:
I also tried to do it for the DB classes but got lots of errors so I left them as they are for now

lib/Controller/ShareApiController.php
```diff
@@ -55,7 +55,7 @@
 
 class ShareApiController extends OCSController {
 	private IUser $currentUser;
-
+
```
susnux, line 58:
```suggestion

```

susnux: (white space - best run `cs:fix` locally)

lib/Service/SubmissionService.php
```diff
@@ -60,66 +58,24 @@
...
+		private IL10N $l10n,
+		private LoggerInterface $logger,
+		private IUserManager $userManager,
 		IUserSession $userSession,
-		IMailer $mailer,
+		private IMailer $mailer,
 		private ITempManager $tempManager,
 		private FormsService $formsService,
 		private IUrlGenerator $urlGenerator,
 		private UploadedFileMapper $uploadedFileMapper,
```
susnux, line 77: Not a blocker but as you already changed the order here, maybe make it a bit nicer and reorder the `UploadedFileMapper` to the other mappers (so e.g. below the answer mapper)?

susnux (approved):
Except for the whitespace issue this looks good!

susnux (approved):

PR #2277, opened 2024-08-11
feat: add warning about removing legacy links

### LINKED ISSUE #2071: Deprecation of `legacyLink` access
Mark `legacyLink` access as deprecated in the docs and in the frontend

Chartman123:
This fixes #2071 by adding a warning NoteCard to submit/create view for forms with legacyLink enabled.

grafik

### DIFF
--- a/src/components/SidebarTabs/SharingSidebarTab.vue
+++ b/src/components/SidebarTabs/SharingSidebarTab.vue
@@ -130,7 +130,7 @@
 				v-tooltip="
 					t(
 						'forms',
-						'For compatibility with the old Sharing, the internal link is still usable as Share link. We recommend replacing the link with a new Share link.',
+						'For compatibility with the old Sharing, the internal link is still usable as Share link. Please replace the link with a new Share link. The internal sharing link will not work anymore starting with Forms 5.0',
 					)
 				"
 				class="share-div__legacy-warning">
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -62,6 +62,19 @@
 				<p v-if="infoMessage" class="info-message">
 					{{ infoMessage }}
 				</p>
+				<!-- TODO: remove with Forms 5.0
+				 Show info about legacyLink that will be removed -->
+				<NcNoteCard
+					v-if="form.access.legacyLink"
+					type="warning"
+					:heading="t('forms', 'Legacy link in use')">
+					{{
+						t(
+							'forms',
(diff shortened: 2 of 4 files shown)


src/components/SidebarTabs/SharingSidebarTab.vue
```diff
@@ -130,7 +130,7 @@
 				v-tooltip="
 					t(
 						'forms',
-						'For compatibility with the old Sharing, the internal link is still usable as Share link. We recommend replacing the link with a new Share link.',
+						'For compatibility with the old Sharing, the internal link is still usable as Share link. Please replace the link with a new Share link. The internal sharing link won\'t work anymore starting with Forms 5.0',
```
susnux, line 133: I am not sure about wording, do we use abbreviations like `won't` vs `will not` ?

src/views/Submit.vue
```diff
@@ -62,6 +62,19 @@
 				<p v-if="infoMessage" class="info-message">
 					{{ infoMessage }}
 				</p>
+				<!-- TODO: remove with Forms 5.0
+				 Show info about legacyLink that will be removed -->
+				<NcNoteCard
```
susnux, line 67: Should we show this in the Submit view? Maybe only show it here if the current user has edit permissions?

susnux (approved):
Looks good, two comments but no blockers

src/views/Submit.vue
Chartman123: I was unsure about it, too. Here's why I added in the end: if it's an old form that still has legacyLink enabled, it probably also hasn't been modified for a longer time, so the message doesn't get read... So submit and/or results does make sense to get the attention.

Alternatively we could also add a Readme file to show the changes to the users after updating.

susnux: Yes having a changelog for this would be great, we can use the user-visible one now: https://docs.nextcloud.com/server/latest/developer_manual/app_development/info.html#changelog

Chartman123: Yes, that's what I thought of ;)

src/components/SidebarTabs/SharingSidebarTab.vue
Chartman123: Fixed :)

src/views/Submit.vue
Chartman123: As the Changelog for users only works in 29+, I'd like to keep the warning in the submit page. :) But I added it for making our changes more prominent 👍🏻

PR #2280, opened 2024-08-11
fix: add `target="_blank"` to links in description

Chartman123:
This closes #1680 by adding a default target `_blank` to the markdown link renderer.

Code is from example here: https://github.com/markdown-it/markdown-it/blob/master/docs/architecture.md#renderer

Signed-off-by: Christian Hartmann <chris-hartmann@gmx.de>

PR #2283, opened 2024-08-14
test: Add e2e tests for form description

susnux:
Added general tests and a regression test for https://github.com/nextcloud/forms/pull/2280

BTW if you wonder what the `annotation` in the test case description is good for: It provides this nice link if you view the results and wonder why that test was added ^^

image


Chartman123 (approved):

susnux:
> As I just noticed: should we change all files to the new copyright headers?

Some day yes, but no priority. Goal would be to be REUSE compatible

PR #2291, opened 2024-08-20
fix(submit): `access` is unset for public forms so check for existance first

susnux:
On public forms `form.access` is undefined to optional chain the access to `legacyLink`


AndyScherzinger (approved):
# 🐘

Chartman123 (approved):

PR #2293, opened 2024-08-20
fix: DI `userId` instead of `userSession`

susnux:
* Resolves https://github.com/nextcloud/forms/pull/2270#issuecomment-2299647167

The `ActivityManager` only needs the user ID so we just can inject the `userId`. Also make sure that it can be null, because the `ActivityManager` is DI in the `FormsService` and the `FormsService` is used also for public forms where no user is logged in.


nickvergessen:
Config service has a similar issue, didn't check open PRs before sending https://github.com/nextcloud/forms/pull/2294

Chartman123:
> Config service has a similar issue, didn't check open PRs before sending #2294

Yes, we're already aware... 🙂 Perhaps @susnux can fix it already, otherwise I'll have a look at it this evening.

susnux:
Fix is included now

Chartman123 (approved):

PR #2294, opened 2024-08-21
fix: Allow guests to open public pages again

nickvergessen:
### Steps

1. Create a public form
2. Open it as a guest
3. :boom: 

```
Cannot assign null to property OCA\Forms\Service\ConfigService::$currentUser of type OCP\IUser

// After fixing that
Cannot assign null to property OCA\Forms\Activity\ActivityManager::$currentUser of type OCP\IUser
```

Not sure if you want to apply some hardening to prevent methods from calling things on `currentUser` which can be null now, so please take over @susnux


nickvergessen:
Fixed via #2293

PR #2306, opened 2024-08-31
enh: make show to all users an admin setting

Chartman123:
- Permit to all can be controlled by an admin setting
- Makes sense to me to control also "show to all" using an admin setting

Replaces original PR #1906 by @akhil1508 where we didn't get any response anymore

### DIFF
--- a/src/FormsSettings.vue
+++ b/src/FormsSettings.vue
@@ -56,6 +56,18 @@
 				@update:checked="onAllowPermitAllChange">
 				{{ t('forms', 'Allow sharing to all logged in accounts') }}
 			</NcCheckboxRadioSwitch>
+			<NcCheckboxRadioSwitch
+				ref="switchAllowShowToAll"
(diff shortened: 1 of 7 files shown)


src/FormsSettings.vue
```diff
@@ -124,6 +136,12 @@ export default {
 			await this.saveAppConfig('allowPermitAll', newVal)
 			el.loading = false
 		},
+		async onAllowShowToAllChange(newVal) {
+			const el = this.$refs.switchAllowShowToAll
+			el.loading = true
```
susnux, line 141: thats hacky :sweat_smile: 
(I know we do the same above^^)

susnux (approved):
makes sense

src/FormsSettings.vue
Chartman123: We can improve it later on :)

PR #2314, opened 2024-09-09
chore: Refactor lastUpdated

Chartman123:
This commit refactors the form creation and update logic in the `ApiController` class. It removes the unnecessary setting of the `created` and `lastUpdated` timestamps in the `Form` entity, as these values are now automatically set in the `FormMapper` class. This improves code readability and reduces redundancy.

The changes also include updates to the `FormMapper` class, where the `insert` and `update` methods now automatically set the `created` and `lastUpdated` timestamps respectively.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -145,7 +145,6 @@ class ApiController extends OCSController {
 			// Create Form
 			$form = new Form();
 			$form->setOwnerId($this->currentUser->getUID());
-			$form->setCreated(time());
 			$form->setHash($this->formsService->generateFormHash());
 			$form->setTitle('');
 			$form->setDescription('');
@@ -157,7 +156,6 @@ class ApiController extends OCSController {
 			$form->setShowExpiration(false);
 			$form->setExpires(0);
 			$form->setIsAnonymous(false);
-			$form->setLastUpdated(time());
 
 			$this->formMapper->insert($form);
 		} else {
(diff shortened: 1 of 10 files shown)


Chartman123:
@susnux do you have any idea why I get an Error 500 as soon as I open the Forms app?

```
Type: Error
Code: 0
Message: Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '512' frames
File: /srv/www/htdocs/nextcloud/lib/private/AppFramework/Utility/SimpleContainer.php
Line: 104
```

Chartman123:
@susnux I was able to resolve my issue (thanks to @come-nc)

This can now be reviewed. I currently needed a little hack in the ApiControllerTest to make them pass. (`state` and `lastUpdated` don't have the expected values in line 499. `null` instead of `0` and `0` instead of `123456789`).

Koc (approved):

Chartman123:
@Koc thanks for the approval. Do you perhaps have an idea how to fix the test in a better way? I'm no expert in tests 🙈😂

Koc:
Usually it is recommended to rely on `Psr\Clock\ClockInterface` instead of plain `time()` and use  [`MockClock`](https://symfony.com/doc/current/components/clock.html#mockclock) implementation in tests.

Another variant - you can mock native `time()` by [defining mocked function](https://stackoverflow.com/questions/39706887/phpunit-mock-an-inbuilt-function) inside same namespace.

But I think we can live with current approach.

Chartman123:
Yes we can leave it for now and add a to-do comment. What bothers me the most is that the time() mocking worked before but doesn't anymore. Probably because the timestamp generation is now moved from the ApiController to the FormMapper

PR #2319, opened 2024-09-11
feat: Ask for restarting submission if form was changed

Koc:
This is a follow-up for #2312 based on a comment https://github.com/nextcloud/forms/pull/2312#issuecomment-2336789918

https://github.com/user-attachments/assets/f1684beb-93b7-4de4-a2a9-3999d5fc78c0

### :gift: Bonus
Added reset button

### DIFF
--- a/src/views/Submit.vue
+++ b/src/views/Submit.vue
@@ -145,17 +145,31 @@
 						@keydown.ctrl.enter="onKeydownCtrlEnter"
 						@update:values="(values) => onUpdate(question, values)" />
 				</ul>
-				<NcButton
-					alignment="center-reverse"
-					class="submit-button"
-					:disabled="loading"
-					native-type="submit"
-					type="primary">
-					<template #icon>
-						<NcIconSvgWrapper :svg="IconSendSvg" />
-					</template>
-					{{ t('forms', 'Submit') }}
-				</NcButton>
+				<div class="form-buttons">
+					<NcButton
+						alignment="center-reverse"
+						class="submit-button"
+						:disabled="!hasAnswers"
+						native-type="reset"
+						type="tertiary-no-background"
+						@click.prevent="showClearFormDialog = true">
+						<template #icon>
+							<NcIconSvgWrapper :svg="IconRefreshSvg" />
+						</template>
+						{{ t('forms', 'Clear form') }}
+					</NcButton>
+					<NcButton
+						alignment="center-reverse"
+						class="submit-button"
+						:disabled="loading"
+						native-type="submit"
+						type="primary">
+						<template #icon>
+							<NcIconSvgWrapper :svg="IconSendSvg" />
+						</template>
+						{{ t('forms', 'Submit') }}
+					</NcButton>
+				</div>
 			</form>
 
 			<!-- Confirmation dialog if form is empty submitted -->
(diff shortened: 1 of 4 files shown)


src/views/Submit.vue
```diff
@@ -654,13 +695,20 @@ export default {
 			}
 		},
 
+		onResetSubmission() {
+			this.deleteFormFieldFromLocalStorage()
+			window.removeEventListener('beforeunload', this.beforeWindowUnload)
+			this.$router.go(this.$router.currentRoute)
```
Koc, line 701: we should do hard reload instead of just calling `this.resetData()` here because `values` prop passed into `Question*` components [parsed inside `mounted` hook](https://github.com/Koc/nextcloud-forms/blob/dfc949eb8132a425415d587c42c6d3a799d3ddf6/src/components/Questions/QuestionDropdown.vue#L179-L184) instead of using normal reactivity.

src/views/Submit.vue
```diff
@@ -179,6 +179,19 @@
 				:buttons="confirmLeaveFormButtons"
 				:can-close="false"
 				:close-on-click-outside="false" />
+			<!-- Confirmation dialog if form was changed -->
+			<NcDialog
+				:open.sync="showRestartSubmissionDialog"
+				:name="t('forms', 'Restart submission')"
```
Chartman123, line 185:
```suggestion
				:name="t('forms', 'Clear form')"
```

src/views/Submit.vue
```diff
@@ -179,6 +179,19 @@
...
+				:message="
+					t(
+						'forms',
+						'Warning the form has changed since you last visit, do you want to restart?',
```
Chartman123, line 189:
```suggestion
						'The form has changed since your last visit. Do you want to clear all answers?',
```

src/views/Submit.vue
```diff
@@ -437,17 +474,25 @@ export default {
...
+		}
+
 		// Public Views get their form by initial-state from parent. No fetch necessary.
 		if (this.publicView) {
 			this.isLoadingForm = false
 		} else {
-			this.fetchFullForm(this.form.id)
+			await this.fetchFullForm(this.form.id)
+
+			// Check if the form was changed since the last submission
```
Chartman123, line 488:
```suggestion
			// Check if the form has changed since storing the data to the local storage
```

src/views/Submit.vue
Chartman123: Ok, does it make sense to refactor the loading from local storage to use normal reactivity instead?

Chartman123 (commented):
Hi @Koc 

nice work. A few wordings changed :) Not tested yet, will do this this evening :)

src/views/Submit.vue
Koc: yeah, I will try. I'm happy that main issue with form submission already fixed, so we have time to play with refactoring

Chartman123: Would you mind to postpone this feature to Forms 5.0? We could then use the extra time to not only bring this warning but also use the reset logic to show a button next to the Submit button to manually clear all fields :) At least Google Forms has this function, too.

Koc:
:spiral_notepad: Changes:
- fixed code review comments
- added reset button
- added :movie_camera: video with demo

cc @Chartman123

src/views/Submit.vue
```diff
@@ -145,17 +145,31 @@
...
+						alignment="center-reverse"
+						class="submit-button"
+						:disabled="!hasAnswers"
+						native-type="reset"
+						type="tertiary-no-background"
+						@click.prevent="showRestartSubmissionDialog = true">
+						<template #icon>
+							<NcIconSvgWrapper :svg="IconRefreshSvg" />
+						</template>
+						{{ t('forms', 'Restart submission') }}
```
Chartman123, line 159: We should use the "clear form" string here as well to be more consistent.

```suggestion
						{{ t('forms', 'Clear form') }}
```

src/views/Submit.vue
```diff
@@ -413,6 +456,29 @@ export default {
...
+		/**
+		 * Buttons for the "confirm restart submission" dialog
+		 */
+		confirmRestartSubmissionButtons() {
+			return [
+				{
+					label: t('forms', 'Abort'),
+					icon: IconCancelSvg,
+					callback: () => {},
+				},
```
Chartman123, line 469: I think tertiary is enough here.
```suggestion
				{
					label: t('forms', 'Abort'),
					icon: IconCancelSvg,
					type: tertiary,
					callback: () => {},
				},
```

Chartman123 (commented):
Two little comments, will test it later :)

src/views/Submit.vue
Koc: ok, but we have other modals with same "Abort" button and without this `type="tertiary"`

Chartman123: True... Didn't check all dialogs. Then let's keep secondary 👍🏻

Koc: reverted to be consistent with other dialogs

Chartman123 (approved):

PR #2333, opened 2024-09-23
feat: Allow to reorder options for "multiple" question type in backend

Chartman123:
This uses backend part of the original PR #2092 by @susnux to include it in the 4.3 release. The frontend still needs some more love and will be part of 5.0

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -764,12 +764,22 @@ class ApiController extends OCSController {
 			throw new OCSBadRequestException();
 		}
 
+		// Retrieve all options sorted by 'order'. Takes the order of the last array-element and adds one.
+		$options = $this->optionMapper->findByQuestion($questionId);
+		$lastOption = array_pop($options);
+		if ($lastOption) {
+			$optionOrder = $lastOption->getOrder() + 1;
+		} else {
+			$optionOrder = 1;
+		}
+
 		$addedOptions = [];
 		foreach ($optionTexts as $text) {
 			$option = new Option();
 
 			$option->setQuestionId($questionId);
 			$option->setText($text);
+			$option->setOrder($optionOrder++);
 
 			try {
 				$option = $this->optionMapper->insert($option);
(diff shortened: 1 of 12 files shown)


Chartman123:
First part for closing #354

Chartman123:
@Koc @susnux I think this is ready now, please review :) I would then create a new release 4.3 :)

lib/Migration/Version040300Date20240420155356.php
```diff
@@ -0,0 +1,65 @@
...
+ * Add "order" column for options
+ */
+class Version040300Date20240420155356 extends SimpleMigrationStep {
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
```
hamza221, line 45: Add a unique index for (questionID, order)?

hamza221:
Maybe also a repair step to add order to existing options instead of setting it as NULL ?

lib/Controller/ApiController.php
```diff
@@ -889,11 +899,104 @@ public function deleteOption(int $formId, int $questionId, int $optionId): DataR
 		}
 
 		$this->optionMapper->delete($option);
+
+		// Reorder the remaining options
+		$options = array_values($this->optionMapper->findByQuestion($questionId));
```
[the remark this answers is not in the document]
Chartman123: You're inside the `deleteOption` function... if you have options 1, 2, 3, 4 and delete option 3, this returns 1, 2, 3 as the new order and not 1, 2, 4

lib/Migration/Version040300Date20240420155356.php
Chartman123: Yes, makes sense. We already have an index for questions with (formId, order)

Chartman123:
> Maybe also a repair step to add order to existing options instead of setting it as NULL ?

Good idea :) But maybe as a follow up in #2092.

lib/Controller/ApiController.php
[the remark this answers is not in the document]
hamza221: My bad, Github's diff UI still confuses me somethimes

lib/Controller/ApiController.php
```diff
@@ -764,12 +764,22 @@ public function newOption(int $formId, int $questionId, array $optionTexts): Dat
 			throw new OCSBadRequestException();
 		}
 
+		// Retrieve all options sorted by 'order'. Takes the order of the last array-element and adds one.
+		$options = $this->optionMapper->findByQuestion($questionId);
+		$lastOption = array_pop($options);
+		if ($lastOption) {
+			$optionOrder = $lastOption->getOrder() + 1;
```
hamza221, line 771: I have a problem with this, if the repair step is not added, this can be NULL+1. 
But since you said it will be added in the next PR I will approve

hamza221 (approved):

lib/Controller/ApiController.php
Chartman123: Yes, I see. But that code is also copied from question reordering and works there without any problems?!

PR #2335, opened 2024-09-23
chore: Refactor checking if user is allowed to edit the form

Chartman123:


susnux (approved):
did not test this in depth but code looks fine to me!

Chartman123:
Yes, I just searched where the two checks from getFormIfAllowed occured and replaced them with it :)

PR #2343, opened 2024-10-04
fix(export): remove new lines from form title in the exported filename

tcitworld:
See https://github.com/nextcloud/forms/pull/1758#issuecomment-1931252574

Even though new lines are valid POSIX, let's avoid them (because it seems Nextcloud Files really doesn't like them in filenames).


hamza221 (approved):

Chartman123 (approved):
Great, thank you :)

PR #2353, opened 2024-10-05
chore: use attributes to define routes

Chartman123:
This pull request introduces significant changes to the API routing and controller structure by centralizing constants and updating route definitions. The most important changes include moving the route definition from routes.php to the Controller attributes, and the introduction of constants for API base paths and requirements.

These changes enhance maintainability by centralizing configuration and leveraging modern PHP features for cleaner and more readable code.


Chartman123:
@susnux ready for review now :)

appinfo/routes.php
```diff
@@ -25,83 +25,14 @@
...
-		// Internal Form Link
-		['name' => 'page#internal_link_view', 'url' => '/{hash}', 'verb' => 'GET'],
-		// App Root
-		['name' => 'page#index', 'url' => '/', 'verb' => 'GET'],
-	],
-
 	'ocs' => [
 		// CORS Preflight
-		['name' => 'api#preflightedCors', 'url' => $apiBase . '{path}', 'verb' => 'OPTIONS', 'requirements' => [
+		['name' => 'api#preflightedCors', 'url' => Constants::API_BASE . '{path}', 'verb' => 'OPTIONS', 'requirements' => [
```
susnux, line 33: this can not be migrated?

Chartman123: I didn't find a corresponding documentation... 🙈

lib/Constants.php
```diff
@@ -26,6 +26,20 @@
 use OCP\Share\IShare;
 
 class Constants {
+	/**
+	 * Constants for API Controllers
+	 */
+	public const API_BASE = '/api/{apiVersion}/';
+	public const API_V3_REQUIREMENTS = [
```
susnux, line 33: Should we also have the version in the constant name? I mean if we some when go with v4 we would have to rename all places?
Maybe just

```suggestion
	public const API_REQUIREMENTS = [
```

susnux (approved):
Just one cosmetic comment, otherwise this is super nice :)

lib/Constants.php
Chartman123: I would merge it like this for now because we need to remove the constants again for making the openapi-extractor work. So this name will only last for a short period of time

PR #2355, opened 2024-10-05
chore: remove legacy link support

Chartman123:


lib/Constants.php
```diff
@@ -68,16 +68,13 @@ class Constants {
 	public const FORM_ACCESS_NOPUBLICSHARE = 0;
 	public const FORM_ACCESS_PERMITALLUSERS = 1;
 	public const FORM_ACCESS_SHOWTOALLUSERS = 2;
+	// still needed for Migrations
```
susnux, line 71:
```suggestion
	/** @deprecated 5.0.0 still needed for Migrations
```

lib/Migration/Version050000Date20241005173955.php
```diff
@@ -0,0 +1,40 @@
...
+	
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		$qbUpdate = $this->db->getQueryBuilder();
+
+		$qbUpdate->update('forms_v2_forms')
```
susnux, line 33: you also need to execute this then

Chartman123: Good point 🙈😂

Chartman123:
@susnx all comments should be fixed now, thanks for the review :)

susnux (approved):
looks good and a quick tests also worked :)

PR #2397, opened 2024-11-05
(fix) default timezone for export

AIlkiv:
In the server/lib/base.php file of the Nextcloud server, UTC is always set as the default timezone:

https://github.com/nextcloud/server/blob/4a44d6a6774d79a8c91cab5f94acafa271b6d4d6/lib/base.php#L609

The main argument for this choice seems to be to simplify log reading, as discussed in this issue: https://github.com/nextcloud/server/issues/3553

However, this hardcoding of UTC affects all instances where date_default_timezone_get is used, resulting in UTC even when it would be more intuitive to use the server's timezone or the one set in config.php under default_timezone.

This pull request corrects the timestamp values in export files, aligning them with the expected timezone settings. The fix is informed by the solutions implemented in other parts of the system where similar timezone issues are handled, as seen in references like:

https://github.com/search?q=org%3Anextcloud+%22%3EgetDefaultTimeZone%28%22&type=code

### DIFF
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -207,7 +207,7 @@ class SubmissionService {
 		$submissionEntities = array_reverse($submissionEntities);
 
 		$questions = $this->questionMapper->findByForm($form->getId());
-		$defaultTimeZone = date_default_timezone_get();
+		$defaultTimeZone = $this->config->getSystemValueString('default_timezone', 'UTC');
 
 		if (!$this->currentUser) {
 			$userTimezone = $this->config->getUserValue($form->getOwnerId(), 'core', 'timezone', $defaultTimeZone);
--- a/tests/Unit/Service/SubmissionServiceTest.php
+++ b/tests/Unit/Service/SubmissionServiceTest.php
@@ -579,7 +579,11 @@ file2.txt"
 				return $questionEntities;
 			}));
 
-		date_default_timezone_set('Europe/Berlin');
+		$this->config->expects($this->once())
+			->method('getSystemValueString')
+			->with('default_timezone', 'UTC')
+			->willReturn('Europe/Berlin');
+
 		$this->config->expects($this->once())
 			->method('getUserValue')
 			->with('currentUser', 'core', 'timezone', 'Europe/Berlin')


Chartman123:
@AIlkiv We've discussed the usage of timezones quite a few times in the past and then agreed on using UTC for the exports as far as I remember...

@susnux any other thoughts on this one?

Either way, you'll have to fix the errors first... ;)

susnux:
> We've discussed the usage of timezones quite a few times in the past and then agreed on using UTC for the exports as far as I remember...

I fully agree, exports are data and thus should be UTC for convenience, so you can do what ever you like afterwards with it while you are *processing* the data.
**But** in this case we are already formatting the date using the user timezone and fallback to server timezone (which is always UTC in this case), so I would say to correct this existing behavior we should fix it as suggested in this PR.

lib/Service/SubmissionService.php
susnux, line 1: Please run `composer cs:fix` locally.

Chartman123:
> as suggested in this PR

ok, also the import of OCA\DAV? this will introduce another dependency on an app. Is there no OCP import available for that?

lib/Service/SubmissionService.php
```diff
@@ -36,6 +36,7 @@
 use OCA\Forms\Db\QuestionMapper;
 use OCA\Forms\Db\SubmissionMapper;
 use OCA\Forms\Db\UploadedFileMapper;
+use OCA\DAV\CalDAV\TimezoneService;
```
susnux, line 39: Oh did not noticed this was from OCA. No please do not use OCA here, that is private API and should not be used by other apps.

lib/Service/SubmissionService.php
```diff
@@ -207,7 +209,7 @@ public function getSubmissionsData(Form $form, string $fileFormat, ?File $file =
 		$submissionEntities = array_reverse($submissionEntities);
 
 		$questions = $this->questionMapper->findByForm($form->getId());
-		$defaultTimeZone = date_default_timezone_get();
```
susnux, line 210: This is private API, please directly use something like

```
$this->config->getSystemValueString('default_timezone', 'UTC')
```

susnux (changes requested):

susnux (approved):

Chartman123 (approved):

Issue #2165, opened 2024-05-13
Allow synchronous form answering and spreadsheet editing


artonge:
**Nextcloud (please complete the following information):**
- Nextcloud-Version: [e.g. 19.0.0]
- Forms-Version: [e.g. 2.0.0-beta4]

**Is your feature request related to a problem? Please describe.**

When you open the spreadsheet of a form, the file gets locked, and it prevents people from answering the form.

**Describe the solution you'd like**

- Store the submitted data in the database
- Try to add the data in the spreadsheet later in a background job

@susnux

Chartman123:
@artonge could you verify which error message you get, when the file is locked? There's another open issue, where we still don't know exactly where the error comes from (#2067)

alexanderdd:
maybe related: https://github.com/nextcloud/forms/issues/1970

alexanderdd:
I just had this problem in 29.0.6

Ideally, the data is not added after manual editing is finished, but live, during editing. Then you have live monitoring of answers coming in =)

Chartman123:
> Ideally, the data is not added after manual editing is finished, but live, during editing. Then you have live monitoring of answers coming in =)

@alexanderdd I think that this can't be done easily as we don't know if someone uses an online office like Collabora/Nextcloud office or OnlyOffice. And even if they have the app installed the file could still be locked by someone working on it through a client locally.

susnux:
> @alexanderdd I think that this can't be done easily as we don't know if someone uses an online office like Collabora/Nextcloud office or OnlyOffice. And even if they have the app installed the file could still be locked by someone working on it through a client locally.

If we do it async (background job) we can check for locking as those apps should lock the file for modifications.
So nothing we need to special handle, but just catch locked exceptions and postpone writing the results.

Issue #2067, opened 2024-04-15
Failed to export Submissions: Not allowed to write to file


chriscroome:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**

A Nextcloud instance that did have a working Forms app now generates an error on form submission.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the form
2. Complete it
3. Click _"Submit"_
5. The use has the error _"There was an error submitting the form"_

**Expected behavior**

A successfully submission. 

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 28.0.4.1
- Forms-Version: 4.2.0

**Desktop (please complete the following information):**
 - OS: Debian
 - Browser Firefox

**Smartphone (please complete the following information):**
 - Browser Firefox/
 - Version 126.0

**Browser log**
```
[ERROR] forms: Error while submitting the form 
Object { app: "forms", level: 2, error: {…} }
NcSettingsSection-NSrCZ23G-6nKKlVz9.mjs:2:337806
    value https://cloud.example.org/apps/forms/js/NcSettingsSection-NSrCZ23G-6nKKlVz9.mjs:2
    value https://cloud.example.org/apps/forms/js/NcSettingsSection-NSrCZ23G-6nKKlVz9.mjs:2
    onConfirmedSubmit https://cloud.example.org/apps/forms/js/Submit-PS6lxYc-.mjs:2
```

**Nextcloud log**

The following is written to the Nextcloud log:

> Failed to export Submissions: Not allowed to write to file

**Server logs**

There is nothing in the Apache or PHP-FPM logs.

**TMPDIR**

The `tempdirectory` is writable.

**Additional context**

The error that is produced appears to be here:

https://github.com/nextcloud/forms/blob/1ba3a1404c3a23eedd8010175b60f10c5a70f688/lib/Service/SubmissionService.php#L226

The change to the server (the forms were working prior to this change), was updating from Debian Bullseye to Debian Bookworm, the instance uses PHP-FPM 8.3.

I have tried to duplicate this on another server with the same configuration but haven't been able to.

Chartman123:
Did you restart your php processes? Are you using php-fpm? We had lots of errors in the past after upgrades due to cached code...

susnux:
> Browser log

[...]

PR #2408, opened 2024-11-13
feat: Refactor form sync to run as a background job with retry

AIlkiv:
issue: [#2165](https://github.com/nextcloud/forms/issues/2165) , https://github.com/nextcloud/forms/issues/2067

I implemented a logic to synchronize the form with the file, allowing multiple attempts if the first one fails. The process runs in the background with delays, ensuring it does not impact the user experience.

### DIFF
--- a/lib/BackgroundJob/SyncSubmissionsWithLinkedFileJob.php
+++ b/lib/BackgroundJob/SyncSubmissionsWithLinkedFileJob.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ * @copyright Copyright (c) 2024 Andrii Ilkiv <ailkiv@users.noreply.github.com>
+ *
+ * @author Andrii Ilkiv <ailkiv@users.noreply.github.com>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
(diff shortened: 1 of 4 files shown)


AIlkiv:
Added a fix for the issue described in https://github.com/nextcloud/forms/issues/2067

Chartman123 (approved):
Looks good to me :) I'd like to wait with the merging for @susnux's review

thanks for this PR, we can also do a backport to stable4, but you'll need to adjust the ApiController part there manually because of the duplicated routes.

Chartman123:
The Codecov upload error isn't specific to Forms: https://github.com/codecov/codecov-action/issues/1643

lib/BackgroundJob/SyncSubmissionsWithLinkedFileJob.php
```diff
@@ -0,0 +1,108 @@
...
+use Throwable;
+
+class SyncSubmissionsWithLinkedFileJob extends QueuedJob {
+	public const MAX_ATTEMPTS = 10;
+
+	public function __construct(
+		private FormMapper $formMapper,
+		private FormsService $formsService,
+		private SubmissionService $submissionService,
+		ITimeFactory $time,
```
susnux, line 47: I think it is better to have the inherited first

```suggestion
		ITimeFactory $time,
		private FormMapper $formMapper,
		private FormsService $formsService,
		private SubmissionService $submissionService,
```

susnux (approved):
I am not sure we need the max attempts, but otheriwse fine with me.

Please fix the code style in the constructor

lib/BackgroundJob/SyncSubmissionsWithLinkedFileJob.php
Koc: fixed

Koc:
@susnux @Chartman123 I did requested changes. All pipelines are green except 2: PHPUnit SQLite + coverage upload. But tests pass

Chartman123:
Might be because you made changes on a branch in a fork that doesn't belong to you. It could also help to rebase to current main

Koc:
Already did by using Github UI button :man_shrugging:

susnux:
It is only the codecov action that is broken for forks, tests are fine

Koc:
pipeline is completely green now :heavy_check_mark:

Koc:
I will take care about backport to `stable4` later today

Chartman123:
Nice, thank you :)

Koc:
here we go #2424

PR #2466, opened 2024-12-17
fix: failing tests

hamza221:
https://github.com/nextcloud/server/pull/48152
https://github.com/nextcloud/server/blob/master/lib/public/RichObjectStrings/IValidator.php#L60

### DIFF
--- a/tests/Unit/FormsMigratorTest.php
+++ b/tests/Unit/FormsMigratorTest.php
@@ -83,11 +83,6 @@ class FormsMigratorTest extends TestCase {
 	public function setUp(): void {
 		parent::setUp();
 
-		// UserMigration is not available below NC24, skip all tests here.
-		if (\OC_Util::getVersion()[0] < 24) {
-			$this->markTestSkipped('UserMigration not available below NC24');
-		}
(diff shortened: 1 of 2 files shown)


hamza221:
So if we use `OCP\ServerVersion` stable30 fails and if we use OC_Util::getVersion() master fails

tests/Unit/FormsMigratorTest.php
```diff
@@ -80,11 +81,16 @@ class FormsMigratorTest extends TestCase {
...
+
 	public function setUp(): void {
 		parent::setUp();
 
+		$serverVersion = \OCP\Server::get(ServerVersion::class);
+
 		// UserMigration is not available below NC24, skip all tests here.
-		if (\OC_Util::getVersion()[0] < 24) {
+		if ($serverVersion->getMajorVersion() < 24) {
 			$this->markTestSkipped('UserMigration not available below NC24');
```
Chartman123, line 94: We can remove the check here completely, NC24 is no longer supported :)

Chartman123 (approved):

PR #2470, opened 2024-12-17
Fix: Typo broke background sync

toad:
#2408 does not work because of a typo (wrong null coalescing operator). So on our existing installation, we get an undefined array key error here.


toad:
@Koc I created a pull request as you suggested. I guess since this passed tests when you committed it it needs more?

Chartman123 (approved):

Koc (approved):

PR #2479, opened 2025-01-03
feat: integration of unified search

### LINKED ISSUE #2392: Connect Forms to Unified Search by Implementing Search Provider

-   Nextcloud-Version: [30.0.1]
-   Forms-Version: [4.3.2]

On a nextcloud instance I'm using as a user, there is often a large number of both polls and forms that I can/am requested to fill out. Scrolling through the list to find the right one is sometimes a bit tiresome, especially if I have forgotten if what I'm looking for is a poll or a form.

I'd like to be able to search for forms by their title and ideally also their description text.
AFAIK one "only" needs to implement the `OCP\Search\IProvider` interface for that and register it. 

There is also `\OCP\Search\IFilteringProvider` which would allow to add additional filters like "already participated" or "expires soon". Not sure about the benefit of that though …

**Additional context**
The polls app is already doing this: https://github.com/nextcloud/polls/blob/04a447e81021882f93ea895616dde8dbab643f5d/lib/Provider/SearchProvider.php

Apart from that, thank you for the great app! It's a pleasure to use it :)

Chartman123:
This closes #2392 by implementing a SearchProvider that let's you search the titles and descriptions of the forms that are available to you.

### DIFF
--- a/lib/Db/FormMapper.php
+++ b/lib/Db/FormMapper.php
@@ -108,16 +108,16 @@ class FormMapper extends QBMapper {
 		// share type user and share with current user
 		$memberships->add(
 			$qbShares->expr()->andX(
-				$qbShares->expr()->eq('shares.share_type', $qbShares->createNamedParameter(IShare::TYPE_USER)),
-				$qbShares->expr()->eq('shares.share_with', $qbShares->createNamedParameter($userId, IQueryBuilder::PARAM_STR)),
+				$qbShares->expr()->eq('shares.share_type', $qbShares->createNamedParameter(IShare::TYPE_USER, IQueryBuilder::PARAM_STR, ':share_type_user')),
+				$qbShares->expr()->eq('shares.share_with', $qbShares->createNamedParameter($userId, IQueryBuilder::PARAM_STR, ':share_with_user')),
 			),
 		);
 		// share type group and one of the user groups
 		if (!empty($groups)) {
 			$memberships->add(
 				$qbShares->expr()->andX(
-					$qbShares->expr()->eq('shares.share_type', $qbShares->createNamedParameter(IShare::TYPE_GROUP)),
-					$qbShares->expr()->in('shares.share_with', $qbShares->createNamedParameter($groups, IQueryBuilder::PARAM_STR_ARRAY)),
+					$qbShares->expr()->eq('shares.share_type', $qbShares->createNamedParameter(IShare::TYPE_GROUP, IQueryBuilder::PARAM_STR, ':share_type_group')),
+					$qbShares->expr()->in('shares.share_with', $qbShares->createNamedParameter($groups, IQueryBuilder::PARAM_STR_ARRAY, ':share_with_groups')),
 				),
 			);
 		}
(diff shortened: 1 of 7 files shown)


Chartman123:
@hamza221 It's basically working, but apparently it doesn't return all matching forms. Perhaps you can find some time and  have a look

Chartman123:
@hamza221 I found the culprit: I used a simple like-operation for the DB query instead of the case insensitive iLike :)

hamza221:
I spent an hour on this and I'm ,not sure what's happening. I'm sure I'm missing something obvious 
The shared search is not working for me (when I remove the lines you added it returns all shared as expected) 
your code looks correct but checking the db logs I found this 
my search term is `%adham%`

`SELECT * 
FROM oc_forms_v2_forms 
WHERE (
    id IN (
        SELECT DISTINCT forms.id 
        FROM oc_forms_v2_forms forms 
        LEFT JOIN oc_forms_v2_shares shares ON forms.id = shares.form_id 
        WHERE (
            (
                (shares.share_type = '0' AND shares.share_with = 'admin') 
                OR (shares.share_type = '1' AND shares.share_with IN ('admin'))
            ) 
            OR (access_enum IN (2))
        ) 
        AND forms.owner_id <> 'admin'
    )
) 
AND (
    title COLLATE utf8mb4_general_ci LIKE '0' 
    OR description COLLATE utf8mb4_general_ci LIKE 'admin'
) 
ORDER BY last_updated DESC, created DESC;`

**See how the like value is wrong** 
Compared to the owned forms search 👇

`SELECT * FROM oc_forms_v2_forms 
WHERE (owner_id = 'admin') 
AND (title COLLATE utf8mb4_general_ci LIKE '%adham%' 
     OR description COLLATE utf8mb4_general_ci LIKE '%adham%') 
ORDER BY last_updated DESC, created DESC;`

Chartman123:
@hamza221 Which database are you using? My dev instance uses MariaDB as database... Perhaps it's related to the db type

Chartman123:
Ok, I found the error: https://github.com/nextcloud/forms/blob/f69a1161920f22505140e98e8a426c20f7d9e95e/lib/Db/FormMapper.php#L166

We set the parameters from `qbShares` to be the parameters of `qbForms`. So they get mixed up when I set the new parameters for the search...

lib/Db/FormMapper.php
```diff
@@ -155,17 +156,26 @@ public function findSharedForms(string $userId, array $groups = [], array $teams
 			)
 			->addOrderBy('last_updated', 'DESC')
 			->addOrderBy('created', 'DESC');
-
-		// We need to add the parameters from the shared forms IDs select to the final select query
-		$qbForms->setParameters($qbShares->getParameters(), $qbShares->getParameterTypes());
-
+	
+		if ($queryTerm) {
+			$qbForms->andWhere($qbForms->expr()->iLike('title', $qbForms->createNamedParameter('%' . $this->db->escapeLikeParameter($queryTerm) . '%', IQueryBuilder::PARAM_STR, ':query_term_title')) .
```
hamza221, line 161: Nit: use orX

lib/Db/FormMapper.php
```diff
@@ -177,6 +187,11 @@ public function findAllByOwnerId(string $ownerId): array {
 			->addOrderBy('last_updated', 'DESC')
 			->addOrderBy('created', 'DESC');
 
+		if ($queryTerm) {
+			$qb->andWhere($qb->expr()->iLike('title', $qb->createNamedParameter('%' . $this->db->escapeLikeParameter($queryTerm) . '%')) .
```
hamza221, line 191: here too

hamza221 (approved):
LGTM

lib/Db/FormMapper.php
Chartman123: Nice catch... Was too obviously 😆 I also extracted the search term now into a variable for better readability

PR #2490, opened 2025-01-14
feat: allow cloning archived forms

### LINKED ISSUE #2019: Allow to clone archived forms
Follow up on #1925 

It should be able to clone archived forms, the clone should not have any state set (default to `0` -> active).

Chartman123:
This closes #2019 by showing the button and handling the events in the frontend and by adjusting the backend to unset the state of the cloned form. It also adds a change to unset the expires timestamp when a form gets cloned.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -180,6 +180,7 @@ class ApiController extends OCSController {
 			unset($formData['id']);
 			unset($formData['created']);
 			unset($formData['lastUpdated']);
+			unset($formData['state']);
 			$formData['hash'] = $this->formsService->generateFormHash();
 			// TRANSLATORS Appendix to the form Title of a duplicated/copied form.
 			$formData['title'] .= ' - ' . $this->l10n->t('Copy');
--- a/src/Forms.vue
+++ b/src/Forms.vue
@@ -131,7 +131,10 @@
 		</template>
 
 		<!-- Archived forms modal -->
-		<ArchivedFormsModal :open.sync="showArchivedForms" :forms="archivedForms" />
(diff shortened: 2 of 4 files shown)


lib/Controller/ApiController.php
```diff
@@ -180,6 +180,8 @@ public function newForm(?int $fromId = null): DataResponse {
 			unset($formData['id']);
 			unset($formData['created']);
 			unset($formData['lastUpdated']);
+			unset($formData['state']);
+			unset($formData['expires']);
```
susnux, line 184: State of course, but not 100% sure about `expires`.
What if I want to clone forms to have the same expiry date?
(Not necessarily archived forms).

Chartman123: Ok, so only unset it if it's in the past?

susnux: Hm I am not sure, what if you want to clone a manually closed form?
Cloning implies for me that everything is the same.

So I would drop the `unset(expires)` at all and for `state` maybe some logic like `state = is state archived ? null : state`. (but unset might also be ok depending on your thought about the notes above).

Chartman123: I think for most of the use cases, unset would be the way to go to have a generally working copy of a form without having to adjust everything. I mean you already have to do the sharing and so on for the copies again. 

But for me it's ok to just drop the `expires` for this PR :)

susnux (approved):

lib/Controller/ApiController.php
susnux: Makes sense, lets see what users think of it.
If they complain we can revert / add that logic

PR #2505, opened 2025-01-20
chore: silence eslint warning

Chartman123:
This removes an eslint warning for the available named export of isCancel in @nextcloud/axios as we also need other methods from this package.

Signed-off-by: Christian Hartmann <chris-hartmann@gmx.de


susnux (commented):
why not import the named one?

Chartman123:
We also need axios.get... Can we import the named one and still use that?

susnux:
> We also need axios.get... Can we import the named one and still use that?

Sure:
```ts
import axios, { isCancel } from '@nextcloud/axios'
```

susnux (approved):

PR #2513, opened 2025-01-23
fix: Improve compatibility with Windows for uploaded files

### LINKED ISSUE #2409: The Forms folder structure is not sanitized which can lead to synchronization issues when uploading files

**Describe the bug**
Uploaded files are seemingly saved in a folder structure `Forms/{id of form} - {title of form}/{id of submission}/{id of question} - {title of question}` but the titles do not seem to get trimmed in regards to trailing white spaces, which results in syncing problems (at least on Windows 11 with NTFS).

**To Reproduce**

1. Go to Forms and create a new form
2. Call it `test ` with a trailing whitespace
3. Add a file upload option and call it `test ` as well
4. Upload a file

**Expected behavior**
I expect the fields that are used for filenames to be trimmed in regard to whitespace and maybe also sanitized regarding other invalid characters like `:`

**Screenshots**
Image

-   Nextcloud-Version: 30.0.2
-   Forms-Version: 4.3.3

-   OS: Windows 11 with NTFS file system
- Nexcloud Desktop v3.14.3

Koc:
Fixes #2409

### DIFF
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -785,6 +785,6 @@ class FormsService {
 	}
 
 	private static function normalizeFileName(string $fileName): string {
-		return str_replace([...mb_str_split(\OCP\Constants::FILENAME_INVALID_CHARS), "\n"], '-', $fileName);
+		return trim(str_replace(Constants::FILENAME_INVALID_CHARS, '-', $fileName));
 	}
 }
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -155,6 +155,19 @@ class Constants {
 		'x-office/spreadsheet',
 	];
 
+	public const FILENAME_INVALID_CHARS = [
+		"\n",
+		'/',
+		'\\',
+		':',
+		'*',
+		'?',
+		'"',
+		'<',
+		'>',
+		'|',
+	];
+
 	/**
 	 * !! Keep in sync with src/mixins/ShareTypes.js !!
 	 */


lib/Service/FormsService.php
```diff
@@ -44,6 +44,19 @@
  * @psalm-import-type FormsShare from ResponseDefinitions
  */
 class FormsService {
+	private const FILENAME_INVALID_CHARS = [
```
Chartman123, line 47: Could you move this to the Constants? And how about using the OCP constants and just combining it with the remaining ones?

Chartman123 (changes requested):

lib/Service/FormsService.php
susnux: Neither of it just use `OCP\Files\IFilenameValidator` and I do not think we should replace them but forbid upload of files containing them.
But also do it in the fronend:

[nextcloud-files validateFilename](https://nextcloud-libraries.github.io/nextcloud-files/functions/index.validateFilename.html)

Chartman123: Ok, didn't know about that :) But yes, then I agree, we should already use the validation upon uploading the files in the frontend and the backend, so that we don't have to manipulate the filenames afterwards.

Chartman123: But one thing remains: The folder structure is based on the question titles. So if you for example put a `:` in the title of a file question, the corresponding folders can't be sync'ed on Windows devices.

Koc: ok, so I've moved `FILENAME_INVALID_CHARS` into `Constants` class. Regarding other hints:

- combining `OCP\Constants::FILENAME_INVALID_CHARS` + `\OCA\Forms\Constants::FILENAME_INVALID_CHARS` has no sense for me because I've defined invalid chars as an array but global invalid characters defined as string. Array is better here because we don't need call `mb_str_split` all the time
- `IFilenameValidator` - we're already calling `$userFolder->getStorage()->verifyPath($path, $uploadedFile['name']);` during file upload. Is it not enough? Even more, as already mentioned, uploaded file can have correct name, but due to form/question name path can be invalid

Anyway, feel free to force-push any changes into my branch.

Koc: @Chartman123 @susnux mates, can you check one more time?

Chartman123 (approved):

PR #2530, opened 2025-01-28
Fix: Add disabled state to PillMenu and update Results view when there are no submissions

### LINKED ISSUE #2529: Can't unlink spreadsheet if there are no responses

**Describe the bug**
If there are no responses to a form, you currently can't unlink the spreadsheet file, because the action menu is not visible.

**To Reproduce**

1. Go to a form with no responses
2. Click on results tab
3. See button to link/open spreadsheet, but no menu to unlink the file.

**Expected behavior**
You should be able to unlink a file if there are no responses

**Screenshots**

- Nextcloud-Version: 30.0.5
- Forms-Version: 4.3.5

Chartman123:
This fixes #2529 by enabling the user to unlink a spreadsheet file if there are currently no submissions

### DIFF
--- a/src/components/PillMenu.vue
+++ b/src/components/PillMenu.vue
@@ -10,6 +10,7 @@
 			:key="option.id"
 			:aria-label="isMobile ? option.ariaLabel : null"
 			:checked="active.id"
+			:disabled="disabled"
 			class="pill-menu__toggle"
 			:class="{ 'pill-menu__toggle--icon-only': isMobile && option.icon }"
 			button-variant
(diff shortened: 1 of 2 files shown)


susnux (approved):
Works, but please note that is is not good for accessibility.
Screen reader users will usually now do not get any information about the pill menu.
(disabled will hide it) But I do not see an easy solution for now.

Chartman123:
> Works, but please note that is is not good for accessibility. Screen reader users will usually now do not get any information about the pill menu. (disabled will hide it) But I do not see an easy solution for now.

This shouldn't matter too much, because it is only disabled as long as it's not needed at all. But if it's better from an a11y perspective, we can also put it behind a `v-if` and only show it when there are submissions?

PR #2533, opened 2025-01-29
feat: Improve error messages for invalid submissions

Koc:
Right now it is really hard to understand for end users why form can't be submitted. This PR aims to improve validation messages.

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -1255,9 +1255,6 @@ class ApiController extends OCSController {
 		if (is_string($isSubmissionValid)) {
 			throw new OCSBadRequestException($isSubmissionValid);
 		}
-		if ($isSubmissionValid === false) {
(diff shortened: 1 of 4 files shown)


Chartman123 (approved):
Nice one :)

susnux (changes requested):
I dont think this makes sense, what is the use case?
It is nothing shown to end users, that should already be done in the frontend and if it is not then there is a bug in the frontend that should be fixed.

So this is for developers that use the API?
In this case we should refactor the function at all (`validateSubmission`) to be `void` and throw instead on error.
Either `InvalidArgumentException` if we keep this developers only I would prefer this.
Otherwise an `\OCP\HintException` which would allow translated error messages for frontend - but here again we should never allow to submit something invalid so a frontend validation is wrong.

Koc:
@susnux yeah, it's mostly for API usage. I've refactored method to throw an `InvalidArgumentException` as you suggested.

susnux (approved):

PR #2543, opened 2025-02-04
test(ci): speed up playwright by using container with browsers

max-nextcloud:


max-nextcloud:
Initializing the container is not significantly faster then just installing the browsers.
Plus the container does not include composer - to it fails to install the php dependencies.

PR #2544, opened 2025-02-04
test(ci): only install headless shell chromium for playwright

max-nextcloud:
Speeds up the installation of the browsers drastically:
| scenario | 1st run | 2nd run | 3rd run | avg |
| ---- | ---- | ---- | ---- | ---- |
|all browsers | 53 sec| 74 sec| 59 sec| ~ 62 sec|
|only chromium  --with-deps| 42 sec| 29 sec| 31 sec|  ~ 34 sec|
|chromium  --with-deps --only-shell| 22 sec| 18 sec | | ~ 20 sec|
|chromium --only-shell| 4 sec| 5 sec| |~ 4.5 sec|

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -44,7 +44,7 @@ jobs:
           TESTING=true npm run build --if-present
 
       - name: Install Playwright Browsers
-        run: npx playwright install --with-deps
+        run: npx playwright install chromium --only-shell
 
       - name: Run Playwright tests
         run: npx playwright test


Chartman123 (approved):

susnux:
I would like to revert this. Speed is no metric here but coverage. 

We had many regressions in Safri and co meaning browser specific. So we should test all different browsers.

Chartman123:
Ok for me. :) or are there any other improvements on the performance possible that include the use of other browsers?

max-nextcloud:
@susnux Even before this PR we the test were only run with chromium as that's the only browser defined in https://github.com/nextcloud/forms/blob/main/playwright.config.ts#L40-L46

I don't think Safari was installed before this PR either. My understanding is that it was firefox, webkit, chromium and headless chromium - of which only the latter was actually used.

susnux:
@max-nextcloud  sorry wrong project :see_no_evil: I was somehow thinking we test Safari but that nc-vue not forms where we test Safari.

PR #2545, opened 2025-02-05
fix(deps): use @nextcloud/e2e-test-server

max-nextcloud:
It replaces `@nextcloud/cypress` and drops the cypress peer dependency.


susnux:
Where can the source be found?

susnux (approved):

PR #2571, opened 2025-02-15
chore: Refactor imports to use the new component paths in Vue files

Chartman123:


src/Forms.vue
```diff
@@ -146,14 +146,14 @@ import axios from '@nextcloud/axios'
...
 import { useIsMobile } from '@nextcloud/vue'
-import NcAppContent from '@nextcloud/vue/dist/Components/NcAppContent.js'
-import NcAppNavigation from '@nextcloud/vue/dist/Components/NcAppNavigation.js'
-import NcAppNavigationCaption from '@nextcloud/vue/dist/Components/NcAppNavigationCaption.js'
-import NcAppNavigationNew from '@nextcloud/vue/dist/Components/NcAppNavigationNew.js'
-import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
-import NcContent from '@nextcloud/vue/dist/Components/NcContent.js'
-import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
-import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
+import NcAppContent from '@nextcloud/vue/components/NcAppContent'
```
Koc, line 149: during my work with Nextcloud I always wondering: why can't we use components auto-import instead of writing tons of this imports?

Chartman123: I just took it for granted...

susnux: What so you mean with auto import @Koc ?

susnux (approved):

src/Forms.vue
Koc: @susnux 

- https://blog.logrocket.com/how-to-automatically-import-register-vue-components/
- https://nuxt.com/docs/guide/concepts/auto-imports

susnux: It makes it more explicit which components will be used and especially where they come from.
Basically the same reasons why I hate mixins: They hide the origin of properties and methods (and other reasons).

I personally even do not like some of the pre-compiler magic they built into script setup recently, like this:
https://vuejs.org/guide/components/props.html#reactive-props-destructure
It makes you think things are in a way (this example reactive objects) they are not (e.g. the prop is not reactive but its just sugar for `props.foo` which is not reactive but only the value).

You can always hide the import block in your IDE but at least the full recipe of your component is stated with all its dependencies.

PR #2574, opened 2025-02-17
fix: positioning of validation type menu toggle in QuestionShort component

Chartman123:
This fixes #2547 by adjusting the position of the toggle

Signed-off-by: Christian Hartmann <chris-hartmann@gmx.de>

PR #2578, opened 2025-02-18
Fix misaligned form elements

### LINKED ISSUE #2403: Elements are not consistently aligned

**Describe the bug**
Currently the elements are not consistently aligned

**To Reproduce**

1. Go to submit/create/results view and see misaligned elements

**Expected behavior**
There should be no misalignments

**Screenshots**
1. Form title/description and Questions in Submit view:
Image

2. Question title and options too wide:
Image

3. Form title/description and Results:
Image

Elsensee:
Closes #2403

I’m very sorry for the delay. This was a tough one to get right, this should probably not get backported to v29, as it might break there.

| Before | After |
| ------- | ----- |
| Screenshot 2025-02-18 at 05-14-03 Test - Formulare - Nextcloud | Screenshot 2025-02-18 at 04-51-02 Test - Formulare - Nextcloud |
| Screenshot 2025-02-18 at 05-14-24 Test - Formulare - Nextcloud | Screenshot 2025-02-18 at 04-50-50 Test - Formulare - Nextcloud |

Notice the error fitting with some room to breath, and the validation menu next to the short answer „reappearing“

Very sorry again for the delay, I know you have a release date! I hope this is still early enough. :)

### DIFF
--- a/src/components/Icons/IconOverlay.vue
+++ b/src/components/Icons/IconOverlay.vue
@@ -21,8 +21,8 @@
 
 .icon-overlay {
 	position: absolute;
-	inset-inline-start: 2px;
-	inset-block-start: 2px;
+	inset-inline-start: 0;
+	inset-block-start: -2px;
 
 	transform: scale(0.7);
 }
--- a/src/components/Questions/AnswerInput.vue
+++ b/src/components/Questions/AnswerInput.vue
@@ -251,7 +251,7 @@ export default {
 	}
 
 	.question__input {
-		width: 100%;
+		width: calc(100% - var(--default-clickable-area));
 		position: relative;
 		inset-inline-start: -12px;
 		margin-inline-end: -12px !important;
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -375,7 +375,7 @@ export default {
 	&__drag-handle {
 		position: absolute;
 		display: flex;
-		inset-inline-start: 0;
+		inset-inline-start: var(--default-grid-baseline);
 		flex-direction: column;
 		justify-content: center;
 		gap: 12px;
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -451,8 +451,13 @@ export default {
 
 			&__warning {
 				margin-block: auto;
-				margin-inline: 4px 12px;
+				margin-inline: 8px;
 				color: var(--color-error);
+			}
+
+			&__menu {
+				margin-block: auto;
+				margin-inline-end: 12px;
 			}
 		}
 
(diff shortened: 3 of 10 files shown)


Chartman123:
As already stated in the issue, you could also have a look at my tries on fixing the alignment and integrate them in this one here if you want to :)

#2577 
#2574 
#2572

Chartman123:
> next to the short answer „reappearing“

This was originally an overlay over the input field. I tried to move it back there in #2574 

> Notice the error fitting with some room to breath

Nice 🤩

Elsensee:
I integrated them. :)

For #2577 I just cherry-picked tbh,
For #2574 I changed it to using the flow-relative `inset` properties. That seemed to be the better option for i18n and I have used that before. I didn’t know the menu was supposed to be an overlay, but it made much more sense that way!

#2572, part of it was already done with my PR. And for the other part I think I was able to improve on your work. :) In addition to your work, I also aligned the toggle buttons with the general "start of text" on that page.

| current main | this PR |
| -------------- | -------- |
| Screenshot 2025-02-18 133712 | Screenshot 2025-02-18 133336 |
| Screenshot 2025-02-18 133656 | Screenshot 2025-02-18 133405 |

Chartman123:
@marcoambrosini just a quick design question: in the submit view (https://github.com/nextcloud/forms/pull/2578#issuecomment-2665616451 last row in the table), should we align the text outside of the fields with the borders of the input fields like in the right screenshot or with the text inside the input fields like in the left screenshot? In the left picture, the header of the question should then be moved further to the right of course (or the input fields further to the left).

@Elsensee apart from that question, really nice job 👍🏻  thank you very much :)

Elsensee:
That’s alright! :) I’m not a designer, I just do what feels/looks good. And basically I just wanted to continue your work regarding positions – originally I didn’t thought about fixing anything else than the Create view, because that had the most fixing needed imho. :D 

When looking through the issues, I noticed that something else was still not quite right, so I fixed that too now. I don’t know if that’s how it was supposed to be, but this is how it looked in v29 so I thought I’d recreate it here then.

| current main | this PR |
| -------------- | -------- |
| Screenshot 2025-02-19 153218 |  Screenshot 2025-02-19 153151 |

But now I ought to be done. ;)

nimishavijay (approved):
Super nice work! 🚀 Looks waaay better now!

Definitely not a blocker, but it would be nice to move the ❗ error symbol to the left of the question field and extend it all the way to the end, what do you think? :) it also could be a separate PR, so I'm approving this one to not block :)

Chartman123:
> Definitely not a blocker, but it would be nice to move the ❗ error symbol to the left of the question field and extend it all the way to the end, what do you think? :)

@nimishavijay thanks for the review. I'd leave for now and think more of making use of the invalid state of the field itself to have the red border if there's no title entered. But as you already said as a follow up

Chartman123 (approved):

PR #2579, opened 2025-02-18
feat: make options draggable

Chartman123:
This is a follow-up on #2092


src/components/Questions/AnswerInput.vue
```diff
@@ -356,31 +366,46 @@ export default {
 
 	&__pseudoInput {
 		color: var(--color-primary-element);
-		margin-inline-start: calc(-1 * var(--default-grid-baseline));
+		margin-inline-start: -2px;
```
susnux, line 369: unrelated?

src/components/Questions/AnswerInput.vue
```diff
@@ -145,6 +152,10 @@ export default {
 			})
 		},
 
+		optionDragMenuId() {
+			return 'q' + this.answer.questionId + 'o' + this.answer.id + '__drag_menu'
```
susnux, line 156: I prefer template string, but no blocker ;)
```suggestion
			return `q${this.answer.questionId}o${this.answer.id}__drag_menu`
```

src/components/Questions/QuestionDropdown.vue
```diff
@@ -169,4 +190,22 @@ export default {
...
+.options-list-transition-move,
+.options-list-transition-enter-active,
+.options-list-transition-leave-active {
+	transition: all var(--animation-slow) ease;
+}
+
+.options-list-transition-enter-from,
+.options-list-transition-leave-to {
+	opacity: 0;
+	transform: translateX(44px);
```
susnux, line 203: Random number?

src/components/Questions/AnswerInput.vue
```diff
@@ -356,31 +366,46 @@ export default {
...
+
+	.option__drag-handle {
+		opacity: 0.5;
+		cursor: grab;
+
+		&:hover,
+		&:focus,
+		&:focus-within {
+			opacity: 1;
+		}
```
susnux, line 393: I do not think we can do it like this for a11y.
Probably better to use:
`color-text-maxcontrast` when not focused and `color-main-text` when focused to keep contrast ratios.

src/components/Questions/AnswerInput.vue
Chartman123: no, otherwise the input fields are no longer aligned with the rest. In current main you can see that they are a little bit off after the reorder PR.

src/mixins/QuestionMultipleMixin.ts
```diff
@@ -37,33 +37,45 @@ export default defineComponent({
...
-			// Only shuffle options if not in editing mode (and shuffling is enabled)
-			if (this.readOnly && this.extraSettings?.shuffleOptions) {
-				return this.shuffleArray(this.options)
-			}
-
-			// Ensure order of options always is the same
-			const options = [...this.options].sort((a, b) => {
-				if (a.order === b.order) {
-					return a.id - b.id
+		sortedOptions: {
```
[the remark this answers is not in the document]
Chartman123: That was the only way I got it working... Is there a better way?

src/components/Questions/AnswerInput.vue
Chartman123: This is the styling of the question drag handle. Should we change it there, too?

src/components/Questions/AnswerInput.vue
susnux: But the value is exactly the same both will be -2px

src/components/Questions/AnswerInput.vue
susnux: would say so

src/components/Questions/AnswerInput.vue
```diff
@@ -26,43 +26,43 @@
...
+				:aria-label="t('forms', 'Move option actions')"
+				class="option__drag-handle option__actions-button"
+				type="tertiary-no-background">
+				<template #icon>
+					<IconDragIndicator :size="20" />
+				</template>
+				<NcActionButton
+					ref="buttonOptionUp"
 					:disabled="index === 0"
-					size="small"
```
susnux, line 36: this changes cause the buttons to be too big again, see focus visible outline.
fixed that on the reorder pr to make the buttons contained within the input

src/components/Questions/AnswerInput.vue
```diff
@@ -26,43 +26,43 @@
...
-					type="tertiary"
 					@click="onMoveDown">
 					<template #icon>
 						<IconArrowDown :size="20" />
 					</template>
-				</NcButton>
-				<NcButton
-					class="option__actions-button"
-					:aria-label="t('forms', 'Delete answer')"
-					size="small"
```
susnux, line 58: this changes cause the buttons to be too big again, see focus visible outline.
fixed that on the reorder pr to make the buttons contained within the input

src/components/Questions/AnswerInput.vue
```diff
@@ -356,31 +366,46 @@ export default {
...
 	}
 
 	.option__actions {
 		display: flex;
 		position: absolute;
 		gap: var(--default-grid-baseline);
-		inset-inline-end: 16px;
-		height: var(--default-clickable-area);
+		inset-inline-end: 12px;
+		height: 100%;
```
susnux, line 378: this is related to the button size which is too big now.

src/components/Questions/AnswerInput.vue
Chartman123: No, default grid baseline is 4px. So yours is - 4px

src/components/Questions/AnswerInput.vue
Chartman123: I tried the two colors but can't see a real difference between focused/not focused.

Chartman123: I had to apply them to a deeper element...

susnux (approved):

PR #2581, opened 2025-02-18
fix: remove linked file from cloned form

Chartman123:
A cloned form kept the file link from the original form. This fixes it by unsetting the fields in the copy.


Chartman123 (commented):
/backport to stable4

susnux (approved):
Thank you for the quick fix!

PR #2584, opened 2025-02-19
feat: replace drag icon for questions

### LINKED ISSUE #2576: Switch drag handles for questions
Other than that, for the left drag handle I would suggest this icon:
https://fonts.google.com/icons?selected=Material+Symbols+Outlined:drag_indicator:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=drag&icon.size=24&icon.color=%23e8eaed
Screenshot 2025-02-17 at 14 25 34

This drag button looks like it has too much right margin, I'd reduce that to `--default-grid-baseline`

_Originally posted by @marcoambrosini in https://github.com/nextcloud/forms/issues/2092#issuecomment-2663140514_

Chartman123:
This closes #2576 by adding IconDragIndicator and replacing IconDragHorizontalVariant with it.

| old icon | new icon |
|---|---|
| grafik | grafik |

### DIFF
--- a/src/components/Icons/IconDragIndicator.vue
+++ b/src/components/Icons/IconDragIndicator.vue
@@ -1,0 +1,45 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Christian Hartmann <chris-hartmann@gmx.de>
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<span
+		:aria-hidden="!title"
+		:aria-label="title"
+		class="material-design-icon drag-indicator-icon"
+		role="img"
+		v-bind="$attrs"
+		@click="$emit('click', $event)">
+		<svg
+			:fill="fillColor"
+			class="material-design-icon__svg"
+			:height="size"
+			:width="size"
+			viewBox="0 -960 960 960">
+			<path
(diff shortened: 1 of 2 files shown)


src/components/Icons/IconDragIndicator.vue
```diff
@@ -0,0 +1,45 @@
...
+		role="img"
+		v-bind="$attrs"
+		@click="$emit('click', $event)">
+		<svg
+			:fill="fillColor"
+			class="material-design-icon__svg"
+			:height="size"
+			:width="size"
+			viewBox="0 -960 960 960">
+			<path
```
Koc, line 20: But material design icons set already contains something similar

image

Chartman123: Yes, similar, but not the same... And Nextcloud wants to move to the newer Material Symbols in general. I didn't find a suiting icon anywhere else so I just created an own Icon component.

Koc (approved):

nimishavijay (approved):
Nice! Agreed with Marco that this is consistent with how other apps show drag indication :)

Chartman123:
@nimishavijay thanks for the feedback. I'm currently implementing a drag feature for the options, too. should we use the same indicator there?

#2579

nimishavijay:
I'd say so yes. Ref Google Forms which also uses the same drag handle for both, but with a different orientation (and I'd assume the different orientation is because it's above the question, not on the left like on ours)

PR #2588, opened 2025-02-20
fix(e2e): use login and createRandomUser from package

max-nextcloud:
`@nextcloud/e2e-test-server` has these since version 0.2.1.

Follow up for #2560


Chartman123 (approved):
Nice, thank you!

PR #2609, opened 2025-02-28
WIP: feat: add linear scale questions

### LINKED ISSUE #2605: [QuestionType] Linear scale
Add a new question type for linear/Likert scales

Chartman123:
This is a first attempt to implement a new question type. It will close #2605

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -75,6 +75,7 @@ class Constants {
 	public const ANSWER_TYPE_DATETIME = 'datetime';
 	public const ANSWER_TYPE_TIME = 'time';
 	public const ANSWER_TYPE_FILE = 'file';
+	public const ANSWER_TYPE_LINEARSCALE = 'linearscale';
 
 	// All AnswerTypes
 	public const ANSWER_TYPES = [
@@ -87,13 +88,15 @@ class Constants {
 		self::ANSWER_TYPE_DATETIME,
 		self::ANSWER_TYPE_TIME,
 		self::ANSWER_TYPE_FILE,
+		self::ANSWER_TYPE_LINEARSCALE,
 	];
 
 	// AnswerTypes, that need/have predefined Options
 	public const ANSWER_TYPES_PREDEFINED = [
 		self::ANSWER_TYPE_MULTIPLE,
 		self::ANSWER_TYPE_MULTIPLEUNIQUE,
-		self::ANSWER_TYPE_DROPDOWN
+		self::ANSWER_TYPE_DROPDOWN,
+		self::ANSWER_TYPE_LINEARSCALE
 	];
 
 	// AnswerTypes for date/time questions
@@ -153,6 +156,13 @@ class Constants {
 		'x-office/document',
 		'x-office/presentation',
 		'x-office/spreadsheet',
+	];
+
+	public const EXTRA_SETTINGS_LINEARSCALE = [
+		'optionsStart' => ['integer'],
+		'optionsEnd' => ['integer'],
+		'optionsLabelBest' => ['string'],
+		'optionsLabelWorst' => ['string'],
 	];
 
 	public const FILENAME_INVALID_CHARS = [
--- a/src/components/Icons/IconLinearScale.vue
+++ b/src/components/Icons/IconLinearScale.vue
@@ -1,0 +1,45 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Christian Hartmann <chris-hartmann@gmx.de>
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<span
+		:aria-hidden="!title"
+		:aria-label="title"
+		class="material-design-icon drag-indicator-icon"
+		role="img"
+		v-bind="$attrs"
+		@click="$emit('click', $event)">
+		<svg
+			:fill="fillColor"
+			class="material-design-icon__svg"
+			:height="size"
+			:width="size"
+			viewBox="0 -960 960 960">
+			<path
+				d="M672-288q-71 0-123.38-44.36Q496.24-376.73 482.9-444H281q-11 26-35 43t-54 17q-40.32 0-68.16-27.77Q96-439.55 96-479.77 96-520 123.84-548q27.84-28 68.16-28 30 0 54 17t35.46 43H484q13.21-67.28 65.1-111.64Q601-672 672-672q79.68 0 135.84 56.23 56.16 56.22 56.16 136Q864-400 807.84-344 751.68-288 672-288Zm0-72q50 0 85-35t35-85q0-50-35-85t-85-35q-50 0-85 35t-35 85q0 50 35 85t85 35Z" />
+			<title v-if="title">{{ title }}</title>
+		</svg>
+	</span>
+</template>
+
+<script>
+export default {
+	name: 'IconLinearScale',
+	props: {
+		title: {
+			type: String,
+			default: '',
+		},
+		fillColor: {
+			type: String,
+			default: 'currentColor',
+		},
+		size: {
+			type: Number,
+			default: 20,
+		},
+	},
+}
+</script>
--- a/src/components/Questions/QuestionLinearScale.vue
+++ b/src/components/Questions/QuestionLinearScale.vue
@@ -1,0 +1,182 @@
+<!--
+  - SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<Question
+		v-bind="questionProps"
+		:title-placeholder="answerType.titlePlaceholder"
+		:warning-invalid="answerType.warningInvalid"
+		v-on="commonListeners">
+		<template #actions>
+			<NcActionInput
+				type="multiselect"
+				:options="[0, 1]"
+				:v-model="lowestOption">
+				{{ t('forms', 'Lowest value') }}
+			</NcActionInput>
+			<NcActionInput
+				type="multiselect"
+				:options="[2, 3, 4, 5, 6, 7, 8, 9, 10]"
+				:v-model="highestOption">
+				{{ t('forms', 'Highest value"') }}
+			</NcActionInput>
+			<NcActionInput>
+				{{ t('forms', 'Label for lowest value"') }}
+			</NcActionInput>
+			<NcActionInput>
+				{{ t('forms', 'Label for highest value"') }}
+			</NcActionInput>
+		</template>
+
+		<fieldset :name="name || undefined" :aria-labelledby="titleId">
+			<NcNoteCard v-if="hasError" :id="errorId" type="error">
+				{{ errorMessage }}
+			</NcNoteCard>
+			<NcCheckboxRadioSwitch
+				v-for="option in scaleOptions"
+				:key="option"
+				:disabled="!readOnly"
+				:aria-errormessage="hasError ? errorId : undefined"
+				:aria-invalid="hasError ? 'true' : undefined"
+				:checked="questionValues"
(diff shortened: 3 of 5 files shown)


susnux:
I also saw this sometimes with labels for each option like:

```
strongly agree        agree      neutral      disagree strongly     disagree
          [  ]                       [  ]             [  ]                       [   ]                       [  ]
```

Chartman123:
> I also saw this sometimes with labels for each option like:
> 
> ```
> strongly agree    agree    neutral   disagree     strongly disagree
>     [ ]            [ ]       [ ]           [ ]            [ ]
> ```

Yes, there are different implementations. I used Google Forms as a "template", Microsoft does it more like in your comment.

jancborchardt (commented):
Nice! So as defaults, we could set a scale of 1 to 5, since that is what the usual Likert scale is as well.
The labels we could also set to a default of "Strongly disagree" for lowest and "Strongly agree" for the highest, as @susnux mentions – probably best to leave out the other labels, except if we default to those as labels and don’t allow customization. (Or we do an either/or: Either you get numbers, or you get labels.)

Google Forms defaults to 1 for lowest and 5 for highest, and allows only 0 or 1 for lowest as well as only 2 to 10 for highest, which makes sense.

nimishavijay:
Awesome! Looks great already! 

For reference this is what Google Forms does, but I actually love how we can see the scale in your screenshot
image

Here are some ideas to improve the look of this 
- [ ] Since the lowest and highest value is an important attribute of the question, we can move it outside to below the scale 
- [ ] Maybe the scale can be a more direct "preview". What would the scale look like in view mode? Could that be replicated here?
- [ ] The labels for the lowest and highest value can also be outside using the inputs with the floating label, what do you think?

Then with the preview it would be like:
```
             1    2    3    4    5
not so good [ ]  [ ]  [ ]  [ ]  [ ] very good

Lowest value  [ 1 ] [ not so good  ]
Highest value [ 5 ] [ very good    ]
```

What do you think?

Chartman123:
> Maybe the scale can be a more direct "preview". What would the scale look like in view mode? Could that be replicated here?

@nimishavijay The scale in the edit mode adjusts directly to the changed values in the actions menu and it will look the same in edit and view mode :)

> Since the lowest and highest value is an important attribute of the question, we can move it outside to below the scale
> The labels for the lowest and highest value can also be outside using the inputs with the floating label, what do you think?

You mean to move the settings from the menu directly to the content of the question? This would be feasible, but also break the UX regarding where to configure the settings of the questions.

> Google Forms defaults to 1 for lowest and 5 for highest, and allows only 0 or 1 for lowest as well as only 2 to 10 for highest, which makes sense.

@jancborchardt those are also the defaults for us here :)

> probably best to leave out the other labels, except if we default to those as labels and don’t allow customization. (Or we do an either/or: Either you get numbers, or you get labels.)

Google allows to customize all labels for matrix questions. So for that use case you could then use a matrix question with just one row.

nimishavijay:
> The scale in the edit mode adjusts directly to the changed values in the actions menu and it will look the same in edit and view mode :)

Oh nice! Do you think they can be radio buttons like in Google forms then? I took a quick look at general likert scale forms in other software and radio buttons seems to ne the general standard.

> You mean to move the settings from the menu directly to the content of the question? This would be feasible, but also break the UX regarding where to configure the settings of the questions.

Yes, I understand. I see this as rather the settings of the options rather than settings of the question itself (eg. Required question, technical question, etc). 
The question types so far were simple enough that they didn't need many options up front, but I'd say this is more complex, what do you think?

Chartman123:
> I see this as rather the settings of the options rather than settings of the question itself (eg. Required question, technical question, etc).

I understand that, but this will break with our current philosophy to show the form in edit view in the same way (except for the borders) as it will be in the submit/view mode. I'm not really sure about that.

> Do you think they can be radio buttons like in Google forms then? I took a quick look at general likert scale forms in other software and radio buttons seems to ne the general standard.

Yes, sure:

image

But I also quite like the button variant, could be some sort of "unique selling point". 😉

susnux:
> I understand that, but this will break with our current philosophy to show the form in edit view in the same way (except for the borders) as it will be in the submit/view mode. I'm not really sure about that.

We could make the labels text boxes just like we do for other question types.
And the options could have a + or -

So nothing in the popover menu needed.

Chartman123:
> We could make the labels text boxes just like we do for other question types. And the options could have a + or -
> 
> So nothing in the popover menu needed.

@nimishavijay @jancborchardt Ferdinand and I decided to keep the min/max options in the actions menu, because it would overload the UI, but to move the labels to question content.

src/components/Icons/IconLinearScale.vue
susnux, line 1: why do we need this component? can we not just directly use the material-design-icons ?

Chartman123: But we could of course use the svg wrapper component, if that's better

nimishavijay (commented):
Nice! Looks really cool! I see that the width of the label elements is lesser than it could be. Could the space between the radio buttons be reduced (maybe by 2px or 4px) and the width of the labels increased? 

Also, what would it look like on mobile?

Chartman123:
> Also, what would it look like on mobile?

Good point... 😀  Currently not really usable at all. But I think I'll need more help here by someone who knows how to fiddle with the CSS.

> I see that the width of the label elements is lesser than it could be. Could the space between the radio buttons be reduced (maybe by 2px or 4px) and the width of the labels increased?

It needs that space if you have options from 0 to 10. But same as above, with more CSS expertise one could probably improve that a lot.

src/components/Icons/IconLinearScale.vue
susnux: No its not important, its designed by you? Then just drop the "drag-indicator-icon" class its unused, no?

Otherwise from MDI we could use RayStartVertexEnd icon

Chartman123: No, it's not designed by me. I took the Material Symbol from Google. The class is probably a remnant of the other component that I just copied and exchanged the SVG part.

Chartman123:
@Elsensee would you like to work on the remaining CSS/design issues here?

Chartman123:
@nimishavijay I improved the layout on mobile devices :) Probably not the best solution but it works. Do you know someone who could help here? Or do you think we can leave it as it is right now?

nimishavijay:
Looks good in mobile :) What do you think about moving the fields to next to the associated radio buttons? Like this:

```
[] 1    [ strongly disagree ]
[] 2
[] 3
[] 4
[] 5    [ strongly agree    ]
```

If that's not possible then I'd suggest simply adding around 4px spacing above the lowest value label to separate it from the question fields :)

Chartman123:
@nimishavijay I can try that :) Google shows the labels above and below, that's where I got the inspiration from. But I also like the idea of putting it next to the lowest/highest value

src/components/Questions/QuestionLinearScale.vue
```diff
@@ -0,0 +1,336 @@
...
+					? 'question__content'
+					: 'question__content question__content__edit'
+			">
+			<NcTextArea
+				v-if="!readOnly"
+				ref="lowest"
+				v-model="optionsLabelLowest"
+				class="label-input-field"
+				:label="t('forms', 'Label (optional)')"
+				:aria-label="t('forms', 'Label for lowest value')"
```
susnux, line 51: Should be the same

susnux:
```suggestion
				:label="t('forms', 'Label for lowest value')"
				:placeholder="t('forms', 'Label (optional)')"
```

src/components/Questions/QuestionLinearScale.vue
```diff
@@ -0,0 +1,336 @@
...
+				resize="none"
+				@input="resizeLabel('lowest')"
+				@blur="onBlur('lowest')">
+			</NcTextArea>
+			<div v-else-if="optionsLabelLowest !== ''" class="label-lowest">
```
susnux, line 56: Add some id here

src/components/Questions/QuestionLinearScale.vue
```diff
@@ -0,0 +1,336 @@
...
+				{{ optionsLabelLowest }}
+			</div>
+			<div class="question__content__options">
+				<NcCheckboxRadioSwitch
+					v-for="option in scaleOptions"
```
susnux, line 61: if first option then:
`aria-describedby="id from above"`

src/components/Questions/QuestionLinearScale.vue
```diff
@@ -0,0 +1,336 @@
...
+				:label="t('forms', 'Label (optional)')"
+				:aria-label="t('forms', 'Label for lowest value')"
+				resize="none"
+				@input="resizeLabel('lowest')"
+				@blur="onBlur('lowest')">
+			</NcTextArea>
+			<div v-else-if="optionsLabelLowest !== ''" class="label-lowest">
+				{{ optionsLabelLowest }}
+			</div>
+			<div class="question__content__options">
```
susnux, line 59:
```suggestion
			<fieldset class="question__content__options">
```

susnux: Maybe:

```suggestion
			<fieldset class="question__content__options">
				<legend class="hidden-visually">
					{{ t('forms', 'From {firstOption} to {lastOption}', { firstOption: optionsLabelLowest, lastOption: optionsLabelHighest }) }}
				</legend>
```

Chartman123:
@nimishavijay I used the easier way and added a `padding-block: var(--default--grid-baseline)` to the labels in mobile view both for edit and view mode.

lib/Constants.php
```diff
@@ -87,13 +88,15 @@ class Constants {
...
+		self::ANSWER_TYPE_LINEARSCALE,
 	];
 
 	// AnswerTypes, that need/have predefined Options
 	public const ANSWER_TYPES_PREDEFINED = [
 		self::ANSWER_TYPE_MULTIPLE,
 		self::ANSWER_TYPE_MULTIPLEUNIQUE,
-		self::ANSWER_TYPE_DROPDOWN
+		self::ANSWER_TYPE_DROPDOWN,
+		self::ANSWER_TYPE_LINEARSCALE
```
susnux, line 99:
```suggestion
		self::ANSWER_TYPE_LINEARSCALE,
```

make it git safe ;)

src/components/Questions/QuestionLinearScale.vue
```diff
@@ -0,0 +1,362 @@
+<!--
+  - SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
```
susnux, line 2:
```suggestion
  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

susnux (approved):
Tested and works fine.
But should here not also the label of the highest option be shown? Currently only the lowest is shown:

grafik

src/components/Results/ResultsSummary.vue
```diff
@@ -86,14 +86,59 @@ export default {
...
 	computed: {
+		questionTypeLabel() {
+			const label = this.answerTypes[this.question.type].label
+			const labelLowest = this.question.extraSettings?.optionsLabelLowest ?? ''
+			const labelHighest =
+				this.question.extraSettings?.optionsLabelHighest ?? ''
+			const optionsLowest =
+				this.question.extraSettings?.optionsLowest?.toString() ?? '1'
+			const optionsHighest =
+				this.question.extraSettings?.optionsHighest?.toString() ?? '1'
```
susnux, line 97: Should be 5 no?

```suggestion
				this.question.extraSettings?.optionsHighest?.toString() ?? '5'
```

Chartman123: Probably copy/paste 🙈

PR #2646, opened 2025-03-17
feat(date): allow restrictions for dates and query date ranges

### LINKED ISSUE #2396: Date field: Range restriction and range responses

- Nextcloud-Version: [30.0.1]
- Forms-Version: [4.3.2]

I'm organizing a week-long event. In the registration form I want people to tell me on which day they'll join and on which day they'll leave. I can use the date field for that, but that allows to add options that are not even in the scope of the event. That can cause some overhead of cleaning up the data and getting back to people about their invalid registration.

* I'd like to be able to restrict the date field input to dates in a specified range. E.g. the user can select only a date between 2024-01-01 and 2024-01-07
* Ideally I'd also like to have the user select a date *range*. That way, instead of having two fields for "date of arrival" and "date of departure", I'd have one field "When will you join us?" und the user can select e.g. the range 2024-01-02 - 2024-01-6. The cherry on top would be if a minimum date range would be possible - thank of hotel bookings that may require a minimum of 2 nights.

Keep doing as is and clean up the data manually :)

**Additional context**

Chartman123:
This closes  #2396 by allowing to set minimum and maximum dates and to query date ranges

Set question settings for minimum/maximum dates and if the user must enter a date range:
image

Restricted dates and date range picker in submit view:
grafik

Results in summary / responses view:
grafik 
grafik


Chartman123:
Currently the Calendar of the NcDateTimePicker is not shown correctly in the menu:

image

@susnux is this a bug in the `@nextcloud/vue` lib?

src/components/Questions/QuestionDate.vue
```diff
@@ -9,6 +9,37 @@
...
+		<template v-if="answerType.pickerType === 'date'" #actions>
+			<NcActionCheckbox
+				:checked="extraSettings?.dateRange"
+				@update:checked="onDateRangeChange">
+				{{ t('forms', 'Query date range') }}
+			</NcActionCheckbox>
+			<NcActionInput
+				v-model="dateMin"
+				type="date"
+				:label="t('forms', 'Pick minimum date')"
```
jancborchardt, line 21:
```suggestion
				:label="t('forms', 'Earliest date')"
```

src/components/Questions/QuestionDate.vue
```diff
@@ -9,6 +9,37 @@
 		:title-placeholder="answerType.titlePlaceholder"
 		:warning-invalid="answerType.warningInvalid"
 		v-on="commonListeners">
+		<template v-if="answerType.pickerType === 'date'" #actions>
+			<NcActionCheckbox
+				:checked="extraSettings?.dateRange"
+				@update:checked="onDateRangeChange">
+				{{ t('forms', 'Query date range') }}
```
jancborchardt, line 16:
```suggestion
				{{ t('forms', 'Limit date range') }}
```

src/components/Questions/QuestionDate.vue
```diff
@@ -9,6 +9,37 @@
...
+				is-native-picker
+				:max="dateMax">
+				<template #icon>
+					<Pencil :size="20" />
+				</template>
+			</NcActionInput>
+			<NcActionInput
+				v-model="dateMax"
+				type="date"
+				:label="t('forms', 'Pick maximum date')"
```
jancborchardt, line 33:
```suggestion
				:label="t('forms', 'Latest date')"
```

src/components/Questions/QuestionDate.vue
```diff
@@ -9,6 +9,37 @@
...
+				v-model="dateMin"
+				type="date"
+				:label="t('forms', 'Pick minimum date')"
+				hide-label
+				:formatter="extraSettingsFormatter"
+				is-native-picker
+				:max="dateMax">
+				<template #icon>
+					<Pencil :size="20" />
+				</template>
```
jancborchardt, line 28:
```suggestion
```

src/components/Questions/QuestionDate.vue
```diff
@@ -9,6 +9,37 @@
...
+				v-model="dateMax"
+				type="date"
+				:label="t('forms', 'Pick maximum date')"
+				hide-label
+				:formatter="extraSettingsFormatter"
+				is-native-picker
+				:min="dateMin">
+				<template #icon>
+					<Pencil :size="20" />
+				</template>
```
jancborchardt, line 40:
```suggestion
```

jancborchardt (changes requested):
Looks nice! :) Added some wording suggestions to make it a bit more easy to understand / less technical.
And I’d say the icons can be removed as they are the same and thus mostly visual noise? It’s a sub-item of the checkbox anyway so it’s fine to not have an icon and thus look a bit indented.

src/components/Questions/QuestionDate.vue
Chartman123: See comment below

src/components/Questions/QuestionDate.vue
Chartman123: See comment below

Chartman123:
@jancborchardt thanks for the review :) 

> It’s a sub-item of the checkbox anyway

No, they are separate from the checkbox and always shown. The restrictions of which date can be choses apply also to single date questions. That's why I added the icons. Without them it felt wrong to me.

src/components/Questions/QuestionDate.vue
Chartman123: I changed the label for the checkbox to "Use date range" to make it clearer

jancborchardt (commented):
As icons, how about:
- "Today" for Earliest date
- "Event" for Latest date

https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%235f6368&icon.query=event

Chartman123:
> As icons, how about:
> 
> * "Today" for Earliest date
> * "Event" for Latest date
> 
> https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%235f6368&icon.query=event

I've changed the icons and updated the screenshot in the OP :)

openapi.json
```diff
@@ -418,6 +418,17 @@
...
                     },
+                    "dateMax": {
+                        "type": "integer",
+                        "format": "int64"
+                    },
+                    "dateMin": {
+                        "type": "integer",
+                        "format": "int64"
+                    },
+                    "dateRange": {
```
susnux, line 429: Maye it makes sense to have `ANSWER_TYPE_DATE` <-> `ANSWER_TYPE_DATE_RANGE`, `ANSWER_TYPE_DATETIME` <->`ANSWER_TYPE_DATETIME_RANGE` ?

But also fine with this.

jancborchardt: @susnux how would that look UI-wise, or what would be the choices in plain English? :D

Chartman123: As our UI elements support the range and non-range function in one and also only a few modifications were needed, I think we can leave it this way :) and I think it doesn't make much difference if we query the question.type vs. the extraSetting. Should I include the same for time questions into this PR or a follow-up?

Chartman123:
> sorry, for some reason the icon link didn’t save the "fill" attribute. :\ The icons should use the filled versions, as we always do.

@jancborchardt fixed :)

jancborchardt (approved):
Really nice! :)

susnux (approved):
I think this is simpler than 4 date types.

Bibo-Joshi:
This is awesome, thanks a ton! 🎊

PR #2649, opened 2025-03-19
Fix(migration): Rename primary key of uploaded_files table

### LINKED ISSUE #2648: Error while enabling Forms: Primary index name "id" for table "oc_forms_v2_uploaded_files" collides with the constraint on table "oc_bookmarks".

**Describe the bug**
Downloaded and trying to enable Forms app. results in error:
```
Primary index name "id" for table "oc_forms_v2_uploaded_files" collides with the constraint on table "oc_bookmarks".
```
**To Reproduce**
Go to **App bundles > Public sector bundle**. For **Forms** app click **Enable**.

**Expected behavior**
Forms app is enabled.

- Nextcloud-Version: Nextcloud Hub 9 (30.0.7)
- Forms-Version: 5.0.3

- OS: Debian Sid
- Browser: Firefox
- Version 128.8esr

**Browser log**

Nothing in browser console

Chartman123:
This fixes #2648. It changes the primary key name parameter in the original migration and adds a new migration for existing installations that drops the old primary key and creates a new one with the new name.

### DIFF
--- a/lib/Migration/Version040300Date20240523123456.php
+++ b/lib/Migration/Version040300Date20240523123456.php
@@ -59,7 +59,7 @@ class Version040300Date20240523123456 extends SimpleMigrationStep {
 				'notnull' => false,
 				'comment' => 'unix-timestamp',
 			]);
-			$table->setPrimaryKey(['id'], 'id');
+			$table->setPrimaryKey(['id'], 'forms_upload_files_id');
 		}
 
 		return $schema;
--- a/lib/Migration/Version050004Date20250319180638.php
+++ b/lib/Migration/Version050004Date20250319180638.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Forms\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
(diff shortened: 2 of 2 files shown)


lib/Migration/Version040300Date20240523123456.php
```diff
@@ -59,7 +59,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 				'notnull' => false,
 				'comment' => 'unix-timestamp',
 			]);
-			$table->setPrimaryKey(['id'], 'id');
+			$table->setPrimaryKey(['id']);
```
susnux, line 62: we probably should also rename it for existing installations

Chartman123:
For the backport we need to rename the migration to match the corresponding v4 version

Koc (approved):

lib/Migration/Version050004Date20250319180638.php
```diff
@@ -0,0 +1,40 @@
...
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('forms_v2_uploaded_files')) {
+			$table = $schema->getTable('forms_v2_uploaded_files');
+			$table->dropPrimaryKey();
+			$table->setPrimaryKey(['id'], 'forms_upload_files_id');
```
susnux, line 35: Or simply no custom key name

Chartman123: This does not work, because the then chosen default is too long and triggers an error

susnux: 🙈

PR #2674, opened 2025-04-02
fix: quote cell values to prevent formula evaluation

Chartman123:

### DIFF
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -312,6 +312,11 @@ class SubmissionService {
 						->setWrapText(true);
 				} else {
 					$activeWorksheet->setCellValue([$column, $row], $value);
+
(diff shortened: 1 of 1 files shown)


susnux (approved):

lib/Service/SubmissionService.php
```diff
@@ -312,6 +312,11 @@ private function exportData(array $header, array $data, string $fileFormat, ?Fil
 						->setWrapText(true);
 				} else {
 					$activeWorksheet->setCellValue([$column, $row], $value);
+
+					// Quote cell values that start with '=' to prevent evaluation of formulas
+					if (is_string($value) && str_starts_with($value, '=')) {
```
susnux, line 317: Is the value trimmed?

Chartman123: I just tested it with ` =1+1` and it was quoted in the exported xlsx file

PR #2685, opened 2025-04-08
fix(validation): Update validationType handling with getter/setter and simplify change logic

Chartman123:

### DIFF
--- a/src/components/Questions/QuestionShort.vue
+++ b/src/components/Questions/QuestionShort.vue
@@ -50,9 +50,10 @@
 						validationTypeObject, validationTypeName
 					) in validationTypes"
 					:key="validationTypeName"
-					:checked="validationType === validationTypeName"
-					:name="validationTypeName"
-					@update:checked="onChangeValidationType(validationTypeName)">
+					:model-value="validationType"
+					:name="`${id}_validationMenu`"
+					:value="validationTypeName"
+					@update:model-value="onChangeValidationType(validationTypeName)">
 					{{ validationTypeObject.label }}
 				</NcActionRadio>
 				<NcActionInput


src/components/Questions/QuestionShort.vue
```diff
@@ -50,9 +50,9 @@
 						validationTypeObject, validationTypeName
 					) in validationTypes"
 					:key="validationTypeName"
-					:checked="validationType === validationTypeName"
-					:name="validationTypeName"
-					@update:checked="onChangeValidationType(validationTypeName)">
+					v-model="validationType"
+					:value="validationTypeName"
+					:name="`${id}_validationMenu`">
```
susnux, line 55:
```suggestion
					:name="`${id}_validationMenu`"
					:value="validationTypeName">
```

src/components/Questions/QuestionShort.vue
```diff
@@ -124,8 +124,23 @@ export default {
...
+					this.onExtraSettingsChange({
+						validationType: value,
+						validationRegex: this.validationRegex,
+					})
+				} else {
+					this.isValidationTypeMenuOpen = false
+					this.onExtraSettingsChange({
+						validationType: value === 'text' ? undefined : value,
+					})
+				}
```
susnux, line 142: Not a fan of putting any API calls in computed values

susnux (approved):
works but not a fan of computed setters with API calls within

src/components/Questions/QuestionShort.vue
susnux: Yes because you use `v-mode`, but it then should be `:model-value="..."`

PR #2689, opened 2025-04-10
chore(QuestionMultiple): update "Other" answer handling to use v-model and add change handler

Chartman123:

### DIFF
--- a/src/components/Questions/QuestionMultiple.vue
+++ b/src/components/Questions/QuestionMultiple.vue
@@ -106,7 +106,8 @@
 						class="question__input"
 						:label="placeholderOtherAnswer"
 						:required="otherAnswer !== undefined"
-						:value.sync="otherAnswerText" />
+						:model-value="cachedOtherAnswerText"
+						@update:model-value="onOtherAnswerTextChange" />
 				</div>
 			</fieldset>
 		</template>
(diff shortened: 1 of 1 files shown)


src/components/Questions/QuestionMultiple.vue
```diff
@@ -291,34 +292,8 @@ export default {
...
-											QUESTION_EXTRASETTINGS_OTHER_PREFIX,
-										),
-								),
-								prefixedValue,
-							],
-				)
-			},
+		otherAnswerText() {
+			return this.cachedOtherAnswerText
 		},
```
susnux, line 297:
```suggestion
```

src/components/Questions/QuestionMultiple.vue
```diff
@@ -106,7 +106,8 @@
 						class="question__input"
 						:label="placeholderOtherAnswer"
 						:required="otherAnswer !== undefined"
-						:value.sync="otherAnswerText" />
+						:model-value="otherAnswerText"
```
susnux, line 109:
```suggestion
						:model-value="cachedOtherAnswerText"
```

susnux (commented):
we can just remove one of both then, no?

Chartman123:
> we can just remove one of both then, no?

Yes, makes sense, good catch! I must have overlooked that...

susnux (approved):

PR #2710, opened 2025-04-18
Add pagination for submissions

Koc:
This PR adds pagination for submissions list page. Few notes about implementation:
- It's not affect "Summary" tab
- Pagination works on the server
- There is an extra "search" field that used for filtering submissions

<details><summary>:mag: Preview</summary>
<p>

image
image

</p>
</details>


src/views/Results.vue
```diff
@@ -201,12 +201,79 @@
 
 			<!-- Responses view for individual responses -->
 			<section v-else>
+				<div class="pagination-block" :class="{ 'large-width': isMobile }">
```
Koc, line 204: Stolen from tables :ninja: . In theory we can extract this as a separate component to UI library

Chartman123 (commented):
Hi @Koc 

nice start, here are a few comments:

- I think we should split up the filtering from the rest of the pagination block and move it up to the right of the summary/responses switch
grafik
- The page dropdown is placed too high, there should be more spacing on top of it
- `floating` style looks odd to me, should be just embedded into the page
- page selection should be below the submissions or above and below but not only above them.
- A max of 50 items seems too big for me as a default. You'd still have to scroll a lot before you reach the end of the page
- At best we should make the item limit configurable (user setting)

Regarding the pagination in general: I think we should only fetch the currently shown submissions from the API. But in this case the filtering will no longer be able to work only in the front-end. So more adjustments needed then...

src/views/Results.vue
```diff
@@ -201,12 +201,79 @@
...
+					<div class="pagination-items">
+						<NcButton
+							type="tertiary"
+							:disabled="totalPages === 1 || pageNumber <= 1"
+							:aria-label="t('tables', 'Go to first page')"
```
Koc, line 209: copy-paste detected :see_no_evil:

src/views/Results.vue
```diff
@@ -348,6 +427,10 @@ export default {
 			picker: null,
 			showConfirmDeleteDialog: false,
 
+			submissionSearch: '',
+			pageNumber: 1,
+			rowsPerPage: 50,
```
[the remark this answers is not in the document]
Koc: set to 20 by default

Koc:
Hey @Chartman123 , thank you for quick feedback, very appreciated!

I've did some changes and updated screenshots in the PR description. So, I've moved pagination to backend and added extra parameter for full text search. For some reason it not works for me :confused: . Please help, feel free to push.

lib/Db/SubmissionMapper.php
```diff
@@ -106,40 +113,61 @@ public function hasFormSubmissionsByUser(Form $form, string $userId): bool {
...
 		if ($limit !== -1) {
 			$query->setMaxResults($limit);
 		}
 
 		$result = $query->executeQuery();
-		$row = $result->fetch();
 		$result->closeCursor();
 
-		return (int)($row['num_submissions'] ?? 0);
+		return $result->rowCount();
```
Koc, line 170: this not works in SQLite :confused:

lib/Controller/ApiController.php
```diff
@@ -1161,9 +1164,11 @@ public function getSubmissions(int $formId, ?string $fileFormat = null): DataRes
 
 		// Load submissions and currently active questions
 		if (in_array(Constants::PERMISSION_RESULTS, $this->formsService->getPermissions($form))) {
-			$submissions = $this->submissionService->getSubmissions($formId);
+			$submissions = $this->submissionService->getSubmissions($formId, null, $query, $limit, $offset);
+			$totalSubmissionsCount = $this->submissionMapper->countSubmissions($formId, null, $query);
```
Chartman123, line 1168: Can be removed
```suggestion
			$submissions = $this->submissionService->getSubmissions($formId, null, $query, $limit, $offset);
```

docs/API_v3.md
```diff
@@ -741,7 +747,8 @@ Get all Submissions to a Form
       "options": [],
       "extraSettings": {}
     }
-  ]
+  ],
+  "totalSubmissionsCount": 40
```
Chartman123, line 751: Can be reverted, is now part `GET /forms/{formId}
```suggestion
  ]
```

lib/Controller/ApiController.php
```diff
@@ -1161,9 +1164,11 @@ public function getSubmissions(int $formId, ?string $fileFormat = null): DataRes
...
 		} else {
-			$submissions = $this->submissionService->getSubmissions($formId, $this->currentUser->getUID());
+			$submissions = $this->submissionService->getSubmissions($formId, $this->currentUser->getUID(), $query, $limit, $offset);
+			$totalSubmissionsCount = $this->submissionMapper->countSubmissions($formId, null, $query);
```
Chartman123, line 1171: Same here:
```suggestion
			$submissions = $this->submissionService->getSubmissions($formId, $this->currentUser->getUID(), $query, $limit, $offset);
```

lib/Controller/ApiController.php
```diff
@@ -1196,6 +1201,7 @@ public function getSubmissions(int $formId, ?string $fileFormat = null): DataRes
 		$response = [
 			'submissions' => $submissions,
 			'questions' => $questions,
+			'totalSubmissionsCount' => $totalSubmissionsCount,
```
Chartman123, line 1204: and here

lib/ResponseDefinitions.php
```diff
@@ -77,7 +77,8 @@
  *
  * @psalm-type FormsSubmissions = array{
  *   submissions: list<FormsSubmission>,
- *   questions: list<FormsQuestion>
+ *   questions: list<FormsQuestion>,
+ *   totalSubmissionsCount: int
```
Chartman123, line 81: Same here
```suggestion
 *   questions: list<FormsQuestion>
```

openapi.json
```diff
@@ -592,7 +592,8 @@
                 "type": "object",
                 "required": [
                     "submissions",
-                    "questions"
+                    "questions",
+                    "totalSubmissionsCount"
```
Chartman123, line 596: And here:
```suggestion
                    "questions"
```

openapi.json
```diff
@@ -606,6 +607,10 @@
                         "items": {
                             "$ref": "#/components/schemas/Question"
                         }
+                    },
+                    "totalSubmissionsCount": {
+                        "type": "integer",
+                        "format": "int64"
                     }
```
Chartman123, line 614: Can be removed

src/views/Results.vue
```diff
@@ -478,22 +544,25 @@ export default {
...
 
 				// Append questions & submissions
-				this.submissions = loadedSubmissions
-				this.questions = loadedQuestions
+				this.submissions = this.formatDateAnswers(
+					data.submissions,
+					data.questions,
+				)
+				this.questions = data.questions
+				this.totalSubmissionsCount = data.totalSubmissionsCount
```
Chartman123, line 565:
```suggestion
				this.submissions = this.formatDateAnswers(
					data.submissions,
					data.questions,
				)
				this.questions = data.questions
```

tests/Integration/Api/ApiV3Test.php
```diff
@@ -1116,7 +1116,8 @@ public function dataGetSubmissions() {
 							]
 						]
 					],
-					'questions' => $this->dataGetFullForm()['getFullForm']['expected']['questions']
+					'questions' => $this->dataGetFullForm()['getFullForm']['expected']['questions'],
+					'totalSubmissionsCount' => 3,
```
Chartman123, line 1120:
```suggestion
					'questions' => $this->dataGetFullForm()['getFullForm']['expected']['questions']
```

tests/Integration/Api/ApiV3Test.php
```diff
@@ -1236,6 +1237,7 @@ public function testExportToCloud() {
 	public function dataDeleteSubmissions() {
 		$submissionsExpected = $this->dataGetSubmissions()['getSubmissions']['expected'];
 		$submissionsExpected['submissions'] = [];
+		$submissionsExpected['totalSubmissionsCount'] = 0;
```
Chartman123, line 1240: Can be removed

tests/Integration/Api/ApiV3Test.php
```diff
@@ -1368,6 +1370,7 @@ public function dataDeleteSingleSubmission() {
 	 * @param array $submissionsExpected
 	 */
 	public function testDeleteSingleSubmission(array $submissionsExpected) {
+		$submissionsExpected['totalSubmissionsCount'] = $submissionsExpected['totalSubmissionsCount'] - 1;
```
Chartman123, line 1373: Can be removed

tests/Unit/Controller/ApiControllerTest.php
```diff
@@ -245,6 +245,7 @@ public function dataGetSubmissions() {
 							'extraSettings' => new \stdClass(),
 						],
 					],
+					'totalSubmissionsCount' => 1,
```
Chartman123, line 248: Can be removed

tests/Unit/Controller/ApiControllerTest.php
```diff
@@ -265,6 +266,7 @@ public function dataGetSubmissions() {
...
```
Chartman123, line 269: Can be removed

tests/Unit/Controller/ApiControllerTest.php
```diff
@@ -298,6 +300,11 @@ public function testGetSubmissions(array $submissions, array $questions, array $
 			->with(1)
 			->willReturn($questions);
 
+		$this->submissionMapper->expects($this->once())
+			->method('countSubmissions')
+			->with(1)
+			->willReturn(1);
+
```
Chartman123, line 307: Can be removed

src/views/Results.vue
```diff
@@ -344,13 +396,18 @@ export default {
 
 			questions: [],
 			submissions: [],
+			totalSubmissionsCount: 0,
```
Chartman123, line 399: Can be removed

src/views/Results.vue
```diff
@@ -202,15 +235,27 @@
...
 					:submission="submission"
 					:questions="questions"
+					:highlight="submissionSearch"
 					:can-delete-submission="canDeleteSubmission(submission.userId)"
 					:can-edit-submission="canEditSubmission(submission.userId)"
 					@delete="deleteSubmission(submission.id)" />
+
+				<PaginationToolbar
+					class="bottom-pagination"
+					:total-items-count="totalSubmissionsCount"
```
Chartman123, line 256:
```suggestion
					:total-items-count="form.submissionCount"
```

Chartman123 (commented):
submissionCount is now part of the Form, you can just rely on it. I needed it for the allowEditSubmission PR :)

lib/Db/SubmissionMapper.php
Chartman123: This can be fixed with my commit :)

Koc:
@Chartman123 mate, the problem is that we have 2 different counts.

image

Chartman123:
@Koc ah yes you're right... 🙈

Chartman123:
But your count is wrong at the moment for shared forms:

grafik

You're counting all submissions and not only those by the current user if the user doesn't have the see all results permission

Chartman123:
Count looks good now :) But I'm not happy with the name `totalSubmissionsCount` as it's not the total of all submissions available to the user. What about renaming it to `filteredSubmissionsCount`?

And the search box is not properly aligned to the other items in the block. You could also use the `var(--default-grid-baseline)` for the margin instead of the `4px`.

Koc:
@Chartman123 done :heavy_check_mark:

src/views/Results.vue
```diff
@@ -869,4 +938,13 @@ export default {
 		}
 	}
 }
+
+.search-wrapper {
+	margin-top: calc(-1 * var(--default-grid-baseline));
+	margin-left: var(--default-grid-baseline);
```
Chartman123, line 944: Let's change this to margin-inline-start/end and margin-block-start so that it also works in RTL mode. `auto` makes it aligned on the right and the end margin is needed because the header and the question sections are not of the same width.

```suggestion
	margin-block-start: calc(-1 * var(--default-grid-baseline));
	margin-inline-start: auto;
	margin-inline-end: var(--default-clickable-area);
```

Chartman123 (approved):
Two more little comments, but rest looks fine for me :)

I posted this PR to Design team chat and asked if someone could have a look at it design-wise...

jancborchardt (commented):
Looks good! Just 2 comments:

Usually the pagination element is on the bottom of the oage rather than the top. It can be made sticky on the bottom if wanted.

> At best we should make the item limit configurable (user setting)

@Chartman123 just for info – something unimportant as this should not be a setting but instead have a proper value, as per design guidelines. :)

Chartman123:
@jancborchardt Thanks for the review 👍🏻 

> Usually the pagination element is on the bottom of the oage rather than the top.

@Koc could you then please remove the page selection from above the results?

Koc:
@Chartman123 removed

PR #2712, opened 2025-04-19
feat(time): Add time restrictions and range support

Chartman123:
This is a follow-up on #2646 to support restrictions and range for time questions.

Set question settings for minimum/maximum time values and if the user must enter a time range:
image

Restricted times and time range picker in submit view:
grafik

Result in summary / responses view:
grafik
grafik


Chartman123:
@susnux the NcDateTimePickerNative doesn't show the icon on the right to open the picker anymore. Is this a bug in the vue lib? But it did work when I started working on this... So not sure... 😉 

@nimishavijay what do you think about the icons for the actions? I'm not really happy with them, but didn't find better ones...

nimishavijay:
Regarding the icons, I agree it's a tricky one. What do you think of this [clock loader 20](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&selected=Material+Symbols+Outlined:clock_loader_20:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=time) and [clock loader 80](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&selected=Material+Symbols+Outlined:clock_loader_80:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=time)? 

And IIRC we are more and more adopting the native time picker by the browser and not the time picker component in the screenshot, which we previously had (for example, in the Calendar app we stopped using the time picker component). I would suggest using 2 native time picker fields with a "to" or "-" in the middle to indicate range and the allowed times in a subline so that it looks up-to-date :)

Chartman123:
@nimishavijay regarding the picker type: the NcDateTimePicker will stay around for ranges, that's at least what I found in the docs. So I'd like to stick with it as long as this is still possible. :)

> What do you think of this [clock loader 20](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&selected=Material+Symbols+Outlined:clock_loader_20:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=time) and [clock loader 80](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&selected=Material+Symbols+Outlined:clock_loader_80:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=time)?

They would have been my second choice, too

nimishavijay:
>  regarding the picker type: the NcDateTimePicker will stay around for ranges, that's at least what I found in the docs. So I'd like to stick with it as long as this is still possible. :)

No problem! In that case I'd suggest thinking about how we can differentiate the start and end times more, as now it is just one field and seemingly one large picker. Also FYI it looks pretty different in the [style guide ](https://nextcloud-vue-components.netlify.app/#/Components/NcPickers?id=ncdatetimepicker) (I'm not sure when this change happened, maybe @susnux knows more)

Chartman123:
> No problem! In that case I'd suggest thinking about how we can differentiate the start and end times more, as now it is just one field and seemingly one large picker. Also FYI it looks pretty different in the [style guide ](https://nextcloud-vue-components.netlify.app/#/Components/NcPickers?id=ncdatetimepicker) (I'm not sure when this change happened, maybe @susnux knows more)

Yes, I think the style guide is already based on v9 if you don't specify v8 explicitely :)

susnux:
> NcDateTimePickerNative doesn't show the icon on the right to open the picker anymore

This is browser dependent as this is the native browser input and cannot be controlled by us.

Chartman123:
@nimishavijay I've changed the icons now and I think that they look much better. 👍🏻 (screenshot updated in OP)

susnux (approved):
codewise it seems to be fine :white_check_mark:

Chartman123:
@nimishavijay There's now a new feature request about switching the date/time pickers to the native ones in #2719. So we'll change that for all date/time related questions in a separate PR 👍🏻

PR #2721, opened 2025-04-29
chore: unify debounce timing for input handling

Chartman123:
This unifies the debounce timing for user inputs to 400 ms.


src/components/Questions/AnswerInput.vue
```diff
@@ -174,7 +174,7 @@ export default {
 		// As data instead of method, to have a separate debounce per AnswerInput
 		this.debounceOnInput = debounce((event) => {
 			return this.queue.add(() => this.onInput(event))
-		}, 500)
+		}, 400)
```
Koc, line 177: let's introduce global constant for that? like `INPUT_DEBOUNCE_MS`

Chartman123: Where should we put the constant?

Chartman123: Yes sure, but we already have some files that provide constants. Should we also merge them at least partially?

susnux: Only if does not match the current locations. Should be rather grouped by use case than by "type"

src/models/Constants.ts
susnux, line 1: I am not sure in general if grouping "constants" makes sense.
Often its better to just have them where they are needed / belong logically to not look into many places but directly find it in the same code context.

Especially `FILE_SIZE_UNITS` which is **only** used by the files type and now you have to lookup another file to find the definition of it.
Same for `QUESTION_EXTRASETTINGS_OTHER_PREFIX` which is only used by question multiple

susnux (approved):
works but dont agree with moving things out of context

Koc (approved):

src/models/Constants.ts
Chartman123: Yes, FILE_SIZE_UNITS could stay in the component itself. I agree 🙂 

QUESTION_EXTRASETTING_OTHER_PREFIX can be put in the Constants.ts file. It also exists in the backend and is defined in Constants.php.

Issue #2700, opened 2025-04-16
Display question ID in the Forms UI for easier Windmill automation


ericfischereu:
**Nextcloud (please complete the following information):**

- Nextcloud-Version: 31
- Forms-Version: 

**Is your feature request related to a problem? Please describe.**
When using Windmill, getting answers from a from submission is only working by querying the question ID. The ID however is not visible in the form.

**Describe the solution you'd like**
Similar to tables: When editing a column, there's an info icon displaying table and column id. The same should be implemented in Forms: Through the "three dot menu" of a question the question ID should be visible.

**Describe alternatives you've considered**
It's possible to set a technical name for each question. This however is not returned by the API. If the technical name would be available in the API response, the ID would be obsolete.

**Additional context**
Example output of `/ocs/v2.php/apps/forms/api/v3/forms/{formId}/submissions`
```
{
    "submission": {
        "id": 27,
        "formId": 3,
        "userId": "admin",
        "answers": [
            {
                "id": 105,
                "text": "Laptop",
                "fileId": null,
                "questionId": 7,
                "submissionId": 27
            },
            {
                "id": 106,
                "text": "Super Laptop 5",
                "fileId": null,
                "questionId": 8,
                "submissionId": 27
            }
        ],
        "timestamp": 1744785175,
        "userDisplayName": "admin"
    }
}
```

Chartman123:
@ericfischereu you can get the required information by querying the other endpoints of our API. Especially the `GET /api/v3/forms/{formId}` where you can find all the information related to that form

ericfischereu:
@Chartman123 Or by just running a workflow once, then you see the what's sent. But that's not really user friendly, we're talking about a Low-Code automation solution. I don't think that we can expect from non developers to call endpoints without a template (right now there's not script for getting form information).

Just make it clear: there are solutions for this, but it's not really user friendly. Taking into account that this should be simple to implement, I think it's worth it. But it's also hot a high priority task of course.

susnux:
We should rather use the technical name for that. Thats exactly what it is there for.

Chartman123:
But we don't need the technical question name in the submissions endpoint ourselves and it's available through the other endpoints. I don't think that we should extend our API for some third-party tools. Especially if we don't know anything about the use case.

@ericfischereu please explain in detail what you're trying to do. A screen recording could also help to fully understand your feature request.

ericfischereu:
I have a workflow, which takes form submissions and writes data into a table.

[...]

PR #2723, opened 2025-04-30
feat: add question name to form answer output

susnux:
* Resolves https://github.com/nextcloud/forms/issues/2700

This allows to identify question from within the UI, as the user can see the name assigned to a question to identify it when using the submissions in external places like workflows.

### DIFF
--- a/docs/DataStructure.md
+++ b/docs/DataStructure.md
@@ -164,19 +164,21 @@ A submission-object describes a single submission by a user to a form.
 
 The actual answers of users on submission.
 
-| Property     | Type    | Restrictions  | Description                                     |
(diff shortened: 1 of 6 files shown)


Chartman123:
@susnux can we get this ready?

Chartman123 (approved):
Working now :)

@susnux other question: should we make the technical question name unique per form? currently you can set the same name for multiple questions...

susnux:
> other question: should we make the technical question name unique per form? currently you can set the same name for multiple questions...

Oh yes this would make sense!

Chartman123:
Ok, I'll add it as a follow-up :)

PR #2737, opened 2025-05-11
feat: add form locking mechanism and update API documentation

Chartman123:
- Implemented form locking functionality to prevent concurrent edits.
- Updated API responses to include `lockedBy` and `lockedUntil` fields.
- Enhanced permission checks to ensure only the form owner can transfer ownership.
- Revised documentation to reflect changes in form data structure and API behavior.


hamza221:
#936

Chartman123:
@susnux @hamza221 should be ready for a first review :)

susnux (approved):
Generally makes sense and looks good!
I would personally have it inverted meaning using `lockedSince` instead of `lockedUntil` as then we can change the locking time easier on updates.
But makes not really a difference.

PR #2739, opened 2025-05-13
fix: disable edit switch when the form is archived

Chartman123:

### DIFF
--- a/src/components/SidebarTabs/SettingsSidebarTab.vue
+++ b/src/components/SidebarTabs/SettingsSidebarTab.vue
@@ -23,6 +23,7 @@
 		</NcCheckboxRadioSwitch>
 		<NcCheckboxRadioSwitch
 			:model-value="form.allowEditSubmissions"
+			:disabled="formArchived"
 			type="switch"
 			@update:model-value="onAllowEditSubmissionsChange">
 			{{ t('forms', 'Allow editing own responses') }}


Koc (approved):

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -23,6 +23,7 @@
 		</NcCheckboxRadioSwitch>
 		<NcCheckboxRadioSwitch
 			:model-value="form.allowEditSubmissions"
+			:disabled="formArchived"
```
Koc, line 26: do we have validation on backend side for that?

Chartman123: yes, this is already implemented here: 
https://github.com/nextcloud/forms/blob/1c6f53f9891bb961e8c1b758c44bdcb77f49bb92/lib/Controller/ApiController.php#L1390

and then here:
https://github.com/nextcloud/forms/blob/1c6f53f9891bb961e8c1b758c44bdcb77f49bb92/lib/Controller/ApiController.php#L1730

and finally here:
https://github.com/nextcloud/forms/blob/1c6f53f9891bb961e8c1b758c44bdcb77f49bb92/lib/Service/FormsService.php#L516

It's the same logic as for new submissions

Chartman123: But good point: I just noticed that we never implemented a check in `updateForm` if the form is archived and then only allow to update the form state by the form owner. 🙈

Chartman123: See #2740

PR #2741, opened 2025-05-13
fix: prevent modifications to archived forms in ApiController and ShareApiController

### LINKED ISSUE #2740: `updateForm` and ShareController don't check if form is archived
We currently do not check if a form is archived when the backend receives requests to change a form attribute or shares (new, update, delete).

If a form is archived, the form owner (or someone with `edit` permissions, too?) should only be able to update the form state to set the form active again.

No-one should be able to add, update, delete shares for archived forms.

Chartman123:
This fixes #2740 by adding checks if the form is archived and not currently in the process of "un-archiving"

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -268,6 +268,17 @@ class ApiController extends OCSController {
 		]);
 
 		$form = $this->getFormIfAllowed($formId);
+		if (
+			$this->formsService->isFormArchived($form)
+			&& !(
+				sizeof($keyValuePairs) === 1
+				&& key_exists('state', $keyValuePairs)
+				&& $keyValuePairs['state'] === Constants::FORM_STATE_CLOSED
+			)
+		) {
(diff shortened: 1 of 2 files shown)


lib/Controller/ApiController.php
```diff
@@ -268,6 +268,17 @@ public function updateForm(int $formId, array $keyValuePairs): DataResponse {
 		]);
 
 		$form = $this->getFormIfAllowed($formId);
+		if (
+			$this->formsService->isFormArchived($form)
+			&& !(
+				sizeof($keyValuePairs) === 1
+				&& key_exists('state', $keyValuePairs)
```
Koc, line 275: is FE sending only changed keys and not all?

Koc (approved):

lib/Controller/ApiController.php
Chartman123: Yes, only the single one :)

PR #2748, opened 2025-05-18
feat: add color question type

### LINKED ISSUE #1617: Color picker, color presets
In some of my forms i used to make the user choose a color from a picker and sometimes also from a list of preset colors or even both, so it would be a useful feature to add a customizable color picker that can be set to pick color from picker or by hex code, pick color from a list or both. The answer can be saved as hex value or rgb based on form setting.

Chartman123:
This closes #1617


Chartman123:
The advanced fields of the NcColorPicker only work when there is already a color chosen: https://github.com/nextcloud-libraries/nextcloud-vue/issues/4037#issuecomment-2889170222

jancborchardt (commented):
Seems nice! :) I'm just not sure about showing the Hex code (at least directly) as that is very nerdy. Would be nicer to have the color boxes on the left so they are all aligned.

jancborchardt:
Also cc @raxp122 for review since they requested the feature. :)

Chartman123:
@jancborchardt thanks for the review. I'll then turn the color box and the hex code around. I think it should be visible so that you can copy/paste the value. In the csv/spreadsheet export it will also only be the text representation as this is how we store it.

raxp122:
It looks so nice, how can i test this?

Chartman123:
@raxp122 thanks for your feedback :) Do you have a test instance? If not, I don't recommend testing it on your prod environment. 

- You can download the files of this Pull Request here: [forms.tar.gz](https://github.com/user-attachments/files/20547684/forms.tar.gz)
- Unzip it and move the extracted files into your apps folder replacing your current forms directory.
- You will then also get all the other changes already merged for the upcoming 5.2 release
- Beware that this will change your database: Make a backup of your database and your forms directory

PR #2761, opened 2025-06-08
feat: make full-text search case-insensitive and search by submission author as well

Koc:


lib/Db/SubmissionMapper.php
```diff
@@ -61,15 +61,18 @@ public function findByForm(int $formId, ?string $userId = null, ?string $query =
...
 			$qb->join(
 				'submissions',
 				$this->answerMapper->getTableName(),
 				'answers',
 				$qb->expr()->eq('submissions.id', 'answers.submission_id')
 			)
 				->andWhere(
-					$qb->expr()->like('answers.text', $qb->createNamedParameter('%' . $query . '%'))
+					$qb->expr()->orX(
+						$qb->expr()->iLike('submissions.user_id', $qb->createNamedParameter($searchString)),
```
Chartman123, line 73: Why didn't you add the `%` here, too?

lib/Db/SubmissionMapper.php
```diff
@@ -156,7 +159,10 @@ protected function countSubmissionsWithFilters(int $formId, ?string $userId = nu
 				$qb->expr()->eq('submissions.id', 'answers.submission_id')
 			)
 				->andWhere(
-					$qb->expr()->like('answers.text', $qb->createNamedParameter('%' . $searchString . '%'))
+					$qb->expr()->orX(
+						$qb->expr()->iLike('submissions.user_id', $qb->createNamedParameter($searchString)),
```
Chartman123, line 163: Same here of course

Chartman123 (commented):
I think the search should always do the placeholder search for consistency. On the other hand, I'm not really sure if we should really extend the filtering to the username.

lib/Db/SubmissionMapper.php
Koc: IMHO for the usernames it's better to have full-match or at least search by prefix.

Koc:
I've updated PR description with motivation of this change.

lib/Db/SubmissionMapper.php
Chartman123: Ok, but we need to take into account that the shown name in the frontend and the stored user id in the backend don't need to be the same. They can even be totally different, especially in large organizations. In those cases a placeholder search might be very handy.

Koc: added

Chartman123 (approved):

PR #2772, opened 2025-06-25
fix: properly set string values in exports to escape formulas

susnux:
* follow up on https://github.com/nextcloud/forms/pull/2674


Chartman123:
> /backport to stable4

Stable4 is already out of support and has a branch protection in place.

Chartman123 (approved):

PR #2894, opened 2025-09-01
style(icon): update icons to outline variants in TopBar and Results components

Chartman123:


AndyScherzinger (approved):
Nice! I didn't catch them (not the pattern I searched for replacing icons 👍

PR #2824, opened 2025-07-27
Section functionality

shynkaruck:
# Add Section functionality to Nextcloud Forms

## Description

This PR adds a new "Section" element type to Nextcloud Forms that acts as a visual separator to group fields logically, without affecting data storage, validation, or submission logic.

### 1. Section Creation
- New "Section" element type in the form editor
- Only parameter: section name (title) visible to users

### 2. Editor Display
- Sections appear as non-input elements
- Draggable for reordering
- Title displayed in bold/visually distinct style
- No description field for sections
- Bottom separator line for visual distinction

### 3. Form Filling Display
- Sections display as headings/subheadings at their position
- No influence on validation, mandatory fields, or submission flow

### 4. Export/View Responses Behavior
- Sections do not affect CSV export or response viewing
- Not stored in user responses

#### Constants and Types
- **`lib/Constants.php`**: Added `ANSWER_TYPE_SECTION = 'section'` constant
- **`lib/ResponseDefinitions.php`**: Updated `FormsQuestionType` to include `"section"`

#### API Controller (`lib/Controller/ApiController.php`)
- **Section Validation**: Sections cannot be required, have options, or file uploads
- **Answer Storage**: Sections are filtered out from answer storage
- **Export Filtering**: Sections are excluded from submissions export
- **Form Cloning**: Sections maintain `isRequired = false` when cloning forms

#### Submission Service (`lib/Service/SubmissionService.php`)
- **Export Data**: Sections are filtered out from CSV/Excel export data
- **Validation**: Sections are ignored during submission validation

#### Components
- **`src/components/Questions/QuestionSection.vue`**: New component for section rendering
- **`src/components/Questions/Question.vue`**: Updated to support section display and editing
- **`src/models/AnswerTypes.js`**: Added section type with appropriate icon and labels
- **`src/models/Constants.ts`**: Added `ANSWER_TYPE_SECTION` constant

#### Views
- **`src/views/Create.vue`**: Updated to pass question type to components
- **`src/views/Submit.vue`**: Updated to filter sections from validation and storage
- **`src/components/Results/ResultsSummary.vue`**: Updated to handle sections in results display
- **`src/components/Results/Submission.vue`**: Updated to filter sections from submission data

#### ApiControllerTest.php
- `testUpdateQuestion_sectionCannotBeRequired()`: Verifies sections cannot be made required
- `testNewOption_sectionCannotHaveOptions()`: Verifies sections cannot have options
[...]

PR #2913, opened 2025-09-08
feat: introduce Section as a new question type

### LINKED ISSUE #624: Add sections or multiple descriptions
- Nextcloud-Version: 19.0.3
- Forms-Version: 2.0.4

Forms with many questions or with very different questions cannot be divided into sections for better clarity/visibility/readability. It's all one long list of questions.

Add multiple descriptions in the form that do not contain a question and are not part of the results for the sole purpose of dividing questions into groups/sections and provide additional information.

Could be a question with a hidden field.

**Additional context**
Example: registration for a hackathon. There is one section for the submitter (name, email, phone), one section for the team (name, description, previous hackathons), and one section for the project description (name, excerpt, long description, category). Each section contains a short text with some hints on what to look out for while filling the form.

AIlkiv:
Based on https://github.com/nextcloud/forms/pull/2824

@Chartman123  This is an interesting feature because in the future it can be expanded to split one form into pages by section. It can also be expanded to display pages only when a certain condition is met. Much has already been written about this.

Demo:
https://github.com/user-attachments/assets/fc7edb0d-1f24-4493-9c48-0a90dda21052

The only problem I haven't figured out how to solve is that on the mobile layout, the menu partially overlaps the Section (you can see it in the demo).

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -76,6 +76,7 @@ class Constants {
 	public const ANSWER_TYPE_LONG = 'long';
 	public const ANSWER_TYPE_MULTIPLE = 'multiple';
 	public const ANSWER_TYPE_MULTIPLEUNIQUE = 'multiple_unique';
+	public const ANSWER_TYPE_SECTION = 'section';
 	public const ANSWER_TYPE_SHORT = 'short';
 	public const ANSWER_TYPE_TIME = 'time';
 
@@ -95,6 +96,7 @@ class Constants {
 		self::ANSWER_TYPE_LONG,
 		self::ANSWER_TYPE_MULTIPLE,
 		self::ANSWER_TYPE_MULTIPLEUNIQUE,
+		self::ANSWER_TYPE_SECTION,
 		self::ANSWER_TYPE_SHORT,
 		self::ANSWER_TYPE_TIME,
 	];
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -1187,9 +1187,12 @@ class ApiController extends OCSController {
 		}
 		$questions = [];
 		foreach ($this->formsService->getQuestions($formId) as $question) {
+			if ($question['type'] === Constants::ANSWER_TYPE_SECTION) {
+				continue;
+			}
+
 			$questions[$question['id']] = $question;
 		}
-
 
 		// Append Display Names
 		$submissions = array_map(function (array $submission) use ($questions) {
--- a/lib/ResponseDefinitions.php
+++ b/lib/ResponseDefinitions.php
@@ -43,7 +43,7 @@ namespace OCA\Forms;
  *   questionType?: string,
  * }
  *
- * @psalm-type FormsQuestionType = "dropdown"|"multiple"|"multiple_unique"|"date"|"time"|"short"|"long"|"file"|"datetime"|"grid"
+ * @psalm-type FormsQuestionType = "dropdown"|"multiple"|"multiple_unique"|"date"|"time"|"short"|"long"|"file"|"datetime"|"grid"|"section"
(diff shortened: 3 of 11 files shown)


Chartman123:
> The only problem I haven't figured out how to solve is that on the mobile layout, the menu partially overlaps the Section (you can see it in the demo).

@AIlkiv @susnux I could imagine that we move the form title to the top bar like in Talk where the chat name is visible there. In Forms we could do it like this:
- Show the title normally in the app content as long as it's visible
- Move the title to the top bar as soon as it scrolls out of the visible area
- Always place the top bar above the app content
- Stick the section below the top bar.

One problem I can see with your current implementation: If there is a very long description added to a section, you're not able to see the questions anymore.

AIlkiv:
> One problem I can see with your current implementation: If there is a very long description added to a section, you're not able to see the questions anymore.

I can add max-height + overflow-y: auto

AIlkiv:
@Chartman123 I have already added a height restriction for the block.

demo:


https://github.com/user-attachments/assets/13c6eb99-1e6c-44d9-9e8f-a138c76830c8

AIlkiv:
@Chartman123 If the only problem is the menu overlap, then I suggest we merge this PR and discuss the layout change in a new issue, as it goes beyond the scope of this task.

Chartman123:
@AIlkiv no, we should do this in this PR already before merging. Other than that our main branch is currently ready for the 5.2 release and I'd like to have this in the 5.3 release, so we need to wait until we branch off the stable 5.2

I also don't like the scrollbars that also appear when there's still enough space for the description... So I added Jan and Nimisha from the design team to have a look :)

AIlkiv:
> I also don't like the scrollbars that also appear when there's still enough space for the description... So I added Jan and Nimisha from the design team to have a look :)

@Chartman123 Scroll appears only for long descriptions.

jancborchardt (changes requested):
Very nice! Some feedback:
- The divider line is too intense / has too much contrast with the background. Does it use var(--color-border)?
- Indeed there is a risk of too large descriptions, but also having intermittent scrollable areas is not so nice. I would suggest to limit the description to 2 rows upon scrolling down, and you can read it in full again by scrolling up, or tapping it will scroll you up to it?

susnux:
I am not really sure about the implementation as a question type. Its not really a question type but rather some formatting / structuring.
I know we have also feature requests for paged forms, so maybe we should rather think about a besser way to represent such cases then misuse question types for it.
Examples could be 'forms-sections' which could be assigned to a question or similar, something one could brainstorm about.

AIlkiv:
> I am not really sure about the implementation as a question type. Its not really a question type but rather some formatting / structuring. I know we have also feature requests for paged forms, so maybe we should rather think about a besser way to represent such cases then misuse question types for it. Examples could be 'forms-sections' which could be assigned to a question or similar, something one could brainstorm about.

@susnux  In the current context, you are right. But we can expand it and call it “form element” instead of “question.”
I think introducing a new entity will complicate the module more than changing “question.”

To add output to pages in the future, we will simply need to add a “separate page” checkbox to the Section settings in the current implementation.

haraldreingruber:
Is there anything I can help to move this forward? 

If a different approach is preferred, would a proof of concept be a good idea, or is there no consensus yet?

jancborchardt (commented):
@Chartman123 @susnux what's your current view on this, and how to proceed? Just so this is unblocked for @AIlkiv :)

Chartman123:
@jancborchardt I had another look at it... Designwise I'm still not very happy with it and I think your suggestions/requests haven't been integrated yet.

@susnux what about the implementation as question type? Should we leave it like this for now?

@AIlkiv If we keep it as a question type, please add some error handling in the backend, that sections can't take any answers. This should also be present in the tests.

Chartman123:
Closed in favor of #3664

PR #2924, opened 2025-09-14
feat: introduce Grid as a new question type

### LINKED ISSUE #2606: [QuestionType] Multiple choice grid
Create a new question type for multiple choice grids. This is basically the same as for linear scale, but has multiple rows

Koc:
Closes #2606

:construction: Development still in progress

I'm trying to reuse already existent `Option` entity by adding extra column `optionType=row|column`.

## :heavy_check_mark: Implemented
<details><summary>:mag: Cols/Rows management</summary>
<p>

image

</p>
</details> 

<details><summary>:mag: Cols/Rows titles and various input type</summary>
<p>

image

</p>
</details> 

<details><summary>:mag: Very basic table rendering</summary>
<p>

image

</p>
</details> 

## :spiral_notepad: Todo
- [x] Fix total mess with input focus
- [x] Fix items reordering
- [x] Create nested menu to create type of the answer (it is impossible to change answer type later)
- [ ] Proper rendering on answer mode
- [ ] Store and display results
- [ ] Export

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -71,6 +71,7 @@ class Constants {
 	public const ANSWER_TYPE_DATETIME = 'datetime';
 	public const ANSWER_TYPE_DROPDOWN = 'dropdown';
 	public const ANSWER_TYPE_FILE = 'file';
+	public const ANSWER_TYPE_GRID = 'grid';
 	public const ANSWER_TYPE_LINEARSCALE = 'linearscale';
 	public const ANSWER_TYPE_LONG = 'long';
 	public const ANSWER_TYPE_MULTIPLE = 'multiple';
@@ -85,6 +86,7 @@ class Constants {
 		self::ANSWER_TYPE_DATETIME,
 		self::ANSWER_TYPE_DROPDOWN,
 		self::ANSWER_TYPE_FILE,
+		self::ANSWER_TYPE_GRID,
 		self::ANSWER_TYPE_LINEARSCALE,
 		self::ANSWER_TYPE_LONG,
 		self::ANSWER_TYPE_MULTIPLE,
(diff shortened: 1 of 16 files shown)


Chartman123:
@Koc I wouldn't combine checkbox and radio grids into one single question type, but make it two types like we already have for "normal" checkbox/radio questions.

Koc:
@Chartman123 Did I understand your correctly that we should have multiple question types here? Not just grid but "Grid - checkboxes", "Grid - radios", "Grid - dropdowns", etc.

I thought to create nested menu for question creation. Select "Grid" on a first step and then select input type on a second step.

Chartman123:
> Not just grid but "Grid - checkboxes", "Grid - radios", "Grid - dropdowns", etc.

Yes, no dropdown grid, but these two:

- Multiple choice grid
- Checkbox grid

You can use the same component like QuestionMultiple.vue for checkboxes and radios, so in this case QuestionGrid.vue and depending on the prop `unique` it is a radio grid or a checkbox grid, but two different question types `grid` and `grid_unique`.

https://github.com/nextcloud/forms/blob/3424510afd7089ecc4c24e5f8aafcd0dcbd52536/src/models/AnswerTypes.js#L89-L90

And I think that we should keep the options as they are defining just the rows. The number of columns and their labels could be stored in `extraSettings`.

nimishavijay:
Awesome feature! :) I will personally definitely use this :) 

So the main screen which can use some design is the table rendering, I tried a first basic version here simulating real questions:
image

What seemed to work:
- [ ] Always use full width as @jancborchardt mentioned
- [ ] Each row min-height is 34px
- [ ] First row min-height is 44px
- [ ] Each column min-width is 64px
- [ ] First column min-width to 200px
- [ ] Gap between rows is 4px
- [ ] Gap between columns is 8px
- [ ] Inside the table use the MDI radio or checkbox icons
- [ ] Use border-bottom: 2px solid var(--color-border) for the first row to separate the heading 



Other than that some comments while editing:
- [ ] Inside the action menu, not sure if we need a row and column name field as the labels for the rows and columns should ideally be self-explanatory, and we additionally run into the problem that the first cell of the table will have 2 strings to be displayed 
- [ ] Inside the "Add row" or "Add column" field, the icon can be a plus icon instead of the row or column icon
- [ ] Wording changes
  - [ ] Grid rows --> Rows
  - [ ] Grid columns --> Columns 
  - [ ] Add a new answer option --> Add row/Add column
  - [ ] Capitalize the options inside the dropdown in the action menu (radio --> Radio)

Additionally I know that trying to align the layout of view and edit mode is something that is desired generally so we can try a similar layout for this question type as a follow up, if you wish :) It would look something like this:

image

Koc:
@nimishavijay Thank you for styling guides, I will work on it soon.

@Chartman123 I'd prefer to use options because we need ids to associate selected values with rows/columns. User can rename them, so we need rely on id. Also we need to support more complex scenarios than just radio/checkbox. E.g. matrix of numbers, see screen.

image

image


Also we already have components for manage options and this is a great opportunity to reuse them


NB! I plan to reopen this PR in upstream instead of my fork.

Koc:
let's continue in #2945

PR #2944, opened 2025-10-01
fix(Results): enhance response loading logic based on active response view

### LINKED ISSUE #2943: Title: Summary view only reflects the last 20 answers after update to Forms 5.2.0

**Describe the bug**

After updating to Forms 5.2.0 on Nextcloud 30.0.15, we noticed a change in how responses are displayed in the Results tab.

The update introduced pagination for responses (which is a good idea, it makes loading faster with many submissions).

However, the summary view only takes into account the currently displayed page of results (e.g. only the last 20 answers), instead of aggregating all responses.

This makes the summary chart/statistics misleading and basically useless for forms with many answers.

**To Reproduce**

1. Create a form and collect more than 20 responses. 
2. Go to the Results tab.
3. Observe that the summary graphs/charts only represent the 20 visible responses (the current page), not the full dataset.

**Expected behavior**

The summary view should always be calculated on all responses, regardless of pagination. Pagination should only affect the list of individual answers, not the aggregated statistics.

- Nextcloud: 30.0.15
- Forms app: 5.2.0

**Additional context**

Before the update, the summary correctly reflected all submissions. Since the update, we can only trust the CSV export to get real totals.

Chartman123:
This fixes #2943 by always loading all submissions in the summary view

### DIFF
--- a/src/components/PillMenu.vue
+++ b/src/components/PillMenu.vue
@@ -9,7 +9,7 @@
 			v-for="option of options"
 			:key="option.id"
 			:aria-label="isMobile ? option.ariaLabel : null"
-			:checked="active.id"
+			:model-value="active.id"
 			:disabled="disabled || option.disabled"
 			class="pill-menu__toggle"
 			:class="{ 'pill-menu__toggle--icon-only': isMobile && option.icon }"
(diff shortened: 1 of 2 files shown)


Chartman123:
Change in PillMenu is unrelated, but slipped through when I switched from `checked` prop to `modelValue` for NcCheckboxRadioSwitch components

PR #2957, opened 2025-10-10
Refactor code for compliance with new eslint version

Chartman123:


susnux:
I think we now have two options:
- keep prettier and add it to the eslint config
- remove prettier and solely rely on the eslint formatting

Chartman123:
> I think we now have two options:
> 
>     * keep prettier and add it to the eslint config
> 
>     * remove prettier and solely rely on the eslint formatting

I'd say let's keep prettier? Or does this have any bigger downsides?

src/components/Questions/QuestionDropdown.vue
susnux: BTW this warning often comes from using mixins, so as long as we use mixins we can ignore it.

Chartman123:
> why removing prettier ci? In case we do not enforce this on CI, we could also just completely remove prettier and just use eslint for formatting.

Because it's already monitored in ESLint... If you have code, that doesn't comply to prettier rules, eslint will fail. So we don't need it twice

susnux:
> Because it's already monitored in ESLint... If you have code, that doesn't comply to prettier rules, eslint will fail. So we don't need it twice

Ah ok makes sense :)

PR #2972, opened 2025-10-15
refactor: Migrate components for Vue 3 compatibility

Chartman123:


pringelmann:
@Chartman123 please let me know if there is anything specific you would like me to help with. Happy to help push this forward but don't want to duplicate work :)

Chartman123:
@pringelmann I did some changes today that I still need to test this evening.

From my point of view, two things are currently still missing: e2e tests are broken and reordering questions and options with the up/down arrows aren't working.

Chartman123:
@pringelmann I was able to fix the sorting problems, so now only the e2e tests are still failing. Perhaps you can have a look at it

Chartman123:
Looks already quite good. Some things I noticed that are still not working as expected:

- [x] (re)storing values to/from LocalStorage
- [x] date/time picker:
  - [x] time picker shows current time by default
  - [x] date picker doesn't accept date when clicking on it in the picker
- [x] add multiple options layout broken
- [x] add multiple options not working
- [x] Short/Long/Date/Time/File/Linear Scale/Color show missing title indicator even if title is set
- [x] Editing a submitted response from the results view doesn't load values
- [ ] deleting options is broken for radio questions (broken in `main` too)

pringelmann:
Taking a look

Chartman123:
Too slow/quick 😁 wanted to squash the commits first...

Issue #3049, opened 2025-11-17
Only show users in transfer dialog


mahibi:
Technically it is only possible to transfer a form to a user. 

**Actual behavior**
In the transfer dialog more options are offered (e.g. teams). When this is chosen it will lead to an error during the transfer. 

**Expected behavior**
As forms can be only owned by an user, only users should be listed.

artonge:
PR: https://github.com/nextcloud/forms/pull/3050

PR #3050, opened 2025-11-20
fix(Settings): Only list users in transfer ownership

artonge:
Fixed two issues:

## 1. Selecting a result had no effect

https://github.com/user-attachments/assets/414709de-23e3-4b8b-93a7-06e9011c48be

## 2. Non-user results were presented to the user

Fix https://github.com/nextcloud/forms/issues/3049

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/src/components/SidebarTabs/TransferOwnership.vue
+++ b/src/components/SidebarTabs/TransferOwnership.vue
@@ -44,7 +44,9 @@
 					:loading="loading"
 					:options="options"
 					:placeholder="t('forms', 'Search for a user')"
-					@search="asyncSearch">
+					@search="
+						(query) => asyncSearch(query, [SHARE_TYPES.SHARE_TYPE_USER])
+					">
 					<template #no-options>
 						{{ noResultText }}
 					</template>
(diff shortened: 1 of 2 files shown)


artonge:
> #3049 is about the _ownership transfer_ not the sharing. Sharing still needs to be possible with groups and teams.

Oups, indeed, shouldn't have rushed it...

Corrected.

Chartman123:
@artonge @susnux I've also already looked at this bug and tried to fix it with the `filter-by` prop of `NcSelectUser` but wasn't successful. There might be a bug in the vue lib that prevents passing the filter function to the component. But this filter would probably be the easiest way to fix this.

artonge:
> But this filter would probably be the easiest way to fix this.

Yeah, that or a computed property.

The advantage of the current solution is that we do not ask the server for more information than we need. Happy to comply, thought :).

Chartman123:
Yes, that was also my first attempt to fix it, but I didn't succeed in passing the required parameters to the search mixin. Which we should also get rid of during vue3 migration.

susnux (approved):
👍

PR #3051, opened 2025-11-21
feat: shows a confirmation dialog on question deletion

Koc:
We've found that sometimes users can accidentally delete their questions. Let's add confirmation dialog to be more consistent with other deletions.

image

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -114,7 +114,7 @@
 						</template>
 						{{ t('forms', 'Copy question') }}
 					</NcActionButton>
-					<NcActionButton @click="onDelete">
+					<NcActionButton close-after-click @click="onDelete">
 						<template #icon>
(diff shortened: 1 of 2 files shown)


Chartman123:
Just a question: how could you accidentally delete a question? The action is already hidden in the question menu, so you have to actively open the menu and then click on the button. I don't think that we should over complicate our UI and have too many pop up dialogs.

Google doesn't show a confirmation when you delete a question, same for Microsoft. Both already show the delete button directly on the question and not even inside a menu.

susnux:
> Google doesn't show a confirmation when you delete a question, same for Microsoft. Both already show the delete button directly on the question and not even inside a menu.

Agree. At the moment  I do not see why we need this dialog.
Maybe you can clarify how users accidentally delete them?

Koc:
let's close it for now. I don't know how it happened (how user accidentally removed their question)

PR #3060, opened 2025-11-23
fix: transfer ownership confirmation string formatting

AIlkiv:
Before:

https://github.com/user-attachments/assets/19fcddf9-1d82-4cb9-a364-f20cd0832f49

After:

https://github.com/user-attachments/assets/16f951c7-b932-40ab-afbe-fa88f509508e

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -81,6 +81,7 @@
 						:placeholder="t('forms', 'Form title')"
 						required
 						autofocus
+						@keydown.enter.prevent
 						@input="onTitleChange" />
 				</h2>
 				<label class="hidden-visually" for="form-desc">


Chartman123:
@AIlkiv nice catch. But I'm wondering if we shouldn't disallow line breaks in the title in the first place (they should then be cleaned up in the backend, too)

@susnux what do you think?

susnux:
> But I'm wondering if we shouldn't disallow line breaks in the title in the first place (they should then be cleaned up in the backend, too)

Agree!

Chartman123:
@AIlkiv could you please add the changes to disallow line breaks (and possibly tabs) in the title to the front and backend? 

What about existing titles in the database? should we add a migration, too?

AIlkiv:
@Chartman123 If I do this on the backend when saving, will it be okay?

Chartman123:
@AIlkiv I've added an event handler to the title to prevent adding new lines, but I think that we should also adjust `onTitleChange` to handle copy/paste actions that already include line breaks.

If we just sanitize the string on saving in the backend the user might wonder why the title looks different after saving the form. Instead the API could throw an error if the title contains unwanted characters.

AIlkiv:
@Chartman123 Thank you, I understand. I will try to finish it according to your comment.

PR #3127, opened 2026-01-22
chore: adjust metadata for main branch

susnux:
- main is v5.3
- this app is AGPL-3.0-or-later
- update workflows from organization
- support Nextcloud 33 on main branch


susnux:
This is only for main - for stable5 @Chartman123 will add Nextcloud 33 support :)

Issue #3121, opened 2026-01-17
Answers are visible although no view right is shared


piiskop:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**
Although the form is shared with no view right answers of all the respondents are visible to all users.

**To Reproduce**

1. Create a form!
2. Share the form with at least two other accounts no "Vaata vastuseid"-right!
3. Let the owners of these two accounts submit the form answers!
4. Experience that all of them see the answers of each other!

**Expected behavior**
Answers are not visible to respondents if the form is shared with no "Vaata vastuseid"-right to them.

**Screenshots**

**Nextcloud (please complete the following information):**

- Nextcloud-Version: [Nextcloud Hub 25 Autumn](https://nextcloud.com/) (32.0.3)
- Forms-Version: 5.2.3

**Desktop (please complete the following information):**

- OS: Ubuntu
- Browser Chromium
- Version 143.0.7499.169 (Official Build) snap (64-bit)

**Smartphone (please complete the following information):**

**Browser log**

```
Open your console, reload your page and/or do the action leading to this issue and copy/paste the log in this thread.
```

<details>
<summary>How to access your browser console (Click to expand)</summary>

# Chrome

- Press either CTRL + SHIFT + J to open the “console” tab of the Developer Tools.
- Alternative method:
    1. Press either CTRL + SHIFT + I or F12 to open the Developer Tools.
    2. Click the “console” tab.

# Safari

- Press CMD + ALT + I to open the Web Inspector.
- See Chrome’s step 2. (Chrome and Safari have pretty much identical dev tools.)

# IE9

1. Press F12 to open the developer tools.
2. Click the “console” tab.

# Firefox

- Press CTRL + SHIFT + K to open the Web console (COMMAND + SHIFT + K on Macs).
- or, if Firebug is installed (recommended):
    1. Press F12 to open Firebug.
    2. Click on the “console” tab.

# Opera

1. Press CTRL + SHIFT + I to open Dragonfly.
2. Click on the “console” tab.
 </details>

**Additional context**

Chartman123:
@piiskop I just tested it the way you described it and couldn't reproduce it. Only the owner of the form could see both results, the two users only saw their own submissions. so everything works as expected for me.

Could you perhaps add some screenshots? And try to reproduce it with a second form?

[...]

PR #3128, opened 2026-01-22
fix: do not add result permissions to see own submissions

susnux:
* resolves https://github.com/nextcloud/forms/issues/3121

When a user has submitted a form they should not have result permissions, but instead we should only show their own submissions.


Chartman123:
Just checked it, seems to fully fix the issue :) Thanks!

susnux:
> Looks good, just the failing tests :)

Randomly worked before as the parameters were passed positional but in the meantime we added new parameters :see_no_evil: 
fixed :white_check_mark:

PR #3183, opened 2026-02-26
Fix IME composition handling in option inputs

don9x2E:
## Summary

This PR now includes a follow-up fix after the initial IME handling patch.

- Prevent unintended option creation while IME composition is in progress
- Move new option creation for the local "Add a new answer option" row to explicit user intent
  - create only on `Enter` (after composition ends) or explicit `+` button click
- Keep normal live updates for already-created options
- Move focus to the next empty local option after creating one with Enter
- Fix delayed "Delete form" confirmation dialog opening by rendering the dialog outside the actions slot

## Why this follow-up was needed

The previous attempt improved composition guards, but real Korean IME usage still hit edge cases where input-driven side effects could trigger unexpectedly in the local option row.

In practice, automatic creation on input was still risky under composition timing differences across browsers/IME implementations.

## Behavioral change and trade-off

The local empty option row no longer auto-creates an option purely from input events.
Instead, creation is explicit (`Enter` or `+` button).

This slightly reduces automation, but removes IME race/side-effect paths and gives deterministic behavior for CJK input users.

## Repro (before)

- Create form
- Add multiple-choice/dropdown/radio options
- Type Korean with IME in the local "Add a new answer option" field
- Observe unexpected creation timing / side effects in some flows

## Verification (after)

- During IME composition: no automatic option creation
- After composition ends: `Enter` creates exactly one option
- `+` button creates exactly one option
- Focus moves to the newly appended empty local option row
- "Delete form" shows confirmation dialog immediately on click

### DIFF
--- a/src/components/Questions/AnswerInput.vue
+++ b/src/components/Questions/AnswerInput.vue
@@ -22,7 +22,7 @@
 			@input="debounceOnInput"
 			@keydown.delete="deleteEntry"
 			@keydown.enter.prevent="focusNextInput"
-			@compositionstart="onCompositionEnd"
+			@compositionstart="onCompositionStart"
 			@compositionend="onCompositionEnd" />
 
 		<!-- Actions for reordering and deleting the option  -->
@@ -295,7 +295,10 @@ export default {
 		/**
 		 * Request a new answer
 		 */
-		focusNextInput() {
+		focusNextInput(e) {
+			if (this.isIMEComposing || e?.isComposing) {
+				return
+			}
 			if (this.index <= this.maxIndex) {
 				this.$emit('focus-next', this.index, this.optionType)
 			}
@@ -308,6 +311,10 @@ export default {
 		 * @param {Event} e the event
 		 */
 		async deleteEntry(e) {
+			if (this.isIMEComposing || e?.isComposing) {
+				return
+			}
+
 			if (this.answer.local) {
 				return
 			}


Chartman123:
@don9x2E thanks for your fix :) please sign-off your commit

susnux (approved):

don9x2E:
Follow-up update pushed in commit 4b8ae6d6.

What was added after the approved IME patch:
- local option row now creates options only on explicit intent (Enter after composition end or + button)
- focus moves to the next empty local option after creating one
- delete-form confirmation dialog now opens immediately (moved out of actions slot)

This resolves the remaining Korean IME edge-case observed in real usage.

don9x2E:
Design proposal only (no code change): restoring auto-create safely

If we want to bring back auto-create for the local option row later, a safer staged approach is:

1. Keep current explicit mode as default.
2. Re-enable auto-create only for non-IME input paths first.
3. Add a strict guard layer for IME paths:
   - never create on input/composition events
   - allow create only after compositionend + explicit user intent (Enter/click)
4. Add browser/IME matrix tests (Korean/Japanese/Chinese IME, Chromium/Firefox/WebKit).
5. Optional feature flag (or config) to compare behavior before making it default.

Acceptance criteria for any auto-create restoration:
- zero unintended creates during composition
- exactly one create per explicit intent
- stable focus behavior after creation
- no regressions for existing option update flow.

Issue #3099, opened 2025-12-18
Delete form only possible by super weird, click order


indivisualvj:
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**
When i try to delete a form, i have to click the ... menu again in order to open up the delete confirmation dialog.

**To Reproduce**

1. Create Form
2. Click on ... menu then on delete
3. Nothing happens
4. Click on ... menu button again
5. Confirmation dialog will appear and form can be deleted

**Expected behavior**
Directly open the confirmation dialog after i press on delete in the menu

**Screenshots**

<img width="786" height="646" alt="Image" src="https://github.com/user-attachments/assets/c4d1a12c-4486-4824-b51c-cc196e355d04" />

<img width="1796" height="744" alt="Image" src="https://github.com/user-attachments/assets/3e0572d1-91b8-4834-8a0d-b8209f911221" />

**Nextcloud (please complete the following information):**

- Nextcloud-Version: 32.0.3
- Forms-Version: 5.2.3

**Desktop (please complete the following information):**

- OS: Linux (Mint 22.2)
- Browser: Firefox
- Version: 146.0 (64-bit)

**Smartphone (please complete the following information):**
n.a.

**Browser log**

```
Error: Your focus-trap must have at least one container with at least one tabbable node in it at all times
    f focus-trap.esm.js:410
    activate focus-trap.esm.js:864
    useFocusTrap NcModal.mjs:582
    VueJS 37
        me
        t
        _enterCb
        qn
        ji
        Ut
        Ei
        Ei
        Ut
        bind
        on
        qc
        dt
        y
        u
        Uc
        _update
        r
        get
        e
        ys
        $mount
        init
        p
        u
        Uc
        _update
        r
        get
        e
        ys
        $mount
        init
        p
        u
        Uc
        _update
vue.runtime.esm.js:3065:17

```

**Additional context**
n.a.

Chartman123:
@susnux could this be caused by a change in the vue lib? we could "fix" this by removing the `close-after-click` from the delete action, but then the menu is kept open and still accepts clicks while the modal NcDialog is opened.

susnux:
The reported error will be solved with https://github.com/nextcloud-libraries/nextcloud-vue/pull/8093


[...]

PR #3185, opened 2026-02-26
Follow up IME option creation and delete dialog timing fixes

don9x2E:
## Summary

Follow-up fixes after #3183 to address remaining real-world Korean IME issues and related UX regressions.

- Prevent unintended option creation while IME composition is active
- Local "Add a new answer option" row now creates only on explicit intent:
  - `Enter` after composition end, or
  - explicit `+` button click
- Keep live update behavior for already-created options
- After creating an option with Enter, move focus to the next empty local option row
- Fix delayed "Delete form" confirmation dialog by rendering dialog outside the actions slot

## Why this follow-up is needed

The initial IME patch improved composition handling, but in actual Korean IME use there were still edge cases where input-driven side effects could trigger option creation timing unexpectedly.

Switching local-row creation to explicit intent removes those composition timing race paths while preserving normal editing for existing options.

## Trade-off

Auto-create-on-input for the local empty row is intentionally disabled for now to prioritize deterministic IME-safe behavior.

## Validation

- IME composition does not auto-create options
- Exactly one option is created on explicit Enter/click
- Focus advances to the next empty local row after creation
- Delete-form confirmation opens immediately on click

### DIFF
--- a/playwright/e2e/ime-input.spec.ts
+++ b/playwright/e2e/ime-input.spec.ts
@@ -17,7 +17,7 @@ test.beforeEach(async ({ page }) => {
 })
 
 test(
-	'IME input does not trigger new option',
+	'IME input requires explicit intent to create new option',
 	{
 		annotation: {
 			type: 'issue',
@@ -75,7 +75,12 @@ test(
 		await client.send('Input.insertText', {
 			text: 'さ',
 		})
-		// so there were 4 inputs but those should only result in one new option
+		// Committing composition text alone must not create a new option.
+		await expect(question.answerInputs).toHaveCount(0)
+		await expect(question.newAnswerInput).toHaveValue('さ')
+
+		// Explicit intent (Enter) creates exactly one option.
+		await question.newAnswerInput.press('Enter')
 		await expect(question.answerInputs).toHaveCount(1)
 		await expect(question.answerInputs).toHaveValue('さ')
 	},
--- a/src/components/Questions/AnswerInput.vue
+++ b/src/components/Questions/AnswerInput.vue
@@ -11,9 +11,9 @@
 			class="question__item__pseudoInput" />
 		<input
 			ref="input"
+			v-model="localText"
 			:aria-label="ariaLabel"
 			:placeholder="placeholder"
-			:value="answer.text"
 			class="question__input"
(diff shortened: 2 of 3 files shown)


src/components/AppNavigationForm.vue
Chartman123: Could you please remove this unrelated change?

don9x2E:
Thanks for the review! 

- I’ll move the AppNavigationForm.vue formatting change into a separate PR that references #3099.

- I’ll update the add option button to be disabled only during active IME composition (so it works for direct inputs).

- I’ll add the missing JSDoc blocks for the new handlers.

- I’ll push the follow-up updates shortly.

Chartman123:
@don9x2E now the button is always enabled, even if there's no input at all. 😉 

@susnux Can't we get rid of the debounce logic for onInput if we only submit the new options upon pressing enter or clicking the button?

don9x2E:
Addressed: the add button is now disabled when input is empty (kept enabled for non-empty input, only disabled during active IME composition).

Re debounce: we still need it for updating existing options (typing in non-local options triggers PATCH).

For new option creation, it only happens via Enter or the add button.

Chartman123:
@don9x2E the button is still not working as expected: now it stays disabled even if there is some input. It looks like your canCreateLocalAnswer is not executed upon input.

Chartman123:
@don9x2E still the same... there were some issues with reactivity left. I tried to fix it with Copilot. Please have a look at my commit, if it's still working correctly during IME input. non-IME input is working now.

Before we merge it, please rebase your code on current main and also squash all commits into a single one.

susnux:
> Can't we get rid of the debounce logic for onInput if we only submit the new options upon pressing enter or clicking the button?


I guess so

don9x2E:
Rebased on current main and squashed into a single commit (e6dd00a4) with DCO sign-off. Also removed the AppNavigationForm changes (handled separately in #3188).

Chartman123:
@don9x2E you've lost my changes

don9x2E:
Sorry for the missing. Re-applied the reviewer changes and force-pushed the squashed commit (c39eb6f6).

Chartman123:
ok, now just the lint error and then we're good to go :)

Issue #1278, opened 2022-08-04
Voters can rank the options given in order of importance


ilarimikkonen:
**Nextcloud (please complete the following information):**
- Nextcloud-Version: 24 - or "latest"
- Forms-Version: 2.5.1 - or "latest"

**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
As a entity organising voting, I'd like to give participants the possibility to rank options in order of importance (dragging visual elements or other method) so that it's easier/possible for the voters to answer questions where ranking is needed.

**Describe alternatives you've considered**
N/A

**Additional context**

as an idea:

![image](https://user-images.githubusercontent.com/7221736/182840671-bfad7bb8-a7da-4dcc-a83a-62c62075e1f2.png)

Chartman123:
@jotoeri Should this be realized as a new question type? I'd add the corresponding label then...

robertpetry:
The ability to do a ranked ballot would be great.  I appreciate the suggestion for a phone friendly interface where one could make the rank by moving options around, but if that interface is slowing down development just having boxes the users fill with the ranking numbers would be fine for at least my/many organizations.  One could maybe drag rank numbers from a palette if that would be an easier phone interface. A few other comments

1.) In many voting systems it is possible to choose to rank options the same or not rank them entirely.  (In the latter case unranked candidates are considered to have been ranked equally at last.) So for instance a legitimate ballot is
2 option1 
(unranked) option 2
(unranked) option 3
1 option 4
2 option 5
This indicates option 4 is the first preference, options 1 and 5 are equal at second and options 2 and 3 are equal at third (last).  So this type of ballot is easily done with numbers if numerical entry is used.  With the graphical interface originally suggested you could potentially drag option 4 beside option 5.  You would want flags for the question type which might force ranking of all candidates or no equal rankings depending on the voting system.  Anyway my point would be to consider making the ballot flexible enough to accommodate common voting systems with these sorts of ranked ballots.

2.) To make the feature usable so voters were truly anonymous, you could have a random unique identifier created on the form that the person voting could then copy and send via email (or otherwise communicate) to some third party who would record the voter was valid off of their valid voter list.  Those legitimate UID's could then be passed along to the person counting the ballots to reject any ballots that were illegitimate before counting.  This would ensure privacy of the vote.

I have a small organization that has elections for committees from a relatively small pool of voters and this question type would make it easy to do with Forms.

datapumpernickel:
[...]

PR #3247, opened 2026-03-28
Feature/ranking question type

datapumpernickel:
Adds a new 'ranking' question type that allows respondents to drag-and-drop predefined options into their preferred order. Answers are stored as a JSON-encoded array of option IDs in a single Answer row, following the same pattern as the grid type.

It could potentially implement a more advanced ranking algorithm evaluation for the responses. 

Disclaimer: this code has been mostly written by genAI


datapumpernickel:
As proposed in #1278

Chartman123:
@datapumpernickel thanks for the PR 👍🏻 we're currently refactoring Forms to use Vue3. Could you please already base your branch on refactor/vue3 instead of main? we also switched to vue-draggable-plus in that branch instead of vuedraggable.

datapumpernickel:
Ok, I will check that out!

PR #3248, opened 2026-03-29
Feature/ranking question type vue3

datapumpernickel:
As proposed in #1278: Add Ranking question type

Refactored to Vue3, after closing: #3247

This adds a new question type that lets respondents rank options by dragging them into order.

Results are scored using Borda count — first place gets the most points, last place gets the least. The summary shows each option's total score and average rank with a visual bar, along with a short explanation of how scoring works.

For export, ranking questions get one column per option (like grid questions do), with the rank number as the value. Answers are stored as a JSON array of option IDs, which keeps things consistent with how grid answers are stored.

There's also an optional shuffle setting. 

Again disclaimer: heavy use of genAI, I did try to check the code to the best of my abilities and ran it on nextcloud-docker-dev to actually interact with it and export data.

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -76,6 +76,7 @@ class Constants {
 	public const ANSWER_TYPE_LONG = 'long';
 	public const ANSWER_TYPE_MULTIPLE = 'multiple';
 	public const ANSWER_TYPE_MULTIPLEUNIQUE = 'multiple_unique';
+	public const ANSWER_TYPE_RANKING = 'ranking';
 	public const ANSWER_TYPE_SHORT = 'short';
 	public const ANSWER_TYPE_TIME = 'time';
 
@@ -95,6 +96,7 @@ class Constants {
 		self::ANSWER_TYPE_LONG,
 		self::ANSWER_TYPE_MULTIPLE,
 		self::ANSWER_TYPE_MULTIPLEUNIQUE,
+		self::ANSWER_TYPE_RANKING,
 		self::ANSWER_TYPE_SHORT,
 		self::ANSWER_TYPE_TIME,
 	];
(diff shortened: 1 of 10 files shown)


Chartman123:
@datapumpernickel please don't merge the other branch into this one, use `rebase`:)

datapumpernickel:
In the current implementation, respondents drag options into their preferred order. Since there is always an order (the default one), there is no way to distinguish "the respondent actively chose this order" from "the respondent skipped the question."

Making the question required would be meaningless because a valid answer is always present.

To support a true "required" behavior, we would need one of:

1. A **two-box drag-and-drop** (unranked → ranked), as suggested in #1278. This would make it clear whether the respondent actively ranked items. However, it could be bulky and hard to use on mobile or with long option texts.
2. A **confirmation button** that the respondent clicks to confirm they agree with the current ordering before it counts as answered.

For now, ranking questions simply always submit the current order, defaulting to the original option order if unchanged.

datapumpernickel:
Ok, it is now implemented as a tap-and-drag interface, which allows to set a required-flag or leave it blank. 🎊 
Sorry, for the forth-and-back. 

See attached video for demonstration: https://github.com/user-attachments/assets/9745e93c-87db-49ae-990f-ed305827deee

Chartman123:
Looks good :)

One thing I noticed: in submit view you can't change the order by keyboard. And for consistency we should move the drag handle to the end of the options like in create view.

datapumpernickel:
Hah, good catch. I tried to be more consistent with the design now. Keyboard should work now, but I do find the whole keyboard use a little unreliable still. I think I might be out of my depth for that.

Chartman123:
@datapumpernickel the underlying vue3 is now merged into main... You can rebase onto main again and then reopen this PR :) sorry for the inconvenience

Issue #1418, opened 2022-12-02
email validation for anonymous respondents


KiaraGrouwstra:
i am having my form respondents fill out an email field, so we can contact them after.
however, i would also like to feel confident that when a submission mentions a given email address, that the respondent actually owns that address.

to this end, i would like to be able to have forms offer an email field that allows validating the email address upon submission.
once the user clicks the link in the received email, their submission would be considered validated.

BarBaar44:
Totally agree with this one! But would like to broaden this request a bit:

Add the option to validate input. For example:
- valid email (the initial request)
- Number range
- Valid date
- Valid country/state/city
- etc

Chartman123:
@BarBaar44 you can already add a validation to text fields... Other than that, this is offtopic here. So I will hide your comment and mine, too :)

Issue #525, opened 2020-07-28
Feature request: Confirmation Mail


derhagen:
**Describe the solution you'd like**
If a user fills out a form, they often expect to receive a confirmation mail. E.g. after filling out a job application form, you often receive an automated mail saying "Thanks for your application, here is an overview of the data you submitted: ....", possibly including extra information about the further process. It would be nice to be able to configure such an automated mail in Nextcloud Forms. Building blocks for the mail body would be:
* Overview of submitted data
* Custom text
* Access to form field data in the custom text, e.g. "Hello {name}, thank you for contacting ..."

This requires including a special Email-field in the form for responding (depends on #177) and maybe an optional checkbox "I want to receive a confirmation mail"

**Possible implementation variants**

**a) Standalone**
Define a mail body template and subject per each form, select one of the mail accounts configured in "Mail" as sender (alternatively provide SMTP data) per each form

**b) Using Nextcloud Flow**
I didn't use Nextcloud Flow before, but if it is what I think it is, I could imagine it brings all the necessary building blocks for this, if integrated into forms.
I could imagine defining a flow as follows: when new form answer arrives, use the message body defined in "template.md", replace "{name}" placeholder by value from field "{name}" and "{data}" placeholder by the data overview. Send email to the address from field {email}, using the account "whatever@example.org" from the "Mail" app.

**c) Using External Scripts**
Provide a REST-API/Webhooks/etc. so I can have an external 10 lines Python script to take care of it. Maybe this is already possible? (waiting for #451 to answer this)

skjnldsv:
Hey! Thanks for your suggestion!
That is interesting! @jancborchardt what do you think?

jancborchardt:
Sure, but let’s keep it dead-simple:

At the end of every form, just before the submit button, there could be a checkbox:

> - [ ] Get a copy of the responses via mail to: `[ mail@example.com ]`

And that’s all. It would be unchecked by default, and if it’s checked you can fill in your email address.
If the person filling it is a Nextcloud user, their email from the profile can be prefilled here.

(Nothing the admin or form creator has to decide about this since it’s a feature purely for user convenience. It will use the standard Nextcloud mail template and we’re not going into the rabbit hole of customization here. ;))

ateuber:
This would be a really useful feature for us.

loscil06:
> Sure, but let’s keep it dead-simple:
> 
> At the end of every form, just before the submit button, there could be a checkbox:
> 
[...]

PR #3252, opened 2026-03-31
feat(forms): add respondent confirmation emails with optional email verification

QuiteBitter:
## What this changes

- Adds respondent confirmation mails for short questions that use email validation and are explicitly marked as the confirmation recipient in `extraSettings`.
- Adds an optional email verification flow for that recipient question. When verification is enabled, submissions stay unverified until the recipient uses the mailed verification link.
- Sends the confirmation mail only after the submission is verified, so verified flows do not acknowledge unverified addresses.
- Adds the public verification route, token persistence, and unit/integration coverage for the full submit -> verify -> confirm path.

## Related issues

- #1418
- #525

## Upstream context

- #3098 already proposes confirmation mails. This branch intentionally follows the review direction there by using an explicit recipient flag on the question instead of auto-picking an email field, and then extends that flow with email verification.
- `forms` currently supports Nextcloud 31-33. `OCP\\Mail\\IEmailValidator` is only available from Nextcloud 32 onward, so this implementation keeps `IMailer::validateMailAddress()` for 31 compatibility.

## Scope / non-goals

- Uses the configured system mail account and a fixed branded template.
- Does not add per-form sender selection or custom subject/body templating yet.
- Leaves generated locale catalogs out of the PR because Nextcloud translations are handled through Transifex.

## Validation

- Nextcloud 33: unit and integration suites passed.
- Nextcloud 31: unit and integration suites passed.

### DIFF
--- a/docs/API_v3.md
+++ b/docs/API_v3.md
@@ -439,6 +439,7 @@ Update a single or multiple properties of a question-object.
 - Restrictions:
     - It is **not allowed** to update one of the following key-value pairs: _id, formId, order_.
     - `extraSettings.confirmationRecipient` can only be enabled for short questions with `extraSettings.validationType` set to `email`.
+    - `extraSettings.requireEmailVerification` can only be enabled for short questions with `extraSettings.validationType` set to `email` and `extraSettings.confirmationRecipient` set to `true`.
 - Response: **Status-Code OK**, as well as the id of the updated question.
 
 ```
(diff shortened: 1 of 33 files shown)


Chartman123:
Thanks for your PR! Please understand that we're currently migrating to Vue3 and won't merge new features until this has finished to minimize extra work.

Other than that: #525 is already addressed in #3098

QuiteBitter:
> Thanks for your PR! Please understand that we're currently migrating to Vue3 and won't merge new features until this has finished to minimize extra work.
> 
> 
> 
> Other than that: #525 is already addressed in #3098

Okey, I will wait for this to complete :) good luck! After completion, I'll check what's needed to be done on my side.

Chartman123:
@QuiteBitter the other PR is now merged

QuiteBitter:
Ty for the Note

PR #3262, opened 2026-04-01
feat: ranking question type

### LINKED ISSUE #1278: Voters can rank the options given in order of importance

datapumpernickel:
As proposed in #1278: Add Ranking question type

Refactored to Vue3, after closing: #3247

Rebased to main after closing: #3248

This adds a new question type that lets respondents rank options by dragging them into order.

Results are scored using Borda count — first place gets the most points, last place gets the least. The summary shows each option's total score and average rank with a visual bar, along with a short explanation of how scoring works.

For export, ranking questions get one column per option (like grid questions do), with the rank number as the value. Answers are stored as a JSON array of option IDs, which keeps things consistent with how grid answers are stored.

There's also an optional shuffle setting.

Again disclaimer: heavy use of genAI, I did try to check the code to the best of my abilities and ran it on nextcloud-docker-dev to actually interact with it and export data.

Ok, it is now implemented as a tap-and-drag interface, which allows to set a required-flag or leave it blank. 🎊
Sorry, for the forth-and-back.

See attached video for demonstration: https://github.com/user-attachments/assets/9745e93c-87db-49ae-990f-ed305827deee

### DIFF
--- a/src/components/Questions/QuestionRanking.vue
+++ b/src/components/Questions/QuestionRanking.vue
@@ -10,7 +10,6 @@
 		:warningInvalid="answerType.warningInvalid"
 		:contentValid="contentValid"
 		:shiftDragHandle="shiftDragHandle"
-		:allowRequired="false"
 		v-on="commonListeners">
 		<template #actions>
 			<NcActionCheckbox
@@ -26,13 +25,38 @@
 			</NcActionButton>
 		</template>
 
-		<!-- Submit mode: drag to rank -->
+		<!-- Submit mode -->
 		<div
 			v-if="readOnly"
 			class="question__content"
 			role="list"
 			:aria-labelledby="titleId"
 			:aria-describedby="description ? descriptionId : undefined">
+			<!-- Unranked pool (visible when items remain) -->
+			<div v-if="unrankedOptions.length > 0" class="ranking-unranked">
+				<p class="ranking-unranked__label">
+					{{ t('forms', 'Tap to rank') }}
+				</p>
+				<button
+					v-for="option in unrankedOptions"
+					:key="option.id"
+					class="ranking-unranked__item"
+					@click="rankOption(option)">
+					{{ option.text }}
+				</button>
+			</div>
+
+			<!-- Empty state when nothing ranked yet -->
+			<p v-if="rankedOptions.length === 0" class="ranking-empty">
+				{{ t('forms', 'Tap items above to rank them') }}
+			</p>
+
+			<!-- Ranked list header (to separate from pool) -->
+			<p v-if="rankedOptions.length > 0" class="ranking-ranked__label">
+				{{ t('forms', 'Your ranking') }}
+			</p>
+
+			<!-- Ranked list -->
 			<Draggable
 				v-model="rankedOptions"
 				:animation="200"
@@ -49,6 +73,14 @@
 						<IconDragHorizontalVariant :size="20" />
 					</span>
 					<span class="ranking-item__text">{{ option.text }}</span>
+					<NcButton
+						variant="tertiary"
+						:ariaLabel="t('forms', 'Remove from ranking')"
+						@click="unrankOption(option)">
+						<template #icon>
+							<IconClose :size="20" />
+						</template>
+					</NcButton>
 				</div>
 			</Draggable>
 		</div>
@@ -107,7 +139,9 @@
 import { VueDraggable as Draggable } from 'vue-draggable-plus'
 import NcActionButton from '@nextcloud/vue/components/NcActionButton'
 import NcActionCheckbox from '@nextcloud/vue/components/NcActionCheckbox'
+import NcButton from '@nextcloud/vue/components/NcButton'
 import NcLoadingIcon from '@nextcloud/vue/components/NcLoadingIcon'
+import IconClose from 'vue-material-design-icons/Close.vue'
 import IconContentPaste from 'vue-material-design-icons/ContentPaste.vue'
 import IconDragHorizontalVariant from 'vue-material-design-icons/DragHorizontalVariant.vue'
 import OptionInputDialog from '../OptionInputDialog.vue'
@@ -123,10 +157,12 @@ export default {
 	components: {
 		AnswerInput,
 		Draggable,
+		IconClose,
 		IconContentPaste,
 		IconDragHorizontalVariant,
 		NcActionButton,
 		NcActionCheckbox,
+		NcButton,
 		NcLoadingIcon,
 		OptionInputDialog,
 		Question,
@@ -141,6 +177,7 @@ export default {
 			isLoading: false,
 			isOptionDialogShown: false,
 			rankedOptions: [],
+			unrankedOptions: [],
 			OptionType,
 		}
 	},
@@ -176,7 +213,7 @@ export default {
 
 	methods: {
 		/**
-		 * Initialize ranked options from existing values or default order
+		 * Initialize ranked/unranked options from existing values or default order
 		 */
 		initRankedOptions() {
 			const sorted = this.sortOptionsOfType(this.options, OptionType.Choice)
@@ -187,28 +224,64 @@ export default {
 				this.rankedOptions = this.values
 					.map((id) => byId[parseInt(id)])
 					.filter(Boolean)
+				this.unrankedOptions = sorted.filter(
+					(o) => !this.rankedOptions.some((r) => r.id === o.id),
+				)
+			} else if (this.readOnly) {
+				// Submit mode: start with all options unranked
+				this.rankedOptions = []
+				this.unrankedOptions = [...sorted]
 			} else {
+				// Edit mode: show all options in default order
 				this.rankedOptions = [...sorted]
+				this.unrankedOptions = []
 			}
+		},
 
-			// In submit mode, emit the initial ranking so the answer is always
-			// recorded – even when the user agrees with the default order.
-			if (this.readOnly && this.rankedOptions.length > 0) {
+		/**
+		 * Move an option from the unranked pool to the ranked list
+		 *
+		 * @param {object} option The option to rank
+		 */
+		rankOption(option) {
+			this.unrankedOptions = this.unrankedOptions.filter(
+				(o) => o.id !== option.id,
+			)
+			this.rankedOptions.push(option)
+			this.emitValues()
+		},
+
+		/**
+		 * Move an option from the ranked list back to the unranked pool
+		 *
+		 * @param {object} option The option to unrank
+		 */
+		unrankOption(option) {
+			this.rankedOptions = this.rankedOptions.filter((o) => o.id !== option.id)
+			this.unrankedOptions.push(option)
+			this.emitValues()
+		},
+
+		/**
+		 * Emit the current ranking after a drag reorder
+		 */
+		onRankingEnd() {
+			this.emitValues()
+		},
+
+		/**
+		 * Emit the current values based on ranking state
+		 */
+		emitValues() {
+			if (this.rankedOptions.length === 0) {
+				// Nothing ranked — emit empty to signal unanswered
+				this.$emit('update:values', [])
+			} else {
 				this.$emit(
 					'update:values',
 					this.rankedOptions.map((o) => o.id),
 				)
 			}
-		},
-
-		/**
-		 * Emit the new ranking after a drag ends
-		 */
-		onRankingEnd() {
-			this.$emit(
-				'update:values',
-				this.rankedOptions.map((o) => o.id),
-			)
 		},
 	},
 }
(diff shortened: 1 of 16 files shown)


src/models/AnswerTypes.js
```diff
@@ -12,6 +12,7 @@ import IconFile from 'vue-material-design-icons/FileOutline.vue'
 import IconGrid from 'vue-material-design-icons/Grid.vue'
 import IconNumeric from 'vue-material-design-icons/Numeric.vue'
 import IconRadioboxMarked from 'vue-material-design-icons/RadioboxMarked.vue'
+import IconReorderHorizontal from 'vue-material-design-icons/ReorderHorizontal.vue'
```
Chartman123: Please use this icon, as Microsoft Forms also uses this.
```suggestion
import IconSwapVertical from 'vue-material-design-icons/SwapVertical.vue'
```

Chartman123 (commented):
Did a first review of your code and added some comments/suggestions.

Regarding the design:
- I think the current represantation of the unranked/ranked options is too massive.
- Please try to use CSS variable defined in core for pixel values
- I don't like that the page "jumps" as soon as all unranked options are selected (same when one of them is removed again from the ranked list
- Now that I see it, I'm no longer sure if we should really move the drag handle to the end.

For a more profound design review, cc @nextcloud/designers

datapumpernickel:
Thank you for taking the time to do a first review - appreciated! 

I incorporated your comments and also re-wrote the layout to a two column layout. This seemed the logical solution to the "jumping" page situation. I was worried that on mobile this might look horrible, but see attached video for an okish-interaction I would say. 

https://github.com/user-attachments/assets/76c3c223-c9ba-4a7c-8f96-3f0b7e0e15d2

Chartman123:
I liked the single column layout. But perhaps we should just keep the area for unranked options visible and just display a text that says that no more options are available/all options are ranked

datapumpernickel:
I liked single column also a bit more, but even when you leave the area visible, the area will change size noticeably, when options go from one row to two rows. So unless we leave the area for unranked options very large, which in edge cases might prevent filling out the form, I think two column layout is better for the non-jumpy behaviour. But I am really not passionate about either, just did not have a better idea for evading jumping behaviour (which I agree is annoying).

Chartman123:
True, didn't consider two (or potentially more) rows. So let's revert to the old behavior :) and see what the designers will say about the rest.

datapumpernickel:
Ok, reverted to single column layout. See attached video for two row behaviour. 

https://github.com/user-attachments/assets/aad70895-213f-40d2-9a22-c7e121adffba

datapumpernickel:
Ok, not sure how I fucked up all the tests. will fix sometime this weekend :)

nimishavijay:
Super interesting! I also like the one column layout better and generally it looks really good! I do have some questions:
1. Is there a way to rank without using the mouse? That would be important for accessibility (keyboard users). 
2. Just thinking out loud here: Explicitly mentioning how to do an interaction (like "Tap to rank") is generally avoided, but removing that wording indication may leave users confused. Could we allow dragging into the ranked section as well?
3. Follow up to previous point: I'm also wondering if we can have placeholder boxes in the ranked section if empty, so people can drag their option to the place they would like. Eg: For `Rank fruits in order of preference: mango, banana, strawberry, apple and oranges` my thinking may be "I hate oranges so that's definitely last, and I love mangoes so that's definitely first". And it seems like by this tap to rank interaction it may be difficult to put directly our lowest preference. What are your thoughts about this?

datapumpernickel:
Hi @nimishavijay,

sorry for the long delay in answering and thanks for taking the time to look at this! 

1) It currently implements sorting via keyboard the same way other next cloud questions do. So it should be possible, yes. (see attached screen recording)
2) Dragging also works now. I would then remove the "tap to rank" hint. (see attached screen recording)

3) With this I am not sure. I totally understand the logic behind this. Intuitively I would also approach it this way, but my use-case at my org already implements some pretty long options, which would mean, displaying all the options, plus all the placeholders would potentially move the placeholder off the screen, so I could not even put it in last place without scrolling while dragging, which is a bit awkward always. I know, this is not great form design and possibly an edge case, but on mobile we would see this quite quickly, right? 

https://github.com/user-attachments/assets/ac5978d8-d540-4b0e-9d2d-0cf10d4fd404

https://github.com/user-attachments/assets/6328a3ed-e4b5-4db6-8d16-1f16f379b757

Chartman123:
> 3\. With this I am not sure. I totally understand the logic behind this. Intuitively I would also approach it this way, but my use-case at my org already implements some pretty long options, which would mean, displaying all the options, plus all the placeholders would potentially move the placeholder off the screen, so I could not even put it in last place without scrolling while dragging, which is a bit awkward always. I know, this is not great form design and possibly an edge case, but on mobile we would see this quite quickly, right?

I second that. For only a few options this might still be feasible, but for longer lists it will add lots of space. Especially for non-mandatory questions.

Chartman123:
@datapumpernickel now that we moved the drag handle back to the front and given that our current implementation for options was used in the backend where we also already had a drag handle for re-ordering questions, we could improve the design further:

- move the drag handle all the way to the block start of the ranked options and before the rank number

datapumpernickel:
I put the drag handle all the way to the block start, but tbh I liked the ranking up front a little better. it somehow looks less tidy now. But thats more of an uninformed feeling, less of an actual opinion. If you prefer this, I can totally make the change and commit it! 

Screenshot 2026-04-16 at 00 12 29

Chartman123:
@datapumpernickel I like the handle at the front better. The order of the ranked items is about the options and not the handles ;) 

Could you please also adjust AnswerInput.vue so that it doesn't show the radio button in edit mode? I'm unsure if we should just leave it empty or also just use the handle icon in this case. We probably need some icon so that the existing options and the new option text are correctly aligned.

I also rebased your branch on current main :) You'll have to `git pull --rebase` before you continue :)

datapumpernickel:
Got it. Moved the handlebar to the front and implemented the drag-icon in edit mode. However note, that I had to change the default size for the indicators in `AnswerInput.vue` because the pseudoIcon IconDragIndicator has a different size and made everything misalign. 

In more technical terms: `pseudoIcon` in `AnswerInput.vue` renders icons dynamically via `<component :is="pseudoIcon">` without passing a `:size` prop. The vue-material-design-icons (`RadioboxBlank`, `CheckboxBlankOutline`, `Plus`) all default to `size: 24`, while `IconDragIndicator` defaults to `size: 20`. Adding an explicit `:size="24"` ensures all pseudo icons render at a consistent size regardless of their individual defaults, preventing the layout shift for the ranking question type.

datapumpernickel:
Just wanted to check-in, whether there is anything needed from my side to move this forward? 

Please dont misunderstand as me wanting to rush this, just making sure its not hanging on my side. I understand the limits of open source development and time!

Chartman123:
> anything needed from my side to move this forward?

In general it looks very good now. I just wondered if it would look better if we have some kind of hover effect for the already ranked options. Could you please try how it looks if you add the styling that is used for `tertiary` NcButtons?

datapumpernickel:
Done! Hover is a great improvement actually! 
Screenshot 2026-04-23 at 22 03 25

Chartman123:
@datapumpernickel I now had some more time to have a look at it and found some issues:

- [x] restoring unsubmitted values from local storage doesn't work
- [x] clear form button doesn't work for ranking questions
- [x] validate function is missing (see QuestionMultiple)

I also rebased your branch onto current main. Please `git pull --rebase` before you continue :)

jancborchardt (approved):
Nice, looks good design-wise :)

datapumpernickel:
Hi @Chartman123,

thank you for taking the time to review! That sounds like quite some bugs still... 😬 Sorry about that!  I will look at it sometime this week and push again when I fixed them. Not sure I understand the first issue yet, but I will circle back with questions if it stays that way once I looked at the actual code for that!

Chartman123:
I think 1 and 2 are connected and probably easy to fix. 

When fill out a form, but don't submit it before you leave the page, the current values will be stored in your browser in the so called local storage. The values should then be restored upon your next visit on that form. I confirmed that storing these values works as expected for the new question type. But restoring does not, you have to start from scratch to rank the options.

datapumpernickel:
Hi @Chartman123 

thanks for the hints. I think it is implemented. At least in my tests it looked good. Also tried adding into the playwright test suite, which now worked on my computer. Getting the hang of rebase too. ;-) 

If it is required, all options must be ranked. 

edit: the error validation wording was a bit misleading with "you must rank at least one option", which is never really true. You must always rank all options, otherwise results would require other metrics (which could be implemented in future features)

https://github.com/user-attachments/assets/6e2bcdbc-e09a-4619-9ce9-927dce89ca37

Chartman123:
@datapumpernickel great :) I think it's ready to merge. I requested a review from two other Forms devs, too.

In the meantime you could squash all your commits into a single one (`git rebase -i main` and then prefix all commits after the first one with `f` instead of `pick`). 

ah, and please add the new question type to the docs/DataStructure.md

Chartman123 (approved):

src/components/Questions/QuestionRanking.vue
```diff
@@ -0,0 +1,537 @@
...
+			:aria-labelledby="titleId"
+			:aria-describedby="description ? descriptionId : undefined">
+			<!-- Unranked pool -->
+			<div class="ranking-unranked">
+				<Draggable
+					v-show="unrankedOptions.length > 0"
+					v-model="unrankedOptions"
+					class="ranking-unranked__pool"
+					:animation="200"
+					:group="{ name: 'ranking', pull: true, put: true }"
```
Chartman123: I found a bug: If you have multiple ranking questions, you could move options from one question to another. I think we need to have unique group names here. My suggestion adds the question id to the group name:
```suggestion
					:group="'ranking_' + id"
```

src/components/Questions/QuestionRanking.vue
```diff
@@ -0,0 +1,537 @@
...
+			<!-- Ranked list -->
+			<div class="ranking-ranked">
+				<p class="ranking-section__label">
+					{{ t('forms', 'Your ranking') }}
+				</p>
+				<Draggable
+					v-model="rankedOptions"
+					class="ranking-ranked__list"
+					:animation="200"
+					:group="{ name: 'ranking', pull: true, put: true }"
```
Chartman123: Same here for the ranked options
```suggestion
					:group="'ranking_' + id"
```

src/components/Questions/QuestionRanking.vue
```diff
@@ -0,0 +1,537 @@
...
+					<button
+						v-for="option in unrankedOptions"
+						:key="option.id"
+						class="ranking-unranked__item"
+						@click="rankOption(option)">
+						{{ option.text }}
+					</button>
+				</Draggable>
+				<p
+					v-show="unrankedOptions.length === 0"
```
Chartman123: Please use v-if instead of v-show, otherwise the text will always be rendered to the DOM.

```suggestion
					v-if="unrankedOptions.length === 0"
```

src/components/Questions/QuestionRanking.vue
```diff
@@ -0,0 +1,537 @@
...
+							variant="tertiary"
+							:ariaLabel="t('forms', 'Remove from ranking')"
+							@click="unrankOption(option)">
+							<template #icon>
+								<IconClose :size="20" />
+							</template>
+						</NcButton>
+					</div>
+				</Draggable>
+				<p v-show="rankedOptions.length === 0" class="ranking-ranked__empty">
```
Chartman123:
```suggestion
				<p v-if="rankedOptions.length === 0" class="ranking-ranked__empty">
```

Chartman123 (changes requested):

src/components/Questions/QuestionRanking.vue
datapumpernickel: Should I just accept this change or implement it in the local repo?

Chartman123: You can do both :) If you do it locally you will get fewer new commits... Please also have a look at the other v-shows too and replace them with v-if/v-else :)

src/components/Questions/QuestionRanking.vue
```diff
@@ -0,0 +1,537 @@
...
+					<button
+						v-for="option in unrankedOptions"
+						:key="option.id"
+						class="ranking-unranked__item"
+						@click="rankOption(option)">
+						{{ option.text }}
+					</button>
+				</Draggable>
+				<p
+					v-if="unrankedOptions.length === 0"
```
Chartman123:
```suggestion
					v-if="unrankedOptions.length > 0"
					v-model="unrankedOptions"
					class="ranking-unranked__pool"
					:animation="200"
					:group="'ranking_' + id"
					@end="emitValues">
					<button
						v-for="option in unrankedOptions"
						:key="option.id"
						class="ranking-unranked__item"
						@click="rankOption(option)">
						{{ option.text }}
					</button>
				</Draggable>
				<p
					v-else
```

src/components/Questions/QuestionRanking.vue
```diff
@@ -0,0 +1,537 @@
...
+
+			if (this.isRequired && this.rankedOptions.length === 0) {
+				this.errorMessage = t('forms', 'You must rank all options')
+				return false
+			}
+
+			if (
+				this.rankedOptions.length > 0
+				&& this.rankedOptions.length !== optionsCount
+			) {
```
Chartman123:
```suggestion
			if (
				(this.isRequired && this.rankedOptions.length === 0)
				|| (this.rankedOptions.length > 0
					&& this.rankedOptions.length !== optionsCount)
			) {
```

Chartman123 (commented):
We're getting really close :) Two small comments on further improving the code of the new question component.

And another bug: restoring the ranked options from an already submitted response does not work. The options are loaded unranked.

datapumpernickel:
Thank you for the patient reviews and checking of all these bugs! I had not considered that. I tried now reloading both submitted and unsubmitted answers and it worked. :)

Chartman123 (approved):
Great :) no more stoppers from my side.

Chartman123:
Resolved the merge conflicts and changed the icons to Material Symbols as the other PR got merged earlier

sharko789:
~~This is great! Not sure if this is the right place to ask this, but are there options (e.g. maximum options to be ranked) for the question type? If so, could you send a screenshot @datapumpernickel?
(unfortunately I am unable to install the alpha version on my NC33.0.3 server, otherwise I would test this myself)~~

I still can't get it to install even on a NC34 docker container I spun up for testing, but I found the time to look at the code and opened a new issue for this (mentioned below).

PR #3289, opened 2026-04-14
feat: Add confirmation email for form respondents

### LINKED ISSUE #525: Feature request: Confirmation Mail

dtretyakov:
Closes #525 - Send confirmation emails to form respondents after submission.

Features:
- Add confirmation email settings (enabled, subject, body) to Form entity.
- Implement email sending with placeholder replacement:
  - {formTitle}, {formDescription}
  - Field placeholders based on name or label (e.g. {name}, {email}).
- Allow form creators to explicitly select the recipient email field when multiple email fields are present.
- Add UI in Settings sidebar to configure confirmation emails and recipient selection.
- Replace basic email validation with Nextcloud's internal IEmailValidator.
- Integration with activity notifications and background jobs for file syncing.

Technical changes:
- Database migration for new Form properties.
- Enhanced FormsService with email sending logic and validation.
- Extensive unit and integration tests covering the new functionality.
- Updated API documentation and OpenAPI spec.


dtretyakov:
@Chartman123 I also tried to streamline UX for the feature.

Now the form looks like that:

Снимок экрана 2026-04-14 в 09 04 36

Chartman123:
@dtretyakov looks good :)

I'll ping the design specialists to see what they think about it :)

dtretyakov:
`OCP\Mail\IEmailValidator` breaks backward compatibility for this PR. The interface is not available in nextcloud/ocp on `stable31`.

For app code here, `IMailer::validateMailAddress()` is the safer choice: it is already available across the supported server branches and covers the same validation use case without introducing a branch-specific compatibility break.

Chartman123:
No problem, just ignore this. For the next release we will no longer support NC31

dtretyakov:
@Chartman123 the branch is rebased on the fresh main and uses `OCP\Mail\IEmailValidator`.
Please let me know if you're missing something to push it forward and make NC customers happier.

lib/Migration/Version050301Date20260413233000.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
Chartman123:
```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

lib/Migration/Version050301Date20260413233000.php
```diff
@@ -0,0 +1,65 @@
...
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Add confirmation email fields to forms
+ */
+class Version050301Date20260413233000 extends SimpleMigrationStep {
```
Chartman123:
```suggestion
class Version050300Date20260413233000 extends SimpleMigrationStep {
```

lib/Constants.php
```diff
@@ -147,6 +147,7 @@ class Constants {
 	];
 
 	public const EXTRA_SETTINGS_SHORT = [
+		'confirmationEmailRecipient' => ['boolean'],
```
Chartman123: Not sure, if we should add the extraSetting to the whole short question type or if we'd better have it only for `*_SHORT_EMAIL`. Otherwise it could also be set for e.g. number fields.

@susnux @pringelmann what do you think?

lib/ResponseDefinitions.php
```diff
@@ -141,8 +142,11 @@
  *   shares: list<FormsShare>,
  *   submissionCount?: int,
  *   submissionMessage: ?string,
+ *   confirmationEmailEnabled: bool,
+ *   confirmationEmailSubject: ?string,
+ *   confirmationEmailBody: ?string,
+ *   confirmationEmailRecipient: ?int,
  * }
```
Chartman123, line 148:
```suggestion
 * }
 *
```

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -194,6 +194,132 @@
...
+			<p class="confirmation-email__hint">
+				{{ t('forms', 'Set up confirmation emails in three steps:') }}
+			</p>
+			<ol class="confirmation-email__steps">
+				<li>{{ t('forms', 'Add an email field to the form.') }}</li>
+				<li>
+					{{
+						t(
+							'forms',
+							'Select which email field receives the confirmation email.',
```
Chartman123: The wording is kind of strange here... the field doesn't receive the confirmation mail, but the address entered in that field.

So perhaps better something like:

```suggestion
							'Select which email field is used for confirmation emails.',
```

But I'm not very happy with my text either...

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -194,6 +194,132 @@
...
+				" />
+			<NcNoteCard
+				v-else-if="requiresConfirmationEmailRecipientSelection"
+				type="error"
+				:text="
+					t(
+						'forms',
+						'Select which email field should receive confirmation emails before finishing this setup.',
+					)
+				" />
```
Chartman123: Perhaps you could only use one NcNoteCard for all errors and use a single computed prop that returns the error text based on the error?

src/components/SidebarTabs/SettingsSidebarTab.vue
```diff
@@ -194,6 +194,132 @@
...
+						@change="onConfirmationEmailRecipientSelectionChange">
+						<option value="">
+							{{ t('forms', 'Select an email field') }}
+						</option>
+						<option
+							v-for="question in confirmationEmailQuestions"
+							:key="question.id"
+							:value="question.id">
+							{{ confirmationEmailQuestionLabel(question) }}
+						</option>
```
Chartman123: Why not use NcSelect here? this would already bring the correct styling. Same for the other input fields

Chartman123 (commented):
I did a quick review and found some parts that should be changed in my opinion.

Please also add the new fields to FormsMigrator.php.

Also not sure about the frontend part, especially in the sidebar as I'm no vue expert. I added @susnux and @pringelmann as reviewers :)

Btw no need to hurry, the next release is not yet directly around the corner. We'll get it merged once it's ready :)

lib/Migration/Version050300Date20260413233000.php
Chartman123, line 1: Please also change the file name

lib/Service/FormsService.php
```diff
@@ -1003,6 +1003,15 @@ public function areExtraSettingsValid(array $extraSettings, string $questionType
...
 
+		// confirmationEmailRecipient is only valid for short questions with email validation.
+		if (
+			$questionType === Constants::ANSWER_TYPE_SHORT
+			&& !empty($extraSettings['confirmationEmailRecipient'])
+			&& ($extraSettings['validationType'] ?? null) !== 'email'
+		) {
+			return false;
+		}
+
```
Chartman123: Please move this block down into the ANSWER_TYPE_SHORT block

lib/Constants.php
dtretyakov: We could also consider an alternative option.

From the domain perspective the form now has the following settings:
* `replyEmail`
* `subject`
* `body`

In the `body` we could reference form fields.

What if the same logic will be applied to the `replyEmail`?

In this case:
* we don't need to use extraSettings at all
* we'll be storing email field id in the replyEmail form field

Reference: https://github.com/nextcloud/forms/pull/3289#issuecomment-4241929568

pringelmann: I like this. It's more uniform than the FK approach and removes the special-case `extraSetting` entirely. Three template strings, one resolver, done.

One thing to think through: validation gets a bit fuzzier. With an FK we can verify at save time that the referenced question exists and is email-validated. With a placeholder string we either resolve lazily at send time (silent failure if the field was renamed/deleted) or eagerly validate the template against the current question set (which then needs to re-validate on every question edit).

I'd lean toward eager validation in the API layer plus a clear UI affordance in the sidebar - e.g. a dropdown that writes the placeholder for the user rather than letting them type `{whatever}` freehand. That keeps the "free-form template" flexibility for power users while making the common path foolproof.

Worth doing IMO.

pringelmann:
One thing I want to flag before this lands: as written, this turns any public form into an unauthenticated outbound mailer. A submitter controls the recipient address (it's just whatever they typed into the email field), so an attacker can script submissions with a victim's address and use the instance to mailbox-bomb them. Worst case it also burns the operator's MX reputation.

A few options, not mutually exclusive:
- Rate limit per recipient address and/or per form.
- Restrict the feature to authenticated submitters only (or at least gate it behind a warning when the form is public).
- Move sending into a queued job so the queue can be paced.

I'd be more comfortable with at least one of these in place before we ship. Could also be worth a line in the PR description spelling out which mitigations apply, since this is a new outbound channel.

lib/Service/FormsService.php
```diff
@@ -739,6 +745,170 @@ public function notifyNewSubmission(Form $form, Submission $submission): void {
 		}
 
 		$this->eventDispatcher->dispatchTyped(new FormSubmittedEvent($form, $submission));
+
+		// Send confirmation email if enabled
+		$this->sendConfirmationEmail($form, $submission);
```
pringelmann: This runs inline here, so any latency lands directly on the user's submission response. A slow MX, a TLS handshake hiccup, or a transient timeout all become user-visible delays. The `try/catch` swallows errors but not the time spent. Can lead to poor UX.

Perhaps worth moving into a queued job carrying `formId` + `submissionId` and resolving the rest there? Side benefit: it makes the throttling issue for mailbox spamming essentially free, since the job queue is already the right place to pace outbound mail.

Chartman123:
> One thing I want to flag before this lands: as written, this turns any public form into an unauthenticated outbound mailer.

@pringelmann Perhaps we should also add an admin setting that can turn the feature on/off and that should be off by default?

lib/Migration/Version050300Date20260413233000.php
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+		$table = $schema->getTable('forms_v2_forms');
+
+		if (!$table->hasColumn('confirmation_email_enabled')) {
+			$table->addColumn('confirmation_email_enabled', Types::BOOLEAN, [
+				'notnull' => false,
+				'default' => 0,
```
pringelmann, line 38: This means existing rows on upgrade might end up as `NULL` rather than `false`. It happens to look right because of the casting, but the column doesn't actually reflect the domain - it's a flag, not tri-state. I'd just make it `notnull => true, default => false` then we can drop all the bool casting.

pringelmann:
> > One thing I want to flag before this lands: as written, this turns any public form into an unauthenticated outbound mailer.
> 
> @pringelmann Perhaps we should also add an admin setting that can turn the feature on/off and that should be off by default?

Yeah good idea, default-off admin toggle is a sensible safety net and I'd vote for adding it.

That said, I don't think it replaces a per-form/per-recipient mitigation. Once an admin flips it on (which many likely will, because it is a very useful feature for users), the abuse surface is fully open again on every public form on the instance. So it protects the instances that don't need the feature but does nothing for the ones that do.

I'd still suggest at least one of: rate limit per recipient, gate it behind authentication, or pace via a queued job. The admin toggle then becomes a global kill switch on top of that, which is genuinely useful (e.g. an admin can disable the feature instance-wide if they spot abuse in their mail logs)

lib/Service/FormsService.php
Chartman123: > Perhaps worth moving into a queued job carrying `formId` + `submissionId` and resolving the rest there? Side benefit: it makes the throttling issue for mailbox spamming essentially free, since the job queue is already the right place to pace outbound mail.

Good idea 👍🏻 a background job sounds better

lib/Migration/Version050300Date20260413233000.php
Chartman123: We have `default => 0` everywhere for booleans. I think this is because one of the DB systems (sqlite?)...

pringelmann: Ah ok, good to know thnx :)

lib/Constants.php
dtretyakov: I implemented that change in the last commit, please check whether it look OK.

dtretyakov:
To mitigate the risks associated with automated emails now we have implemented a 4-layer defense strategy:

   1. **Administrative Control**: A global "kill switch" (`allowConfirmationEmail`) is available in the admin settings (defaulting to off), allowing instance operators to opt-in or disable the feature entirely.
   2. **Burst Protection**: We utilize Nextcloud's built-in `AnonRateLimit` and `UserRateLimit` middleware on the submission endpoint to prevent rapid-fire scripting attacks from a single IP.
   3. **Recipient Safeguard**: A strict per-recipient rate limit (max 3 emails per form/recipient pair per 24 hours) is enforced using a distributed cache with a "silent skip" strategy. This prevents a malicious actor from using a form to target a specific email address repeatedly, even when using multiple source IPs.
   4. **Asynchronous Processing**: Email sending is offloaded to a `QueuedJob` (`SendConfirmationMailJob`). This decouples the SMTP transaction from the HTTP response, preventing server hangs due to mail server latency and ensuring natural pacing of outgoing traffic.
   
If you see this combination is excessive please let me know and we could relax it.

dtretyakov:
@Chartman123 or @pringelmann please let me know if it requires some extra changes.

docs/DataStructure.md
```diff
@@ -21,6 +21,10 @@ This document describes the Object-Structure, that is used within the Forms App
...
 | submissionMessage    | String                               | max. 2048 ch.                           | Optional custom message, with Markdown support, to be shown to users when the form is submitted (default is used if set to null) |
+| confirmationEmailEnabled | Boolean                          |                                         | If enabled, send a confirmation email to the respondent after submission                                                         |
+| confirmationEmailSubject | String                           | max. 255 ch.                            | Optional confirmation email subject template (supports placeholders)                                                              |
+| confirmationEmailBody    | String                           |                                         | Optional confirmation email body template (plain text, supports placeholders)                                                    |
+| confirmationEmailRecipient | Integer                          |                                         | The id of the question belonging to the form, that holds the email address of the respondent                                     |
```
come-nc:
```suggestion
| confirmationEmailQuestionId | Integer                          |                                         | The id of the question belonging to the form, that holds the email address of the respondent                                     |
```

I would make it clearer in the naming that it holds a question id. It helps when reading the code.

lib/Controller/ApiController.php
```diff
@@ -1353,6 +1381,8 @@ public function deleteAllSubmissions(int $formId): DataResponse {
 	 *
 	 * 201: empty response
 	 */
+	#[AnonRateLimit(limit: 3, period: 3600)]
+	#[UserRateLimit(limit: 10, period: 3600)]
```
come-nc: Is that 3/h and 10/h? I understand the limitation for email sending, but for form submission it’s a severe restriction.
I use forms that see a higher usage rate than this.

lib/Migration/Version050300Date20260413233000.php
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		if (!$table->hasColumn('confirmation_email_recipient')) {
+			$table->addColumn('confirmation_email_recipient', Types::INTEGER, [
+				'notnull' => false,
+				'default' => null,
+			]);
+		}
+
+		return $schema;
```
come-nc: You should only return $schema if changes are needed, and otherwise return `null`.

lib/Service/FormsService.php
```diff
@@ -739,6 +754,178 @@ public function notifyNewSubmission(Form $form, Submission $submission): void {
...
+
+			$fieldKey = !empty($questionName) ? $questionName : $questionText;
+			$fieldKey = strtolower(preg_replace('/[^a-zA-Z0-9]/', '', $fieldKey));
+
+			if (!empty($answerMap[$questionId])) {
+				$answerValue = implode('; ', $answerMap[$questionId]);
+				$replacements['{' . $fieldKey . '}'] = $answerValue;
+				if (!empty($questionName)) {
+					$replacements['{' . strtolower(preg_replace('/[^a-zA-Z0-9]/', '', $questionName)) . '}'] = $answerValue;
+				}
```
come-nc:
```suggestion
```
This is redundant

lib/Service/FormsService.php
```diff
@@ -739,6 +754,178 @@ public function notifyNewSubmission(Form $form, Submission $submission): void {
...
+		if ($recipientEmail === null || !$this->emailValidator->isValid($recipientEmail)) {
+			$this->logger->debug('No valid email address found in submission for confirmation email', [
+				'formId' => $form->getId(),
+				'submissionId' => $submission->getId(),
+			]);
+			return;
+		}
+
+		$cacheKey = 'email_rl_' . hash('sha256', $form->getId() . ':' . strtolower($recipientEmail));
+		$cache = $this->cacheFactory->createDistributed('forms_confirmation_email');
```
come-nc: The cache object should be created in the constructor and stored in a property I believe.

lib/Service/FormsService.php
```diff
@@ -739,6 +754,178 @@ public function notifyNewSubmission(Form $form, Submission $submission): void {
...
+		if ($recipientQuestionId === null) {
+			return null;
+		}
+
+		foreach ($questions as $questionData) {
+			if (($questionData['id'] ?? null) !== $recipientQuestionId) {
+				continue;
+			}
+
+			if ($this->isConfirmationEmailQuestion($questionData)) {
```
come-nc:
```suggestion
			if ($this->isEmailQuestion($questionData)) {
```

lib/Service/FormsService.php
```diff
@@ -739,6 +754,178 @@ public function notifyNewSubmission(Form $form, Submission $submission): void {
...
+			return null;
+		}
+
+		return null;
+	}
+
+	/**
+	 * @param FormsQuestion $question
+	 */
+	private function isConfirmationEmailQuestion(array $question): bool {
```
come-nc:
```suggestion
	private function isEmailQuestion(array $question): bool {
```

lib/Service/FormsService.php
```diff
@@ -874,7 +1061,11 @@ public function areExtraSettingsValid(array $extraSettings, string $questionType
 			}
 
 			// Special handling of short input for validation
-		} elseif ($questionType === Constants::ANSWER_TYPE_SHORT && isset($extraSettings['validationType'])) {
+		} elseif ($questionType === Constants::ANSWER_TYPE_SHORT) {
+			if (!isset($extraSettings['validationType'])) {
+				return true;
+			}
```
come-nc: What does this change do? It looks to me like it does not change anything.

lib/Controller/ApiController.php
Chartman123: Yes, this shouldn't be applied to the submission endpoint itself.

lib/Controller/ApiController.php
```diff
@@ -206,12 +206,15 @@ public function newForm(?int $fromId = null): DataResponse {
 			$formData['showExpiration'] = false;
 			$formData['expires'] = 0;
 			$formData['isAnonymous'] = false;
+			$formData['confirmationEmailQuestionId'] = null;
```
Chartman123: Just use `unset()` like for the other variables that shouldn't get copied from the old form.

lib/Controller/ApiController.php
```diff
@@ -220,6 +223,10 @@ public function newForm(?int $fromId = null): DataResponse {
 				$newQuestion = Question::fromParams($questionData);
 				$this->questionMapper->insert($newQuestion);
 
+				if ($oldConfirmationEmailQuestionId === $oldQuestion->getId()) {
```
Chartman123:
```suggestion
				if (isset($oldConfirmationEmailQuestionId) && $oldConfirmationEmailQuestionId === $oldQuestion->getId()) {
```

lib/Service/ConfigService.php
```diff
@@ -50,6 +50,10 @@ public function getRestrictCreation(): bool {
 		return json_decode($this->config->getAppValue($this->appName, Constants::CONFIG_KEY_RESTRICTCREATION, 'false'));
 	}
 
+	public function getAllowConfirmationEmail(): bool {
+		return json_decode($this->config->getAppValue($this->appName, Constants::CONFIG_KEY_ALLOWCONFIRMATIONEMAIL, 'false'));
+	}
```
Chartman123, line 57: I know that we're currently using the deprecated `$this->config->getAppValue()` here, but could you please already use the new IAppConfig interface?

see https://docs.nextcloud.com/server/stable/developer_manual/digging_deeper/config/appconfig.html

We'll change the other occurences later in a follow-up.

dtretyakov:
Thank you all for having the detailed review, I applied the fixes.
Also there are refactoring to make simplify the support of new feature:
* `SendConfirmationMailJob` now uses email template.
* `ConfirmationEmailService` now contains the new logic to process form send requests and triggers the Job.
* Cleanup in the `SettingsSidebarTab` to use NC components.

Снимок экрана — 2026-04-29 в 21 17 55

lib/Controller/ApiController.php
```diff
@@ -206,12 +208,15 @@ public function newForm(?int $fromId = null): DataResponse {
 			$formData['showExpiration'] = false;
 			$formData['expires'] = 0;
 			$formData['isAnonymous'] = false;
+			unset($formData['confirmationEmailQuestionId']);
```
Chartman123: Could you please move this line up to the other `unset`s?

src/FormsSettings.vue
Chartman123, line 1: please rebase your branch on current main, @susnux fixed a bug in the admin settings. You need to adjust your code accordingly :)

Chartman123:
Other question: should we add a check if sending mails is generally available for the server? In my dev instance, I don't have the mail settings configured, but could enable the confirmation mails...

dtretyakov:
@Chartman123 I rebased the branch and applied the changes. 

> Other question: should we add a check if sending mails is generally available for the server? In my dev instance, I don't have the mail settings configured, but could enable the confirmation mails...

Now there is a check whether Email is configured by introducing `isMailConfigured` config property.
It should prevent enabling and using this feature.

lib/Migration/Version050300Date20260413233000.php
Chartman123: But yes, `notnull => true` would make sense. Every row should have a value here.

Chartman123:
@dtretyakov from my side it looks good now, just the little suggestion for the migration step regarding `notnull`.

src/FormsSettings.vue
```diff
@@ -169,5 +236,10 @@ export default {
 	&__creation__multiselect {
 		width: 100%;
 	}
+
+	&__rate-limit {
+		margin-top: calc(var(--default-grid-baseline) * 3);
+		max-width: 300px;
+	}
```
Chartman123, line 242: This max width causes the helper text to be split into new lines within the word boundaries. Do we really need this css rule?

lib/Migration/Version050300Date20260413233000.php
dtretyakov: changed `'notnull' => true`

src/FormsSettings.vue
dtretyakov: Thanks for noticing, fixed

lib/Migration/Version050300Date20260413233000.php
Chartman123: why did you change it back?

pringelmann (approved):
nice work, thank you for your contribution!

Chartman123 (approved):
@dtretyakov before merging, could you please squash your commits?

one last question: why did you change notnull back to false?

dtretyakov:
> could you please squash your commits?

I did that.

> why did you change notnull back to false?

@Chartman123 In the commit where we made `notnull=true, default=0` [MySQL tests were failing with the following](https://github.com/nextcloud/forms/actions/runs/25263649671/job/74101022462?pr=3289):
```
Nextcloud was successfully installed
Column "oc_forms_v2_forms"."confirmation_email_enabled" is type Bool and also NotNull, so it can not store "false".
Error: Process completed with exit code 1.
```
Which is emitted by assert inside the migrator: https://github.com/nextcloud/server/issues/45576


As I see there are no usages of combination `Types::BOOLEAN, notnull=true, default=0|false` in the server migrations and similar problems were mitigated by making `notnull=false` (https://github.com/nextcloud/mail/pull/6515/changes), so I suspect that it's not supported in all DBs: 
https://github.com/search?q=repo%3Anextcloud%2Fserver+Types%3A%3ABOOLEAN&type=code

So it would be great to ask someone from the NC core team to get more context.

Chartman123:
Thanks for explaining :) Then one last CI check and I'll merge this 👍🏻 Thanks again for your contribution!

--
Edit:
The limitation comes from Oracle databases: https://github.com/nextcloud/server/issues/45576#issuecomment-2143982917

PR #3296, opened 2026-04-15
Security: Raw exception messages are returned to clients in throttling middleware

tomaioo:
## Summary

Security: Raw exception messages are returned to clients in throttling middleware

## Problem

**Severity**: `Medium` | **File**: `lib/Middleware/ThrottleFormAccessMiddleware.php:L21`

In `afterException`, the middleware returns `$exception->getMessage()` to the client for `NoSuchFormException`. If exception messages include internal identifiers, query context, or untrusted input, this can lead to information disclosure and potentially reflected content issues in consuming clients.

## Solution

Return a fixed, generic error message (e.g., 'Form not found') and log detailed exception context server-side only.

## Changes

- `lib/Middleware/ThrottleFormAccessMiddleware.php` (modified)

### DIFF
--- a/lib/Middleware/ThrottleFormAccessMiddleware.php
+++ b/lib/Middleware/ThrottleFormAccessMiddleware.php
@@ -25,7 +25,7 @@ class ThrottleFormAccessMiddleware extends Middleware {
 		}
 
 		$response = new DataResponse(
-			$exception->getMessage(),
+			'Form not found',
 			$exception->getCode(),
 		);
 		$response->throttle(['action' => 'form']);


pringelmann:
Hi @tomaioo, thanks for the report, but I don't think this one holds up.

I checked every throw site of `NoSuchFormException` (`lib/Service/FormsService.php`, `lib/Controller/ApiController.php`) and all the messages are static hardcoded strings. Nothing gets interpolated from form IDs, user input, or query context, so the "reflected untrusted input" risk in the description doesn't actually exist in the code today.

There is a smaller, related concern: the different messages do give away whether a form is missing vs. forbidden, which chips away at what the throttle is trying to prevent. But the PR only rewrites the body and still forwards `$exception->getCode()`, so 404 vs 403 leaks the same thing the message change hides. You'd also end up sending "Form not found" as the body of a 403, which just isn't true.

If you think you've found a real security issue in a Nextcloud project, the right path is HackerOne via the [security policy](https://github.com/nextcloud/server/security/policy), not a public PR.

I'd lean towards closing this. If you want to tighten the enumeration surface as a hardening change, it'd need to normalize the status code too and come with a test - happy to review that version.

PR #3304, opened 2026-04-19
Chore: Update PHP from 8.1 to 8.2 in composer files

Chartman123:
update php deps


Chartman123:
Forgot to look at the PHP requirements for NC32 that we're still supporting

PR #3311, opened 2026-04-23
Allow custom public share tokens for form links

### LINKED ISSUE #2815: Readable links when sharing

- [Nextcloud Hub 10](31.0.6)
- Forms-Version: 5.1.2 but all others to, I guess

When sharing a suvey, the link is cryptic. It could be pleasant when the owner is able to edit this "Lable" 

f.e. 
instead of this:  https://sub.domain.tld/index.php/apps/forms/s/NDiryNis7HjnLSmiHPykpdNN
this: https://sub.domain.tld/index.php/apps/forms/s/Feedback

Integration in this App: 
https://apps.nextcloud.com/apps/cfg_share_links

**Additional context**

alexander-rebello:
This adds admin-gated custom tokens for public Forms share links. By default the feature is disabled, so existing instances keep the current random-token behavior. When enabled by an admin, form owners can edit the token of an existing public link directly in the sharing sidebar, save it explicitly, and the old URL becomes invalid immediately.

It also adds the necessary backend support for token updates, keeps public-link routing compatible with custom tokens, and includes tests plus API documentation updates.

### DIFF
--- a/.github/workflows/lint-eslint.yml
+++ b/.github/workflows/lint-eslint.yml
@@ -28,12 +28,6 @@ jobs:
       src: ${{ steps.changes.outputs.src}}
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          persist-credentials: false
-          fetch-depth: 0
-
       - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
         id: changes
         continue-on-error: true
--- a/.github/workflows/lint-eslint.yml
+++ b/.github/workflows/lint-eslint.yml
@@ -28,6 +28,12 @@ jobs:
       src: ${{ steps.changes.outputs.src}}
 
     steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
       - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
         id: changes
         continue-on-error: true
--- a/docs/API_v3.md
+++ b/docs/API_v3.md
@@ -631,7 +631,12 @@ Update a single or all properties of an option-object
   | Parameter | Type | Description |
   |------------------|----------|-------------|
   | _keyValuePairs_ | Array | Array of key-value pairs to update |
-- Restrictions: Currently only the _permissions_ can be updated.
+- Restrictions:
+  - Allowed keys are _permissions_ and _token_.
+  - _token_ updates are only available when the admin setting _allowCustomPublicShareTokens_ is enabled.
+  - _token_ can only be updated on link shares.
+  - _token_ must be unique among link shares and only contain alphanumeric characters.
+  - _token_ length must be between 8 and 256 characters.
 - Response: **Status-Code OK**, as well as the id of the share object.
 
 ```
@@ -648,30 +653,6 @@ Update a single or all properties of an option-object
   | _formId_ | Integer | ID of the form containing the share |
   | _shareId_ | Integer | ID of the share to delete |
 - Response: **Status-Code OK**, as well as the id of the deleted share.
-
-```
-"data": 5
-```
-
-### Update a Public Share Token
-
-- Endpoint: `/api/v3/forms/{formId}/shares/{shareId}/token`
-- Method: `PATCH`
-- Url-Parameters:
-  | Parameter | Type | Description |
-  |-----------|---------|-------------|
-  | _formId_ | Integer | ID of the form containing the share |
-  | _shareId_ | Integer | ID of the public link share to update |
-- Parameters:
-  | Parameter | Type | Description |
-  |-----------|---------|-------------|
-  | _token_ | String | New token for the public share link |
-- Restrictions:
-  - Only available when the admin setting _allowCustomPublicShareTokens_ is enabled.
-  - Only link shares can be updated.
-  - Token must be unique among link shares and only contain alphanumeric characters.
-  - Token length must be between 8 and 256 characters.
-- Response: **Status-Code OK**, as well as the id of the updated share.
 
 ```
 "data": 5
--- a/docs/API_v3.md
+++ b/docs/API_v3.md
@@ -653,6 +653,30 @@ Update a single or all properties of an option-object
 "data": 5
 ```
 
+### Update a Public Share Token
+
+- Endpoint: `/api/v3/forms/{formId}/shares/{shareId}/token`
+- Method: `PATCH`
+- Url-Parameters:
+  | Parameter | Type | Description |
+  |-----------|---------|-------------|
+  | _formId_ | Integer | ID of the form containing the share |
+  | _shareId_ | Integer | ID of the public link share to update |
+- Parameters:
+  | Parameter | Type | Description |
+  |-----------|---------|-------------|
+  | _token_ | String | New token for the public share link |
+- Restrictions:
+  - Only available when the admin setting _allowCustomPublicShareTokens_ is enabled.
+  - Only link shares can be updated.
+  - Token must be unique among link shares and only contain alphanumeric characters.
+  - Token length must be between 8 and 256 characters.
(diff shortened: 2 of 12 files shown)


docs/API_v3.md
```diff
@@ -653,6 +653,30 @@ Update a single or all properties of an option-object
...
+  - Only available when the admin setting _allowCustomPublicShareTokens_ is enabled.
+  - Only link shares can be updated.
+  - Token must be unique among link shares and only contain alphanumeric characters.
+  - Token length must be between 8 and 256 characters.
+- Response: **Status-Code OK**, as well as the id of the updated share.
+
+```
+"data": 5
+```
+
```
Chartman123: I wouldn't add a new endpoint for that, please use the already existing update share endpoint

lib/Controller/PageController.php
```diff
@@ -40,6 +40,7 @@
 #[OpenAPI(scope: OpenAPI::SCOPE_IGNORE)]
 class PageController extends Controller {
 	private const TEMPLATE_MAIN = 'main';
+	private const PUBLIC_SHARE_HASH_REQUIREMENT = '[a-zA-Z0-9]{8,256}';
```
Chartman123: Could this be moved to our central Constants.php file?

lib/Controller/PageController.php
```diff
@@ -145,7 +146,7 @@ public function internalLinkView(string $hash): Response {
 	#[NoAdminRequired()]
 	#[NoCSRFRequired()]
 	#[PublicPage()]
-	#[FrontpageRoute(verb: 'GET', url: '/s/{hash}', requirements: ['hash' => '[a-zA-Z0-9]{24,}'])]
+	#[FrontpageRoute(verb: 'GET', url: '/s/{hash}', requirements: ['hash' => self::PUBLIC_SHARE_HASH_REQUIREMENT])]
```
Chartman123: Would probably be good if we can decide here wether custom share tokens are allowed on that instance. But IIRC it's not working with dynamically defined requirements.

lib/Controller/ShareApiController.php
```diff
@@ -301,6 +301,85 @@ public function updateShare(int $formId, int $shareId, array $keyValuePairs): Da
...
+	 * @throws OCSForbiddenException Not allowed to update token on non-link share
+	 * @throws OCSForbiddenException This form is not owned by the current user
+	 * @throws OCSNotFoundException Could not find share
+	 *
+	 * 200: the id of the updated share
+	 */
+	#[CORS()]
+	#[NoAdminRequired()]
+	#[ApiRoute(verb: 'PATCH', url: '/api/v3/forms/{formId}/shares/{shareId}/token')]
+	public function updateShareToken(int $formId, int $shareId, string $token): DataResponse {
```
Chartman123: See other comment... Should be moved to already existing endpoint

Chartman123 (dismissed):
Added some review comments on the PHP part

lib/Controller/PageController.php
alexander-rebello: Yes, makes sense

lib/Controller/PageController.php
alexander-rebello: Meaning it would be good, but can't be done in this instance? Or should I try?

alexander-rebello:
Whats the holdup? Could we check if this version works?

.github/workflows/lint-eslint.yml
Chartman123, line 1: Please remove this change... Workflows are coming from templates and shouldn't be changed in pull requests

lib/Controller/ShareApiController.php
```diff
@@ -256,22 +260,64 @@ public function updateShare(int $formId, int $shareId, array $keyValuePairs): Da
...
 
-		//Don't allow to change other properties than permissions
-		if (count($keyValuePairs) > 1 || !array_key_exists('permissions', $keyValuePairs)) {
-			$this->logger->debug('Not allowed to update other properties than permissions');
-			throw new OCSForbiddenException('Not allowed to update other properties than permissions');
+		$allowedKeys = ['permissions', 'token'];
+		foreach (array_keys($keyValuePairs) as $key) {
+			if (!in_array($key, $allowedKeys, true)) {
+				$this->logger->debug('Not allowed to update unknown properties');
+				throw new OCSForbiddenException('Not allowed to update unknown properties');
```
Chartman123: I'd prefer some other wording here. The properties aren't "unknown", just not allowed.
```suggestion
				$this->logger->debug('Not allowed to update other properties than permissions or token');
				throw new OCSForbiddenException('Not allowed to update other properties than permissions or token');
```

Chartman123:
Please don't merge the `main` branch into your dev branch, always rebase your branch on current main. :)

Chartman123:
Oh and I just saw it: please use some other branch name for your changes in your forked repository...

src/FormsSettings.vue
Chartman123, line 1: Please rebase your branch on our current main and adjust this file accordingly to the changes merged into main lately. We had a bug in this file that prevented sending the requests to the server.

Chartman123:
@alexander-rebello could you please rework the front-end so that it looks/works more like for the file sharing?

grafik

The field should also auto-save upon changes and don't have to rely on clicking a save button in the menu. The button on the right to go back to some auto-generated token is also nice and should be implemented here, too. And I think that we should also hide the token input behind some "customize share token" (instead of the "advanced settings") section.

lib/Controller/PageController.php
```diff
@@ -145,7 +145,7 @@ public function internalLinkView(string $hash): Response {
 	#[NoAdminRequired()]
 	#[NoCSRFRequired()]
 	#[PublicPage()]
-	#[FrontpageRoute(verb: 'GET', url: '/s/{hash}', requirements: ['hash' => '[a-zA-Z0-9]{24,}'])]
+	#[FrontpageRoute(verb: 'GET', url: '/s/{hash}', requirements: ['hash' => Constants::PUBLIC_SHARE_HASH_REQUIREMENT])]
```
pringelmann, line 148: Going from `{24,}` to `{8,}` is a steep drop. afaik `publicLinkView` has no brute force protection (neither does embeddedFormView, which still has no min length at all). Without throttling, that's brute-forceable with a moderate request budget against any instance that turns this on. Could you keep a higher minimum (16+) and add `#[BruteForceProtection]` to both view controllers? Either alone would be a big improvement.

lib/Controller/ShareApiController.php
```diff
@@ -256,22 +260,64 @@ public function updateShare(int $formId, int $shareId, array $keyValuePairs): Da
...
+			}
+
+			if ($token !== $formShare->getShareWith()) {
+				$this->validatePublicShareToken($token);
+
+				try {
+					$existingShare = $this->shareMapper->findPublicShareByHash($token);
+					if ($existingShare->getId() !== $formShare->getId()) {
+						$this->logger->debug('Share hash already exists.');
+						throw new OCSBadRequestException('Share hash exists, please retry.');
```
pringelmann, line 302: This response confirms a token is in use somewhere on the instance, even when the caller has no other way to know that link exists. An authenticated user can probe arbitrary strings against this endpoint and enumerate tokens of private forms they don't own. Better to return the same generic "Invalid share token" that above validation already throws, so "taken" and "invalid format" look identical from the outside.

lib/Controller/PageController.php
Chartman123: What's the min length for share tokens in Files? I think we should use the same value.

alexander-rebello: My own nextcloud instance even allows one character tokens in Files

src/FormsSettings.vue
alexander-rebello: Sure, on it

lib/Controller/PageController.php
susnux: I think short values make sense for e.g. public campaigns.
But we must add bruteforce protection here then.

lib/Controller/ShareApiController.php
alexander-rebello: makes sense, will do!

Chartman123:
@alexander-rebello just wanted to check on your progress :) I think we really need a rebase on latest main here, because we now had quite some changes merged lately. Do you think you can adjust your code accordingly in the next few days? Otherwise I'd like to postpone this feature to the 5.4 release.

Chartman123:
I saw that you created a new branch in your fork... So I'll close this PR. Please create a new one from the new branch.

PR #3316, opened 2026-04-30
fix(settings): properly handle loading state

susnux:


Chartman123 (approved):
Tested and works, thanks for the quick fix!

PR #3321, opened 2026-05-02
refactor(config): replace IConfig with IAppConfig for improved type safety and clarity

Chartman123:

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -15,7 +15,7 @@ use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\FrontpageRoute;
 use OCP\AppFramework\Http\Attribute\OpenAPI;
 use OCP\AppFramework\Http\DataResponse;
-use OCP\IConfig;
+use OCP\AppFramework\Services\IAppConfig;
 use OCP\IRequest;
 
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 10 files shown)


lib/Controller/ConfigController.php
```diff
@@ -65,8 +65,27 @@ public function updateAppConfig(string $configKey, $configValue): DataResponse {
 			return new DataResponse('Mail server is not configured', Http::STATUS_BAD_REQUEST);
 		}
 
-		// Set on DB
-		$this->config->setAppValue($this->appName, $configKey, json_encode($configValue));
```
pringelmann, line 69: We will likely need two new cases here for the new confirmation email keys

Chartman123: Yes, and also for allowing comments on which I'm currently working after susnux's  comment in the respective feature request.

Do you see a better solution than adding cases here?

pringelmann: Maybe a type map? Something like this:

```php
public const CONFIG_KEY_TYPES = [
    self::CONFIG_KEY_ALLOWPERMITALL => 'bool',
    self::CONFIG_KEY_ALLOWPUBLICLINK => 'bool',
    self::CONFIG_KEY_ALLOWSHOWTOALL => 'bool',
    self::CONFIG_KEY_RESTRICTCREATION => 'bool',
    self::CONFIG_KEY_ALLOWCONFIRMATIONEMAIL => 'bool',
    self::CONFIG_KEY_CREATIONALLOWEDGROUPS => 'array',
    self::CONFIG_KEY_CONFIRMATIONEMAILRATELIMIT => 'int',
];
```

and then try -> match -> catch

But not necessary for this PR :)

Chartman123: Oh yes, that sounds better :) thanks

pringelmann (approved):

PR #3322, opened 2026-05-04
refactor: replace `@var` tags with typed properties

Chartman123:

### DIFF
--- a/lib/Db/Share.php
+++ b/lib/Db/Share.php
@@ -23,13 +23,9 @@ use OCP\AppFramework\Db\Entity;
  * @method void setShareWith(string $value)
  */
 class Share extends Entity {
-	/** @var int */
 	protected $formId;
-	/** @var int */
 	protected $shareType;
-	/** @var string */
 	protected $shareWith;
-	/** @var string */
 	protected $permissionsJson;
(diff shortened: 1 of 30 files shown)


tests/Unit/Activity/ActivityManagerTest.php
```diff
@@ -23,9 +23,9 @@
 class ActivityManagerTest extends TestCase {
 
 	private ActivityManager $activityManager;
-	private IManager&MockObject $manager;
-	private IGroupManager&MockObject $groupManager;
-	private CirclesService&MockObject $circlesService;
```
pringelmann, line 28: Why is the type loosening needed here?

Chartman123: I don't know... I just got some errors without this change and we didn't have the `&MockObject` anywhere else.

pringelmann (approved):

tests/Unit/Activity/ActivityManagerTest.php
Chartman123: Strange, I reverted these changes and now I don't get the errors any more.

PR #3323, opened 2026-05-05
feat: improve handling of option limits

### LINKED ISSUE #3257: No indication that option limits are set for checkbox questions
Currently there's no hint for the respondent, that he has to pay attention to limits when he responds to a checkbox question.

We should show the limits to the user and also automatically make the question required as soon as the limit for the minimum number of options is set.

Chartman123:
This closes #3257 by adding a NcNoteCard that shows the current limits. It also fixes a bug where the limits could be greater than the available options. If the lower limit is set, to question will automatically get required.

### DIFF
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -147,6 +147,9 @@
 					v-html="computedDescription" />
 				<!-- eslint-enable vue/no-v-html -->
 			</div>
+			<NcNoteCard v-if="hasInfo" :id="infoId" type="info">
+				{{ infoMessage }}
+			</NcNoteCard>
 			<NcNoteCard v-if="hasError" :id="errorId" type="error">
 				{{ errorMessage }}
 			</NcNoteCard>
@@ -263,6 +266,11 @@ export default {
 			type: String,
 			default: null,
 		},
+
+		infoMessage: {
+			type: String,
+			default: null,
+		},
 	},
 
 	emits: [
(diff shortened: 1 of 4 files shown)


src/mixins/QuestionMixin.js
```diff
@@ -217,8 +217,16 @@ export default {
...
+		hasInfo() {
+			return !!this.infoMessage
+		},
+
 		errorId() {
-			return 'q' + this.index + '_error'
+			return `q${this.index}_error`
+		},
+
+		infoId() {
```
Koc, line 228: looks like unused 🤔

src/components/Questions/Question.vue
```diff
@@ -147,6 +147,9 @@
 					v-html="computedDescription" />
 				<!-- eslint-enable vue/no-v-html -->
 			</div>
+			<NcNoteCard v-if="hasInfo" :id="infoId" type="info">
```
Koc, line 150: why do we need `id` here?

src/mixins/QuestionMixin.js
Chartman123: it's the same like errorId...

src/components/Questions/Question.vue
Chartman123: same as for error...

susnux: errorid is used for semantically describe the error, but infoid is unused.
You need to add a `aria-describedby` for the infoId similar like whats there for `aria-errormessage` for the errorId

Chartman123: ok, thanks for the explanation :)

Chartman123:
the `infoId` review comments are resolved now :) at least I think they are...

Koc (approved):

PR #3327, opened 2026-05-09
chore: begin migration to TypeScript

Chartman123:
This is a first PR to migrate our codebase to TypeScript.

I used Copilot to assist me with the migration. First tests on my dev instance didn't show any errors in the console.


src/router.ts
```diff
@@ -0,0 +1,54 @@
...
+const Create = () => import('./views/Create.vue')
+const Results = () => import('./views/Results.vue')
+const Submit = () => import('./views/Submit.vue')
+const EmptyContent = () => import('./FormsEmptyContent.vue')
+
+const routes = [
+	{
+		path: '/',
+		name: 'root',
+		components: { default: EmptyContent },
```
pringelmann, line 20: This seems to change runtime behavior, not just types? The root route went from component-less to `components: { default: EmptyContent }`, and createWebHistory lost its second arg. Might have been a copilot mistake - can you please check what the implications of this are? (since we already mount empty content via `emptyContent.ts`)

src/utils/OcsResponse2Data.ts
```diff
@@ -7,11 +7,9 @@
  * Extract actual data from Axios OCS response
  * Just a small wrapper for nice code.
  *
- * @param {object} response response returned by axios
- * @return {object} The actual data out of the ocs response
+ * @param response response returned by axios
+ * @return The actual data out of the ocs response
  */
-function OcsResponse2Data(response) {
+export default function OcsResponse2Data(response: unknown): unknown {
```
pringelmann: `response: unknown` makes `response.data.ocs.data` a type error, and returning `unknown` forces every caller into `as any` , which spreads `any` back through the sharing code. This is stricter but weaker than the old JS. Maybe we can make it generic instead:

```ts
export default function OcsResponse2Data<T = unknown>(
  response: { data: { ocs: { data: T } } },
): T {
  return response.data.ocs.data
}
```

Then the `as any` casts at the call sites can go.

src/router.ts
Chartman123: `createWebHistory` or better `generateUrl` takes these args:

 * @param {string} url address (can contain placeholders e.g. /call/{token} would replace {token} with the value of params.token
 * @param {object} params parameters to be replaced into the url
 * @param {UrlOptions} options options for the parameter replacement

So I think we really don't need the `''` anymore.

The `default: EmptyContent` was added by me manually to get rid of a vue-router warning in the console, that there should always be a component for the root route:

`[Vue Router warn]: Record with path "/" is either missing a "component(s)" or "children" property.`

src/utils/OcsResponse2Data.ts
susnux: Or just use OCSResponse from `@nextcloud/typing`?

pringelmann (approved):

PR #3329, opened 2026-05-11
feat: add support for comments in forms

### LINKED ISSUE #1655: Ability to add Comment section for user (like polls)
- Nextcloud-Version: [25]
- Forms-Version: [3.3]

Hi,
it would be nice if form users could comment and basically have a conversation on certain forms (like the comment section in polls where every user can see eachothers comments). This option should only be available if activated.

Usecase: 
We use forms to organize daytrips and if the users could use the comment section to organize rideshares it would really help.

Thx

Chartman123:
This closes #1655 by integrating the comments app.

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -29,6 +29,7 @@ use OCP\AppFramework\Http\Response;
 use OCP\AppFramework\Http\Template\PublicTemplateResponse;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\AppFramework\Services\IInitialState;
+use OCP\Comments\ICommentsManager;
 use OCP\IL10N;
 use OCP\IRequest;
 use OCP\IURLGenerator;
@@ -51,6 +52,7 @@ class PageController extends Controller {
 		private FormsService $formsService,
 		private IAccountManager $accountManager,
 		private IInitialState $initialState,
+		private ICommentsManager $commentsManager,
 		private IL10N $l10n,
 		private IUrlGenerator $urlGenerator,
 		private IUserManager $userManager,
@@ -66,7 +68,9 @@ class PageController extends Controller {
 	#[NoCSRFRequired()]
 	#[FrontpageRoute(verb: 'GET', url: '/')]
 	public function index(?string $hash = null, ?int $submissionId = null): TemplateResponse {
-		Util::addScript($this->appName, 'forms-main');
+		// Ensure the Comments client is available and load comments resources
+		$this->commentsManager->load();
+		Util::addScript($this->appName, 'forms-main', 'comments');
 		Util::addStyle($this->appName, 'forms');
 		Util::addStyle($this->appName, 'forms-style');
 		$this->insertHeaderOnIos();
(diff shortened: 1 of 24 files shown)


Chartman123:
Front-end code was created with assistance by Copilot. There's probably room for improvement.

I'm also facing the problem that sometimes the comments tab loads, but doesn't show existing comments. Probably some kind of race condition.

lib/Controller/PageController.php
```diff
@@ -66,7 +68,9 @@ public function __construct(
 	#[NoCSRFRequired()]
 	#[FrontpageRoute(verb: 'GET', url: '/')]
 	public function index(?string $hash = null, ?int $submissionId = null): TemplateResponse {
-		Util::addScript($this->appName, 'forms-main');
+		// Ensure the Comments client is available and load comments resources
+		$this->commentsManager->load();
```
susnux, line 72: Is this needed? I never did this before 👀

Chartman123: This adds the script to the page. this is also how announcementcenter adds the comments :)

https://github.com/nextcloud/announcementcenter/blob/7d7f283091643bbebafa349bdcc1cc1bc63af2aa/lib/Controller/PageController.php#L67

lib/Listener/CommentsEntityListener.php
```diff
@@ -0,0 +1,43 @@
...
+
+	#[\Override]
+	public function handle(Event $event): void {
+		if (!$event instanceof CommentsEntityEvent) {
+			return;
+		}
+
+		// Register the 'forms' entity collection so the Comments app can
+		// check whether a given form id allows comments.
+		$event->addEntityCollection('forms', function ($formId) {
```
pringelmann, line 36: As far as I can tell this callback is the only thing the comments layer checks before letting someone read or write. Checking just `allowComments` means any logged-in user can hit `/comments/forms/<id>` and post or read comments on any form with comments enabled, no share access required. Maybe we need the some kind of per-user check here?

susnux: yes this makes sense, we should check here if the current user:
1. is the owner
2. or the form is shared with the user

Chartman123: We shouldn't limit comments to the owner or shared admins but to all logged in users with access to the form. The use case of the feature request was for allowing communication between respondents.

pringelmann:
Should we also clean up comments when a form is deleted? (`deleteForm()` in `FormMapper.php`) Otherwise it will likely lead to many of orphan table rows over time

lib/Listener/CommentsEntityListener.php
susnux: yes that what I meant we still would need the check:
1. is the form shared with anyone?
2. otherwise: is the user the owner
3. otherwise: is the form shared with the user

Because otherwise some user could read comments of a form where he does not have access to

Chartman123: Yes that's right. I first thought that you meant only shared edit permissions

Chartman123:
Fixed the review comments now :)

pringelmann (approved):

susnux (approved):

PR #3343, opened 2026-05-16
fix: Prevent infinity loop on form deletion

Koc:
Reproducer:
1. Create form
2. Open this form
3. Try to delete it

Actual result: form has been deleted but we have some kind of the infinity loop and tons of errors.

image

This PR aims to fix that


Chartman123:
Strange, this was already fixed some time ago iirc. And I deleted a form on my dev instance without this fix and couldn't see the loading loop. Are you sure that you're on current main?

Chartman123:
Ok, I deleted another form right now and hit the loop...

Chartman123 (approved):

PR #3360, opened 2026-05-25
feat: add question at position

### LINKED ISSUE #2873: Option to insert new question anywhere in the form

- Nextcloud-Version: 31.0.8
- Forms-Version: 5.1.2

Its frustrating and buggy when i want to insert a question at the top of my form, but the new question is inserted at the very end of the form. And with a form of 50 questions i have to drag or move the one question up to the right position... the move/drag buttons also glitch our of position after 20 questions or so, and it becomes harder to find the hitbox for these buttons misaligned.

Id like for the new question button behaviour to change so that new question are inserted after the question being actively edited

I tried to drag and scroll but the viewport doesn't scroll when you drag to it's top or bottom

**Additional context**
an image of the glitching buttons:
Image

Chartman123:
This closes #2873 by adding a new NcActions menu between existing questions. You can then choose to insert a new question or copy the question above. The menu is visually hidden and only shown when your focusing the question above. It's accessible with the keyboard, too.

Inserting a new question re-uses the existing menu for adding new questions.

https://github.com/user-attachments/assets/2ea1de97-67d1-4ba9-9ab4-51d4aba8d3b5

Mobile view (button in 0.3 opacity as no hover available):

image

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -116,108 +116,103 @@
 				<!-- Questions list -->
 				<Draggable
 					v-model="form.questions"
-					:animation="200"
-					tag="transition-group"
-					:componentData="{
-						name: isDragging
-							? 'no-external-transition-on-drag'
-							: 'question-list',
-					}"
+					:animation="300"
+					target=".sort-target"
+					direction="vertical"
+					invertSwap
 					handle=".question__drag-handle"
 					@change="onQuestionOrderChange"
-					@start="isDragging = true"
-					@end="isDragging = false">
-					<component
-						:is="answerTypes[question.type].component"
-						v-for="(question, index) in form.questions"
-						:key="question.id"
-						:ref="setQuestionRef"
-						v-bind="form.questions[index]"
-						:canMoveDown="index < form.questions.length - 1"
-						:canMoveUp="index > 0"
-						:answerType="answerTypes[question.type]"
-						:index="index + 1"
-						:maxStringLengths="maxStringLengths"
-						@update:text="(val) => (form.questions[index].text = val)"
-						@update:description="
-							(val) => (form.questions[index].description = val)
-						"
-						@update:isRequired="
-							(val) => (form.questions[index].isRequired = val)
-						"
-						@update:name="(val) => (form.questions[index].name = val)"
-						@update:extraSettings="
-							(val) => (form.questions[index].extraSettings = val)
-						"
-						@update:options="
-							(val) => (form.questions[index].options = val)
-						"
-						@clone="cloneQuestion(question)"
-						@delete="deleteQuestion(question.id)"
-						@moveDown="onMoveDown(index)"
-						@moveUp="onMoveUp(index)" />
+					@start="onDragStart"
+					@end="onDragEnd">
+					<TransitionGroup
+						tag="ul"
+						:name="isDragging ? undefined : 'question-list'"
+						class="sort-target">
+						<component
+							:is="answerTypes[question.type].component"
+							v-for="(question, index) in form.questions"
+							:key="question.id"
+							:ref="(el) => setQuestionRef(el, question)"
+							v-bind="form.questions[index]"
+							:canMoveDown="index < form.questions.length - 1"
+							:canMoveUp="index > 0"
+							:answerType="answerTypes[question.type]"
+							:index="index + 1"
+							:maxStringLengths="maxStringLengths"
+							@update:text="
+								(val) => (form.questions[index].text = val)
+							"
+							@update:description="
+								(val) => (form.questions[index].description = val)
+							"
+							@update:isRequired="
+								(val) => (form.questions[index].isRequired = val)
+							"
+							@update:name="
+								(val) => (form.questions[index].name = val)
+							"
+							@update:extraSettings="
+								(val) => (form.questions[index].extraSettings = val)
+							"
+							@update:options="
+								(val) => (form.questions[index].options = val)
+							"
+							@clone="cloneQuestion(question, index)"
+							@delete="deleteQuestion(question.id)"
+							@moveDown="onMoveDown(index)"
+							@moveUp="onMoveUp(index)">
+							<template
+								v-if="index < form.questions.length - 1"
+								#insert>
+								<div
+									class="question-insert"
+									:class="[
+										{
+											'is-open':
+												insertMenuOpenedIndex === index,
+										},
+										{
+											'is-mobile': isMobile,
+										},
+									]">
+									<AddQuestionMenu
+										:menuName="t('forms', 'Insert question')"
+										:aria-label="
+											t(
+												'forms',
+												'Insert question after question {index}',
+												{ index: index + 1 },
+											)
+										"
+										variant="tertiary"
+										:position="index"
+										:isLoadingQuestions="isLoadingQuestions"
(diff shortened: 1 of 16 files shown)


Chartman123:
Note for the review: I used copilot especially for the frontend part. Please have a closer look on the changes made in Create.vue and see if there's a better way for doing it. The questionRefsMap was needed for focusing the newly inserted/cloned questions correctly.

nimishavijay:
Really cool! I would 100% use this :) So design wise, the add button between questions is the right call. What do you think of the following changes:

- [x] I see that the "Clone question" has been moved here, but I think it was great to have it in the action menu like before. It makes it clear which question you are actually cloning. I'd also change the wording to be "Copy question" like before. However once you click on copy the clone can be created right below instead of at the end. What do you think?
- [x] Removing the "Clone" option also makes the add button have only one option -- add here, so then we could directly open the question type menu. 
- [x] And either way it would be good to have this as a full-width tertiary button :)

Let me know what you think :)

Chartman123:
> * I see that the "Clone question" has been moved here, but I think it was great to have it in the action menu like before. It makes it clear which question you are actually cloning. I'd also change the wording to be "Copy question" like before. However once you click on copy the clone can be created right below instead of at the end. What do you think?

@nimishavijay the clone action also stayed in the question menu to clone to the end of the form. I'll fix the wording

> * And either way it would be good to have this as a full-width tertiary button :)

good idea :)

jancborchardt (changes requested):
Very cool! 
- [x] Agree with @nimishavijay that it’s better to have the "Copy question" separate, also because then we can directly open the type selector on first click
- [x] The type selector should also open in place, instead of opening it on the bottom left, as that moves focus
- [x] When the question gets added, it’s kind of unclear what happened (at least from the video, maybe due to fps?) – some quick animation might be nice. :)

Chartman123:
> [ ] Agree with @nimishavijay that it’s better to have the "Copy question" separate, also because then we can directly open the type selector on first click

@jancborchardt but how do you want to distinguish where to add the copied question if we only have it inside the question menu? We'd need two actions in the menu then...

> [ ] The type selector should also open in place, instead of opening it on the bottom left, as that moves focus

It was the easiest way for me to just re-use the existing menu. Not sure if I can mount it directly on the new button...

> * When the question gets added, it’s kind of unclear what happened (at least from the video, maybe due to fps?) – some quick animation might be nice. :)

The newly added question gets added and the UI then scrolls to the new location. So it's basically already animated.

Chartman123:
@nimishavijay @jancborchardt I reworked the insert menu now and only kept the copy question action in the question menu. I changed the behaviour to always copy a question below the existing one. Full width NcActions is currently not directly supported by the vue lib, so I added a name to the menu for now.

src/views/Create.vue
```diff
@@ -737,4 +777,39 @@ export default {
...
+.question-list-move,
+.question-list-enter-active,
+.question-list-leave-active {
+	transition: all var(--animation-slow) ease;
+}
+
+.question-list-enter-from,
+.question-list-leave-to {
+	opacity: 0;
+	transform: translateX(44px);
```
pringelmann: Maybe we can use a token here?


```suggestion
    transform: translateX(var(--default-clickable-area, 44px));
```

Chartman123: What about `--clickable-area-large` which is `48px`. Default is only `34px`.

Chartman123: Yes, makes sense... also have this already in QuestionRanking. So I'd use default for options and large for questions :)

src/views/Create.vue
```diff
@@ -737,4 +776,39 @@ export default {
...
+
+.question-insert {
+	/* closer to the question above */
+	position: relative;
+	margin-block-end: -34px;
+	inset-block-end: -16px;
+	display: flex;
+	justify-content: center;
+	opacity: 0;
+	transition: opacity 0.12s ease;
```
pringelmann, line 786: On touch devices there's no hover and `:focus-within` only fires after you've already tabbed into the button. That leaves no way to insert a question between two existing ones on a phone or tablet - you'd have to add at the bottom and drag it up. Could we keep these visible at low contrast by default, or surface them on a tap somewhere?

Chartman123: We could use `useIsMobile` and add a class in this case

src/views/Create.vue
```diff
@@ -116,117 +116,97 @@
...
+									class="question-insert"
+									:class="[
+										{
+											'is-open':
+												insertMenuOpened
+												&& pendingInsertIndex === index + 1,
+										},
+									]">
+									<AddQuestionMenu
+										:menuName="t('forms', 'Insert question')"
```
pringelmann, line 179: "Insert question" doesn't tell a screen reader user where it inserts. After reading question 3 and its actions they hit "Insert question" with no anchor. Something like 

```js
t('forms', 'Insert question after question {index}', { index: index + 1 })
```
would match the labelling style already used elsewhere

pringelmann (approved):
Very nice! Just a small comment on mobile experience :)

src/views/Create.vue
Chartman123: NcActions doesn't use the aria-label if menuName is set. I think that the text would be too long if we use this for the menuName. Could we leave it this way until NcActions provide a `wide` setting? Then we could use the `+` icon only without the name and the aria-label would be used.

pringelmann: sound good!

src/views/Create.vue
```diff
@@ -300,6 +286,12 @@ export default {
 			questionMenuOpened: false,
 			activeQuestionType: null,
 			questionRefs: [],
+			questionRefsMap: {},
```
pringelmann, line 291: You don't need to keep both. `questionRefs` (the array) is never written in practice: every question in the v-for has a server-assigned id by mount time, so `setQuestionRef` always takes the map branch. The positional fallbacks in `insertQuestion` are unreachable for the same reason. The map alone covers every case.

Vue 3 calls function refs with `null` on unmount, which also lets you drop the `beforeUpdate` reset and have the map clean itself up per item:

```js
setQuestionRef(el, question) {
  if (el) {
    this.questionRefsMap[question.id] = el
  } else {
    delete this.questionRefsMap[question.id]
  }
}
```

After this, `insertQuestion`'s focus logic collapses to `this.questionRefsMap[newQuestionObj.id]?.focus()` in both branches, and `questionRefs` + `beforeUpdate` go away.

src/views/Create.vue
```diff
@@ -300,6 +286,12 @@ export default {
...
+
+			// when set to a number, the next created question will be inserted at this index
+			pendingInsertIndex: null,
```
pringelmann: The position is known per-menu at template time. If `AddQuestionMenu` takes it as a prop and includes it in the emit:

```vue
<AddQuestionMenu :position="index + 1" @addQuestion="addQuestion" />
```

```js
addQuestion(type, subtype, position) { ... }
```

then `pendingInsertIndex`, `onInsertMenuUpdate`, and the null-out timing in the close handler all go away. The position lives with the menu that owns it. Same fix covers the clone button (it can emit its own position directly).

jancborchardt (approved):
Looks reaaally nice, and is very useful! :)

PR #3363, opened 2026-05-26
fix: only send a single submission notification per user

Chartman123:
This deduplicates submission notifications by aggregating affected users and sending a single shared-submission activity per unique user, excluding the form owner so they only get the main submission notification.

### DIFF
--- a/lib/Service/FormsService.php
+++ b/lib/Service/FormsService.php
@@ -713,10 +713,9 @@ class FormsService {
 	 * Creates activities for new submissions on a form
 	 *
 	 * @param Form $form Related Form
-	 * @param string $submitter The ID of the user who submitted the form. Can also be our 'anon-user-'-ID
+	 * @param Submission $submission the current submission
 	 */
 	public function notifyNewSubmission(Form $form, Submission $submission): void {
-		$shares = $this->getShares($form->getId());
 		try {
 			$this->activityManager->publishNewSubmission($form, $submission->getUserId());
 		} catch (\Exception $e) {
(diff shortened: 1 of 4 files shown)


lib/Service/FormsService.php
```diff
@@ -728,12 +727,43 @@ public function notifyNewSubmission(Form $form, Submission $submission): void {
 			);
 		}
 
+		$shares = $this->getShares($form->getId());
+		$sharedUserIds = [];
 		foreach ($shares as $share) {
 			if (!in_array(Constants::PERMISSION_RESULTS, $share['permissions'])) {
 				continue;
 			}
+			switch ($share['shareType']) {
```
pringelmann: The group/circle resolution moved out of the try/catch. If `getCircleUsers` or `groupManager->get()` throws, the whole submission breaks now. Could we wrap this loop (or each resolution) so activity errors stay non-fatal like before?

Chartman123: good catch, thanks

tests/Unit/Service/FormsServiceTest.php
```diff
@@ -1165,9 +1165,26 @@ public static function dataNotifyNewSubmission() {
...
+				2
+			],
+			'circle-share' => [
+				[[
+					'shareWith' => 'sharedCircle',
+					'shareType' => IShare::TYPE_CIRCLE,
+					'permissions' => [ Constants::PERMISSION_RESULTS ]
+				]],
+				2
+			],
```
pringelmann, line 1187: Could we add a case where the same user appears in multiple shares (e.g. direct user share + group share containing the same user), asserting only one notification fires? That's the actual regression this PR fixes and right now nothing pins it down.

Maybe something like this?

```php
'overlapping-shares' => [
    [
        [
            'shareWith' => 'user2',
            'shareType' => IShare::TYPE_USER,
            'permissions' => [ Constants::PERMISSION_RESULTS ]
        ],
        [
            'shareWith' => 'sharedGroup',
            'shareType' => IShare::TYPE_GROUP,
            'permissions' => [ Constants::PERMISSION_RESULTS ]
        ]
    ],
    2
],
```

Chartman123: It should be enough to add `ownerUser` to `$sharedUsers` as the owner should also get filtered.

pringelmann (approved):

tests/Unit/Service/FormsServiceTest.php
Chartman123: But testing the overlap makes sense, too :) I'll add that

PR #3368, opened 2026-05-29
fix(export): type numeric answers as numbers in exports

solracsf:
Radio and other choice-type answers are always PHP strings, even when the option labels are pure numbers like "1"-"5".

`setCellValue()` wrote every string via `setValueExplicit()` with the default `TYPE_STRING`, so numeric answers landed in Text-formatted cells and could not be averaged in the spreadsheet application.

For XLSX/ODS, route purely numeric values through `setCellValue()` so they are stored with a numeric data type. A value is only treated as numeric when it is `is_numeric()`, does not start with a formula-trigger character (`= + - @`), and round-trips exactly `((string)(+$value) === $value)`.

The round-trip guard keeps leading-zero phone numbers, scientific notation, trailing-zero decimals and oversized ids as text to avoid silent data loss, since all export values arrive as strings from `Answer::getText()`.

The anti-formula string protection and CSV escaping are unchanged.

### DIFF
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -33,6 +33,7 @@ use OCP\IUser;
 use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\Mail\IMailer;
+use PhpOffice\PhpSpreadsheet\Cell\DataType;
 use PhpOffice\PhpSpreadsheet\IOFactory;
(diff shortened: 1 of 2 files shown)


solracsf:
@Chartman123 we've manually backported it as codebase is not the same.

Chartman123:
You're absolutely right, my fault... so please re-open the manual backport :)

Chartman123 (approved):

PR #3380, opened 2026-06-04
Add export/import functionality for forms

### LINKED ISSUE #1425: Add export/import functionality for forms (transferability)
- Nextcloud-Version: 25.0.2
- Forms-Version: 3.0.1

I cannot transfer a form from one Nextcloud instance to another. 

I would like to be able to export a form (or all forms or selected forms) to an editable file, e. g., XML or JSON, and then import the form(s) via that file into another or even the same Nextcloud instance. 

None

**Additional context**
None

timedin-de:
building export and import via json
Download Button:
- fetches Forms Data (questions and settings, removes ids and hashes) => Downloads as json blob 
- adds version of the app => on import, warn if downgrade
- modal to ask if answers should get exported as well? (!handle file uploads, not exported)

#1425

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -23,11 +23,13 @@ use OCA\Forms\Db\SubmissionMapper;
 use OCA\Forms\Db\UploadedFile;
 use OCA\Forms\Db\UploadedFileMapper;
 use OCA\Forms\Exception\NoSuchFormException;
+use OCA\Forms\Helper\FilePathHelper;
 use OCA\Forms\ResponseDefinitions;
 use OCA\Forms\Service\ConfigService;
 use OCA\Forms\Service\ConfirmationEmailService;
 use OCA\Forms\Service\FormsService;
 use OCA\Forms\Service\SubmissionService;
+use OCP\App\IAppManager;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\IMapperException;
 use OCP\AppFramework\Http;
@@ -76,6 +78,7 @@ class ApiController extends OCSController {
 		IRequest $request,
 		IUserSession $userSession,
 		private readonly AnswerMapper $answerMapper,
+		private readonly FilePathHelper $filePathHelper,
 		private readonly FormMapper $formMapper,
 		private readonly OptionMapper $optionMapper,
 		private readonly QuestionMapper $questionMapper,
@@ -86,6 +89,7 @@ class ApiController extends OCSController {
 		private readonly SubmissionService $submissionService,
 		private readonly IL10N $l10n,
 		private readonly LoggerInterface $logger,
+		private readonly IAppManager $appManager,
 		private readonly IUserManager $userManager,
 		private readonly IRootFolder $rootFolder,
 		private readonly UploadedFileMapper $uploadedFileMapper,
@@ -251,7 +255,8 @@ class ApiController extends OCSController {
 	 * Read all information to edit a Form (form, questions, options, except submissions/answers)
 	 *
 	 * @param int $formId Id of the form
-	 * @return DataResponse<Http::STATUS_OK, FormsForm, array{}>
+	 * @param ?bool $download if the form should be downloaded
+	 * @return DataResponse<Http::STATUS_OK, FormsForm, array{}>|DataDownloadResponse<Http::STATUS_OK, 'application/json', array{}>
 	 * @throws OCSBadRequestException Could not find form
 	 * @throws OCSForbiddenException User has no permissions to get this form
 	 *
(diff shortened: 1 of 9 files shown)


Chartman123:
@timedin-de Please don't merge the main branch into your branch, always rebase your's on main. Also please sign-off your commits.

timedin-de:
@Chartman123 i dont think i did merge the main branch into mine, did sth slip through? Yes, will do, sorry for forgetting it everytime ^^

Chartman123:
Now the history looks clean, don't now what it was then...

src/Forms.vue
```diff
@@ -137,24 +148,52 @@
...
 import IconArchive from '@material-symbols/svg-400/outlined/archive.svg?raw'
+import IconUpload from '@material-symbols/svg-400/outlined/upload.svg?raw'
 import axios from '@nextcloud/axios'
 import { showError } from '@nextcloud/dialogs'
 import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { loadState } from '@nextcloud/initial-state'
 import moment from '@nextcloud/moment'
 import { generateOcsUrl } from '@nextcloud/router'
-import { useIsMobile } from '@nextcloud/vue'
+import { NcActionButton, NcDialog, useIsMobile } from '@nextcloud/vue'
```
Chartman123: Please revert this line and import NcActionButton and NcDialog like the other components below

src/Forms.vue
```diff
@@ -440,6 +484,121 @@ export default {
...
+				})
+				reader.readAsText(file)
+			})
+		}
+
+		const closeModal = () => {
+			showVersionMismatch.value = false
+			formForImport = undefined
+		}
+		const onDownloadForm = async (id) => {
```
Chartman123, line 606: why don't you adjust the API endpoint and introduce an "export" or "download" flag there? Then you could already remove the unneeded props and also directly trigger the download response from the server.

lib/Controller/ApiController.php
```diff
@@ -157,14 +159,19 @@ public function getForms(string $type = 'owned'): DataResponse {
...
 		// Check if user is allowed
 		if (!$this->configService->canCreateForms()) {
 			$this->logger->debug('This user is not allowed to create Forms.');
 			throw new OCSForbiddenException('This user is not allowed to create Forms.');
 		}
 
-		if ($fromId === null) {
+		// Validate mutually exclusive parameters
+		if ($fromId !== null && $import === true) {
+			throw new OCSBadRequestException('Cannot use both fromId and import parameters');
```
Chartman123, line 184: please add this to the docblock

lib/Controller/ApiController.php
```diff
@@ -183,10 +190,24 @@ public function newForm(?int $fromId = null): DataResponse {
 
 			$this->formMapper->insert($form);
 		} else {
-			$oldForm = $this->formsService->getFormIfAllowed($fromId, Constants::PERMISSION_EDIT);
+			// Fill variables from json or database
+			if ($import) {
+				if (!isset($formData['questions']) || !\is_array($formData['questions'])) {
+					throw new OCSBadRequestException('Invalid form data: missing questions');
```
Chartman123, line 209: same here

lib/Controller/ApiController.php
```diff
@@ -213,26 +236,35 @@ public function newForm(?int $fromId = null): DataResponse {
...
 
-			// Get Questions, set new formId, reinsert
-			$questions = $this->questionMapper->findByForm($oldForm->getId());
-			$oldConfirmationEmailQuestionId = $oldForm->getConfirmationEmailQuestionId();
-
 			foreach ($questions as $oldQuestion) {
-				$questionData = $oldQuestion->read();
+				if ($import) {
+					if (!isset($oldQuestion['id'])) {
+						throw new OCSBadRequestException('Invalid question data: missing id');
```
Chartman123, line 268: and here

Chartman123 (commented):
I've had a quick look at the PR and added a few comments :)

src/Forms.vue
timedin-de: so i wanted to touch the PHP-Backend as little as needed ^^.

Ok so i'll just do a window.open with the file directly downloading from server.

Chartman123: All things about the data model should be handled in the backend so that the frontend part, in theory, is exchangable.

You can look at the action button that is used for downloading the results :)

timedin-de:
@Chartman123 i added the fixup commits for your comments, i'll rebase them if you accept them

One other thing I would improve is validation/sanitasation of import/export.

I'd make whitelist for allowed properties and remove others (or fail the request?) and copy the validations for questions: areExtraSettingsValid and `array_search($type, Constants::ANSWER_TYPES)`
and options:

Chartman123:
@timedin-de are you still working on this?

timedin-de:
> @timedin-de are you still working on this?

yes, im currently less active. Do you have any additions missing in this? And a decision for sanitazation?

Chartman123:
> And a decision for sanitazation?

I vote for the whitelist and throw an error if other properties are part of the request.

> Do you have any additions missing in this?

No, don't think so :)

Chartman123:
@timedin-de Did you merge main again? Please always use rebase... And can you please remove the issue id from the commit message? It creates a new log record on every push in the issue.

timedin-de:
> @timedin-de Did you merge main again? Please always use rebase... And can you please remove the issue id from the commit message? It creates a new log record on every push in the issue.

I'm sorry, i dont know what, but something got messed up while rebasing (promise, i didnt merge :grimacing: ) Now it should be clean.

Edit: it seems it messed up while doing the interactive rebase, I'll need to recheck the git docs on that.

Chartman123:
I think you need to make sure that your forked main is up-to-date before doing the rebase :)

timedin-de:
> I think you need to make sure that your forked main is up-to-date before doing the rebase :)

thanks, yeah that and i ran `git rebase -i HEAD~10` but should've `git rebase -i main` ^^

Now it should be clean

lib/Controller/ApiController.php
Chartman123, line 1: I had another look at your changes in this file: I think it would be more readable and easier to handle if you used a separate if-branch for import. (so in the end three branches: create new, copy from id, import). If you want to deduplicate code you could extract common lines into helper functions.

timedin-de: for the defaults, are these still valid and needed for both import and dupe?
grafik
Could you explain why you set these defaults? Like if i need submitMultiple for a form, i probably would need it also for  the duped or reimported form

Chartman123: The defaults have been set to be consistent with a newly created form. In most cases you would expect to only copy or export/import the structure of the form and not the settings. So I'd say yes: we should keep them for the import, too.

samin-z:
@timedin-de im going over the pr currently, few points first before getting into the review:
1. i think its better to rename 'download form' to 'export form' since the import button is already 'import form' so we keep it consistent and easier for users to understand they can import this downloaded form (so it's the export for it).
2. it throws 400 error when 'extrasettings' include min and max date
```
"extraSettings": {
                    "dateMin": "2026-09-08",
                    "dateMax": "2026-09-12"
 },
```
`"message": "Invalid question data: invalid extraSettings"`

timedin-de:
> it throws 400 error when 'extrasettings' include min and max date

Thanks for your tests, it seems to work now, maybe you had an export from the old version? The current one downloads dates as number
grafik

timedin-de:
@samin-z  @Chartman123 

Added your changes, rebased and pushed clean

Chartman123:
> The current one downloads dates as number

We've changed the way min/max dates are stored in the db now. Old values can still be stored as unix timestamp/integer. Newer values will be stored as date string 'yyyy-mm-dd'

timedin-de:
@Chartman123 @samin-z 

> it throws 400 error when 'extrasettings' include min and max date

I see, i had an old frontend cache, now i tested with string dates but it worked for me. 

Can you confirm its still there? As the import function is using the same formsService->areExtraSettingsValid from normal form creation it should always be in sync with (date) format changes

PR #3405, opened 2026-06-10
chore: install rector

Chartman123:
Install rector and use the following rules/sets:

- NEXTCLOUD_32
- php81

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
 		"test:unit": "phpunit --display-all-issues -c tests/phpunit.xml",
 		"test:integration": "phpunit --display-all-issues -c tests/phpunit.integration.xml",
 		"psalm": "psalm --no-cache",
-		"openapi": "generate-spec"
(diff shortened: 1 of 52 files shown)


Koc:
wow, nice! Do we want add it into pipeline?

Koc (approved):

Chartman123:
There's a workflow template here that runs rector once a week and creates a PR with the resulting changes.

https://github.com/nextcloud/.github/blob/master/workflow-templates%2Frector-apply.yml

However I didn't find an org template for adding it to the CI pipeline

Chartman123:
@pringelmann should we backport those changes to stable5.3? Would make backporting of following bugfixes for 5.3 easier, but on the other hand those are bigger changes than you'd normally expect in a patch release.

pringelmann:
@Chartman123 I think its fine. 5.3 is still very fresh, and it seems theres very low regression risk. Reduced future backport friction is worth it imo

PR #3410, opened 2026-06-10
fix: Fix unit selection in file type

### LINKED ISSUE #3408: Unit selection for files question not working correctly
While testing file upload questions, I noticed another issue that may be related to usability.

When configuring the maximum file size for a file upload question, the unit selector (kB / MB / GB) does not work correctly. The dropdown opens, but I am unable to change the unit. It remains unchanged and appears to result in an error.

I am not sure whether this is a bug or related to my setup, but I wanted to mention it here in case others experience the same problem.

_Ursprünglich gepostet von @lichtenfelsjw-gif in https://github.com/nextcloud/forms/issues/3407#issuecomment-4671467167_

Koc:
Closes #3408

## 🤖 AI (if applicable)
- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/src/FormsSettings.vue
+++ b/src/FormsSettings.vue
@@ -22,7 +22,7 @@
 				:placeholder="t('forms', 'Select groups')"
 				class="forms-settings__creation__multiselect"
 				label="displayName"
-				@input="onCreationAllowedGroupsChange" />
+				@update:modelValue="onCreationAllowedGroupsChange" />
 		</NcSettingsSection>
 		<NcSettingsSection
 			:name="t('forms', 'Confirmation emails')"
--- a/src/components/Questions/Question.vue
+++ b/src/components/Questions/Question.vue
@@ -102,7 +102,7 @@
 						:labelOutside="false"
 						:showTrailingButton="false"
 						:modelValue="name"
-						@input="onNameChange">
+						@update:modelValue="onNameChange">
 						<template #icon>
 							<NcIconSvgWrapper :svg="IconIdentifier" />
 						</template>
@@ -374,8 +374,8 @@ export default {
 			this.$emit('update:description', target.value)
 		},
 
-		onNameChange({ target }) {
-			this.$emit('update:name', target.value)
+		onNameChange(name) {
+			this.$emit('update:name', name)
 		},
 
 		onRequiredChange(isRequired) {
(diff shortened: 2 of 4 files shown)


Chartman123:
Nice, thanks :) do we have to adjust the other `@input`'s in the menu, too?

Koc:
I thought about that as well :handshake: but it works fine with even with `@input` :man_shrugging: . So I've decided to reduce amount of changes.

If you want - I can adjust all of them, just say

Chartman123:
Yes, would be nice :)

`@input` is no longer listed in the events section of the style guide. so might stop working sometime in the future :)

https://nextcloud-vue-components.netlify.app/#/Components/NcActions

Please also have a look at the custom file extensions, in my prod instance the field only takes numbers

Koc:
@Chartman123 I did more changes, but...

1. there are short answer question type that relies on `event?.isComposing` (used for Chineese/Japanese). Sorry, but I have no any idea how to combine this with `update:modelValue`, it seems like we still need `@input` for that
2. extension selection not works for me. It works fine for [vue-select](https://vue-select.org/guide/values.html#tagging) but not works for [Nextcloud](https://vue-select.org/guide/values.html#tagging) wrapper. Looks like a bug on `nextcloud-libraries/nextcloud-vue` side

feel free to push to my branch

Chartman123:
> 1. it seems like we still need `@input` for that

Yes, I think so, too 👍🏻 

I could fix the extension selection by adding a key to the NcActionInput. Please test it :) If it's working on your instance, too, you can squash the commits and rebase on main :)

src/components/Questions/QuestionShort.vue
```diff
@@ -60,7 +60,7 @@
 					ref="regexInput"
 					:label="t('forms', 'Regular expression for input validation')"
 					:modelValue="validationRegex"
-					@input="onInputRegex"
+					@update:modelValue="onInputRegex"
```
Koc: IMHO this will not work. 

Before: `onInputRegex` receives an event with properties, including `isComposing`
Now: `onInputRegex` receives string primitive

Koc:
wow, with `key` helped. But how it works? Was the problem caused by reusing of another component?

I've reverted changes in the `QuestionShort.vue` (see comment)

Chartman123:
Copilot suggested that it might have come from and interfered with some other type value in the attrs passed to the component...

Chartman123 (approved):

PR #3382, opened 2026-06-05
fix: add validation method and error message

Chartman123:
This fixes #1586 and fixes #1255 by adding a validate() function to all question types.

This improves the error message handling for required questions that have not been answered.

Signed-off-by: Christian Hartmann <chris-hartmann@gmx.de>

PR #3413, opened 2026-06-12
fix: show validation error for QuestionShort in NcNoteCard 

Chartman123:
After adding custom validation with `@invalid.prevent="validate"` in #3382 the validation in QuestionShort didn't work as expected anymore.

The validation was refactored to also make use of the NcNoteCard to fix the issue.

While fixing this I realized that the aria-messages about errors were also missing in the above mentioned PR and so I added them here as well.

### DIFF
--- a/src/components/Questions/QuestionShort.vue
+++ b/src/components/Questions/QuestionShort.vue
@@ -15,6 +15,8 @@
 				ref="input"
 				:aria-labelledby="titleId"
 				:aria-describedby="description ? descriptionId : undefined"
+				:aria-errormessage="hasError ? errorId : undefined"
+				:aria-invalid="hasError ? 'true' : undefined"
 				:placeholder="submissionInputPlaceholder"
 				:disabled="!readOnly"
 				:name="name || undefined"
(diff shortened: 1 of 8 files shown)


src/components/Questions/QuestionShort.vue
```diff
@@ -159,44 +163,42 @@ export default {
...
+				return false
+			}
+
 			this.errorMessage = null
 			return true
 		},
 
-		onInput() {
-			/** @type {HTMLObjectElement} */
+		onInput: debounce(async function () {
```
pringelmann: Debouncing the whole handler delays `update:values` by 400ms, not just validation. Submit reads from `answers`, which only updates on this emit, but the native submit validates the live DOM. So typing a value and submitting within 400ms passes validation and sends the stale value. Can we maybe emit synchronously and debounce only `validate()`?

playwright/support/sections/SubmitSection.ts
```diff
@@ -46,6 +46,8 @@ export class SubmitSection {
 	): Promise<void> {
 		const question = this.getQuestion(questionName)
 		await question.getByRole('textbox').fill(value)
+		// Wait for the debounced input handler to fire (INPUT_DEBOUNCE_MS = 400ms)
+		await this.page.waitForTimeout(500)
```
pringelmann: Hard 500ms sleep will flake and adds half a second to every fillInput. It's for covering the debounce below right? Decouple the emit from validate and this can go; otherwise we should wait on the value/error, not a fixed timeout.

src/components/Questions/QuestionShort.vue
Chartman123: would be fine for me :) I just wanted to avoid constant showing/hiding of the NoteCard on every keystroke.

playwright/support/sections/SubmitSection.ts
Chartman123: Yes, this was introduced to cover the debounce timeout...

src/components/Questions/QuestionShort.vue
pringelmann: Debouncing the validation will also avoid that, since it only shows on error

pringelmann (approved):

PR #3426, opened 2026-06-16
fix: disallow update accessEnum directly

### LINKED ISSUE #3424: updateForm accepts the raw accessEnum key
### Describe the bug

`updateForm` in `PATCH .../forms/{formId}` applies updates with a dynamic setter loop in:

```php
foreach ($keyValuePairs as $key => $value) {
    $method = 'set' . ucfirst($key);
    $form->$method($value);
}
```

That loop will call any matching setter on the `Form` entity. `accessEnum` is the raw integer column that `access` wraps, and it's only meant to be written through `setAccess()`, which normalises the boolean `access` payload and runs the config checks in `checkAccessUpdate()`. Passing `accessEnum` straight to the loop hits `setAccessEnum()` and skips all of that.

`checkForbiddenKeys()` already blocks the other internal columns, but its list doesn't include `accessEnum`:

```php
$forbiddenKeys = [
    'id', 'hash', 'ownerId', 'created', 'lastUpdated', 'lockedBy', 'lockedUntil'
];
```

### Expected behaviour

Clients should set form access through the `access` key only. `accessEnum` is an internal column and shouldn't be writable from the update payload.

### Fix

Add `accessEnum` to the `forbiddenKeys` list in `checkForbiddenKeys()` so the loop rejects it.

Chartman123:
This fixes #3424 and adds tests

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -1875,7 +1875,7 @@ class ApiController extends OCSController {
 	 */
 	private function checkForbiddenKeys(array $keyValuePairs): void {
 		$forbiddenKeys = [
-			'id', 'hash', 'ownerId', 'created', 'lastUpdated', 'lockedBy', 'lockedUntil'
+			'id', 'hash', 'ownerId', 'created', 'lastUpdated', 'lockedBy', 'lockedUntil', 'accessEnum'
 		];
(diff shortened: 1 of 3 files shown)


pringelmann (approved):
Thank you for this!

PR #3432, opened 2026-06-17
feat: remember last results view in localStorage

### LINKED ISSUE #483: Remember last selection of Summary/Responses tab on refresh
Related to Add summary response visualization #460

In the "Responses" view there’s soon 2 tabs: Summary and individual Responses. The default tab is "Summary", but it might be nice to remember the last open tab possibly per-form, as this will be useful for small surveys where you want individual replies?

What do you think @jotoeri @skjnldsv?

Chartman123:
This implements #483 by storing the last used results view to localStorage. Values are stored per form per user, values for forms that are no longer available to the user will be removed upon loading forms.

## 🤖 AI (if applicable)

### DIFF
--- a/src/views/Results.vue
+++ b/src/views/Results.vue
@@ -498,6 +498,7 @@ export default {
 		// Reload results when form changes
 		async hash() {
 			await this.fetchFullForm(this.form.id)
+			this.loadActiveResponseViewFromLocalStorage()
 			this.loadFormResults()
 			SetWindowTitle(this.formTitle)
 		},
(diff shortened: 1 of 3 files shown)


jancborchardt (approved):
Very nice! All works correctly, either when refreshing the page, or navigating between the different View/Edit/Results tabs.

One detail:
The focus is initially always on "Summary", so there is a brief flash of blue over there until it switches to "Responses". Is it possible to have an "indeterminate" state so that until we know which one is selected, none will be visually highlighted?

jancborchardt:
And alternatively – @Chartman123 does it make sense to also add this part of the path at the end of the URL? Then you can specifically bookmark it too, for opening on other devices.

Chartman123:
> And alternatively – @Chartman123 does it make sense to also add this part of the path at the end of the URL? Then you can specifically bookmark it too, for opening on other devices.

Yes, that's what I already thought about too :)

> Is it possible to have an "indeterminate" state so that until we know which one is selected, none will be visually highlighted?

I'll have another look at it.

Thanks for the review! 👍🏻

Chartman123:
@jancborchardt updated the code so that the current results view is also reflected in a query param ~~(`?summary` or `?responses`)~~ (`?view=summary` or `view=responses`)

src/views/Results.vue
```diff
@@ -521,15 +528,183 @@ export default {
...
+			}
+
+			try {
+				await this.$router.push({
+					name: 'results',
+					params: {
+						hash: this.form.hash,
+					},
+					query: {
+						[view.id]: null,
```
samin-z: its better to use empty string instead of null for query params: "[view.id]: '' " to match test expectations ("?summary" and not "?summary = null").

src/views/Results.vue
```diff
@@ -379,7 +380,7 @@ export default {
 
 	data() {
 		return {
-			activeResponseView: responseViews[0],
+			activeResponseView: {},
```
samin-z: this is an empty object, when it check newView?.id it returns 'undefined' while the object exist but no 'id' property exist.
its better to have 'activeResponseView: null', then set it in 'beforeMount()' through 'await this.syncActiveResponseViewFromRoute()'.

samin-z (approved):

PR #3492, opened 2026-07-14
fix: improve user validation on file question type

samin-z:
when submitting the form where question type if file, check with token and question id if the same user who uploaded the file is submitting it.


Chartman123 (approved):
Looks good, just one little comment: please also update the documentation in https://github.com/nextcloud/forms/blob/main/docs/API_v3.md#upload-a-file

samin-z:
thank you @Chartman123 
ill fix the pipelines and then update the documentation

PR #3494, opened 2026-07-14
fix: remove invalid characters constant and use filename validator for sanitization

### LINKED ISSUE #3458: Upload failed with period at the end of the question
File upload fails if there is a period at the end of the question and Windows file name compatibility is enabled.

**To Reproduce**

1. Enable Windows compatible filenames. (`/settings/admin`)
2. Create a new form with a file upload field. Name it for example: `Upload a file.` Period is important!
3. Try to upload a file: Internal Server Error (The folder cannot be found.)

**Expected behavior**

Upload Successful.

**Nextcloud:**

- Nextcloud-Version: 33.0.6
- Forms-Version: 5.3.2

Chartman123:
This fixes #3458. Now that we're at min version 32, we can use `sanitizeFilename()` from `OCP\Files\IFilenameValidator` to cleanup the names of uploaded files.

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -229,19 +229,6 @@ class Constants {
 		self::ANSWER_GRID_TYPE_RADIO,
 	];
 
-	public const FILENAME_INVALID_CHARS = [
-		"\n",
-		'/',
-		'\\',
-		':',
-		'*',
-		'?',
-		'"',
-		'<',
-		'>',
-		'|',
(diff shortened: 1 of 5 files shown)


Koc (approved):
It seems like tests require adjustment

Chartman123:
> It seems like tests require adjustment

They're taking like forever to run today... 🙈

PR #3504, opened 2026-07-16
fix(api): default option type to choice for non-grid questions

### LINKED ISSUE #3505: API: options created without optionType are stored but not rendered
### Steps to reproduce

1. Create a form and a `multiple_unique` (or `dropdown` / `multiple`) question via the API.
2. Add options **without** specifying `optionType`:

```
POST /ocs/v2.php/apps/forms/api/v3/forms/{formId}/questions/{questionId}/options
Content-Type: application/json

{"optionTexts": ["Yes", "No"]}
```

### Expected behaviour

The options are selectable in the rendered form. `optionType` is documented as optional, and `DataStructure.md` states *"`choice` is the default for normal option lists."*

### Actual behaviour

The options are stored with `optionType: null`. The API returns them in `GET .../forms/{id}`, but the frontend does not render them. The question appears with no selectable answers; if it is required, the form cannot be submitted.

The failure is silent: no error, options present in the API response, just not shown.

### Root cause

`ApiController::newOption()` never applies a default `optionType`. The `option_type` migration (`Version050300Date20250914000000`) backfills existing null types to `choice`, but new inserts through the API keep a null type again.

### Server configuration

- Nextcloud: 32.0.9
- Forms: 5.3.3

Fix proposed in #3504.

josefbehr:
### The problem

Creating an option through the API without an `optionType` stores it with a **null** type:

```
POST /ocs/v2.php/apps/forms/api/v3/forms/{formId}/questions/{questionId}/options
{"optionTexts": ["Yes", "No"]}
```

The options are saved and returned by the API, but the frontend does not render them. The question then appears with no selectable answers. If it is a required question, the form cannot be submitted at all. Verified on Nextcloud 32.0.9 / Forms 5.3.3 with `multiple_unique` questions: the options existed in `GET .../forms/{id}` but the rendered form showed an empty question.

This is easy to hit because `optionType` is documented as optional, and the failure is silent: no error, options present in the API response, just not rendered.

### Why 'choice' is the right default

- The `option_type` migration (`Version050300Date20250914000000`) already backfills existing null types to `'choice'`.
- `DataStructure.md` states: *"`choice` is the default for normal option lists."*

So `'choice'` is already the documented and migrated default. `newOption()` simply never applied it to new inserts, so options created after the migration slip through with a null type again.

### The fix

Default `optionType` to `'choice'` for any non-grid question when the caller omits it. Grid questions are deliberately left untouched, since their options require an explicit `row`/`column` type. Adds the missing `Option::OPTION_TYPE_CHOICE` constant next to the existing `OPTION_TYPE_ROW` / `OPTION_TYPE_COLUMN`.

The existing integration test `testCreateNewOption` asserted `optionType => null`, which codified the broken behaviour; updated to expect `'choice'`.

### Context

Found while building a feedback form entirely through the API. `optionType` was omitted on two `multiple_unique` questions, and both rendered without any options.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -938,20 +938,22 @@ class ApiController extends OCSController {
 			throw new OCSBadRequestException('This question is not part ot the given form');
 		}
 
-		// Options of non-grid questions default to the 'choice' type, mirroring
-		// the column default applied by the option_type migration. Without this,
-		// options created through the API keep a null type and are not rendered.
-		if ($optionType === null && $question->getType() !== Constants::ANSWER_TYPE_GRID) {
-			$optionType = Option::OPTION_TYPE_CHOICE;
-		}
-
 		// Retrieve all options sorted by 'order'. Takes the order of the last array-element and adds one.
+		// The lookup keeps the caller-provided $optionType (null means "all options of the question"),
+		// so the ordering is not affected by the default applied below.
 		$options = $this->optionMapper->findByQuestion($questionId, $optionType);
 		$lastOption = array_pop($options);
 		if ($lastOption) {
 			$optionOrder = $lastOption->getOrder() + 1;
 		} else {
 			$optionOrder = 1;
+		}
+
+		// Options of non-grid questions default to the 'choice' type, mirroring the
+		// value applied by the option_type migration. Without a type the stored
+		// option is not rendered by the frontend.
+		if ($optionType === null && $question->getType() !== Constants::ANSWER_TYPE_GRID) {
+			$optionType = Option::OPTION_TYPE_CHOICE;
 		}
 
 		$addedOptions = [];
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -939,21 +939,12 @@ class ApiController extends OCSController {
 		}
 
 		// Retrieve all options sorted by 'order'. Takes the order of the last array-element and adds one.
-		// The lookup keeps the caller-provided $optionType (null means "all options of the question"),
-		// so the ordering is not affected by the default applied below.
 		$options = $this->optionMapper->findByQuestion($questionId, $optionType);
 		$lastOption = array_pop($options);
 		if ($lastOption) {
 			$optionOrder = $lastOption->getOrder() + 1;
 		} else {
 			$optionOrder = 1;
-		}
-
-		// Options of non-grid questions default to the 'choice' type, mirroring the
-		// value applied by the option_type migration. Without a type the stored
-		// option is not rendered by the frontend.
-		if ($optionType === null && $question->getType() !== Constants::ANSWER_TYPE_GRID) {
-			$optionType = Option::OPTION_TYPE_CHOICE;
 		}
 
 		$addedOptions = [];
@@ -963,12 +954,18 @@ class ApiController extends OCSController {
 			$option->setQuestionId($questionId);
 			$option->setText($text);
 			$option->setOrder($optionOrder++);
-			$option->setOptionType($optionType);
+			// Only set the type when the caller provides one, so the 'choice'
+			// column default applies for normal option lists. Setting a null
+			// here would override that default and leave the option unrendered.
+			if ($optionType !== null) {
+				$option->setOptionType($optionType);
+			}
 
 			try {
 				$option = $this->optionMapper->insert($option);
-				// Add the stored option to the collection of added options
-				$addedOptions[] = $option->read();
+				// Re-read so the response reflects the column default that was
+				// applied when no type was set explicitly.
+				$addedOptions[] = $this->optionMapper->findById($option->getId())->read();
 			} catch (IMapperException $e) {
 				$this->logger->error("Failed to add option: {$e->getMessage()}");
 				// Optionally handle the error, e.g., by continuing to the next iteration or returning an error response
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -894,7 +894,7 @@ class ApiController extends OCSController {
 	 * @param int $formId id of the form
 	 * @param int $questionId id of the question
 	 * @param list<string> $optionTexts the new option text
-	 * @param string|null $optionType the new option type (e.g. 'row')
+	 * @param string|null $optionType the new option type (e.g. 'row'), defaults to 'choice'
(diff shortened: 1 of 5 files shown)


lib/Controller/ApiController.php
```diff
@@ -938,6 +938,13 @@ public function newOption(int $formId, int $questionId, array $optionTexts, ?str
 			throw new OCSBadRequestException('This question is not part ot the given form');
 		}
 
+		// Options of non-grid questions default to the 'choice' type, mirroring
+		// the column default applied by the option_type migration. Without this,
+		// options created through the API keep a null type and are not rendered.
+		if ($optionType === null && $question->getType() !== Constants::ANSWER_TYPE_GRID) {
+			$optionType = Option::OPTION_TYPE_CHOICE;
+		}
+
```
Chartman123: Please remove this addition here and adjust the default in line 911

Chartman123 (changes requested):
Please fix the workflows and see my comment.

BTW: please create a bug report first before creating PRs.

lib/Controller/ApiController.php
josefbehr: Thanks, and sorry for jumping straight to PRs. I filed #3505 as the bug report.

On the placement: moving the default up to the signature (so it applies before `findByQuestion`) changes the option ordering. That lookup filters by option type, so a freshly defaulted `choice` stops matching options that were inserted without a type (e.g. the integration fixture, or rows created before the `option_type` migration), and the order resets to 1. The failing CI on the first push was exactly this.

I pushed 754d826 which applies the default *after* the ordering lookup instead: `findByQuestion` still gets the caller-provided type, so ordering is unchanged, and the inserted option still gets `choice`. Happy to switch to a signature default plus a fixture update if you'd prefer that shape.

lib/Controller/ApiController.php
```diff
@@ -954,6 +949,13 @@ public function newOption(int $formId, int $questionId, array $optionTexts, ?str
 			$optionOrder = 1;
 		}
 
+		// Options of non-grid questions default to the 'choice' type, mirroring the
+		// value applied by the option_type migration. Without a type the stored
+		// option is not rendered by the frontend.
+		if ($optionType === null && $question->getType() !== Constants::ANSWER_TYPE_GRID) {
+			$optionType = Option::OPTION_TYPE_CHOICE;
+		}
+
```
Chartman123: Please see my previous comment. This is exactly the same as before.

Chartman123 (changes requested):

lib/Controller/ApiController.php
Chartman123: We should add a new migration step then that adds `choice` to the options without a value. we could also add the default value to the column config.

josefbehr: Done, switched to that approach in f66dd9b:

- New migration gives `option_type` a `choice` default and backfills the remaining null rows.
- `newOption()` no longer writes an explicit null, so the column default applies. It re-reads the inserted option so the POST response reflects the stored value rather than the in-memory null.
- Grid options are unaffected since callers always pass an explicit row/column type.

Fixtures now receive `choice` from the column default, so I updated the affected expectations. One caveat: I can't run the PHP suite locally, so I'm relying on CI to confirm the fixture/default interaction. Happy to adjust if you'd rather not re-read in the controller.

lib/Db/Option.php
Chartman123, line 1: we can keep this

lib/Migration/Version050300Date20260716000000.php
```diff
@@ -0,0 +1,65 @@
...
+		$table = $schema->getTable('forms_v2_options');
+
+		if ($table->hasColumn('option_type')) {
+			$column = $table->getColumn('option_type');
+			if ($column->getDefault() === null) {
+				$column->setDefault('choice');
+			}
+		}
+
+		return $schema;
```
Chartman123:
```suggestion
		$schema = $schemaClosure();
		$table = $schema->getTable('forms_v2_options');
		$changed = false;

		if ($table->hasColumn('option_type')) {
			$column = $table->getColumn('option_type');
			if ($column->getDefault() === null) {
				$column->setDefault('choice');
				$changed = true;
			}
		}

		return $changed ? $schema : null;
```

lib/Controller/ApiController.php
Chartman123:
```suggestion
	 * @param string|null $optionType the new option type (e.g. 'row'), defaults to 'choice'
```

Chartman123:
```suggestion
	public function newOption(int $formId, int $questionId, array $optionTexts, ?string $optionType = 'choice'): DataResponse {
```
You can also use the Constant if you keep it.

lib/Controller/ApiController.php
```diff
@@ -954,12 +954,18 @@ public function newOption(int $formId, int $questionId, array $optionTexts, ?str
 			$option->setQuestionId($questionId);
 			$option->setText($text);
 			$option->setOrder($optionOrder++);
-			$option->setOptionType($optionType);
+			// Only set the type when the caller provides one, so the 'choice'
+			// column default applies for normal option lists. Setting a null
+			// here would override that default and leave the option unrendered.
+			if ($optionType !== null) {
+				$option->setOptionType($optionType);
+			}
```
Chartman123: can be reverted with the above suggestions now that the migration exists

lib/Controller/ApiController.php
```diff
@@ -954,12 +954,18 @@ public function newOption(int $formId, int $questionId, array $optionTexts, ?str
...
 
 			try {
 				$option = $this->optionMapper->insert($option);
-				// Add the stored option to the collection of added options
-				$addedOptions[] = $option->read();
+				// Re-read so the response reflects the column default that was
+				// applied when no type was set explicitly.
+				$addedOptions[] = $this->optionMapper->findById($option->getId())->read();
```
Chartman123: shouldn't be needed with the above suggestions

Chartman123 (changes requested):

josefbehr:
Applied all four suggestions in 99ea1cf: signature default `'choice'`, dropped the guard and the re-read, early return in the migration, and the docblock note. Thanks for the guidance.

Chartman123 (approved):
Great, looks good now :)

PR #3522, opened 2026-07-20
fix: file share

samin-z:
## 🤖 AI (if applicable)

- [-] The content of this PR was partly or fully generated using AI


Chartman123 (approved):
Looks good as far as I can tell.

Should we add a repair step to cleanup left-overs?

PR #3547, opened 2026-07-30
fix: results route handling in dev mode

Chartman123:
This fixes a regression introduced in #3432. When you reloaded a results page with F5 or directly called a subview when the app was running with `npm run dev`, the results view got duplicated over and over into the router-view.

As a fix/refactoring I moved the routes from url params to direct child routes defined in router.ts.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -117,6 +117,16 @@ class PageController extends Controller {
 	 */
 	#[NoAdminRequired()]
 	#[NoCSRFRequired()]
+	#[FrontpageRoute(verb: 'GET', url: '/{hash}/results/{view}', requirements: ['hash' => '[a-zA-Z0-9]{16,}'])]
+	public function resultsViews(string $hash): TemplateResponse {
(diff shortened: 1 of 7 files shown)


samin-z (approved):

pringelmann:
@Chartman123 sorry I was meaning to do a review this morning. I found a bug where switching to different result views contains stale data:

image
Two things here, both needed:


1. `props: { default: true }` is on the parent record, which has no `components`, so it never applies. Move it to both children or the `hash` prop stays empty forever.
2. The `hash` watcher only calls `fetchFullForm()` now. It used to hit `loadFormResults()` through `syncActiveResponseViewFromRoute()`. Add it back, the comment above it still says "Reload results when form changes".

Unrelated: the view only gets saved to localStorage from the `$route.name` watcher, which doesn't fire on mount, so a reloaded `/results/responses` isn't remembered. Saving in `beforeMount` fixes it, but that flips your new redirect test to `/results/responses`.

PR #3563, opened 2026-08-05
fix: enhance results view handling

Chartman123:
This fixes the review comments (https://github.com/nextcloud/forms/pull/3547#issuecomment-5192763898)


pringelmann (approved):
Thank you!

PR #3577, opened 2026-08-11
chore: migrate mixins to composables

Chartman123:
This starts the migration of our frontend codebase from Vue's Option API to the Composition API.

As first step, the current mixins are migrated to composables.

## 🤖 AI (if applicable)


src/models/Permissions.ts
susnux, line 1: This makes not much sense as a composable, should be rather just some constants one imports

Chartman123: So move it to models?

src/composables/useShareLink.ts
```diff
@@ -0,0 +1,99 @@
...
+		const path = isEmbeddingAllowed(share)
+			? `/apps/forms/embed/${share.shareWith}`
+			: `/apps/forms/s/${share.shareWith}`
+		return new URL(generateUrl(path), window.location).href
+	}
+
+	const copyLink = async (event: Event, link: string): Promise<void> => {
+		try {
+			await navigator.clipboard.writeText(link)
+			showSuccess(t('forms', 'Form link copied'))
```
samin-z, line 70: i see 't' is used but not imported from '@nextcloud/l10n'.

src/composables/useQuestion.ts
```diff
@@ -0,0 +1,367 @@
...
+				{
+					keyValuePairs: {
+						[key]: value,
+					},
+				},
+			)
+			emitEvent('forms:last-updated:set', props.formId)
+		} catch (error) {
+			logger.error('Error while saving question', { error })
+			showError(t('forms', 'Error while saving question'))
```
samin-z, line 268: also here 't' is not imported.

src/composables/useUserSearch.ts
```diff
@@ -0,0 +1,230 @@
...
+			&& query.value.trim() !== ''
+			&& query.value.length > minSearchStringLength
+		)
+	})
+
+	const noResultText = computed(() => {
+		if (!query.value) {
+			return t('forms', 'No recommendations. Start typing.')
+		}
+		return t('forms', 'No elements found.')
```
samin-z, line 89: same here.

src/composables/useViewForm.ts
```diff
@@ -0,0 +1,190 @@
...
+		const form = getForm()
+		return markdownit.render(form.description || '') || form.description
+	})
+
+	const isFormLocked = computed(() => {
+		const form = getForm()
+		return (
+			form.lockedUntil === 0
+			|| (Number(form.lockedUntil || 0) > moment().unix()
+				&& form.lockedBy !== getCurrentUser().uid)
```
samin-z: i already tested it manually but no error came up, but old mixin used 'getCurrentUser()?.uid' as optional, maybe it's safer to have the same here too.

AndyScherzinger (approved):

PR #3635, opened 2026-09-09
feat(results): average, median, range and Net Promoter Score for scale questions

global-prog:
### Summary

Scale questions currently show a bar per value, which makes it hard to compare two questions, or two runs of the same form, at a glance.

This adds a small set of figures above the existing bars for `linearscale` questions:

- **count, average, median, lowest, highest**
- for a **0–10** scale, a **Net Promoter Score** with its promoter and detractor shares (the usual 9–10 / 7–8 / 0–6 split)

The median sits next to the average deliberately — at the response counts most forms see, one extreme answer moves an average a long way, and showing both makes that visible.

The NPS line only appears when the scale is actually 0–10, since that is the shape those scales are almost always used for.

### Scope

Purely additive and presentational:

- one file changed, `src/components/Results/ResultsSummary.vue`
- no API change, no schema change, no migration
- questions of other types render exactly as before
- the existing per-value bars are untouched; the figures are shown above them

### Testing

- `npm run lint` and `npm run stylelint` clean
- built against current `main` (`vite --mode production build`) with no errors
- checked with an empty result set (renders nothing), a single response, and a 0–10 scale

Happy to adjust the wording, the choice of figures, or the placement if you would prefer them elsewhere in the summary.

## 🤖 AI (if applicable)


Chartman123:
@global-prog Thanks for your contribution. As said in the other PR: please create an issue/feature request before starting implentation to discuss it. Also please add screenshots to the description with before/after. This should also be added to the new rating quetions in #3637.

Also please add the AI declaration to your description if you used AI for this.

global-prog:
Thanks, noted on opening an issue first. I'll add before/after screenshots to the description. The AI declaration is added there as well.

PR #3636, opened 2026-09-09
feat: copy questions from another form

global-prog:
### Summary

Rebuilding the same set of questions by hand for every new form is tedious. The server already knows how to clone a question together with its options — it simply refuses to do so across forms.

This lifts that restriction for forms the user is allowed to **edit**, and adds a small dialog to pick them.

### Why the restriction could not just be removed

Two things needed handling:

**Permission.** Dropping the same-form test alone would let any question be read out of *any* form by guessing ids, so the source form is now checked with `getFormIfAllowed($sourceFormId, PERMISSION_EDIT)`.

**A latent bug.** `Question::read()` includes the source question's `formId`, and the clone path passes that straight into `Question::fromParams()`. For same-form cloning that is harmless because the value matches; for a cross-form copy the new question would have been created back in the *source* form. The target id is now set explicitly.

### UI

A dialog lists the forms you can edit, then the questions of the chosen one, with select-all.

Copies are issued sequentially because each is appended at the end of the form — in parallel the resulting order would be unpredictable. If some copies succeed and one fails, the message says some were copied rather than implying none were.

### Scope

- no schema change
- no change to the API surface: the `fromId` parameter and its documented behaviour are unchanged for same-form cloning, so `openapi.json` is unaffected
- `lib/Controller/ApiController.php` changes by 10 insertions / 2 deletions

### Testing

- `npm run lint` and `npm run stylelint` clean, `php -l` clean
- checked: copying into an empty form, copying a question with options, copying several at once, and that same-form cloning still behaves as before

Happy to drop the dialog and ship only the backend change if you would rather design the entry point differently.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -570,9 +570,14 @@ class ApiController extends OCSController {
 
 			try {
 				$sourceQuestion = $this->questionMapper->findById($fromId);
-				// Only allow cloning questions that belong to the same form
+				// A question may be cloned from another form, but only from one the user is
+				// allowed to edit. Without that check any question could be read out of any
+				// form by guessing ids.
 				if ($sourceQuestion->getFormId() !== $formId) {
-					throw new OCSBadRequestException('Question doesn\'t belong to given form');
+					$this->formsService->getFormIfAllowed(
+						$sourceQuestion->getFormId(),
+						Constants::PERMISSION_EDIT,
+					);
 				}
 				$sourceOptions = $this->optionMapper->findByQuestion($fromId);
 			} catch (IMapperException) {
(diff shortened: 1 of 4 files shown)


Chartman123:
@global-prog Thanks for your contribution. 👍🏻 However, please create an issue/feature request before starting implementation the next time so that we can discuss it. Also please add the AI declaration to your description if you used AI for this.

global-prog:
Thanks! Understood, I'll open a feature request first for anything further.

While adding tests I found that copying into a form with no questions yet failed with a server error: the copy is appended after the target form's last question, and there isn't one. That couldn't happen while copying was limited to the same form, but it's the ordinary case for this dialog, so it's fixed in the latest commit, along with tests for copying from another form, into an empty form, and from a form without edit rights.

The AI declaration is added to the description.

PR #3637, opened 2026-09-09
feat: rating question type

### LINKED ISSUE #2608: [QuestionType] Rating
Create a question type for ratings. This is basically a linear scale but with stars as options

global-prog:
Closes #2608.

### Summary

A compact rating question, with **stars**, **hearts** or **thumbs up**, configurable from 2 to 10 icons (default 5).

A linear scale already covers 1..N, but renders as a row of radio buttons. A rating is the control people expect for "how would you rate this", takes far less width, and reads at a glance in the results.

### Accessibility

Built from real radio inputs rather than clickable icons, so:

- it stays keyboard navigable, and each option is announced with the value it selects (`3 of 5`) rather than as an unnamed radio
- the hit area is a full `--default-clickable-area` square while the icon itself stays small — an icon-sized target is awkward to hit on a phone
- the hover animation is dropped under `prefers-reduced-motion`
- the icons wrap, since ten of them plus the clear button do not fit one line on a narrow screen

### Storage and validation

The answer is stored as the plain number, so results and CSV export need no special handling.

It is validated **server-side** as a whole number within the configured maximum rather than trusting the client. A rating carries no options, so it is checked on its own rather than as a predefined-option type.

### Scope

- no schema change; `maxRating` and `ratingIcon` live in the existing `extraSettings`
- `FormsQuestionType` is deliberately left alone, matching how `linearscale`, `ranking` and `color` are already handled, so **`openapi.json` is unaffected**

### Testing

- `npm run lint`, `npm run stylelint` and `php -l` clean
- checked: default 5 stars, switching to hearts and thumbs, changing the count, clearing an answer, a required rating left empty, and a submitted value appearing correctly in the results and CSV export

Happy to adjust the default icon count, the icon set, or drop the hearts/thumbs options if you would rather keep it to stars.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -222,11 +222,13 @@ class Constants {
 	];
 
 	/**
-	 * How many icons a rating question offers, and which icon to draw.
-	 * ratingIcon is one of 'star' (default), 'heart' or 'thumb'.
+	 * A rating is a linear scale that always starts at 1 and is drawn as icons, so it
+	 * shares the linear scale's key for its top end (and that key's bounds) rather than
+	 * having one of its own. optionsLowest is deliberately absent: a rating's lowest end
+	 * is always 1. ratingIcon is one of 'star' (default), 'heart' or 'thumb'.
 	 */
 	public const EXTRA_SETTINGS_RATING = [
-		'maxRating' => ['integer', 'NULL'],
+		'optionsHighest' => ['integer', 'NULL'],
 		'ratingIcon' => ['string', 'NULL'],
 	];
 
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -101,6 +101,7 @@ class Constants {
 	public const ANSWER_TYPE_MULTIPLE = 'multiple';
 	public const ANSWER_TYPE_MULTIPLEUNIQUE = 'multiple_unique';
 	public const ANSWER_TYPE_RANKING = 'ranking';
+	public const ANSWER_TYPE_RATING = 'rating';
 	public const ANSWER_TYPE_SHORT = 'short';
 	public const ANSWER_TYPE_TIME = 'time';
 
@@ -121,6 +122,7 @@ class Constants {
 		self::ANSWER_TYPE_MULTIPLE,
 		self::ANSWER_TYPE_MULTIPLEUNIQUE,
 		self::ANSWER_TYPE_RANKING,
+		self::ANSWER_TYPE_RATING,
 		self::ANSWER_TYPE_SHORT,
 		self::ANSWER_TYPE_TIME,
 	];
@@ -217,6 +219,15 @@ class Constants {
 		'columns' => ['array'],
 		'questionType' => ['string'],
 		'rows' => ['array'],
+	];
+
+	/**
+	 * How many icons a rating question offers, and which icon to draw.
+	 * ratingIcon is one of 'star' (default), 'heart' or 'thumb'.
+	 */
+	public const EXTRA_SETTINGS_RATING = [
+		'maxRating' => ['integer', 'NULL'],
+		'ratingIcon' => ['string', 'NULL'],
 	];
 
 	public const EXTRA_SETTINGS_RANKING = [
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -633,6 +633,19 @@ class SubmissionService {
 			}
 
 			// Check if all answers are within the possible options
+			// A rating carries no options, so it is validated on its own rather than as a
+			// predefined-option type: the answer is the number of icons chosen.
+			if ($question['type'] === Constants::ANSWER_TYPE_RATING) {
+				$maxRating = $question['extraSettings']['maxRating'] ?? 5;
+				foreach ($answers[$questionId] as $answer) {
+					if (!ctype_digit((string)$answer)
+						|| (int)$answer < 1
+						|| (int)$answer > $maxRating) {
+						throw new \InvalidArgumentException(sprintf('The answer for question "%s" must be a whole number between 1 and %d.', $question['text'], $maxRating));
+					}
+				}
+			}
+
 			if (in_array($question['type'], Constants::ANSWER_TYPES_PREDEFINED) && empty($question['extraSettings']['allowOtherAnswer'])) {
 				// Normalize option IDs once for consistent comparison (DB may return ints, request may send strings)
 				$optionIds = $this->normalizeOptionIds($question['options'] ?? []);
--- a/lib/Service/SubmissionService.php
+++ b/lib/Service/SubmissionService.php
@@ -633,16 +633,12 @@ class SubmissionService {
 			}
 
 			// Check if all answers are within the possible options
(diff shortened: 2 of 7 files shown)


Chartman123:
@global-prog thanks for the PR :) please add the AI section to the description if you used AI to generate this PR 👍🏻 

some comments on the code itself: as this uses mostly the same logic as linear scale questions, we should re-use the existing extraSettings (especially `maxOption` instead of `maxRating`) and also the validation logic as far as possible. Rest is looking good from a first glance at it. 

Perhaps you could also add a screenshot/recording of how it works to the description so that the @nextcloud/designers could have a look at it without having to build it 👍🏻

global-prog:
Thanks! Done in the latest commit:

- the rating now stores its top end as `optionsHighest` instead of `maxRating`, and gets the linear scale's bounds on it (2 to 10). `optionsLowest` is refused, since a rating always starts at 1.
- the linear scale's range check is moved into one helper that both types call, with the same message and defaults, so linear scale behaviour is unchanged.
- added tests for the rating's settings and its answer validation.

I've added the AI declaration to the description, and I'll add a screenshot of the editor and the filled-in form there as well.

lib/Constants.php
```diff
@@ -219,6 +221,17 @@ class Constants {
 		'rows' => ['array'],
 	];
 
+	/**
+	 * A rating is a linear scale that always starts at 1 and is drawn as icons, so it
+	 * shares the linear scale's key for its top end (and that key's bounds) rather than
+	 * having one of its own. optionsLowest is deliberately absent: a rating's lowest end
+	 * is always 1. ratingIcon is one of 'star' (default), 'heart' or 'thumb'.
+	 */
+	public const EXTRA_SETTINGS_RATING = [
```
Chartman123, line 230: We don't need the doc block here

```suggestion
	public const EXTRA_SETTINGS_RATING = [
```

lib/Service/SubmissionService.php
```diff
@@ -633,18 +633,23 @@ public function validateSubmission(array $questions, array $answers, string $for
...
+			// A rating carries no options, so it cannot go through the predefined-option
+			// branch below, but its answer is a point on a scale exactly as a linear scale's
+			// is, so it is held to the same rule.
+			if ($question['type'] === Constants::ANSWER_TYPE_RATING) {
+				foreach ($answers[$questionId] as $answer) {
+					$this->validateScaleAnswer($question, $answer);
+				}
+			}
+
 			if (in_array($question['type'], Constants::ANSWER_TYPES_PREDEFINED) && empty($question['extraSettings']['allowOtherAnswer'])) {
```
Chartman123, line 645: During the review I noticed that we actually shouldn't keep linearscale in the predefined questions array in Constants.php. As a preparation for a follow-up PR we should already move rating questions to a new type array ANSWER_TYPES_SCALE in Constants and check the question type for being a member of that new array.

```suggestion
			}

			// Check if answers to scale questions are within the limits
			if (in_array($question['type'], Constants::ANSWER_TYPES_SCALE)) {
				foreach ($answers[$questionId] as $answer) {
					$this->validateScaleAnswer($question, $answer);
				}
			}

			// Check if all answers are within the possible options
			if (in_array($question['type'], Constants::ANSWER_TYPES_PREDEFINED) && empty($question['extraSettings']['allowOtherAnswer'])) {
```

lib/Constants.php
```diff
@@ -121,6 +122,7 @@ class Constants {
 		self::ANSWER_TYPE_MULTIPLE,
 		self::ANSWER_TYPE_MULTIPLEUNIQUE,
 		self::ANSWER_TYPE_RANKING,
+		self::ANSWER_TYPE_RATING,
 		self::ANSWER_TYPE_SHORT,
 		self::ANSWER_TYPE_TIME,
 	];
```
Chartman123, line 128:
```suggestion
	];

	// AnswerTypes that use a number scale
	public const ANSWER_TYPES_SCALE = [
		self::ANSWER_TYPE_RATING,
    ];
```

src/components/Questions/QuestionRating.vue
```diff
@@ -0,0 +1,272 @@
...
+import NcButton from '@nextcloud/vue/components/NcButton'
+import NcIconSvgWrapper from '@nextcloud/vue/components/NcIconSvgWrapper'
+import Question from './Question.vue'
+import {
+	QUESTION_EMITS,
+	QUESTION_PROPS,
+	useQuestion,
+} from '../../composables/useQuestion.ts'
+
+/** Matches the linear scale default the server assumes when optionsHighest is unset. */
```
Chartman123, line 93: We could skip this comment here

src/components/Questions/QuestionRating.vue
```diff
@@ -0,0 +1,272 @@
...
+				class="rating__icon"
+				:class="{ 'rating__icon--on': value <= currentValue }">
+				<input
+					class="hidden-visually"
+					type="radio"
+					:name="`rating_${id}`"
+					:aria-label="
+						n('forms', '%n of {max}', '%n of {max}', value, {
+							max: optionsHighest,
+						})
```
Chartman123, line 54: why don't you use just `t` here?

src/components/Questions/QuestionRating.vue
```diff
@@ -0,0 +1,272 @@
...
+		:titlePlaceholder="answerType.titlePlaceholder"
+		:warningInvalid="answerType.warningInvalid"
+		:errorMessage="errorMessage"
+		v-on="commonListeners">
+		<template #actions>
+			<NcActionInput
+				:modelValue="optionsHighest"
+				type="multiselect"
+				:clearable="false"
+				:label="t('forms', 'Number of icons')"
```
Chartman123, line 18: Could we find a better wording here?
```suggestion
				:label="t('forms', 'Value of highest rating')"
```

src/components/Questions/QuestionRating.vue
```diff
@@ -0,0 +1,272 @@
...
+					@change="onPick(value)" />
+				<NcIconSvgWrapper
+					:svg="value <= currentValue ? filledIcon : outlineIcon" />
+			</label>
+			<NcButton
+				v-if="readOnly && currentValue"
+				variant="tertiary"
+				@click="onPick(0)">
+				{{ t('forms', 'Clear') }}
+			</NcButton>
```
Chartman123, line 68: Do we really need this button? Radio/Linear Scale question can't be cleared neither. If we really want to keep it, we should use a separate method `onClear` and really unset the value

src/components/Questions/QuestionRating.vue
Chartman123, line 1: Looks good in general, however I'm not sure about the presentation of the question. Google e.g. uses colored buttons (stars -> yellow, hearts -> red, thumbs -> blue). We could also use the whole width for displaying the icons like we do for linear scale questions. The settings could also be reworked a little and contain the icons and the texts. Google also has labels above the icons.



But I think this is something our designers could review/judge best :)

Chartman123 (commented):
I've add a closer look on the code now and added a few more suggestions and comments. Could you please also rebase your branch on current main? Thanks :)

PR #3638, opened 2026-09-09
feat: image and video blocks

### LINKED ISSUE #624: Add sections or multiple descriptions

global-prog:
### Summary

Two display-only blocks, so a form can show a picture or point at a video between its questions rather than only text.

They are modelled as ordinary question types that happen to carry no answer, which keeps them inside the existing ordering and drag-and-drop with no new concepts to learn. A shared `ANSWER_TYPES_DISPLAY_ONLY` list marks them, so they are skipped when validating a submission and filtered out of exports — without that, each block would add an empty column to every CSV.

### One deliberate choice: video is a link, not an embed

Embedding would make **every respondent's browser contact the video host on page load**, disclosing their IP address and usually setting cookies. On a form marked as anonymous that would quietly undermine the promise being made to respondents, so the block renders as a link they choose to follow.

Images do render inline, since that is the point of an image block, but with `referrerpolicy="no-referrer"` so the form's own address is not passed on. The editor also warns when the address is not on this instance:

> This address is on another site. Loading it tells that site the IP address of everyone who opens the form.

I am happy to add an opt-in embed for video if you would rather have it, but I did not think it should be the default.

### Scope

- no schema change; the address and description live in the existing `extraSettings`
- `FormsQuestionType` is left alone, matching how `linearscale`, `ranking` and `color` are already handled, so **`openapi.json` is unaffected**

### Testing

- `npm run lint`, `npm run stylelint` and `php -l` clean
- checked: a block with no address set, an image on this instance, an image elsewhere (warning shown), a video link, submitting a form containing both blocks, and confirming neither appears as a column in the CSV export

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -100,9 +100,11 @@ class Constants {
 	public const ANSWER_TYPE_LONG = 'long';
 	public const ANSWER_TYPE_MULTIPLE = 'multiple';
 	public const ANSWER_TYPE_MULTIPLEUNIQUE = 'multiple_unique';
+	public const ANSWER_TYPE_IMAGE = 'image';
 	public const ANSWER_TYPE_RANKING = 'ranking';
 	public const ANSWER_TYPE_SHORT = 'short';
 	public const ANSWER_TYPE_TIME = 'time';
+	public const ANSWER_TYPE_VIDEO = 'video';
 
 	public const ANSWER_GRID_TYPE_CHECKBOX = 'checkbox';
 	public const ANSWER_GRID_TYPE_NUMBER = 'number';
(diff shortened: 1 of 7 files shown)


Chartman123:
@global-prog Please look at the corresponding issue at #624 to see if you covered all of what has been discussed there already. Also please add the AI declaration to your description if you used AI for this.

global-prog:
Thanks, I read through #624.

It is mostly about section headings and descriptions to break up a long form, which this PR does not do: it only adds image and video blocks. So I don't think it should close #624 on its own. Section headings are what #2913 is working on.

One thing that may help there: this PR adds a shared `ANSWER_TYPES_DISPLAY_ONLY` list in `Constants`, which `SubmissionService` checks during validation and when building the export. I've also added the backend check you asked for on #2913: an answer submitted to a display-only block is now refused rather than skipped, with tests. If sections land as a question type they could go on the same list and get that behaviour without a second mechanism.

If you decide display content shouldn't be a question type, which is the open question on #2913, I'll follow that decision here too.

The AI declaration is added to the description.

Issue #2643, opened 2025-03-16
Offer a reference form of all available options to first time app users


sunjam:
**Nextcloud (please complete the following information):**

Any

**Is your feature request related to a problem? Please describe.**
Currently it is unclear what the options are in forms at a glance.  I've created a template that visually lays it all out within a form, making it easy to share with others.  [See the live example here](https://cloud.disroot.org/apps/forms/s/mNNHwzSBmWriWH2Cr5xTMJaN)

**Describe the solution you'd like**
I'd suggest including a skeleton form that includes all of the form options spelled out.  I think this is useful as a skeleton form for people to immediately understand how it works.

**Describe alternatives you've considered**
Manually browse through options.

**Additional context**
Screenshots follow:

![Image](https://github.com/user-attachments/assets/ad476e8e-3819-427e-a1b2-8086e68356dc)
![Image](https://github.com/user-attachments/assets/bfa08513-81f3-4a48-b567-0fe94faf666e)

Chartman123:
This will blow up the database especially for large installations. None of the other Form applications (e.g. Google, Microsoft) offer this neither. Most of our features are already self-explaining. So I don't think that we really need this.

@susnux what do you think?

susnux:
Not sure how this is related to the database?
We could add this on the empty view (`/apps/forms`) just as a fake form.

But in general I think it should be self explaining and not needed. Especially it will get much more complex over time.
Also you can just experiment your self with questions, no?

Chartman123:
> We could add this on the empty view (`/apps/forms`) just as a fake form.

Yes, but then it has limited functionallity.

But I'm also not aware that other apps provide something similar... Perhaps we could add a more user facing documentation to our docs? But then again: normal users wouldn't find it directly if we don't link it inside the app.

sunjam:
Adding through user facing documentation makes total sense!

Chartman123:
@sunjam you're absolutely right about where the user docs reside :) they are maintained in this repository: https://github.com/nextcloud/documentation

So if you want you can start contributing there. Any help is much appreciated 🙏🏻

sunjam:
@Chartman123 I cannot find any user facing documentation for Forms in the User Manual.

Looking in https://docs.nextcloud.com/server/latest/user_manual/en/index.html

Do I need to create a whole new section?  Also not finding much in the latest admin manual either.

Chartman123:
[...]

PR #3639, opened 2026-09-09
feat: pick number, phone, email and web address when adding a question

global-prog:
### Summary

Validation for **phone numbers**, **email addresses** and **numbers** already exists — but it can only be reached from the input-type menu *inside* a short-text question that has already been created. Someone adding a question sees only "Short answer" and has no particular reason to know the rest is there.

This makes four of them pickable directly from the add-question menu:

- **Number**
- **Phone number**
- **Email address**
- **Web address**

### How

Each is a **preset**: an ordinary short-text question created with its validation already set, applied client-side after creation. No new backend type is introduced and nothing changes on the server.

**Email is deliberately a preset rather than a type of its own.** `Question::checkEmailType()` recognises an email question only as short text carrying email validation, and the confirmation-email recipient picker depends on that test — a dedicated `email` type would have quietly broken it.

**Web address** uses the existing regex validation, since there is no dedicated URL validator. The pattern is kept simple because it has to be valid in **both JavaScript and PCRE** — the browser and the server each apply it — and it was checked against both engines to behave identically:

| input | result |
|---|---|
| `https://example.org/x` | accepted |
| `http://a.b` | accepted |
| `HTTPS://X.Y` | accepted |
| `not a url` | rejected |
| `ftp://x.y` | rejected |
| `https://has space` | rejected |

The preset mechanism is generic, so further presets cost only an entry in the answer type list.

### Scope

- two files, frontend only
- no schema change, no API change, `openapi.json` unaffected

### Testing

- `npm run lint`, `npm run typecheck` and `prettier --check` all clean
- checked each of the four: created from the menu, validation applied immediately without a reload, an invalid value rejected on submit, and the confirmation-email recipient picker still offering a question created via the Email preset

### DIFF
--- a/src/models/AnswerTypes.ts
+++ b/src/models/AnswerTypes.ts
@@ -7,12 +7,16 @@ import type { Component } from 'vue'
 import type { FormsOption, FormsQuestion } from '../types/Entities.d.ts'
 
 import IconNumeric from '@material-symbols/svg-400/outlined/123.svg?raw'
+import IconNumeric from '@material-symbols/svg-400/outlined/123.svg?raw'
 import IconArrowDownDropCircleOutline from '@material-symbols/svg-400/outlined/arrow_drop_down_circle.svg?raw'
 import IconCalendar from '@material-symbols/svg-400/outlined/calendar_today.svg?raw'
+import IconPhone from '@material-symbols/svg-400/outlined/call.svg?raw'
 import IconCheckboxOutline from '@material-symbols/svg-400/outlined/check_box.svg?raw'
 import IconFile from '@material-symbols/svg-400/outlined/draft.svg?raw'
 import IconGrid from '@material-symbols/svg-400/outlined/grid_view.svg?raw'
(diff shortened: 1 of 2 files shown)


Chartman123:
I don't think that we should add this overhead.

global-prog:
Thanks for taking a look.

So that I address the right thing — do you mean the four extra entries in the add-question menu, or the preset mechanism itself?

What I was trying to fix is discoverability rather than capability. Phone, email and number validation already exist, but they can only be reached from the input-type menu *inside* a short-text question that has already been created, so someone adding a question has no particular reason to know they are there.

If menu length is the concern, two smaller options keep that without four new entries:

- surface the input type at add-time on the existing **Short answer** entry, instead of adding separate ones;
- or add **Number** only, as the one that is arguably a distinct question type rather than a text field with a pattern.

And if you would rather not have any of it, I am happy to close it — no objection from me. Just say which you prefer.

Chartman123:
@global-prog I'd prefer creating a user-facing documentation/manual that describes the features. see #2643

PR #3659, opened 2026-09-13
fix: update dateMin and dateMax to accept string or integer formats and enhance validation logic

### LINKED ISSUE #3617: Start-/End-Limited Datepickers validation incorrect due to timezone
**Bug Report + Workaround**
**Issue with:**
Datepicker in Forms

**Symptom:**
After creating a form that contains a datepicker with a specific end date, the users may have issues submitting the form when the selected date is precisely the end date. Resulting in the error feedback "Date/time is not in the allowed range for Question X"

**Investigation:**
My first idea was "that feels like a timezone issue".
When creating the form, the server received a Unix-Timestamp of the selected start/end date within the local system timezone.
When a user is opening the form, the server is sending this specific Unix-Timestamp to the user.
When a user is sending an answer, it sends a date-string, e.g. "24.08.2026" not a Unix timestamp.

**Regular Users:**
Regular users are from Europe (Germany) and mostly use GMT +2 Berlin summertime as their local timezone.

**Workaround:**
When the creator uses **GMT -2 or lower** on his local system, a regular user **can** submit the end-date without error.
When the creator uses **GMT -1 or higher** on his local system, a regular user **cannot** submit the end-date without error.

**Assumption:**
The server converts the user's date string to a Unix timestamp with the server's timezone and compares it with the start/end Unix timestamp of the form.

**Can it be recreated?**
Create your own form (have local timezone GMT +2 Berlin summertime active).
Add a Date.
Add an end-date limitation.
Test your form and submit the chosen end-date. (Test-Instance should be configured to GMT +0)

**Configuration:**
The (Docker-) Nextcloud instance runs on 33.0.5.1 with the Forms add-on.
Local instance timezone is set to GMT +0

**Potential approach to fix this issue:**
When a form with a datepicker and end date is created, don't save a Unix timestamp. Use a date string instead—just the same as what is sent when answering the form.

Chartman123:
Fixes #3617 by making the validation of date limits timezone-independent

## 🤖 AI (if applicable)


lib/Service/SubmissionService.php
```diff
@@ -752,10 +752,31 @@ private function validateDateTime(array $answers, string $format, ?string $text
...
+					: null;
+				$timeMax = isset($extraSettings['timeMax'])
+					? DateTime::createFromFormat($format, $extraSettings['timeMax'])
+					: null;
+
+				$compareDate = ($dateMin !== null || $dateMax !== null)
+					? (DateTime::createFromFormat('!' . $format, $dateStr) ?: $d)
+					: $d;
+
+				if (($dateMin !== null && $compareDate < $dateMin)
```
samin-z: 'createFromFormat' can return 'false' and not only 'dateTime' or 'null', and if it's false then it can pass the check.
instead you can use
```
if (($dateMin instanceof \DateTimeInterface && $compareDate < $dateMin)
	|| ($dateMax instanceof \DateTimeInterface && $compareDate > $dateMax)
	|| ($timeMin instanceof \DateTimeInterface && $d < $timeMin)
	|| ($timeMax instanceof \DateTimeInterface && $d > $timeMax)
)
```

lib/Service/SubmissionService.php
```diff
@@ -752,10 +752,31 @@ private function validateDateTime(array $answers, string $format, ?string $text
 			$previousDate = $d;
 
 			if ($extraSettings) {
-				if ((isset($extraSettings['dateMin']) && $d < (new DateTime())->setTimestamp($extraSettings['dateMin']))
-					|| (isset($extraSettings['dateMax']) && $d > (new DateTime())->setTimestamp($extraSettings['dateMax']))
-					|| (isset($extraSettings['timeMin']) && $d < DateTime::createFromFormat($format, $extraSettings['timeMin']))
-					|| (isset($extraSettings['timeMax']) && $d > DateTime::createFromFormat($format, $extraSettings['timeMax']))
+				$dateMin = isset($extraSettings['dateMin'])
+					? (is_int($extraSettings['dateMin'])
```
samin-z: 'is_int' do not catch numeric strings like '16900000'. for legacy timestamps something like 'is_numeric' and then casting it to int is safer than just using 'is_int'

lib/Service/SubmissionService.php
```diff
@@ -752,10 +752,31 @@ private function validateDateTime(array $answers, string $format, ?string $text
...
+						? (new DateTime())->setTimestamp($extraSettings['dateMin'])->setTime(0, 0, 0)
+						: DateTime::createFromFormat('!' . $format, $extraSettings['dateMin']))
+					: null;
+				$dateMax = isset($extraSettings['dateMax'])
+					? (is_int($extraSettings['dateMax'])
```
samin-z: same as before 'is_int'

lib/Service/FormsService.php
```diff
@@ -861,9 +861,38 @@ public function areExtraSettingsValid(array $extraSettings, string $questionType
 
 		// Validate extraSettings for specific question types
 		if ($questionType === Constants::ANSWER_TYPE_DATE) {
+			$format = Constants::ANSWER_PHPDATETIME_FORMAT['date'];
+			$dateMinDate = null;
+			$dateMaxDate = null;
+
+			// Validate dateMin format
+			if (isset($extraSettings['dateMin'])) {
+				if (is_int($extraSettings['dateMin'])) {
```
samin-z: same as SubmissionService for 'is_int'

lib/Service/FormsService.php
```diff
@@ -861,9 +861,38 @@ public function areExtraSettingsValid(array $extraSettings, string $questionType
...
+					$dateMinDate = \DateTime::createFromFormat('!' . $format, $dateMinString);
+					if (!$dateMinDate || $dateMinDate->format($format) !== $dateMinString) {
+						return false;
+					}
+				}
+			}
+
+			// Validate dateMax format
+			if (isset($extraSettings['dateMax'])) {
+				if (is_int($extraSettings['dateMax'])) {
```
samin-z: here too.

lib/Service/SubmissionService.php
```diff
@@ -752,10 +752,31 @@ private function validateDateTime(array $answers, string $format, ?string $text
 			$previousDate = $d;
 
 			if ($extraSettings) {
-				if ((isset($extraSettings['dateMin']) && $d < (new DateTime())->setTimestamp($extraSettings['dateMin']))
-					|| (isset($extraSettings['dateMax']) && $d > (new DateTime())->setTimestamp($extraSettings['dateMax']))
-					|| (isset($extraSettings['timeMin']) && $d < DateTime::createFromFormat($format, $extraSettings['timeMin']))
-					|| (isset($extraSettings['timeMax']) && $d > DateTime::createFromFormat($format, $extraSettings['timeMax']))
+				$dateMin = isset($extraSettings['dateMin'])
+					? (is_int($extraSettings['dateMin'])
+						? (new DateTime())->setTimestamp($extraSettings['dateMin'])->setTime(0, 0, 0)
```
samin-z: *** can be fixed in the future but quick mention: 
old timestamp limits can still pick the wrong calendar day because of timezone (server vs browser). when reading legacy int, turn it into a 'Y-m-d' string first and then use the same string comparison path as the new limits instead of 'setTimestamp()->setTime(0, 0, 0)', something like:
```

$dateMin = (new \DateTime('@' . $extraSettings['dateMin']))
	->setTimezone(new \DateTimeZone('UTC'))
	->format('Y-m-d');
$dateMin = \DateTime::createFromFormat('!Y-m-d', $dateMin);
```

lib/Service/SubmissionService.php
```diff
@@ -752,10 +752,31 @@ private function validateDateTime(array $answers, string $format, ?string $text
...
+						: DateTime::createFromFormat('!' . $format, $extraSettings['dateMin']))
+					: null;
+				$dateMax = isset($extraSettings['dateMax'])
+					? (is_int($extraSettings['dateMax'])
+						? (new DateTime())->setTimestamp($extraSettings['dateMax'])->setTime(0, 0, 0)
```
samin-z: same here as ***

samin-z (approved):

samin-z:
one point maybe worth mentioning (maybe as a follow-up), when the range is selected for question as a user answering that question it could be confusing why i can not select outside of the range if i do not know what that range is . it is also possible that in description or in question title more info is given but does it make sense to give more info in that regard to the users?
@Chartman123

Chartman123:
You mean something like for Checkbox questions where we display a NcNotecard if there are existing limits?

samin-z:
> You mean something like for Checkbox questions where we display a NcNotecard if there are existing limits?

yes, something like that, but you have more info on how forms are used so maybe there is no need for that.

Chartman123:
yes and we already have an open issue for #2643

PR #3660, opened 2026-09-13
[stable5.3] fix: update dateMin and dateMax to accept string or integer formats and enhance validation logic

### LINKED ISSUE #3617: Start-/End-Limited Datepickers validation incorrect due to timezone

Chartman123:
Fixes #3617 in stable5.3 by making the validation of date limits timezone-independent.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI (N/A)


Chartman123:
Playwright tests failure is unrelated to the changes here.

